PR #9, opened 2019-06-13
Item 5754: [LabKey UX] Add/remove fields (API) - Automated Testing: Refactor common code. Create a new helper for saveDomain api.

labkey-bpatel:


src/org/labkey/remoteapi/domain/GetDomainCommand.java
```diff
@@ -1,13 +1,10 @@
 package org.labkey.remoteapi.domain;
 
 import org.json.simple.JSONObject;
-import org.labkey.remoteapi.PostCommand;
 
-public class GetDomainCommand  extends PostCommand<DomainResponse>
+public class GetDomainCommand  extends DomainCommand
```
labkey-tchad: I'd leave `GetDomainCommand` an independent class. `getDomain` doesn't require a POST and many of the methods in `DomainCommand` aren't relevant (e.g. `setDomainKind`).

labkey-bpatel: Done!

PR #16, opened 2019-06-19
Item 5779: missing value test automation

labkey-chrisj:
This adds test coverage for saveBatch api
note that the test case using the Java remoteAPI classes is marked 'ignore'- the RemoteAPI objects to be broken beyond the scope of this story

### DIFF
--- a/src/org/labkey/test/tests/SampleSetRemoteAPITest.java
+++ b/src/org/labkey/test/tests/SampleSetRemoteAPITest.java
@@ -523,14 +523,15 @@ public class SampleSetRemoteAPITest extends BaseWebDriverTest
         int imported = 0;
         for(TestDataGenerator dataGen : dataGenerators)
         {
-            new AssayImportPage(getDriver()).setNamedTextAreaValue("TextAreaDataCollector.textArea",
+            AssayImportPage page = new AssayImportPage(getDriver())
+                    .setNamedTextAreaValue("TextAreaDataCollector.textArea",
                     dataGen.writeTsvContents());
             imported++;
 
             if(imported < limit)
-                clickButton("Save and Import Another Run");
+                page.clickSaveAndImportAnother();
             else
-                clickButton("Save and Finish");
+                page.clickSaveAndFinish();
         }
     }
 
--- a/src/org/labkey/test/tests/SampleSetRemoteAPITest.java
+++ b/src/org/labkey/test/tests/SampleSetRemoteAPITest.java
@@ -301,7 +301,8 @@ public class SampleSetRemoteAPITest extends BaseWebDriverTest
                 "      materialOutputs: [\n" +
                 "        // create/update samples in the target sample set\n" +
                 "        { rowId: "+secondAddedRowId+", sampleSet: { name: 'mvAssaySamples' }, properties: { 'mvStringData': 'newer from api', volume: 7 } },\n" +
-                "        { name: 'another really new one', sampleSet: { name: 'mvAssaySamples' }, properties: { 'mvStringData': 'also new from api', volume:17 } }\n" +
+                "        { name: 'another really new one', sampleSet: { name: 'mvAssaySamples' }, " +
+                "           properties: { 'mvStringData': 'also new from api', volume:17, mvstringdata_mvindicator : 'N' } }\n" +
                 "      ],\n" +
                 "\n" +
                 "      properties: {\n" +
@@ -326,6 +327,10 @@ public class SampleSetRemoteAPITest extends BaseWebDriverTest
         assertNotNull("expect a new sample to be added to the set", thirdAddedSample);
         assertEquals(17.0, thirdAddedSample.get("volume"));
         assertEquals("also new from api", thirdAddedSample.get("mvStringData"));
+        assertEquals("N", thirdAddedSample.get("mvstringdata_mvindicator"));
+
+        //TODO: validate the update sample scenario if it ever works
+        // thirdAddedSample receiving no mvIndicator value is a repro of https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=37786
     }
 
 
--- a/src/org/labkey/test/tests/SampleSetRemoteAPITest.java
+++ b/src/org/labkey/test/tests/SampleSetRemoteAPITest.java
@@ -15,17 +15,30 @@
  */
 package org.labkey.test.tests;
 
+import org.json.simple.JSONObject;
 import org.junit.Before;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.labkey.remoteapi.CommandException;
+import org.labkey.remoteapi.assay.Batch;
+import org.labkey.remoteapi.assay.Data;
+import org.labkey.remoteapi.assay.Material;
+import org.labkey.remoteapi.assay.Run;
+import org.labkey.remoteapi.assay.SaveAssayBatchCommand;
+import org.labkey.remoteapi.assay.SaveAssayBatchResponse;
+import org.labkey.remoteapi.query.SelectRowsResponse;
 import org.labkey.test.BaseWebDriverTest;
 import org.labkey.test.Locator;
 import org.labkey.test.SortDirection;
 import org.labkey.test.TestTimeoutException;
 import org.labkey.test.WebTestHelper;
 import org.labkey.test.categories.DailyA;
+import org.labkey.test.components.CustomizeView;
+import org.labkey.test.pages.AssayDesignerPage;
+import org.labkey.test.pages.assay.AssayImportPage;
+import org.labkey.test.pages.assay.AssayRunsPage;
 import org.labkey.test.params.FieldDefinition;
 import org.labkey.test.util.DataRegionTable;
 import org.labkey.test.util.PortalHelper;
@@ -33,10 +46,12 @@ import org.labkey.test.util.TestDataGenerator;
 
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 
 @Category({DailyA.class})
 @BaseWebDriverTest.ClassTimeout(minutes = 10)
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/SampleSetRemoteAPITest.java
```diff
@@ -181,6 +196,328 @@ public void sampleSetWithMissingValueField() throws IOException, CommandExceptio
...
+                        TestDataGenerator.simpleFieldDef("mvStringData", FieldDefinition.ColumnType.String)
+                                .setMvEnabled(true).setLabel("MV Field"),
+                        TestDataGenerator.simpleFieldDef("volume", FieldDefinition.ColumnType.Double)
+                ));
+        dgen.createDomain(createDefaultConnection(true), "SampleSet");
+        dgen.addCustomRow(Map.of("name", "1st", "volume", 13.5));
+        dgen.addCustomRow(Map.of("name", "2nd", "volume", 15.5));
+        dgen.addCustomRow(Map.of("name", "3rd", "volume", 16.5));
+        dgen.addCustomRow(Map.of("name", "4th", "volume", 17.5));
+        dgen.insertRows(createDefaultConnection(true), dgen.getRows());
```
labkey-danield: This is missing a field to create a missing value. If the mvStringData (MV Field) field is to be a missing value field you need to se the mvstringdata_mvindicator field as well. The *_mvindicator field is the one that indicates what the actual missing value is, without setting it you just have an empty field.

Also be aware that the case of the *_mvindicator field is different in MSSQL (_MVIndicator) and Postgres (_mvindicator).

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
```diff
@@ -181,6 +196,328 @@ public void sampleSetWithMissingValueField() throws IOException, CommandExceptio
...
+                "  assayId: "+assayIdStringValue+",\n" +
+                "  batch: {\n" +
+                "    name: 'foo',\n" +
+                "    runs: [{\n" +
+                "      name: 'bar',\n" +
+                "\n" +
+                "      materialOutputs: [\n" +
+                "        // create some new samples in the target sample set if it doesn't exist\n" +
+                "        { name: 'a new one', sampleSet: { name: 'mvAssaySamples' }, properties: { 'mvStringData': 'new from api' } },\n" +
+                "        { name: 'another new one', sampleSet: { name: 'mvAssaySamples' }, properties: { 'mvStringData': 'also new from api' } }\n" +
```
labkey-danield: For a missing value test wouldn't it be useful to set the mvStringData, and mvstringdata_mvindicator appropriately?

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
```diff
@@ -181,6 +196,328 @@ public void sampleSetWithMissingValueField() throws IOException, CommandExceptio
...
+                "\n" +
+                "      dataRows: [{\n" +
+                "        dataFileField: 'S4',\n" +
+                "        sampleId: 'def'\n" +
+                "      }]\n" +
+                "    }]\n" +
+                "  }\n" +
+                "})";
+        executeScript(updateScript);
+        sleep(2500); // ugh, again.
```
labkey-danield: Not sure why you need the sleep, could it be because the update isn't working as expected?
How about calling executeAsyncScript?

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
```diff
@@ -181,6 +196,328 @@ public void sampleSetWithMissingValueField() throws IOException, CommandExceptio
...
+        Map<String, Object> thirdAddedSample = updateResponse.getRows()
+                .stream()
+                .filter(a-> a.get("Name").equals("another really new one"))
+                .findFirst().orElse(null);
+
+        assertNotNull("expect a new sample to be added to the set", thirdAddedSample);
+        assertEquals(17.0, thirdAddedSample.get("volume"));
+        assertEquals("also new from api", thirdAddedSample.get("mvStringData"));
+
+
```
labkey-danield: You may want to add a TODO here as a note to check for the updated sample once that is working. 
If it turns out that update isn't valid then this part of the test adds no value (you already did a create above).

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
```diff
@@ -181,6 +196,328 @@ public void sampleSetWithMissingValueField() throws IOException, CommandExceptio
...
+    }
+
+
+    @Test
+    @Ignore
```
labkey-danield: You may want to add a comment as why this test is ignored.

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
```diff
@@ -181,6 +196,328 @@ public void sampleSetWithMissingValueField() throws IOException, CommandExceptio
...
+        DataRegionTable sampleSetList =  DataRegionTable.DataRegion(getDriver()).withName("SampleSet").waitFor();
+        waitAndClick(Locator.linkWithText(sampleSetName));
+        DataRegionTable materialsList =  DataRegionTable.DataRegion(getDriver()).withName("Material").waitFor();
+
+        // get the samples from the sampleset
+        SelectRowsResponse sampleSetResponse = dgen.getRowsFromServer(createDefaultConnection(true));
+
+        // at this point, the API either doesn't work or requires some hack to make it work.
+        // the materialoutputs are present, but the API does not create samples derived from the runs.
+
```
labkey-danield: I realize this test is ignored but it looks like the intent of this test is to validate that you can derive a sample but doesn't have anything to do with missing values. Correct?

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
```diff
@@ -181,6 +196,328 @@ public void sampleSetWithMissingValueField() throws IOException, CommandExceptio
...
+        return designerPage;
+    }
+
+    /**
+     * generates data (runs) for the gpat assay used by tests in this class.
+     * @param assayName
+     * @return
+     */
+    private List<TestDataGenerator> insertAssayData(String assayName, String assaySubfolder, FieldDefinition.LookupInfo assayLookup)
```
labkey-danield: It doesn't look like the parameter assaySubFolder is used.

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
```diff
@@ -181,6 +196,328 @@ public void sampleSetWithMissingValueField() throws IOException, CommandExceptio
...
+        new AssayImportPage(getDriver()).setNamedTextAreaValue("TextAreaDataCollector.textArea", pasteData1);
+        clickButton("Save and Import Another Run");
+        new AssayImportPage(getDriver()).setNamedTextAreaValue("TextAreaDataCollector.textArea", pasteData2);
+        clickButton("Save and Import Another Run");
+        new AssayImportPage(getDriver()).setNamedTextAreaValue("TextAreaDataCollector.textArea", pasteData3);
+        clickButton("Save and Finish");
+
+        return Arrays.asList(dgen1, dgen2, dgen3);
+    }
+
```
labkey-danield: Would there be any value in making a more generic insertAssayData function? It looks like this is for tests that will always use the same data, which is useful, for future tests it might be useful to make a simpler function that allows for different data/fields:

    private void _insertAssayData(String assayName, List<TestDataGenerator> listDataGen)
    {
        waitAndClickAndWait(Locator.linkWithText(assayName));
        AssayRunsPage assayRunsPage = new AssayRunsPage(getDriver());
        DataRegionTable runsTable = assayRunsPage.getTable();
        runsTable.clickHeaderButton("Import Data");
        clickButton("Next");

        int limit = listDataGen.size();
        int imported = 0;
        for(TestDataGenerator dataGen : listDataGen)
        {
            new AssayImportPage(getDriver()).setNamedTextAreaValue("TextAreaDataCollector.textArea", dataGen.writeTsvContents());
            imported++;
            
            if(imported < limit)
                clickButton("Save and Import Another Run");
            else
        }
        
    }

This might make it more flexible for future tests that could have different columns/column-types.

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
labkey-tchad: I thought the `*_mvindicator` field is automatically created. Or maybe it's just automatic in the UI because the GWT designer knows to create it when that checkbox is checked.

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
```diff
@@ -181,6 +196,328 @@ public void sampleSetWithMissingValueField() throws IOException, CommandExceptio
...
+                "      },\n" +
+                "\n" +
+                "      dataRows: [{\n" +
+                "        dataFileField: 'S3',\n" +
+                "        sampleId: 'def'\n" +
+                "      }]\n" +
+                "    }]\n" +
+                "  }\n" +
+                "})";
+        executeScript(createScript);
```
labkey-tchad: Use `executeAsyncScript`. See comment below for details.

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
labkey-tchad: To use `executeAsyncScript`, you need to provide success and failure callbacks to the `saveBatch` JavaScript config.
Should be as simple as adding this after the `assayId` line in the `updateScript`:
```
"success: callback, failure:callback, \n" +
```
Check the JavaScript docs for more info on [`LABKEY.Experiment.saveBatch(config)`](https://www.labkey.org/download/clientapi_docs/javascript-api/symbols/LABKEY.Experiment.html#.saveBatch)

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
labkey-chrisj: That's actually a good idea.  I like it

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
```diff
@@ -181,6 +196,328 @@ public void sampleSetWithMissingValueField() throws IOException, CommandExceptio
...
+
+        // add a bodyWebPart here to make viewing the sampleset easier while debugging
+        new PortalHelper(getDriver()).addBodyWebPart("Sample Sets");
+        refresh();
```
labkey-tchad: Shouldn't need to refresh. This is probably compensating for not using `executeAsyncScript` to save the batch

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
labkey-tchad: Also, `AssayImportPage` has a method to save and import another run that you could chain together instead of finding the button manually. It even returns a new `AssayImportPage` for you

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
labkey-danield: In the UI when you edit/insert a missing value field there is a drop down that sets the *_mvindicator field. If you don't set it the value in the field is not considered missing.

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
labkey-chrisj: Correct.  I've expanded the comment; this documents the behavior of saveBatch() with regard to the differences between the JavaAPI and the JSAPI.
There's more difference than I'd like to see between the two, but so far that appears to be a 'feature'

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
labkey-chrisj: pulling it out

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
labkey-chrisj: done

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
labkey-tchad: Gotcha. I misunderstood. I was thinking about creating the field, not setting it.

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
labkey-chrisj: fixed

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
labkey-chrisj: done

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
labkey-chrisj: yes, it would.  And it fails; so new bug.

labkey-chrisj:
Requested changes are made.  Bugs found as a result; good catches all.

src/org/labkey/test/tests/SampleSetRemoteAPITest.java
```diff
@@ -483,39 +487,46 @@ private AssayDesignerPage generateAssay(String subfolderName, String assayName)
...
+        for(TestDataGenerator dataGen : dataGenerators)
+        {
+            new AssayImportPage(getDriver()).setNamedTextAreaValue("TextAreaDataCollector.textArea",
+                    dataGen.writeTsvContents());
+            imported++;
+
+            if(imported < limit)
+                clickButton("Save and Import Another Run");
+            else
+                clickButton("Save and Finish");
```
labkey-tchad: `AssayImportPage` has methods for clicking these save buttons

labkey-chrisj: refactored to use the page methods

PR #17, opened 2019-06-20
Add support in setFormElement for HTML5Number element.

labkey-danield:
Added a handler for an HTML5 Number Field.

### DIFF
--- a/src/org/labkey/test/WebDriverWrapper.java
+++ b/src/org/labkey/test/WebDriverWrapper.java
@@ -3208,8 +3208,6 @@ public abstract class WebDriverWrapper implements WrapsDriver
             Actions actions = new Actions(getDriver());
             actions.doubleClick(el).perform();
 
-            log("WARNING: Clear did not remove the text, going to clear the field manually.");
-
             // Not really sure where the cursor is in relation to the text so LEFT_ARROW until we are at the start.
             for(int i = 0; i < length; i++)
             {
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/WebDriverWrapper.java
```diff
@@ -3189,6 +3192,41 @@ private void setHtml5DateInput(WebElement el, Date date)
...
+        if(!valueAsNumber.equals("nan"))
+        {
+
+            int length = valueAsNumber.length();
+
+            // Double click to edit the cell.
+            Actions actions = new Actions(getDriver());
+            actions.doubleClick(el).perform();
+
+            log("WARNING: Clear did not remove the text, going to clear the field manually.");
```
labkey-tchad: There's no need to warn about this. We're just dealing with expected behavior. A warning implies that there is some better method that would work better.

labkey-tchad:
Let's merge this independently of the sample management tests. This is really just bolstering the base functionality of our test API and doesn't need to be tied to those tests in perpetuity.

src/org/labkey/test/WebDriverWrapper.java
labkey-danield: Fixed.

labkey-danield:
Changed the title so this change can be merged with out the other SampleManagement work.

PR #18, opened 2019-06-22
Test changes for workbook delete

bbimber:
Test changes associated with making workbooks use standard container delete pathway

### DIFF
--- a/src/org/labkey/test/tests/FolderTest.java
+++ b/src/org/labkey/test/tests/FolderTest.java
@@ -19,10 +19,18 @@ import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
+import org.labkey.remoteapi.CommandException;
+import org.labkey.remoteapi.Connection;
+import org.labkey.remoteapi.query.DeleteRowsCommand;
+import org.labkey.remoteapi.query.InsertRowsCommand;
+import org.labkey.remoteapi.query.SaveRowsResponse;
+import org.labkey.remoteapi.query.SelectRowsCommand;
+import org.labkey.remoteapi.query.SelectRowsResponse;
 import org.labkey.test.BaseWebDriverTest;
 import org.labkey.test.Locator;
 import org.labkey.test.TestFileUtils;
 import org.labkey.test.TestTimeoutException;
+import org.labkey.test.WebTestHelper;
 import org.labkey.test.categories.DailyB;
 import org.labkey.test.categories.Hosting;
 import org.labkey.test.components.ext4.Window;
@@ -32,15 +40,24 @@ import org.labkey.test.pages.admin.ReorderFoldersPage;
 import org.labkey.test.pages.list.BeginPage;
 import org.labkey.test.util.DataRegionTable;
 import org.labkey.test.util.Ext4Helper;
+import org.labkey.test.util.ListHelper;
 import org.labkey.test.util.LogMethod;
 import org.labkey.test.util.LoggedParam;
+import org.labkey.test.util.PasswordUtil;
+import org.labkey.test.util.PortalHelper;
+import org.labkey.test.util.WorkbookHelper;
 import org.openqa.selenium.WebElement;
 
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 
 @Category({DailyB.class, Hosting.class})
 @BaseWebDriverTest.ClassTimeout(minutes = 25 )
@@ -111,7 +128,7 @@ public class FolderTest extends BaseWebDriverTest
     @Before
     public void preTest()
     {
-        clickProject(getProjectName());
+        goToProjectHome();
         goToFolderManagement();
     }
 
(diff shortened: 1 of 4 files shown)


bbimber:
@labkey-jeckels this is that patch we discussed

labkey-jeckels:
@labkey-sweta Can you review this PR and comment?

Would also be great if you can do manual testing against the PR on platform repo:

https://github.com/LabKey/platform/pull/201

Scenario is enabling workbook deletion with a preview of their contents, like we do for single folder deletion. Suggested scenarios:

1. Delete project.
2. Delete subfolder.
3. Delete workbooks that are direct children of a project.
4. Delete workbooks that are children of a subfolder.

@bbimber any other suggestions for important permutations to validate?

bbimber:
For situations 1 and 2, you should consider:

1a) delete project without children
1b) delete project with children
2a/b) same idea, just with a folder

if we want to catch scenarios that are not allowable:
- delete workbooks that are not direct children of the current container
- containers can be deleted by providing their IDs on the url using 'targets'.  try to delete any kind of container that supports children using this method.  it should be rejected.

There's also one case I'm not sure quite how to best handle:

- Delete workbooks.  This implies the current container is not deleted.
- on success, it navigates to a new page
- hit back on the browser.  it gives an exception, but kind of an ugly box of text.  I presume this is b/c the lack of these containers existing is caught in validate().  Is this the best way to handle this type of situation?  should some other code catch that exception and do something friendlier?

bbimber:
Note: the automated test in this PR does some of that.

src/org/labkey/test/tests/FolderTest.java
```diff
@@ -255,6 +272,119 @@ public void folderTemplateTest()
...
+        row1.put("Col1", 1);
+
+        ir.addRow(row1);
+
+        Connection cn = new Connection(WebTestHelper.getBaseURL(), PasswordUtil.getUsername(), PasswordUtil.getPassword());
+        ir.execute(cn, getProjectName() + "/" + subfolder);
+    }
+
+    @Test
+    public void deleteFoldersTest() throws Exception
```
labkey-tchad: Split this into multiple `@Test` methods. Delete without children, with children, workbooks via UI, and workbooks via API.
The different scenarios don't share any setup so it wouldn't add much if any overhead.

src/org/labkey/test/tests/FolderTest.java
```diff
@@ -255,6 +272,119 @@ public void folderTemplateTest()
...
+            DeleteRowsCommand drc = new DeleteRowsCommand("core", "workbooks");
+            Map<String, Object> row1 = new HashMap<>();
+            row1.put("EntityId", entityId);
+            drc.addRow(row1);
+
+            SaveRowsResponse resp = drc.execute(new Connection(WebTestHelper.getBaseURL(), PasswordUtil.getUsername(), PasswordUtil.getPassword()), getContainerId());
+
+            throw new Exception("That command should have failed");
+        }
+        catch (IOException | CommandException e)
```
labkey-tchad: Don't bother catching `IOException`. If that gets thrown then something more serious has gone wrong.
```suggestion
        catch (CommandException e)
```

labkey-tchad:
Running some tests on TeamCity: https://teamcity.labkey.org/viewLog.html?buildId=772678
Also kicked off the external modules suites because LabModulesTest uses workbooks pretty extensively.

bbimber:
@labkey-tchad thanks for thinking about LabModules; however, there is a one-line change in LaboratoryWorkbookTable (yet another repo involved) that goes with this and it's possible features are broken without that.  That change impacts our modules, not core code, so i was hoping to deal with the core PRs and then I'd just check in my associated change.

src/org/labkey/test/tests/FolderTest.java
bbimber: ok, good point.  the last two do share some setup (workbook via UI and workbook via API).  do you feel strongly that these two must be separated?

src/org/labkey/test/tests/FolderTest.java
bbimber: ok

src/org/labkey/test/tests/FolderTest.java
labkey-tchad: I do not. Three methods seems like a reasonable split.

bbimber:
these changes should be in

bbimber:
@labkey-tchad WorkbookTest should not be working as well.

src/org/labkey/test/util/WorkbookHelper.java
```diff
@@ -75,6 +78,15 @@ public int getWorkbookIdFromUrl(URL url)
         }
     }
 
+    public void deleteWorkbooksFromDataRegion(DataRegionTable dr)
+    {
+        dr.clickHeaderButton("Delete");
+        assertTrue(_test.acceptAlert().contains("Are you sure you want to delete the selected rows?"));
```
labkey-tchad: Needs to support deleting a single workbook. Should also use existing method to log actual alert.
```suggestion
        assertAlertContains("Are you sure you want to delete the selected row");
```

bbimber: @labkey-tchad sorry, i fixed this yesterday but apparently didnt push the commit.  we actually need the plural conditional based on the number of checked rows.

src/org/labkey/test/util/WorkbookHelper.java
```diff
@@ -75,6 +78,16 @@ public int getWorkbookIdFromUrl(URL url)
         }
     }
 
+    public void deleteWorkbooksFromDataRegion(DataRegionTable dr)
+    {
+        int checked = dr.getCheckedCount();
+        dr.clickHeaderButton("Delete");
+        assertTrue(_test.acceptAlert().contains("Are you sure you want to delete the selected row" + (checked == 1 ? "" : "s") + "?"));
```
labkey-tchad: Use `assertAlert` so that the failure message will include the actual alert text.
```suggestion
        assertAlert("Are you sure you want to delete the selected row" + (checked == 1 ? "" : "s") + "?");
```

bbimber: ok, done

bbimber:
@labkey-jeckels @labkey-tchad it seems like this was approved, but never merged, or am i misreading something?  thanks.

labkey-jeckels:
@bbimber nope, that's correct. Now merged.

PR #19, opened 2019-06-24
Test for missing mandatory field error thrown during domain update via save domain api.

labkey-bpatel:


src/org/labkey/test/tests/issues/IssueDomainTest.java
```diff
@@ -0,0 +1,126 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+@Category({Issues.class, DailyA.class, Data.class})
+@BaseWebDriverTest.ClassTimeout(minutes = 20)
```
labkey-jeckels: This looks like it should be a lot faster than 20 minutes. Suggest removing this annotation and letting it use the default.

src/org/labkey/test/tests/issues/IssueDomainTest.java
```diff
@@ -0,0 +1,126 @@
...
+    @Test
+    public void mandatoryFieldTest() throws Exception
+    {
+        log("Remove mandatory field 'NotifyList'");
+        GetDomainCommand getCmd = new GetDomainCommand(DOMAIN_NAME, ISSUES_NAME);
+        DomainResponse getDomainResponse = getCmd.execute(this.createDefaultConnection(false), getContainerPath());
+        List<Map<String, Object>> getDomainCols = getDomainResponse.getColumns();
+
+        int index = 0;
+        for (Map<String, Object> col : getDomainCols)
```
labkey-jeckels: Using an Iterator and calling its .remove() method would be a little cleaner

labkey-bpatel: Good point! done!

src/org/labkey/test/tests/issues/IssueDomainTest.java
labkey-bpatel: that was a copy paste, now changed to 2 mins. On my local machine total time is 1m 46 secs (which includes build time and test time), so 2 mins should be good.

PR #22, opened 2019-06-27
Wait for flow analysis to auto-navigate

labkey-tchad:
Fixes an intermittent TeamCity failure


labkey-tchad:
Passed on TeamCity
https://teamcity.labkey.org/viewLog.html?buildId=771436
https://teamcity.labkey.org/viewLog.html?buildId=771438

PR #36, opened 2019-07-08
Story 5846:  Prevent deletion of samples that are inputs to assay or derivation runs

labkey-susanh:
Currently there seem to be some timing issues for these tests, so I would appreciate input on where I'm being too impatient.

### DIFF
--- a/src/org/labkey/test/util/SampleSetHelper.java
+++ b/src/org/labkey/test/util/SampleSetHelper.java
@@ -317,7 +317,7 @@ public class SampleSetHelper extends WebDriverWrapper
     {
         getSamplesDataRegionTable().clickHeaderButton("Delete");
         Window.Window(getDriver()).withTitle(expectedTitle).waitFor()
-                .clickButton("Yes, Delete", true);
+                .clickButton("Yes, Delete", false);
         Window.Window(getDriver()).withTitleContaining("Delete sample").waitFor();
         _ext4Helper.waitForMaskToDisappear();
         return this;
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/util/SampleSetHelper.java
```diff
@@ -312,6 +313,16 @@ public void bulkImport(List<Map<String, String>> data, String importOption, @Nul
         }
     }
 
+    public SampleSetHelper deleteSamples(String expectedTitle)
+    {
+        getSamplesDataRegionTable().clickHeaderButton("Delete");
+        Window.Window(getDriver()).withTitle(expectedTitle).waitFor()
+                .clickButton("Yes, Delete", true);
```
labkey-danield: I was able to reproduce locally the failure for SampleSetTest.testDeleteSamplesSomeWithDerivedSamples that is happening on TeamCity. I was able to get it to pass if I changed the last parameter of the clickButton call to false. 
What is happening is that this clickButton is waiting until the window closes before it returns. That's causing the next line (withTitleContaining("Delete sample").waitFor() to fail the expected condition. 

Maybe you don't need the these two lines:
Window.Window(getDriver()).withTitleContaining("Delete sample").waitFor();
 _ext4Helper.waitForMaskToDisappear();

It looks like this is what is also causing the SampleSetTest.testDeleteSamplesSomeWithAssayData test to fail as well.

labkey-susanh: Thanks, Dan.  changing the "true" to "false" did the trick.

PR #44, opened 2019-07-15
framework and test updates to enable automation to run against hosted…

labkey-chrisj:
… vms and remote servers

### DIFF
--- a/src/org/labkey/test/TestProperties.java
+++ b/src/org/labkey/test/TestProperties.java
@@ -86,12 +86,12 @@ public abstract class TestProperties
 
     public static boolean isServerRemote()
     {
-        return "true".equals(System.getProperty("labkey.isRemote", "true"));
+        return "true".equals(System.getProperty("labkey.isRemote", "false"));
     }
 
     public static boolean isIgnoreMissingModules()
     {
-        return "true".equals(System.getProperty("webtest.ignoreMissingModules", "true"));
+        return "true".equals(System.getProperty("webtest.ignoreMissingModules", "false"));
     }
 
     public static boolean isLeakCheckSkipped()
--- a/src/org/labkey/test/TestProperties.java
+++ b/src/org/labkey/test/TestProperties.java
@@ -84,6 +84,16 @@ public abstract class TestProperties
         return !undefinedBuildTypeProperty.equals(System.getProperty(buildTypeProperty, undefinedBuildTypeProperty));
     }
 
+    public static boolean isServerRemote()
+    {
+        return "true".equals(System.getProperty("labkey.isRemote", "true"));
+    }
+
+    public static boolean isIgnoreMissingModules()
+    {
+        return "true".equals(System.getProperty("webtest.ignoreMissingModules", "true"));
+    }
+
     public static boolean isLeakCheckSkipped()
     {
         return "false".equals(System.getProperty("memCheck", "true"));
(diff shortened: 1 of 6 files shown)


labkey-chrisj:
Just a matter of capturing this- to run a labkey vm you'll need to be aware of the following test.properties settings:
labkey.port=8000
labkey.contextpath=/
labkey.isRemote=true  <-- this means the instance is running in a vm, or explicitly not the local machine
labkey.devMode=false
webtest.ignoreMissingModules  <-- use this to signal tests that it is okay to ignore missing dumbster or other module that might not be part of the AMI being tested

src/org/labkey/test/TestProperties.java
```diff
@@ -84,6 +84,16 @@ public static boolean isTestRunningOnTeamCity()
...
     }
 
+    public static boolean isServerRemote()
+    {
+        return "true".equals(System.getProperty("labkey.isRemote", "true"));
+    }
+
+    public static boolean isIgnoreMissingModules()
+    {
+        return "true".equals(System.getProperty("webtest.ignoreMissingModules", "true"));
```
labkey-danield: I would not return true by default for these two system properties labkey.isRemote & webtest.ignoreMissingModules.

You will impact individuals local runs in ways they may not expect if they do not have these properties set.

src/org/labkey/test/tests/ChartingAPITest.java
```diff
@@ -91,6 +93,14 @@ public static void doSetup() throws Exception
     public void initProject()
     {
         _containerHelper.createProject(getProjectName(), null);
+        if (TestProperties.isIgnoreMissingModules())
```
labkey-danield: Doesn't this need to be if (!TestProperties.isIgnoreMissingModules())?
As it is if TestProperties.isIgnoreMissingModules is true the test will check if the module is there and then throw if it is not, should it check for the module's presence if isIgnoreMissingModules is false?

I'm not familiar with this test but given the class name I'm not sure why this test would want to continue if chartingapi module is missing, so I would expect it to always check for this module regardless of the property setting.

src/org/labkey/test/util/AbstractContainerHelper.java
```diff
@@ -197,6 +199,11 @@ public void enableModules(List<String> moduleNames)
             }
             catch (NoSuchElementException missingModule)
             {
+                if (TestProperties.isIgnoreMissingModules())
```
labkey-danield: Like the previous comment I think this is backwards. Shouldn't the code throw (if the module is not there) only if you don't want to ignore missing modules?

src/org/labkey/test/tests/ChartingAPITest.java
labkey-chrisj: No, we want this check to happen if we are ignoring missingModules- the objective is to throw assumptionviolated if the module isn't there (which has the effect of muting the test).

or we could just have it look like this: 
 Set<String> installedModules = _containerHelper.getActiveModules(getProjectName());
            if (TestProperties.isIgnoreMissingModules() && !installedModules.contains("chartingapi") )
            {
                throw new AssumptionViolatedException("Module not installed [chartingapi].  Skipping test.");
            }

src/org/labkey/test/util/AbstractContainerHelper.java
labkey-chrisj: Nah, AssumptionViolatedException is kind of a special case when it comes to throwing it in the test.  The intent is to mute the test (not signal failure) if we are ignoring missing modules.

labkey-chrisj: The test will either fail (because we don't tell it we don't care about missing modules) or be ignored because we told it we don't care that the module is there.
The 'ignore but only when we say so' path is intended.

src/org/labkey/test/TestProperties.java
labkey-chrisj: okay.  I will update these to default to 'false'.

PR #47, opened 2019-07-16
Sample Set Parent Column Automation.

labkey-danield:
Added automation for parent column alias feature. Moved all Sample Management test to DailyC.

### DIFF
--- a/src/org/labkey/test/tests/SampleSetParentColumn.java
+++ b/src/org/labkey/test/tests/SampleSetParentColumn.java
@@ -1,5 +1,6 @@
 package org.labkey.test.tests;
 
+import org.jetbrains.annotations.Nullable;
 import org.junit.Assert;
 import org.junit.BeforeClass;
 import org.junit.Test;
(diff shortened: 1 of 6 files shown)


src/org/labkey/test/tests/SampleSetParentColumn.java
```diff
@@ -0,0 +1,619 @@
...
+        checkAllRowsInDataRegion("parentMaterials", "Name", List.of("SE_04", "S_11"));
+        checkAllRowsInDataRegion("parentMaterials", "Run", List.of(" ", " "));
+        checkAllRowsInDataRegion("parentMaterials", "Sample Set", List.of(SAMPLE_SET_NAME, PARENT_CONTAINER_SAMPLE_SET_NAME));
+
+        // Not sure how reliable this test will be, will the order always be "S_11, SE_04"?
+        checkAllRowsInDataRegion("Runs", "Name", List.of("Derive sample from S_11, SE_04"));
+
+    }
+
+    @Test
```
labkey-ians: /**
Check that the alias column does not conflict with parent mappings (Issue 37946)
**/

labkey-danield: Added this comment.

labkey-danield:
Added tests for Data Classes as inputs.

PR #69, opened 2019-08-14
Include new `devtools` module in test distributions

labkey-tchad:
Depends on https://github.com/LabKey/platform/pull/318


labkey-tchad:
Verifying that new module is included in test distributions as expected: https://teamcity.labkey.org/viewQueued.html?itemId=797144

PR #72, opened 2019-08-19
6160: HTML JSP Warnings

RosalineP:


modules/dumbster/src/org/labkey/dumbster/view/mailWebPart.jsp
```diff
@@ -193,7 +193,7 @@ function toggleRecorder(checkbox)
             <td><a onclick="toggleBody('email_body_<%=rowIndex%>'); return false;"><%=h(m.getHeaderValue("Subject"))%></a>
                 <div id="email_body_<%=rowIndex%>" style="display: none;"><br><%=HtmlString.unsafe(body.toString())%></div></td>
             <td><a onclick="toggleBody('email_headers_<%=rowIndex%>'); return false;">View headers</a>
-                <div id="email_headers_<%=rowIndex%>" style="display: none;"><br><%=headers%></div></td>
+                <div id="email_headers_<%=rowIndex%>" style="display: none;"><br><%=HtmlString.unsafe(headers.toString())%></div></td>
```
labkey-adam: I think you can just call unsafe() as a shortcut -- it's on AbstractJspBase, and delegates to HtmlString.unsafe()

RosalineP: Ah, ok -- I've changed it to unsafe(headers.toString())

PR #73, opened 2019-08-19
Item 5902: Sample Set CRUD - automated testing

labkey-tchad:
Related pulls:
https://github.com/LabKey/sampleManagement/pull/54
https://github.com/LabKey/premium/pull/116


src/org/labkey/test/LabKeySiteWrapper.java
```diff
@@ -1265,6 +1265,11 @@ public void stopImpersonating()
      */
     public void stopImpersonating(Boolean goHome)
     {
+        if (!onLabKeyClassicPage()) // Single-page apps don't have impersonation capabilities
+        {
+            goToHome();
+            goHome = false;
+        }
```
labkey-danield: I'd argue against making this change for a couple of reasons:

Presumably since the developer had to navigate to home, and leave the single page app, to start impersonating that is the first hint that they would have to go back to stop impersonating.

This change no longer honors the value of goHome that has been passed in which makes the function misleading. If it is set to false the developer gets unexpected behavior (they were navigated to home even if they said not to).

If you have to leave this in I would suggest that it calls stopImpersonatingHTTP() if not in LabKey classic.

Or alternately could you get the current URL before going home and then navigate back? You could do the same with impersonate which would make the (navigation) behavior of impersonate and stopImpersonating the same in the single page app.

src/org/labkey/test/WebDriverWrapper.java
```diff
@@ -2286,19 +2292,23 @@ public boolean isElementVisible(Locator loc)
         return loc.findElement(getDriver()).isDisplayed();
     }
 
-    public void assertElementPresent(Locator loc)
+    public WebElement assertElementPresent(Locator loc)
     {
-        assertTrue("Element is not present: " + loc.getLoggableDescription(), isElementPresent(loc));
+        Optional<WebElement> optionalElement = loc.findOptionalElement(getDriver());
+        assertTrue("Element is not present: " + loc.getLoggableDescription(), optionalElement.isPresent());
+        return optionalElement.orElse(null);
```
labkey-danield: I think this is a good change. I'm wondering if this begins to deviate from the function name. I'm also wondering about the use of the optionalElement, if the assert passes shouldn't we just return the web element, no need for the .orElse(null).

src/org/labkey/test/components/html/BaseBootstrapMenu.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+
+    public void collapse()
+    {
+        if (isExpanded())
+            elementCache().toggleAnchor.click();
+        WebDriverWrapper.waitFor(()-> !isExpanded(), "Menu did not collapse as expected", WebDriverWrapper.WAIT_FOR_JAVASCRIPT);
+    }
+
```
labkey-danield: Just wondering if isExpanded, expand and collapse should be protected. If the goal of the components is to obfuscate menu specific actions from the test the tests shouldn't need to expand/collapse the menu, only those classes that extend this class would do to that.

src/org/labkey/test/pages/experiment/CreateSampleSetPage.java
```diff
@@ -0,0 +1,113 @@
...
+        int index = importAliasInputs.size() - 1;
+        WebElement aliasInput = importAliasInputs.get(index);
+        WebElement aliasSelect = importAliasSelects.get(index);
+
+        setFormElement(aliasInput, importHeader);
+        selectOptionByTextContaining(aliasSelect, materialInputName);
+
+        return this;
+    }
+
```
labkey-danield: Should add a removeParentColumnAlias.

src/org/labkey/test/pages/experiment/CreateSampleSetPage.java
```diff
@@ -0,0 +1,113 @@
...
+    protected class ElementCache extends LabKeyPage.ElementCache
+    {
+        Input nameInput = Input.Input(Locator.name("name"), getDriver())
+                .timeout(WAIT_FOR_JAVASCRIPT).findWhenNeeded(this);
+        Input nameExpressionInput = Input.Input(Locator.name("nameExpression"), getDriver())
+                .timeout(WAIT_FOR_JAVASCRIPT).findWhenNeeded(this);
+        Input descriptionInput = Input.Input(Locator.name("description"), getDriver())
+                .timeout(WAIT_FOR_JAVASCRIPT).findWhenNeeded(this);
+
```
labkey-danield: Should there be an element, or a locator, for the <div class="labkey-error"> tag? And then by extension a method that gets the error message, maybe after clicking "Create".

src/org/labkey/test/pages/experiment/InsertDataClassPage.java
```diff
@@ -19,9 +19,11 @@
 import org.labkey.test.WebDriverWrapper;
```
labkey-danield: Should this class/file really be named CreateDataClassPage? It would make the naming more consistent with the CreateSampleSetPage.

src/org/labkey/test/pages/experiment/InsertDataClassPage.java
```diff
@@ -81,16 +89,16 @@ protected ElementCache newElementCache()
...
+                .timeout(WAIT_FOR_JAVASCRIPT).findWhenNeeded(this);
+        Input nameExpressionInput = Input.Input(Locator.tagWithName("input","nameExpression"), getDriver())
+                .timeout(WAIT_FOR_JAVASCRIPT).findWhenNeeded(this);
+        Select materialSourceSelect = SelectWrapper.Select(Locator.tagWithName("select","materialSourceId"))
+                .timeout(WAIT_FOR_JAVASCRIPT).findWhenNeeded(this);
 
-        WebElement createBtn = Locator.lkButton("Create").findWhenNeeded(getDriver()).withTimeout(WAIT_FOR_JAVASCRIPT);;
+        WebElement createBtn = Locator.lkButton("Create").findWhenNeeded(getDriver()).withTimeout(WAIT_FOR_JAVASCRIPT);
         WebElement cancelBtn = Locator.lkButton("Cancel").findWhenNeeded(getDriver()).withTimeout(WAIT_FOR_JAVASCRIPT);
     }
```
labkey-danield: Similar comment as for the SampleSet page, should there be an element/locator for the error message?

src/org/labkey/test/params/experiment/SampleSetDefinition.java
```diff
@@ -0,0 +1,94 @@
+package org.labkey.test.params.experiment;
```
labkey-danield: Completely unrelated to this code...
Should we add a read-me to the params folder to help clarify what is appropriate to add here?

src/org/labkey/test/LabKeySiteWrapper.java
labkey-tchad: That's reasonable. I made this change so that the post-test validation and cleanup didn't crap out if a test left off impersonating in a single-page app. I'll move this to the `stopImpersonation` line of `ensureSignedInAsPrimaryTestUser`, which is explicitly intended to be a blunt instrument.

src/org/labkey/test/WebDriverWrapper.java
labkey-tchad: I thought IntelliJ would compain if I just called `optionalElement.get()` but its actually smart enough to know that `assertTrue(optionalElement.isPresent())` makes that a safe call.
Fixed.

src/org/labkey/test/components/html/BaseBootstrapMenu.java
labkey-tchad: `PagingWidget` was calling `isExpanded` but it doesn't really need to. Changing method to `protected`.

src/org/labkey/test/pages/experiment/CreateSampleSetPage.java
labkey-tchad: Yup. I pulled the add code from `SampleSetHelper` but didn't notice the remove method. I'll move the remove method as well.

src/org/labkey/test/pages/experiment/InsertDataClassPage.java
labkey-tchad: You'll have to take that up with `InsertDataClassAction`. I don't know why or when we prefer insert vs create vs import much of the time.

src/org/labkey/test/pages/experiment/CreateSampleSetPage.java
labkey-tchad: This page is largely out of scope. I was just cleaning up existing functionality.

src/org/labkey/test/pages/experiment/InsertDataClassPage.java
labkey-tchad: Possibly, but out of scope for this story.

src/org/labkey/test/params/experiment/SampleSetDefinition.java
labkey-tchad: It should definitely be documented but we can add it to our existing documentation or maybe add a `README` at the repository root

PR #76, opened 2019-08-20
add test coverage for domain designer error validation

labkey-chrisj:
This adds selenium tests for the new domain designer's error presentation

### DIFF
--- a/src/org/labkey/test/BootstrapLocators.java
+++ b/src/org/labkey/test/BootstrapLocators.java
@@ -1,0 +1,9 @@
+package org.labkey.test;
+
+public abstract class BootstrapLocators
+{
+    static public Locator infoAlertLoc = Locator.tagWithClass("div", "alert-info");
+    static public Locator successAlertLoc = Locator.tagWithClass("div", "alert-success");
+    static public Locator dangerAlertLoc = Locator.tagWithClass("div", "alert-danger");
+    static public Locator warningAlertLoc = Locator.tagWithClass("div", "alert-warning");
+}
--- a/src/org/labkey/test/BootstrapLocators.java
+++ b/src/org/labkey/test/BootstrapLocators.java
@@ -2,8 +2,8 @@ package org.labkey.test;
 
 public abstract class BootstrapLocators
 {
-    static public Locator infoAlertLoc = Locator.tagWithClass("div", "alert-info");
-    static public Locator successAlertLoc = Locator.tagWithClass("div", "alert-success");
-    static public Locator dangerAlertLoc = Locator.tagWithClass("div", "alert-danger");
-    static public Locator warningAlertLoc = Locator.tagWithClass("div", "alert-warning");
+    static public Locator infoAlert = Locator.tagWithClass("div", "alert-info");
+    static public Locator successAlert = Locator.tagWithClass("div", "alert-success");
+    static public Locator dangerAlert = Locator.tagWithClass("div", "alert-danger");
+    static public Locator warningAlert = Locator.tagWithClass("div", "alert-warning");
 }
--- a/src/org/labkey/test/components/DomainDesignerPage.java
+++ b/src/org/labkey/test/components/DomainDesignerPage.java
@@ -21,15 +21,30 @@ public class DomainDesignerPage extends LabKeyPage<DomainDesignerPage.ElementCac
         return new DomainDesignerPage(driver.getDriver());
     }
 
-    public DomainDesignerPage clickSaveChanges()
+    public DomainDesignerPage clickSave()
     {
-        elementCache().saveChangesBtn.click();
+        elementCache().saveButton.click();
+        return this;
+    }
+    public WebElement saveButton()
+    {
+        return elementCache().saveButton;
+    }
+
+    public DomainDesignerPage clickSaveAndFinish()
+    {
+        elementCache().saveAndFinishButton.click();
 
         String currentURL = getDriver().getCurrentUrl();
         waitFor(()-> Locator.tagWithClass("div", "alert-success").existsIn(getDriver()) ||
                 getDriver().getCurrentUrl() != currentURL,
                 "domain save did not notify success", WAIT_FOR_JAVASCRIPT);
         return this;
+    }
+
+    public WebElement saveAndFinishButton()
+    {
+        return elementCache().saveAndFinishButton;
     }
 
     public DomainFormPanel fieldProperties()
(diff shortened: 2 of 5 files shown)


labkey-bpatel:
@labkey-chrisj - Overall the test coverage looks solid.
A couple of items that would be great to get covered:
1. Perform drag and drop on an error'd field and see if that error (under Details)  moves with the field (to a new index).
2. The tests are mainly done on new fields, great to add a test on an existing field - ex. update an existing field with a reserved field, a blank, and a duplicate - similar errors should appear as it does for the new fields. Issues domain would be a good one to use here since it actually displays reserved fields unlike list or datasets (so you should be able to see an existing reserved field getting highlighted). Plus it will also test this issue that I just fixed: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=38314

src/org/labkey/test/components/DomainDesignerPage.java
```diff
@@ -21,9 +21,19 @@ public static DomainDesignerPage beginAt(WebDriverWrapper driver, String contain
...
+        return this;
+    }
+    public WebElement saveButton()
+    {
+        return elementCache().saveButton;
+    }
+
+    public DomainDesignerPage clickSaveAndFinish()
+    {
+        elementCache().saveAndFinishButton.click();
```
labkey-tchad: `Save and Finish` usually navigates away. Waiting for a page load would be less prone to weird timing errors:
```suggestion
        clickAndWait(elementCache().saveAndFinishButton);
```

src/org/labkey/test/components/DomainDesignerPage.java
```diff
@@ -72,7 +116,12 @@ DomainFormPanel domainFormPanel(String domainName)
...
         static public Locator.XPathLocator domainDesignerButton(String text)
         {
-            return Locator.tagWithClass("button", "domain-designer-button").withText(text);
+            return Locator.tagWithClass("button", "btn-success-default").withText(text);
         }
+
+        static public Locator infoAlertLoc = Locator.tagWithClass("div", "alert-info");
+        static public Locator successAlertLoc = Locator.tagWithClass("div", "alert-success");
+        static public Locator dangerAlertLoc = Locator.tagWithClass("div", "alert-danger");
+        static public Locator warningAlertLoc = Locator.tagWithClass("div", "alert-warning");
```
labkey-tchad: I've noticed these CSS classes in Sample Management as well. We should probably put these Locators somewhere shared so we don't keep redefining them. I'm thinking a new class, similar to the existing `Locators` class: `org.labkey.test.util.BootstrapLocators`

labkey-chrisj:
cool.  I will add coverage for those cases.

src/org/labkey/test/components/DomainFieldRow.java
```diff
@@ -346,7 +373,7 @@ protected ElementCache newElementCache()
 
     public static class DomainFieldRowFinder extends WebDriverComponentFinder<DomainFieldRow, DomainFieldRowFinder>
     {
-        private final Locator.XPathLocator _baseLocator = Locator.tagWithClass("div", "domain-field-row");
+        private final Locator.XPathLocator _baseLocator = Locator.tagWithClassContaining("div", "domain-field-row");
```
labkey-tchad: `tagWithClassContaining` shouldn't be necessary here. `tagWithClass` doesn't care about extra spaces or additional classes.

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -188,12 +195,21 @@ public void testDeleteDomainField() throws Exception
...
         assertEquals("expect only field in the domain to have been deleted", 0, remainingFields.size());
+
+        // double-check to ensure the column has been deleted
+        SelectRowsResponse rowsResponse = dgen.getRowsFromServer(createDefaultConnection(true));
+        // this column should no longer exist
+        List<Map<String, Object>> deleteMe = rowsResponse.getColumnModel().stream().filter(a -> a.get("dataIndex").equals("deleteMe")).collect(Collectors.toList());
+        assertEquals(0, deleteMe.size());
+        // make sure the name field is still there
+        List<Map<String, Object>> name = rowsResponse.getColumnModel().stream().filter(a -> a.get("dataIndex").equals("Name")).collect(Collectors.toList());
+        assertEquals(1, name.size());
```
labkey-tchad: Collect the list of columns so that assertion error messages can be more informative.
```
SelectRowsResponse rowsResponse = dgen.getRowsFromServer(createDefaultConnection(true));
List<String> columnsAfterDelete = rowsResponse.getColumnModel().stream().map(col -> (String) col.get("dataIndex")).collect(Collectors.toList());
Assert.assertThat("Columns after delete", columnsAfterDelete, CoreMatchers.allOf(CoreMatchers.hasItem("Name"), CoreMatchers.not(CoreMatchers.hasItem("deleteMe"))));
```

src/org/labkey/test/components/DomainDesignerPage.java
labkey-chrisj: Okay, I'll take these and put them in a new BootstrapLocators.java class, sibling to Locators.java

src/org/labkey/test/components/DomainFieldRow.java
labkey-chrisj: I did it this way because when there's a warning or error, the class changes to literally domain-field-row-warning or domain-field-row-error- it's not adding a new class, changing the existing one.

src/org/labkey/test/tests/DomainDesignerTest.java
labkey-chrisj: nice

src/org/labkey/test/components/DomainDesignerPage.java
```diff
@@ -21,9 +21,19 @@ public static DomainDesignerPage beginAt(WebDriverWrapper driver, String contain
         return new DomainDesignerPage(driver.getDriver());
     }
 
-    public DomainDesignerPage clickSaveChanges()
+    public DomainDesignerPage clickSave()
     {
-        elementCache().saveChangesBtn.click();
+        elementCache().saveButton.click();
+        return this;
```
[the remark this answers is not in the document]
labkey-chrisj: it now waits for any banner

src/org/labkey/test/components/DomainDesignerPage.java
labkey-chrisj: done

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -507,6 +511,48 @@ public void testAddFieldsWithReservedNames() throws Exception
...
+        assertTrue("expect error containing ["+expectedError+"] but it was ["+error+"]",
+                error.contains(expectedError));
+
+        // double-check to ensure the column has not been altered on the server side
+        SelectRowsResponse rowsResponse = dgen.getRowsFromServer(createDefaultConnection(true));
+        List<String> columnsAfterSaveAttempt = rowsResponse.getColumnModel().stream().map(col -> (String) col.get("dataIndex")).collect(Collectors.toList());
+        Assert.assertThat("Columns after delete", columnsAfterSaveAttempt,
+                CoreMatchers.allOf(CoreMatchers.hasItem("Name"),
+                        CoreMatchers.hasItem("testCol"),
+                        CoreMatchers.hasItem("extraField"),
```
labkey-tchad: There's a varargs variant of this:
```
CoreMatchers.hasItems("Name", "testCol","extraField")
```

src/org/labkey/test/BootstrapLocators.java
```diff
@@ -0,0 +1,9 @@
+package org.labkey.test;
+
+public abstract class BootstrapLocators
+{
+    static public Locator infoAlertLoc = Locator.tagWithClass("div", "alert-info");
+    static public Locator successAlertLoc = Locator.tagWithClass("div", "alert-success");
+    static public Locator dangerAlertLoc = Locator.tagWithClass("div", "alert-danger");
+    static public Locator warningAlertLoc = Locator.tagWithClass("div", "alert-warning");
```
labkey-tchad: I'd say remove the `"Loc"` suffix from these; its kinda redundant with the class name.

src/org/labkey/test/tests/DomainDesignerTest.java
labkey-chrisj: cool, that's much better

src/org/labkey/test/BootstrapLocators.java
labkey-chrisj: done

PR #82, opened 2019-08-28
Fb domain designer advanced settings

labkey-chrisj:
adds test coverage to verify advanced dialog settings behavior in the react-based domain designer


src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -551,6 +551,287 @@ public void verifySavedFieldCannotBeRenamedReservedName() throws Exception
...
+                    TestDataGenerator.simpleFieldDef("name", FieldDefinition.ColumnType.String),
+                    TestDataGenerator.simpleFieldDef("extraField", FieldDefinition.ColumnType.String),
+                    TestDataGenerator.simpleFieldDef("testCol", FieldDefinition.ColumnType.String)));
+    DomainResponse createResponse = dgen.createDomain(createDefaultConnection(true), "SampleSet");
+    List<Map<String, Object>> createdFields = createResponse.getColumns();
+
+    DomainDesignerPage domainDesignerPage = DomainDesignerPage.beginAt(this, getProjectName(), "exp.materials", sampleSet);
+    DomainFormPanel domainFormPanel = domainDesignerPage.fieldProperties(sampleSet);
+    DomainFieldRow defaultViewRow = domainFormPanel.addField("defaultViewField");
+    defaultViewRow.showFieldOnDefaultView(false);
```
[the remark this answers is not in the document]
labkey-chrisj: I'll just explicitly set 'true' on the existing 'extraField' column and assert that it should appear in the visible fields (which it does).  For now, the one we're setting to not appear still does, and that causes the test to fail as expected.

src/org/labkey/test/components/bootstrap/AdvancedSettingsDialog.java
```diff
@@ -0,0 +1,150 @@
+package org.labkey.test.components.bootstrap;
```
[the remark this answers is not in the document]
labkey-chrisj: I've created a package org.labkey.test.components.domain, which now contains AdvancedSettingsDialog and DomainFieldRow

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -791,6 +806,12 @@ public void setFieldAsMeasure() throws Exception
...
         DomainResponse domainResponse = dgen.getDomain(createDefaultConnection(true));
+
+        Map<String, Object> testColumnProperties = getPropertyPerColumn(domainResponse.getColumns(), "measure",
+                "measureField", "extraField");
+        assertEquals("measureField should have dimension marked true", true, testColumnProperties.get("measureField"));
+        assertEquals("extraField shoudl not have dimension marked true", false, testColumnProperties.get("extraField"));
+
         Map<String, Object> measureCol = domainResponse.getColumns().stream()
                 .filter(a->a.get("name").equals("measureField")).findFirst().orElse(null);
         Map<String, Object> extraFieldCol = domainResponse.getColumns().stream()
```
labkey-tchad: Remove these duplicate asserts.

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -758,6 +768,11 @@ public void setFieldAsDimension() throws Exception
...
 
         DomainResponse domainResponse = dgen.getDomain(createDefaultConnection(true));
+        Map<String, Object> testColumnProperties = getPropertyPerColumn(domainResponse.getColumns(), "dimension",
+                "dimensionField", "extraField");
+        assertEquals("dimensionField should have dimension marked true", true, testColumnProperties.get("dimensionField"));
+        assertEquals("extraField shoudl not have dimension marked true", false, testColumnProperties.get("extraField"));
+
         Map<String, Object> dimensionCol = domainResponse.getColumns().stream()
                 .filter(a->a.get("name").equals("dimensionField")).findFirst().orElse(null);
         Map<String, Object> extraFieldCol = domainResponse.getColumns().stream()
```
labkey-tchad: Remove old, duplicate asserts. They have been superseded by the asserts above.

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -674,20 +680,26 @@ public void setPhiLevel() throws Exception
...
         DomainResponse domainResponse = dgen.getDomain(createDefaultConnection(true));
+        Map<String, Object> testColumnProperties = getPropertyPerColumn(domainResponse.getColumns(), "PHI",
+                "notPHI", "limitedPHI", "fullPHI", "restrictedPHI");
+        assertEquals("NotPHI", testColumnProperties.get("notPHI"));
+        assertEquals("Limited", testColumnProperties.get("limitedPHI"));
+        assertEquals("PHI", testColumnProperties.get("fullPHI"));
+        assertEquals("Restricted", testColumnProperties.get("restrictedPHI"));
 
         Map<String, Object> notPHiField = domainResponse.getColumns().stream().filter(a-> a.get("name").equals("notPHI")).findFirst().orElse(null);
         assertEquals("NotPHI", notPHiField.get("PHI"));
```
labkey-tchad: Remove old, duplicate asserts. They have been superseded by the asserts above.

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -824,12 +845,28 @@ public void setFieldAsVariable() throws Exception
...
+        assertEquals("extraField should not have recommendedVariable marked true", false, testColumnProperties.get("extraField"));
+
         Map<String, Object> variableCol = domainResponse.getColumns().stream()
                 .filter(a->a.get("name").equals("variableField")).findFirst().orElse(null);
         Map<String, Object> extraFieldCol = domainResponse.getColumns().stream()
                 .filter(a->a.get("name").equals("extraField")).findFirst().orElse(null);
-        assertEquals("expect column to have measure enabled", true, variableCol.get("recommendedVariable"));
-        assertEquals("expect column not to have measure enabled", false, extraFieldCol.get("recommendedVariable"));
+        assertEquals("expect column to have recommendedVariable enabled", true, variableCol.get("recommendedVariable"));
+        assertEquals("expect column not to have recommendedVariable enabled", false, extraFieldCol.get("recommendedVariable"));
```
labkey-tchad: Remove old, duplicate asserts. They have been superseded by the asserts above.

src/org/labkey/test/tests/DomainDesignerTest.java
labkey-chrisj: Why is that imperative?
I left those in because they get a complete map of the field- which will be a lot better for later debugging than just the derivative/reductive map we created in getPropertyPerColumn.
GPPC does nicely add better logging for the case where the column itself is missing, but it doesn't get us all the state we might want later

src/org/labkey/test/tests/DomainDesignerTest.java
labkey-chrisj: Yes, they check more or less the same thing.  But having them there costs very little and provides a little bit of data for whoever needs to investigate state later if there are failures

src/org/labkey/test/tests/DomainDesignerTest.java
labkey-tchad: If one of these checks would fail, the previous, identical assertions would have already cought them. The opposite goes for if they would pass. If you want to dig through the response in a debugger, you already have the `domainResponse`stashed.

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -547,10 +551,324 @@ public void verifySavedFieldCannotBeRenamedReservedName() throws Exception
...
+
+    public Map<String, Object> getPropertyPerColumn(List<Map<String, Object>> columns, String property, String... expectedColumnNames)
+    {
+        Map<String, Object> propertyPerColumn = new HashMap<>();
+        columns.forEach(col -> propertyPerColumn.put((String) col.get("name"), col.get(property)));
+        assertThat("Didn't find expected columns", propertyPerColumn.keySet(), hasItems(expectedColumnNames));
+        return propertyPerColumn.keySet().stream()
+                .filter(key -> Arrays.asList(expectedColumnNames).contains(key))
+                .collect(Collectors.toMap(Function.identity(), propertyPerColumn::get));
+    }
```
labkey-tchad: I know this method was my suggestion but now that I see it in use, having a single `String` argument followed by a `String` vararg makes for a confusing method call (e.g. `getPropertyPerColumn(domainResponse.getColumns(), "PHI", "notPHI", "limitedPHI", "fullPHI", "restrictedPHI")`). The `expectedColumns` would probably be better as a list.
```
    public Map<String, Object> getPropertyPerColumn(List<Map<String, Object>> columns, String property, Collection<String> expectedColumnNames)
    {
        Map<String, Object> propertyPerColumn = new HashMap<>();
        columns.forEach(col -> propertyPerColumn.put((String) col.get("name"), col.get(property)));
        assertThat("Didn't find expected columns", propertyPerColumn.keySet(), hasItems(expectedColumnNames.toArray()));
        return propertyPerColumn.keySet().stream()
                .filter(key -> expectedColumnNames.contains(key))
                .collect(Collectors.toMap(Function.identity(), propertyPerColumn::get));
    }
```

src/org/labkey/test/tests/DomainDesignerTest.java
labkey-chrisj: They're gone, I surrender

PR #86, opened 2019-09-03
Avoid problem interacting with project menu after exporting

labkey-tchad:


labkey-tchad:
This should avoid that intermittent error when interacting with the project menu immediately after exporting

labkey-tchad:
We've been hitting this project menu error in other places. Need to figure out a more general fix.

PR #87, opened 2019-09-04
Issue 38298: Errors when indexing .7z files in search

labkey-ians:
- Adjust icon counts and adding a check for the default icon.


labkey-ians:
https://github.com/LabKey/platform/pull/391

PR #88, opened 2019-09-06
Add test support for issues attachments

labkey-tchad:


labkey-tchad:
Related Pulls:
https://github.com/LabKey/platform/pull/405
https://github.com/LabKey/biologics/pull/465

src/org/labkey/test/pages/issues/BaseIssuePage.java
```diff
@@ -232,7 +233,7 @@ public String getComment()
 
                 for (String change : changes)
                 {
-                    Pattern pattern = Pattern.compile("(.+)\uc2bb(.*)");
+                    Pattern pattern = Pattern.compile("(.+)\\uc2bb(.*)");
```
labkey-chrisj: is there a way to do this without a regex?

labkey-tchad: Yeah, but that's no fun :)
We could probably just split on that character (`\uc2bb` is double greater-than). Not sure why I went with the regex. Not going to worry about it for now. Nothing even calls this method

PR #90, opened 2019-09-10
Fix test

labkey-ians:
- Update to use new label text for current option

### DIFF
--- a/src/org/labkey/test/tests/SampleSetParentColumnTest.java
+++ b/src/org/labkey/test/tests/SampleSetParentColumnTest.java
@@ -31,6 +31,7 @@ public class SampleSetParentColumnTest extends BaseWebDriverTest
 
     private static final String PROJECT_NAME = "SampleSetParentAliasProject";
     private static final String SUB_FOLDER_NAME = "ParentAliasSubFolder";
+    private static final String CURRENT_SAMPLE_SET_OPTION = "(Current Sample Set)";
 
     protected static final String PARENT_CONTAINER_SAMPLE_SET_NAME = "PrePopulatedSampleSet";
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/SampleSetParentColumnTest.java
```diff
@@ -668,6 +668,7 @@ public void testAliasNameConflictsWithFieldName()
         final String ALIAS_NAME_CONFLICT = "ConflictName";
         final String GOOD_PARENT_NAME = "P8";
         final String SAMPLE_SET_NAME = "SimpleSampleSet08";
+        final String CURRENT_SAMPLE_SET_OPTION = "(Current Sample Set)";
```
labkey-danield: This should be a constant for the class (not just this test).

src/org/labkey/test/tests/SampleSetParentColumnTest.java
```diff
@@ -691,7 +692,7 @@ public void testAliasNameConflictsWithFieldName()
         log("Add a parent alias column to the sample set that conflicts with a given column name.");
         sampleHelper.goToEditSampleSet(SAMPLE_SET_NAME);
 
-        sampleHelper.addParentColumnAlias(Map.of(ALIAS_NAME_CONFLICT, SAMPLE_SET_NAME));
+        sampleHelper.addParentColumnAlias(Map.of(ALIAS_NAME_CONFLICT, CURRENT_SAMPLE_SET_OPTION));
```
labkey-danield: This works but it needs to be done for all the tests that have failed:
testParentInSameSampleSet
testValidAliasNames
testMaterialInputsWorkWithAliases
testUseThenRemoveAnAliasParentColumn
testMultipleParentColumns

PR #91, opened 2019-09-12
Make `testAbsoluteFilePath` actually verify what it claimed

labkey-tchad:
This test was just finding the webdav path somewhere on the page; the absolute path wasn't actually there.


labkey-tchad:
Running on TeamCity:
https://teamcity.labkey.org/viewQueued.html?itemId=817238
https://teamcity.labkey.org/viewQueued.html?itemId=817240

PR #97, opened 2019-09-19
Fb lookup domain designer

labkey-sweta:
Automation test for look up 
Test failure fix for new UI changes


src/org/labkey/test/components/DomainDesignerPage.java
```diff
@@ -26,24 +26,31 @@ public static DomainDesignerPage beginAt(WebDriverWrapper driver, String contain
...
-        String msg = waitForAnyAlert();
-        log("Clicking save.  Waited until alert with message ["+msg+"] appeared");
+
+        if (isAlertVisible())
+        {
+            String msg = waitForAnyAlert();
+            log("Clicking save.  Waited until alert with message [" + msg + "] appeared");
+        }
+        else
+           sleep(500); //TODO: wait for page load default
```
labkey-chrisj: We should definitely find a way to wait for a page load event, or at least wait for the url to change

src/org/labkey/test/components/domain/DomainFieldRow.java
```diff
@@ -131,6 +133,15 @@ public AdvancedSettingsDialog clickAdvancedSettings()
         return new AdvancedSettingsDialog(this, getDriver());
     }
 
+    public DomainFieldRow clickCancelCross()
+    {
+        expand();
+        getWrapper().waitFor(()-> elementCache().cancelCross.isEnabled(),
+                "Cancel cross icon did not become enabled", 1500);
+        elementCache().cancelCross.click();
```
labkey-chrisj: Should this action wait for some state before returning?

src/org/labkey/test/components/domain/DomainFieldRow.java
```diff
@@ -314,6 +326,45 @@ public boolean getDateShift()
         return elementCache().dateShiftBox.get();
     }
 
+    public DomainFieldRow setFromFolder(String formatString)
```
labkey-chrisj: as a matter of naming convention, it would be nice for these methods to align closely to the keys they set on the fields- thus, lookupSchema, lookupQuery, lookupContainer
...so setFromFolder -> setlookupContainer,
setFromSchema -> setLookupSchema,
setFromTargetTable -> setLookupQuery

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -835,6 +835,58 @@ public void setFieldAsVariable() throws Exception
...
+                .expand()
+                .setFromFolder("Current Folder")
+                .setFromSchema("lists")
+                .setFromTargetTable("lookUpList (Integer)")
+                .setDescription("LookUp in same container")
+                .clickCancelCross();
+
+        assertEquals("Incorrect detail message","/DomainDesignerTest Project > lists > lookUpList",lookUpRow.detailsMessage());
+
+        domainDesignerPage.saveButton().click();
```
labkey-chrisj: Where possible, use domainDesignerPage.clickSave (because it also handles the waiting for following stuff like errors or page navigation) rather than accessing the button and clicking it directly- that way you can reduce the risk of painful race conditions

PR #107, opened 2019-09-26
Disable crawler for single-page apps

labkey-tchad:
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=38601

### DIFF
--- a/src/org/labkey/test/util/Crawler.java
+++ b/src/org/labkey/test/util/Crawler.java
@@ -239,6 +239,11 @@ public class Crawler
             new ControllerActionId("harvest", "formatInvoice"),
             new ControllerActionId("targetedms", "downloadDocument"),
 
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
Thanks Trey. This looks good to me, but will defer on actual approval to Dan or Chris to make sure there's consensus on this approach within the test team.

labkey-chrisj:
IIRC we disabled crawler in biologics a while ago, pending our ability to
figure out how it should behave there.

Another thing we had to do (by way of working around the differences
between single-page and regular labkey) was to override the normal behavior
of WebDriverWrapper.beginAt()- because it expects a page load event, and in
SPA apps you only get one if you aren't starting and landing in the app.
Perhaps for our SPA environments (biologics, samplemanager, + future) we'll
want to push that logic upstream into the base implementation?

PR #111, opened 2019-09-30
Issue 38616: ETL with remote source query fails when trying to resolv…

labkey-jeckels:
…e possible missing value indicator


labkey-jeckels:
19.2 versions of the tests are happy. https://teamcity.labkey.org/viewLog.html?buildId=827430&buildTypeId=LabKey_192ReleaseSnapshot_Premium_ModulesSuites_DataIntegrationPostgres&tab=buildResultsDiv

Could be a problem with the fact that I created the wrong branch name (missed the 19.2 prefix) and/or originally opened the PR against develop

PR #112, opened 2019-10-01
item 6150:  Fb prioritized updates domain designer test

labkey-chrisj:
this includes some synchronizing for more robust behavior on newer chromedriver
this adds testing for the prioritized update behavior
- user is prompted to confirm when navigating away with dirty form
- name field expands field row
this does not automate drag-and-drop re-ordering

### DIFF
--- a/src/org/labkey/test/components/domain/UnsavedChangesModalDialog.java
+++ b/src/org/labkey/test/components/domain/UnsavedChangesModalDialog.java
@@ -1,0 +1,40 @@
+package org.labkey.test.components.domain;
+
+import org.labkey.test.components.DomainDesignerPage;
+import org.labkey.test.components.bootstrap.ModalDialog;
+import org.openqa.selenium.WebDriver;
+
+public class UnsavedChangesModalDialog extends ModalDialog
+{
+
+    public UnsavedChangesModalDialog(ModalDialogFinder finder, WebDriver driver)
+    {
+        super(finder.waitFor().getComponentElement(), driver);
+    }
+
+    public DomainDesignerPage discardChanges()  // should land the user on a domain designer page
+    {
+        dismiss("No, Discard Changes");
+        return new DomainDesignerPage(getDriver());
+    }
+
+    public void saveChanges()   // if successful, should navigate away with same effect as clicking 'save' on domain designer page
+    {
+        dismiss("Yes, Save Changes");
+    }
+
+    protected ElementCache newElementCache()
+    {
+        return new ElementCache();
+    }
+
+    protected class ElementCache extends ModalDialog.ElementCache
+    {
+//        final WebElement yesSaveChangesBtn = Locator.css("button").withText("Yes, Save Changes")
+//                .findWhenNeeded(this);
+//
+//        final WebElement noDiscardChangesBtn = Locator.css("button").withText("No, Discard Changes")
+//                .findWhenNeeded(this);
+    }
+
+}
--- a/src/org/labkey/test/components/domain/UnsavedChangesModalDialog.java
+++ b/src/org/labkey/test/components/domain/UnsavedChangesModalDialog.java
@@ -23,18 +23,4 @@ public class UnsavedChangesModalDialog extends ModalDialog
         dismiss("Yes, Save Changes");
     }
 
-    protected ElementCache newElementCache()
-    {
-        return new ElementCache();
-    }
-
-    protected class ElementCache extends ModalDialog.ElementCache
-    {
-//        final WebElement yesSaveChangesBtn = Locator.css("button").withText("Yes, Save Changes")
-//                .findWhenNeeded(this);
-//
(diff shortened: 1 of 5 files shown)


src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -891,19 +893,122 @@ public void testLookUpFieldList() throws IOException, CommandException
...
+
+        DomainDesignerPage domainDesignerPage = DomainDesignerPage.beginAt(this, getProjectName(), "exp.materials", sampleSetName);
+        DomainFormPanel domainFormPanel = domainDesignerPage.fieldProperties(sampleSetName);
 
+        // now attempt to mark 'manufacturer' field 'required'
+        domainFormPanel.getField("manufacturer")
+                .setRequiredField(true);
+        domainDesignerPage.clickSave();     // expect error warning here; this should warn the user
+        String expectedWarning = domainDesignerPage.waitForAnyAlert();
+        assertTrue(expectedWarning.contains("cannot be required when it contains rows with blank values."));
```
labkey-tchad: Do we want a test case that verifies you **can** mark an existing field as required if there aren't any rows with blank values?

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -891,19 +893,122 @@ public void testLookUpFieldList() throws IOException, CommandException
...
+        getDriver().switchTo().alert().dismiss();
+        // ensure current location
+        assertEquals(currentUrl, getDriver().getCurrentUrl());
+
+        domainDesignerPage = new DomainDesignerPage(getDriver()); // we've been refreshing; things are possibly stale
+        domainDesignerPage.clickCancel().saveChanges();
+    }
+
+    @Test
+    public void verifyWarningForMarkingExistingSampleSetFieldRequired() throws Exception
```
labkey-tchad: Is this special sample set behavior? If we're just using sample sets to verify universal behavior, having `SampleSet` in the name is misleading.

src/org/labkey/test/tests/DomainDesignerTest.java
labkey-chrisj: That's not a bad idea.  I'll add it

src/org/labkey/test/tests/DomainDesignerTest.java
labkey-chrisj: I'm not sure; my understanding was that this needed to work specifically on a sampleset.  It's definitely worth following up to see if this is/should be universal/true for every domain kind

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -891,19 +893,122 @@ public void testLookUpFieldList() throws IOException, CommandException
...
+        // now navigate away
+        getDriver().navigate().to(homeUrl);
+        // dismiss the alert
+        shortWait().until(ExpectedConditions.alertIsPresent());
+        getDriver().switchTo().alert().dismiss();
+        // ensure current location
+        assertEquals(currentUrl, getDriver().getCurrentUrl());
+
+        domainDesignerPage = new DomainDesignerPage(getDriver()); // we've been refreshing; things are possibly stale
+        domainDesignerPage.clickCancel().saveChanges();
```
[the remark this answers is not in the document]
labkey-chrisj: good call.

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -891,19 +893,122 @@ public void testLookUpFieldList() throws IOException, CommandException
...
+
+        // now navigate away
+        getDriver().navigate().to(homeUrl);
+        // dismiss the alert
+        shortWait().until(ExpectedConditions.alertIsPresent());
+        getDriver().switchTo().alert().dismiss();
+        // ensure current location
+        assertEquals(currentUrl, getDriver().getCurrentUrl());
+
+        domainDesignerPage = new DomainDesignerPage(getDriver()); // we've been refreshing; things are possibly stale
```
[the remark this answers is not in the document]
labkey-chrisj: fair enough.  tested to see if it was true, and yes... it was

src/org/labkey/test/components/domain/UnsavedChangesModalDialog.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    protected class ElementCache extends ModalDialog.ElementCache
+    {
+//        final WebElement yesSaveChangesBtn = Locator.css("button").withText("Yes, Save Changes")
+//                .findWhenNeeded(this);
+//
+//        final WebElement noDiscardChangesBtn = Locator.css("button").withText("No, Discard Changes")
+//                .findWhenNeeded(this);
+    }
```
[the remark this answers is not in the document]
labkey-chrisj: done

PR #116, opened 2019-10-08
Item 6167: integrate ITN changes into core codebase

labkey-tchad:
@labkey-jeckels platform and trialShare modules are ready for pull requests

### DIFF
--- a/src/org/labkey/test/tests/core/security/GetReadableContainersAPITest.java
+++ b/src/org/labkey/test/tests/core/security/GetReadableContainersAPITest.java
@@ -1,0 +1,273 @@
+package org.labkey.test.tests.core.security;
+
+import org.json.simple.JSONObject;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.labkey.test.BaseWebDriverTest;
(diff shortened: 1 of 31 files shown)


src/org/labkey/test/tests/core/security/GetReadableContainersAPITest.java
```diff
@@ -0,0 +1,273 @@
...
+    @Override
+    protected BrowserType bestBrowser()
+    {
+        return BrowserType.CHROME;
+    }
+
+    @Override
+    protected String getProjectName()
+    {
+        return null;
```
labkey-jeckels: I assume this is never used? What about throwing UnsupportedOperationException?

labkey-tchad: There are places that call it and know to expect a possible `null` (e.g. `BaseWebDriverTest.doCleanup`). I'd be worried about hitting surprise `NPE`s. This has been our standard pattern for tests that don't create a project.

labkey-tchad:
Related changes:
https://github.com/LabKey/trialShare/pull/66
https://github.com/LabKey/platform/pull/505

PR #117, opened 2019-10-08
6373: Script-based filewatcher automated tests

RosalineP:
So I confirmed that these are the checks we want to do:

1. If a pipeline (defined in the xml) contains a 'help' element, we check that the help text is displayed in the upper blue html help text container on the "Create Pipeline Trigger" page.
2. To test for encoding issues, we check that "null" is not present in the help text container if no 'help' xml element is defined.
3. If a pipeline xml file is in the module, its description is shown as an option in the 'Pipeline Task' select options. (This is pre-existing functionality.)
4. If a pipeline specifies that element triggerConfiguration's allow="false", its description is not shown in the 'Pipeline Task' select options. (This is new functionality.)
5. We do _not_ have to check whether the triggers actually work, because that is outside the scope of the dev product changes, and that functionality is already covered by other tests.

### DIFF
--- a/src/org/labkey/test/pages/admin/ImportFolderPage.java
+++ b/src/org/labkey/test/pages/admin/ImportFolderPage.java
@@ -70,6 +70,8 @@ public class ImportFolderPage extends LabKeyPage<ImportFolderPage.ElementCache> 
         clickButton("Import Folder");
     }
 
+    public void clickManageFileWatcherTriggers() { clickAndWait(elementCache().manageFileWatcherTriggersButton); }
+
     @NotNull
     @Override
     public WebDriver getWrappedDriver()
@@ -116,5 +118,6 @@ public class ImportFolderPage extends LabKeyPage<ImportFolderPage.ElementCache> 
 
         WebElement importFolderButton = Locator.lkButton("Import Folder").findWhenNeeded(this);
         WebElement usePipelineButton = Locator.lkButton("Use Pipeline").findWhenNeeded(this);
+        WebElement manageFileWatcherTriggersButton = Locator.lkButton("Manage File Watcher Triggers").findWhenNeeded(this);
     }
 }
--- a/src/org/labkey/test/tests/FileBasedPipelineTest.java
+++ b/src/org/labkey/test/tests/FileBasedPipelineTest.java
@@ -35,6 +35,7 @@ import org.labkey.test.util.PipelineAnalysisHelper;
 import org.labkey.test.util.PortalHelper;
 import org.labkey.test.util.RReportHelper;
 import org.openqa.selenium.NoSuchElementException;
+import org.openqa.selenium.WebElement;
 
 import java.io.File;
 import java.util.Arrays;
@@ -85,16 +86,14 @@ public class FileBasedPipelineTest extends BaseWebDriverTest
         DataRegionTable pipelineTable = new DataRegionTable("query", getDriver());
         pipelineTable.clickInsertNewRow();
 
-        setFormElement(Locator.name("pipelineId"), pipelineTaskWithHelpText);
-        assertTextPresent(helpText);
+        selectOptionByText(Locator.id("pipelineTaskSelect"), pipelineTaskWithHelpText);
+        Locator.byClass("lk-trigger-help-text").containing(helpText);
 
-        assertTextNotPresent(pipelineTaskInvisible);
+        Locator.id("pipelineTaskSelect").notContaining(pipelineTaskInvisible);
 
-        setFormElement(Locator.name("pipelineId"), pipelineTaskNoHelpText);
-        // assert null is not on page within top bubble
+        selectOptionByText(Locator.id("pipelineTaskSelect"), pipelineTaskNoHelpText);
+        Locator.byClass("lk-trigger-help-text").notContaining("null");
     }
-
-
 
     @Test
     public void testRCopyPipeline()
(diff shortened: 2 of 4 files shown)


src/org/labkey/test/tests/FileBasedPipelineTest.java
```diff
@@ -72,6 +73,28 @@ public void startTest()
         goToProjectHome();
     }
 
+    @Test
+    public void testScriptBasedFilewatchers()
```
labkey-tchad: File watchers require the premium module. This test case need to be in the `premium` module or it will cause community test runs to fail.

src/org/labkey/test/tests/FileBasedPipelineTest.java
```diff
@@ -72,6 +73,28 @@ public void startTest()
...
+
+        goToFolderManagement().goToImportTab();
+        clickAndWait(Locator.lkButton("Manage file watcher triggers"));
+        DataRegionTable pipelineTable = new DataRegionTable("query", getDriver());
+        pipelineTable.clickInsertNewRow();
+
+        selectOptionByText(Locator.id("pipelineTaskSelect"), pipelineTaskWithHelpText);
+        Locator.byClass("lk-trigger-help-text").containing(helpText);
+
+        Locator.id("pipelineTaskSelect").notContaining(pipelineTaskInvisible);
```
labkey-tchad: `Locator`s are just glorified `StringBuilder`s. They don't actually do anything until you attempt to use them to find something on the page. (e.g. `assertElementPresent` or `findElement`)

labkey-chrisj: What is your intent with lines 92 and 95?  You can use these locators to .findElement(scope), or to get a boolean with .existsIn(scope)- but these lines appear to declare the locators without either assigning them to a variable or using it to validate.

src/org/labkey/test/tests/FileBasedPipelineTest.java
```diff
@@ -72,6 +73,28 @@ public void startTest()
...
+    @Test
+    public void testScriptBasedFilewatchers()
+    {
+        String pipelineTaskWithHelpText = "Pipeline task with help text";
+        String helpText = "Hello! I am help text.";
+        String pipelineTaskNoHelpText = "two-steps";
+        String pipelineTaskInvisible = "Task that is not evokable by filewatcher";
+
+        goToFolderManagement().goToImportTab();
+        clickAndWait(Locator.lkButton("Manage file watcher triggers"));
```
labkey-chrisj: this (clicking the 'manage file watcher triggers' button) could be an opportunity to scaffold in a page object, and to implement finding and clicking the button in the ImportTabPage class.
If you don't have bandwidth or scope to do that just now, please let me know- I'd like to track this as future work.

src/org/labkey/test/tests/FileBasedPipelineTest.java
RosalineP: Ah ok! I'll move it over.

src/org/labkey/test/tests/FileBasedPipelineTest.java
RosalineP: Oh, I think I was mistaken on what notContaining() did!
I wanted to check that the pipelineTaskInvisible and the 'null' did not exist within those specific elements. Should I be using existsIn() for that check?

labkey-tchad:
I was poking my nose in here for unrelated reasons. This cook is leaving the kitchen to y'all.

src/org/labkey/test/tests/FileBasedPipelineTest.java
RosalineP: Gotcha, updated!

src/org/labkey/test/tests/FileBasedPipelineTest.java
```diff
@@ -72,6 +73,28 @@ public void startTest()
...
+        DataRegionTable pipelineTable = new DataRegionTable("query", getDriver());
+        pipelineTable.clickInsertNewRow();
+
+        selectOptionByText(Locator.id("pipelineTaskSelect"), pipelineTaskWithHelpText);
+        Locator.byClass("lk-trigger-help-text").containing(helpText);
+
+        Locator.id("pipelineTaskSelect").notContaining(pipelineTaskInvisible);
+
+        selectOptionByText(Locator.id("pipelineTaskSelect"), pipelineTaskNoHelpText);
+        Locator.byClass("lk-trigger-help-text").notContaining("null");
```
labkey-tchad: `org.labkey.test.components.pipeline.PipelineTriggerWizard` encapsulates this functionality.

RosalineP: Cool! Updated to use the PipelineTriggerWizard.

src/org/labkey/test/pages/admin/ImportFolderPage.java
```diff
@@ -116,5 +118,6 @@ protected ElementCache newElementCache()
 
         WebElement importFolderButton = Locator.lkButton("Import Folder").findWhenNeeded(this);
         WebElement usePipelineButton = Locator.lkButton("Use Pipeline").findWhenNeeded(this);
+        WebElement manageFileWatcherTriggersButton = Locator.lkButton("Manage File Watcher Triggers").findWhenNeeded(this);
```
labkey-chrisj: just as an FYI, you can chain a timeout setter on finder functions- .findWhenNeeded(context).withTimeout(msec).
without the timeout, it resolves the element immediately when called (and will fail if it is not in the page at that time).  With it, a timer is started when we start to look.

it's irrelevant if that element is always going to be there by the time code gets to it, but it's a nice thing to know about :-)

src/org/labkey/test/pages/admin/ImportFolderPage.java
```diff
@@ -70,6 +70,8 @@ public void clickImportFolder()
         clickButton("Import Folder");
     }
 
+    public void clickManageFileWatcherTriggers() { clickAndWait(elementCache().manageFileWatcherTriggersButton); }
```
labkey-chrisj: When we have a Page class to encapsulate the destination page, this function will ideally return an instance of it

PR #122, opened 2019-10-14
Fb domain validation

labkey-chrisj:
This adds test coverage for field range validators, conditional formats, regular expression validators, lookup validators, and default values

### DIFF
--- a/src/org/labkey/test/components/domain/AdvancedSettingsDialog.java
+++ b/src/org/labkey/test/components/domain/AdvancedSettingsDialog.java
@@ -4,6 +4,7 @@ import org.labkey.test.Locator;
 import org.labkey.test.components.PropertiesEditor;
 import org.labkey.test.components.bootstrap.ModalDialog;
 import org.labkey.test.components.html.Checkbox;
+import org.labkey.test.pages.LabKeyPage;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.ui.ExpectedConditions;
@@ -65,6 +66,25 @@ public class AdvancedSettingsDialog extends ModalDialog
         getWrapper().waitFor(()-> elementCache().showInDetailsView.get().equals(checked),
                 "showInDetailsView checkbox was not set as expected", 1000);
         return this;
+    }
+
+    // default value options
+    public String getDefaultValueType()
+    {
+        return getWrapper().getFormElement(elementCache().defaultTypeSelect);
+    }
+    public AdvancedSettingsDialog setDefaultValueType(String type)
+    {
+        getWrapper().waitFor(()->  elementCache().defaultTypeSelect.isEnabled(),
+                "phiSelect did not become enabled in time", 1500);
+        getWrapper().setFormElement(elementCache().defaultTypeSelect, type);
+        return this;
+    }
+
+    public LabKeyPage clickDefaultValuesLink()
+    {
+        getWrapper().clickAndWait(Locator.linkWithText("Set Default Values"));
+        return new LabKeyPage(getDriver());  // todo: return more strongly-typed page
     }
 
     public String getPHILevel()
@@ -164,6 +184,10 @@ public class AdvancedSettingsDialog extends ModalDialog
                 Locator.input("domainpropertiesrow-shownInInsertView").findWhenNeeded(this));
         public Checkbox showInDetailsView = new Checkbox(
                 Locator.input("domainpropertiesrow-showInDetailsView").findWhenNeeded(this));
+
+        // default value options
+        public WebElement defaultTypeSelect = Locator.tagWithName("select", "domainpropertiesrow-defaultValueType")
+                .findWhenNeeded(this);
 
         // misc options
         public WebElement phiSelect = Locator.tagWithAttribute("select", "name", "domainpropertiesrow-PHI")
--- a/src/org/labkey/test/components/domain/AdvancedSettingsDialog.java
+++ b/src/org/labkey/test/components/domain/AdvancedSettingsDialog.java
@@ -6,6 +6,7 @@ import org.labkey.test.components.bootstrap.ModalDialog;
 import org.labkey.test.components.html.Checkbox;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.ui.ExpectedConditions;
 
 public class AdvancedSettingsDialog extends ModalDialog
 {
@@ -49,6 +50,8 @@ public class AdvancedSettingsDialog extends ModalDialog
     public AdvancedSettingsDialog showOnUpdateView(boolean checked)
     {
         elementCache().showInUpdateView.set(checked);
+        getWrapper().waitFor(()-> elementCache().showInUpdateView.get().equals(checked),
+                "showInUpdateView checkbox was not set as expected", 1000);
         return this;
     }
 
@@ -59,6 +62,8 @@ public class AdvancedSettingsDialog extends ModalDialog
     public AdvancedSettingsDialog showOnDetailsView(boolean checked)
     {
         elementCache().showInDetailsView.set(checked);
+        getWrapper().waitFor(()-> elementCache().showInDetailsView.get().equals(checked),
+                "showInDetailsView checkbox was not set as expected", 1000);
         return this;
     }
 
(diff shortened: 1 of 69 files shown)


src/org/labkey/test/components/domain/RangeValidatorPanel.java
```diff
@@ -0,0 +1,207 @@
...
+        {
+            _name = name;
+            return this;
+        }
+        public RangeValidatorPanelFinder openedByName(String name)          //finds the currently-expanded
+        {
+            _baseLocator = _baseLocator.withDescendant(Locator.input("domainpropertiesrow-name")
+                    .withAttribute("value", name));
+            return this;
+        }
```
labkey-tchad: Why combine the name and open-ness instead of providing a separate method for each?
e.g.:
`public RangeValidatorPanelFinder withName(String name)`
-and-
`public RangeValidatorPanelFinder opened()`

src/org/labkey/test/components/domain/RangeValidatorPanel.java
```diff
@@ -0,0 +1,207 @@
...
+        return this;
+    }
+
+    public RangeValidatorPanel setName(String name)
+    {
+        expand();
+        elementCache().nameInput().setValue(name);
+        return this;
+    }
+
```
labkey-tchad: We should make this work with the existing `FieldDefinition.RangeValidator` class.
e.g.
```
public RangeValidatorPanel addValidator(RangeValidator validator)
```

src/org/labkey/test/components/domain/DomainFieldRow.java
```diff
@@ -594,20 +650,54 @@ protected Locator locator()
...
+        }
+
+        public WebElement addFormatButton()
+        {
+            return Locator.button("Add Format").waitForElement(getComponentElement(), 2000);
+        }
+        public WebElement editFormatButton()
+        {
+            return Locator.button("Edit Formats").waitForElement(getComponentElement(), 2000);
+        }
```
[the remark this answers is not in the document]
labkey-chrisj: okay.  They do pop the same dialog, even if they appear differently (according to whether or not there are any validators already on the field).  I'll unify them, and their accessor methods

src/org/labkey/test/components/domain/RangeValidatorPanel.java
labkey-chrisj: ...it's because the way to find it by name is different when opened or closed.
When open, the name is in the name field edit.  When closed, it is part of the expression summary (which disappears when opened)

src/org/labkey/test/components/domain/AdvancedSettingsDialog.java
```diff
@@ -67,6 +68,25 @@ public AdvancedSettingsDialog showOnDetailsView(boolean checked)
...
+    // default value options
+    public String getDefaultValueType()
+    {
+        return getWrapper().getFormElement(elementCache().defaultTypeSelect);
+    }
+    public AdvancedSettingsDialog setDefaultValueType(String type)
+    {
+        getWrapper().waitFor(()->  elementCache().defaultTypeSelect.isEnabled(),
+                "phiSelect did not become enabled in time", 1500);
+        getWrapper().setFormElement(elementCache().defaultTypeSelect, type);
```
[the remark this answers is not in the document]
labkey-chrisj: I tried it that way, but selectByVisibleText is a function on the Select class, not on WebDriverWrapper.
So, converting these to be selects and using that method on them to set values.

src/org/labkey/test/components/domain/ConditionalFormatPanel.java
```diff
@@ -0,0 +1,188 @@
...
+import org.openqa.selenium.support.ui.Select;
+
+import static org.labkey.test.components.html.Input.Input;
+
+public class ConditionalFormatPanel extends WebDriverComponent<ConditionalFormatPanel.ElementCache>
+{
+    final WebElement _el;
+    final WebDriver _driver;
+
+    public ConditionalFormatPanel(WebElement element, WebDriver driver)
```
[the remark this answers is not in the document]
labkey-chrisj: I like this idea.  Applying it to the other dialog/panel classes

src/org/labkey/test/components/domain/AdvancedSettingsDialog.java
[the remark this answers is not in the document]
labkey-tchad: Sorry. The method in `WebDriverWrapper` is `selectOptionByText`.

src/org/labkey/test/components/domain/DomainFieldRow.java
```diff
@@ -492,6 +500,54 @@ public boolean hasFieldWarning()
...
+    }
+
+    public ConditionalFormatDialog clickEditFormats()
+    {
+        expand();
+        getWrapper().shortWait().until(ExpectedConditions.elementToBeClickable(elementCache().editFormatButton()));
+        elementCache().editFormatButton().click();
+        return new ConditionalFormatDialog(this, getDriver());
+    }
+
```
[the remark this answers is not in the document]
labkey-chrisj: Yes, that will be nice.  I'm implementing this pattern for RegexValidator, which is easy.
RangeValidator runs into complications converting FieldDefinition.RangeType to Filter.Operator and back.

Because I'm using Filter.Operator.toString() to select operators by value, (and its output equals exactly the url params) I'm weighing the ups and downs of using either one

src/org/labkey/test/components/domain/AdvancedSettingsDialog.java
```diff
@@ -67,6 +68,25 @@ public AdvancedSettingsDialog showOnDetailsView(boolean checked)
         return this;
     }
 
+    // default value options
+    public String getDefaultValueType()
+    {
+        return getWrapper().getFormElement(elementCache().defaultTypeSelect);
+    }
+    public AdvancedSettingsDialog setDefaultValueType(String type)
```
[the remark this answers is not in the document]
labkey-chrisj: Yeah.  In the scenario I did, there were two options, but I don't know if there are any other allowable values here.  I'll generate an enum with those values

src/org/labkey/test/components/domain/RangeValidatorPanel.java
labkey-chrisj: I attempted this, but ran into issues converting RangeValidator and Filter.Operator back and forth.  Using Filter.Operator has distinct advantages- (it is pretty complete, its getUrlKey() function matches to select option values, as well as to URL expressions)
Sure, we could make a lot of older tests work more easily in the short term by not using Filter.Operator- but then we'd end up maintaining both enums, hoping they remain parallel, and selecting options by text and not being able to use the urlKey expressions to predict the expressions as they will appear on the domain field JSON

labkey-chrisj: Perhaps the best option here will be to implement setters that take both Filter.Operator and RangeValidator's params, support them both

labkey-chrisj: I think I have a better solution: pipe the operator through the RangeType enum that FieldDefinition.RangeValidator uses- and pass that to the existing setter

src/org/labkey/test/components/domain/DomainFieldRow.java
```diff
@@ -595,6 +695,7 @@ protected Locator locator()
 
         public Select getFromSchemaInput()
         {
+            expand();
             Select select = SelectWrapper.Select(Locator.name("domainpropertiesrow-lookupSchema")).find(this);
             getWrapper().waitFor(()-> select.getOptions().size() > 0 && !select.getOptions().get(0).getText().equals("Loading..."),
```
labkey-tchad: This `expand` call looks redundant; everywhere that calls this method does their own expand first. Same for `getFromTargetTableInput` and `getLookupValidatorEnabledCheckbox` below. Its also inconsistent with all of the other elements in the ElementCache

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -409,7 +416,7 @@ public void testDeleteFieldInListWithData() throws Exception
         domainDesignerPage.clickSave();
 
         // there should just be the key field (id) now:
-        disablePageUnloadEvents();
+//        disablePageUnloadEvents();
```
labkey-tchad: Delete or uncomment

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -70,7 +77,7 @@ private void doSetup()
     @Before
     public void preTest() throws Exception
     {
-        disablePageUnloadEvents();
+//        disablePageUnloadEvents();
```
labkey-tchad: Delete or uncomment

labkey-chrisj:
build is fixed https://teamcity.labkey.org/viewLog.html?buildId=838341&

Note that we will need to hold off merging until the product changes (platform/fb_domain_validation) are also merged

PR #126, opened 2019-10-21
Move @glass test components out of premium and into testAutomation.

labkey-danield:


labkey-danield:
The change is move. Moving the files is part of this story it was an ask from Dev for some time now to remove dependencies on other modules from Sample Manager, so this removes the dependency on premium.

PR #128, opened 2019-10-25
Remove FireFox preference that is causing many tests to fail

labkey-tchad:
Temprarily(?) backing out part of the domain designer test coverage while we figure out a solution to all these test timeouts.

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=38785


labkey-tchad:
Verifying that fix actually works (using `UserPermissionsTest`):
https://teamcity.labkey.org/viewQueued.html?itemId=841132
and a control to make sure failure doesn't require a whole daily suite to trigger:
https://teamcity.labkey.org/viewQueued.html?itemId=841130

labkey-tchad:
Fix confirmed.

PR #132, opened 2019-10-31
Item 6451: Assay and domain designer test fixes for changes to field designer

cnathe:


src/org/labkey/test/components/DomainDesignerPage.java
```diff
@@ -67,22 +65,22 @@ public boolean isAlertVisible()
...
 
-    public WebElement saveButton()
+    public WebElement finishButton()
     {
-        return elementCache().saveButton();
+        return elementCache().finishButton();
     }
 
-    public DomainFormPanel fieldProperties()
+    public DomainFormPanel fieldsPanel()
```
labkey-jeckels: Maybe a comment to say that this just returns the first one if there are multiple?

cnathe:
@labkey-tchad Trey, Chris said at standup that he can take this one.

PR #133, opened 2019-10-31
New helper for executing external processes with timeout

labkey-tchad:
The old method was leading to test timeouts when the process didn't run correctly

### DIFF
--- a/src/org/labkey/test/util/ProcessHelper.java
+++ b/src/org/labkey/test/util/ProcessHelper.java
@@ -136,5 +136,10 @@ public class ProcessHelper
 
             throw new RuntimeException(cause);
         }
+        finally
+        {
+            if (proc.isAlive())
+                proc.destroyForcibly();
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/util/ProcessHelper.java
```diff
@@ -0,0 +1,140 @@
...
+        StringBuffer output = new StringBuffer();
+        runProcess(output, logOutput);
+        return output.toString();
+    }
+
+    /**
+     * Adapted from `org.labkey.api.reports.ExternalScriptEngine#runProcess()`
+     * Execute the external process in separate thread
+     * @return the exit code for the invocation - 0 if the process completed successfully.
+     */
```
labkey-danield: This looks ok to me (one question below) but I'm depending upon the comment above that this is adapted from org.labkey.api. A well respected and trusted source of high quality software!

src/org/labkey/test/util/ProcessHelper.java
```diff
@@ -0,0 +1,140 @@
...
+        }
+        catch (ExecutionException ex)
+        {
+            // Exception thrown in output collecting thread
+            Throwable cause = ex.getCause();
+            if (cause instanceof IOException)
+                throw new IOException("Failed writing output for process in '" + pb.directory().getPath() + "'.", cause);
+
+            throw new RuntimeException(cause);
+        }
```
labkey-danield: Do you need to try to terminate/destroy proc in these catch blocks?

PR #134, opened 2019-11-01
Issue 38843: Domain designer finish button helper for expected error case

cnathe:


cnathe:
@labkey-chrisj Chris, TeamCity was hitting what looked to be some intermittent errors because of the domain save/finish click not always knowing if it should be waiting for an error message or a page navigation. With this PR, I added an explicit helper for the "I know I should wait for an error message" case. Let me know what you think of this approach.

cnathe:
https://teamcity.labkey.org/project.html?projectId=LabKey_193ReleaseSnapshot&branch_LabKey_193ReleaseSnapshot_Community=domainDesignerTestFixes

PR #140, opened 2019-11-04
stablilize tests and domain designer pages

labkey-chrisj:
These changes add re-tries to get alerts when the element in question is stale (hopefully this will reduce the noise)

### DIFF
--- a/src/org/labkey/test/components/DomainDesignerPage.java
+++ b/src/org/labkey/test/components/DomainDesignerPage.java
@@ -9,6 +9,7 @@ import org.labkey.test.components.domain.DomainFormPanel;
 import org.labkey.test.components.domain.UnsavedChangesModalDialog;
 import org.labkey.test.pages.LabKeyPage;
 import org.labkey.test.util.Maps;
+import org.openqa.selenium.StaleElementReferenceException;
 import org.openqa.selenium.WebDriver;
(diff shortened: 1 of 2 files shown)


cnathe:
> I think our logic in 'clickFinish' assumes there to be no alerts present before clicking finish- I'm seeing staleness issues that suggest we're clicking the button and finding an alert... that becomes stale instantly. If there are alerts present before touching the button, perhaps we need to wait for them to become stale? testing my theory now

@labkey-chrisj @labkey-tchad Chris, are we still seeing these alert staleness issues on release19.3-SNAPSHOT? I just want to make sure we aren't chasing issues on develop which were fixed by the clickFinish vs clickFinishErrorExpected change that went into 19.3.

labkey-chrisj:
I haven't seen issues in 19.3.  I think we should merge your changes forward into develop, and then re-evaluate whether any of mine are necessary any more

cnathe:
Sounds good. @labkey-tchad Trey, do you know when the first merge of release19.3-SNAPSHOT to develop is scheduled to happen?

labkey-chrisj:
I thought Trey was going to do it this AM.  Or are there further
requirements?

labkey-tchad:
19.3-SNAPSHOT is merged to develop

PR #144, opened 2019-11-06
Don't test for a specific version of Rlabkey

labkey-tchad:
This check has only ever made updating Rlabkey more difficult. I hasn't, nor is it likely to, catch any regressions.
I've updated the helpers to print the Rlabkey version for possible debugging purposes.

### DIFF
--- a/data/api/rlabkey-api.xml
+++ b/data/api/rlabkey-api.xml
@@ -3,14 +3,12 @@
         <url>
             <![CDATA[
                 library(Rlabkey)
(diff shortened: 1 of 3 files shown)


labkey-tchad:
Verifying on TeamCity: https://teamcity.labkey.org/viewLog.html?buildId=849508&

labkey-tchad:
Verify on a suite that installs Rlabkey
https://teamcity.labkey.org/viewLog.html?buildId=849531

PR #147, opened 2019-11-07
Automation test : Resolve lookups by Value

labkey-sweta:


src/org/labkey/test/pages/core/admin/ShowAdminPage.java
```diff
@@ -106,6 +106,12 @@ public void clickEmailCustomization()
         clickAndWait(elementCache().emailCustomizationLink);
     }
 
+    public void clickExperimentalFeatures()
+    {
+        goToSettingsSection();
+        clickAndWait(elementCache().experimentalfeaturesLink);
+    }
+
     public void clickNotificationServiceAdmin()
```
labkey-klum: Should we be returning an ExperimentalFeatures page object that has methods to enable and disable experimental features by name? Seems like that would be a generally useful feature for others needing test automation.

src/org/labkey/test/tests/SampleSetTest.java
```diff
@@ -617,6 +646,81 @@ public void deleteLineageParent() throws IOException, CommandException
...
+                        TestDataGenerator.simpleFieldDef("label", FieldDefinition.ColumnType.String)));
+        dgen.addDataSupplier("label", () -> dgen.randomString(10))
+                .withGeneratedRows(10000);
+        dgen.createDomain(createDefaultConnection(true), "SampleSet");
+        dgen.insertRows(createDefaultConnection(true), dgen.getRows());
+
+        log("Waiting for the sample data to get generated");
+        goToProjectHome();
+        waitAndClickAndWait(Locator.linkWithText(sampleSetName));
+        waitForTextWithRefresh(500000, "1 - 100 of 10,000");
```
labkey-susanh: Is it necessary to go to this page only to go somewhere else immediately after all this waiting?  It looks like the insertRows command is a blocking command so it won't return until the samples have been inserted.  Is that not the case? Or does this cause the samples to get cached so other operations can proceed without waiting?

src/org/labkey/test/pages/core/admin/ShowAdminPage.java
labkey-sweta: Created the page and checked in

src/org/labkey/test/tests/SampleSetTest.java
labkey-sweta: There was no other option that i explored to verify if the rows were generated unless i see the number in sample set. 

You made a good point, i rearranged some code so that sample set is visited just once both to verify 10000 records and adding a additional row.

labkey-susanh: A more expedient way to verify that the rows have been inserted is to capture the response from insertRows and make sure there isn't an error there and/or that the number of rows affected is as expected.  Then you can bypass this very costly waiting for the UI altogether.

PR #151, opened 2019-11-09
Release19.3 snapshot

labkey-sweta:


labkey-tchad:
This isn't how we merge release to develop. They've usually diverged too much and TeamCity will never run a build that will verify such a merge. You need to create a feature branch off of develop and merge into that, then open a pull request from that feature branch to develop.

PR #152, opened 2019-11-09
Move drop zone interaction into FileBrowserHelper

labkey-tchad:
Add synchronization to avoid getting blocked by dropzone mask

This should fix the intermittent failure in `FileWatcherDatasetTest.testDatasetWithoutDomainUpdates` where it tries to drop two files in succession.


labkey-tchad:
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=38974

PR #155, opened 2019-11-11
19.3 fb string lookups fix : Automation test

labkey-sweta:


src/org/labkey/test/tests/SampleSetTest.java
```diff
@@ -617,6 +633,85 @@ public void deleteLineageParent() throws IOException, CommandException
...
+                        TestDataGenerator.simpleFieldDef("label", FieldDefinition.ColumnType.String)));
+        dgen.addDataSupplier("label", () -> dgen.randomString(10))
+                .withGeneratedRows(10000);
+        dgen.createDomain(createDefaultConnection(true), "SampleSet");
+        dgen.insertRows(createDefaultConnection(true), dgen.getRows());
+
+        log("Waiting for the sample data to get generated");
+        goToProjectHome();
+        waitAndClickAndWait(Locator.linkWithText(sampleSetName));
+        waitForTextWithRefresh(400000, "1 - 100 of 10,000");
```
labkey-tchad: Why so many rows? That seems like a quantity you'd use for perf or stress testing.

labkey-susanh: Perhaps I failed to submit this comment earlier, but this waiting does not seem to be necessary, even though this "feature" requires 10000 rows in order to be activated. The insertRows call is a blocking call and it looks like it  returns a structure that can be used to verify that the expected number of rows have been inserted.  I hope the rest of the test doesn't require that this UI has been rendered.

labkey-sweta: @labkey-tchad Added the commenting explaining the reason to add 10000 samples

@labkey-susanh Changing the wait condition

src/org/labkey/test/tests/SampleSetTest.java
```diff
@@ -617,6 +633,85 @@ public void deleteLineageParent() throws IOException, CommandException
...
+        goToProjectHome();
+        waitAndClickAndWait(Locator.linkWithText(sampleSetName));
+
+        log("Inserting 10,001 row in the sampleset");
+        DataRegionTable table = new DataRegionTable("Material", getDriver());
+        table.clickInsertNewRow();
+
+        setFormElement(Locator.name("quf_Name"), "Sample1");
+        setFormElement(Locator.name("quf_label"), "Sample1");
+        clickButton("Submit");
```
labkey-tchad: Why not just start with 10,001 rows? This just adds unnecessary and costly navigation if we don't do any kind of pre-check before adding the row.

src/org/labkey/test/tests/SampleSetTest.java
```diff
@@ -617,6 +633,85 @@ public void deleteLineageParent() throws IOException, CommandException
...
+        log("Adding the lookUp field in the list");
+        DomainFieldRow lookUpRow = domainFormPanel.addField("lookUpField")
+                .setType(FieldDefinition.ColumnType.Lookup)
+                .expand()
+                .setFromFolder("Current Folder")
+                .setFromSchema("exp.materials")
+                .setFromTargetTable("10000Samples (Integer)")
+                .setDescription("LookUp in same container with 10000 samples");
+
+        domainDesignerPage.clickFinish();
```
labkey-tchad: Add this field during the initial list creation. No need to mess around in the UI for this test.

src/org/labkey/test/tests/SampleSetTest.java
```diff
@@ -617,6 +633,85 @@ public void deleteLineageParent() throws IOException, CommandException
...
+
+        log("Inserting the new row in the list with the newly created sample as lookup");
+        goToProjectHome();
+        clickAndWait(Locator.linkWithText(listName));
+        table = new DataRegionTable("query", getDriver());
+        table.clickInsertNewRow();
+        setFormElement(Locator.name("quf_id"), "1");
+        setFormElement(Locator.name("quf_name"), "1");
+        setFormElement(Locator.name("quf_lookUpField"), "Sample2");
```
labkey-tchad: Add a comment that this is the input we care about.
```
        setFormElement(Locator.name("quf_lookUpField"), "Sample2"); // This becomes an <input> when there are >10000 options
```

src/org/labkey/test/tests/SampleSetTest.java
```diff
@@ -617,6 +633,85 @@ public void deleteLineageParent() throws IOException, CommandException
...
+        clickButton("Submit");
+
+        String errMsg = Locator.tagWithAttribute("font","class","labkey-error").findElement(getDriver()).getText();
+        assertEquals("Expecpted error is different","Could not convert value: Sample2",errMsg);
```
labkey-tchad: Clean up formatting. There should be spaces between args.
```suggestion
        assertEquals("Expecpted error is different", "Could not convert value: Sample2", errMsg);
```

src/org/labkey/test/tests/SampleSetTest.java
labkey-sweta: The rows(Test data generator)  which get added are random strings and different every time we run the test, so decided to add 10,001 row manually.

PR #162, opened 2019-11-14
Scroll to hidden fields checkbox

labkey-tchad:
NLP tests have been failing because this checkbox for a particular data region ends up a pixel from the bottom of the window and clicks aren't getting through.


labkey-tchad:
Verifying fix: https://teamcity.labkey.org/viewLog.html?buildId=854990

PR #163, opened 2019-11-14
Fix stale element exceptions when defining lookup fields

labkey-tchad:
Also clean up some warnings caused by using `setFormElement` on a `<select>`


labkey-tchad:
SampleManager tests are happy: https://teamcity.labkey.org/viewLog.html?buildId=855064&buildTypeId=LabKey_193ReleaseSnapshot_Premium_ModulesSuites_SampleManagementSqlserver

labkey-tchad:
Other troublesome tests (assayRequest) are also good.

labkey-tchad:
Going to do one last verification against `SampleSetTest`, which has been throwing stale `<option>` exceptions: https://teamcity.labkey.org/viewQueued.html?itemId=858337

PR #164, opened 2019-11-15
Item 6425: selenium test fixes related to SM app user profile changes

cnathe:


cnathe:
> I'm not seeing exactly what changed recently to require this change, but it seems fine.

@labkey-susanh I made a change to the PasswordRule.java class in the matching fb_smUserProfile branch of platform (to use full sentences in the summary text):
https://github.com/LabKey/platform/pull/592/files

cnathe:
@labkey-tchad Trey, I made some more changes to this PR after your review, so requesting re-review. The changes were related to assay import for the SM folder type change to redirect directly to the app for the SM project. This means we need a way to call the importAssayRun without it doing the folder navigation.

cnathe:
FYI, Dan is going to take a look since Trey is out sick today.

PR #166, opened 2019-11-19
Adding check to ignore the dumbster component if it is missing.

labkey-danield:
If test automation is run against some distributions some of the components will be (intentionally) missing. For this to work with TeamCity runs we need to add the parameter ignoreMissingModules to the test.properties file.


labkey-danield:
After reviewing with Trey added the variable webtest.ignoreMissingModules=false to test.properties. This should be the only change necessary.

PR #170, opened 2019-11-21
Issue 39063  use webdav uploader rather than set pipeline root to loc…

labkey-chrisj:
Updates StudyBaseTest to use webDav instead of local filepath, making it possible to run tests against a remote server


labkey-adam:
@labkey-chrisj Looks like this latest commit went into the wrong branch

PR #171, opened 2019-11-23
Issue 39022: GWT refuses to serialize exception when resizing column on SQL Server 2019

labkey-adam:


labkey-adam:
ColumnResizeTest passes on SQL Server 2019 and PostreSQL:

- https://teamcity.labkey.org/viewLog.html?buildId=861311
- https://teamcity.labkey.org/viewLog.html?buildId=861304

PR #173, opened 2019-11-25
Added various getters to return content from a domain designer page.

labkey-danield:
Made some minor changes to the DomainDesigner pages to return list of the panes present (needed to validate behavior in SampleManager).
Fixed an issue in DomainFieldRow.getType.
Added getters for some of the fields in the ReactDesignerPage that are used in the SampleManager tests.

### DIFF
--- a/src/org/labkey/test/components/glassLibrary/components/MultiMenu.java
+++ b/src/org/labkey/test/components/glassLibrary/components/MultiMenu.java
@@ -6,6 +6,7 @@ package org.labkey.test.components.glassLibrary.components;
 
 import org.junit.Assert;
 import org.labkey.test.Locator;
+import org.labkey.test.WebDriverWrapper;
 import org.labkey.test.components.html.BootstrapMenu;
 import org.labkey.test.util.LogMethod;
 import org.labkey.test.util.LoggedParam;
@@ -67,10 +68,13 @@ public class MultiMenu extends BootstrapMenu
         List<String> menuText = new ArrayList<>();
 
         boolean stale;
+        boolean loading;
+        int tries = 1;
 
         do {
 
             stale = false;
+            loading = false;
 
             try
             {
@@ -79,7 +83,15 @@ public class MultiMenu extends BootstrapMenu
                 menuText = new ArrayList<>();
 
                 for (WebElement menuItem : topMenuList) {
+
+                    String menuItemText = menuItem.getText();
+
+                    // Check to see if we are still loading data.
+                    if(menuItemText.toLowerCase().contains("loading"))
+                        loading = true;
+
                     menuText.add(menuItem.getText());
+
                 }
 
             }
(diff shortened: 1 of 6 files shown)


src/org/labkey/test/components/glassLibrary/components/MultiMenu.java
```diff
@@ -60,16 +62,67 @@ public void doMenuAction(List<String> pathToAction)
...
+            if(stale || loading )
+                WebDriverWrapper.sleep(500);
+
+            tries++;
+
+            // Keep trying until we get valid menu text.
+        } while((stale || loading) && (tries <= 10));
+
+        if(tries > 10)
+            throw new RuntimeException("Menu text never returned.");
```
labkey-tchad: Use the stale and loading flags to trigger this error state. It's less prone to off-by-one errors and we could generate more informative error messages (maybe include the menu text if we found any before running out of time). e.g.
```
if (stale)
{
    throw new RuntimeException("Menu items kept going stale");
}
if (loading)
{
    throw new RuntimeException("Menu items still loading. " + menuText.toString());
}
```

src/org/labkey/test/components/glassLibrary/components/MultiMenu.java
```diff
@@ -60,16 +62,67 @@ public void doMenuAction(List<String> pathToAction)
...
+
+                }
+
+            }
+            catch(StaleElementReferenceException staleExc)
+            {
+                // This happens in the time between the change of the menu content from containing "loading"
+                // to having data (like "sample sets").
+                stale = true;
+            }
```
labkey-tchad: Make sure that it wasn't the menu itself that went stale so this doesn't keep looping futilely.
```
catch (..)
{
    [..]
    getComponentElement().isDisplayed(); // will throw an uncaught 'StaleReferenceException' if the entire menu went stale
    stale = true;
}
```

src/org/labkey/test/pages/ReactAssayDesignerPage.java
```diff
@@ -53,12 +52,29 @@ public ReactAssayDesignerPage setName(String name)
...
+    public String getName()
+    {
+        return elementCache().nameInput.get();
+    }
+
+    // I think the name field is the only field that can be enabled/disabled.
+    // For example once an assay is created you can't change it's name so the field will be disabled.
+    public boolean isNameEnabled()
+    {
+        return null == elementCache().nameInput.getComponentElement().getAttribute( "disabled");
```
labkey-tchad: If we're using the `disabled` attribute, I think the standard `isEnabled()` method would work. (Though perhaps not)
```suggestion
        return elementCache().nameInput.getComponentElement().isEnabled();
```

src/org/labkey/test/components/glassLibrary/components/MultiMenu.java
labkey-danield: Updated.

src/org/labkey/test/components/glassLibrary/components/MultiMenu.java
labkey-danield: Updated.

src/org/labkey/test/pages/ReactAssayDesignerPage.java
labkey-danield: Ok, we will give it a shot and see.

PR #178, opened 2019-12-09
SM Workflow Search 6491

labkey-tchad:
Refactored some search and assay helpers.

### DIFF
--- a/src/org/labkey/test/pages/assay/ChooseAssayTypePage.java
+++ b/src/org/labkey/test/pages/assay/ChooseAssayTypePage.java
@@ -1,6 +1,7 @@
 package org.labkey.test.pages.assay;
 
 import org.labkey.test.Locator;
+import org.labkey.test.Locators;
 import org.labkey.test.WebDriverWrapper;
 import org.labkey.test.WebTestHelper;
 import org.labkey.test.components.html.RadioButton;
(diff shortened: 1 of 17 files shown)


src/org/labkey/test/pages/assay/ChooseAssayTypePage.java
```diff
@@ -0,0 +1,76 @@
...
+        clickAndWait(elementCache().cancelButton);
+    }
+
+    public ReactAssayDesignerPage clickNext()
+    {
+        clickAndWait(elementCache().nextButton);
+
+        return new ReactAssayDesignerPage(getDriver());
+    }
+
```
labkey-danield: Don't you need, or shouldn't you have, a clickNextExpectError method?

src/org/labkey/test/pages/assay/plate/PlateTemplateList.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.labkey.test.Locator;
+import org.labkey.test.WebDriverWrapper;
+import org.labkey.test.WebTestHelper;
+import org.labkey.test.pages.LabKeyPage;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+public class PlateTemplateList extends LabKeyPage<PlateTemplateList.ElementCache>
+{
```
labkey-danield: Is this class and the one above, PlateDesignerPage, used anywhere?

src/org/labkey/test/pages/assay/ChooseAssayTypePage.java
labkey-tchad: That easy enough to add now. Didn't realize the page had any error state.

src/org/labkey/test/pages/assay/plate/PlateTemplateList.java
labkey-tchad: Added some JavaDoc explicitly identifying these to page as incomplete.

PR #181, opened 2019-12-18
Item 6426: Test app Profile and Permissions

labkey-tchad:


src/org/labkey/test/BootstrapLocators.java
```diff
@@ -2,8 +2,13 @@
...
 {
-    static public Locator infoAlert = Locator.tagWithClass("div", "alert-info");
-    static public Locator successAlert = Locator.tagWithClass("div", "alert-success");
-    static public Locator dangerAlert = Locator.tagWithClass("div", "alert-danger");
-    static public Locator warningAlert = Locator.tagWithClass("div", "alert-warning");
+    public static final Locator infoAlert = Locator.tagWithClass("div", "alert-info");
+    public static final Locator successAlert = Locator.tagWithClass("div", "alert-success");
+    public static final Locator dangerAlert = Locator.tagWithClass("div", "alert-danger");
+    public static final Locator warningAlert = Locator.tagWithClass("div", "alert-warning");
+
```
labkey-danield: I've changed these in my branch to be *Banner and not *Alert, but I think you have already seen that.
It will just get resolved during merge.

src/org/labkey/test/components/glassLibrary/components/ReactSelect.java
```diff
@@ -303,6 +319,7 @@ public ReactSelect select(String option)
                 .open()
                 .clickOption(option)
                 .getSelections();
+        waitForClosed();
 
         // TODO Comment out this line for the time being. Issue 37897: Sample Management: Lookup field values are not consistently displayed between the edit sample panel and the create sample panel.
 //        assertTrue("Expected '" + option + "' to be selected.  Current selections: " + selections, selections.contains(option));
```
labkey-danield: Not part of your change, but I think these two commented lines should go away. 
Ideally this should be a control used in other apps not just SM, so the first comment is not appropriate. And the bug referenced is TBD so who knows when it would get fixed.
Second I like that there is an assert in the control that checks to make sure it did wha tit was asked to do, but if it can't do that reliably for everyone it shouldn't do it.

src/org/labkey/test/util/DeferredErrorCollector.java
```diff
@@ -37,14 +42,32 @@ public DeferredErrorCollector(@NotNull BaseWebDriverTest test)
...
+    /**
+     * Reset this error checker to
+     */
+    public void resetErrorTypes()
+    {
+        errorTypes.clear();
+        errorTypes.add(AssertionError.class);
+    }
+
+    public void addRecordableErrorType(Class<? extends Throwable> errorType)
```
labkey-danield: JavaDocs comment for public interfaces?

src/org/labkey/test/util/DeferredErrorCollector.java
```diff
@@ -365,9 +418,45 @@ public int getErrorCount()
         }
 
         @Override
-        public String takeScreenShot(String snapShotName)
+        public String takeScreenShot(String screenshotName)
```
labkey-danield: Same comment above about JavaDoc comments and public interfaces.

src/org/labkey/test/util/URLBuilder.java
```diff
@@ -0,0 +1,135 @@
...
+            if (!_resourcePath.startsWith("#"))
+            {
+                url.append("#");
+            }
+            url.append(_resourcePath);
+            appendQueryString(url, _secondaryQuery);
+        }
+        else if (_secondaryQuery != null && !_secondaryQuery.isEmpty())
+        {
+            throw new IllegalArgumentException("Must supply a resource path");
```
labkey-danield: Maybe re-word this error message slightly, it took me a moment to understand the logic. A message that indicates you can't have a SecondQuery value with out a ResourcePath.

src/org/labkey/test/util/URLBuilder.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    private void appendQueryString(StringBuilder url, Map<String, Object> params)
+    {
+        if (params != null)
+        {
+            boolean firstParam = true;
+            for (Map.Entry param : params.entrySet())
+            {
+                if (null != param.getKey())
```
labkey-danield: Would you ever have a null key?

src/org/labkey/test/components/glassLibrary/components/ReactSelect.java
labkey-tchad: Removed. Additionally, such validation isn't valid for all selects. In the the SM permissions page, the selected items show up elsewhere on the page, not in the select itself.

src/org/labkey/test/util/DeferredErrorCollector.java
labkey-tchad: Yup. Got half way through the javadoc for these new methods and forgot to get back to them.

src/org/labkey/test/util/DeferredErrorCollector.java
labkey-tchad: The overridden method is documented. This is in the private wrapper class used for the `fatal` and `withScreenshot` error collectors.

src/org/labkey/test/util/URLBuilder.java
labkey-tchad: Certainly. I also added some JavaDoc to the top of this class since it's pretty bare otherwise.

src/org/labkey/test/util/URLBuilder.java
labkey-tchad: Not generally, but some kinds of `Maps` allow them. This was copied from `WebTestHelper` and I didn't want to mess with its existing functionality.

PR #191, opened 2019-12-30
Fb protocol command remote api

labkey-chrisj:


src/org/labkey/remoteapi/assay/Protocol.java
```diff
@@ -0,0 +1,415 @@
...
+        {
+            for (Object detectionMethod : (JSONArray)json.get("availableDetectionMethods"))
+                _availableDetectionMethods.add((String)detectionMethod);
+        }
+        if (json.containsKey("selectedDetectionMethod"))
+            _selectedDetectionMethod = (String)json.get("selectedDetectionMethod");
+
+        if (json.containsKey("availableMetadataInputFormats"))
+        {
+            _availableMetadataInputFormats = (HashMap<String,String>)json.get("availableMetadataInputFormats");
```
cnathe: Does this casting to a HashMap directly here work? I would have expected you to need to cast it to a JSONObject and then iterate through it to build up the map yourself.

src/org/labkey/remoteapi/assay/Protocol.java
```diff
@@ -0,0 +1,415 @@
...
+            _selectedPlateTemplate = (String)json.get("selectedPlateTemplate");
+
+        if (json.get("protocolTransformScripts") instanceof JSONArray)
+        {
+            for (Object transformScript : (JSONArray)json.get("protocolTransformScripts"))
+                _protocolTransformScripts.add((String)transformScript);
+        }
+        if (json.containsKey("protocolParameters"))
+        {
+            _protocolParameters = (HashMap<String,String>)json.get("protocolParameters");
```
cnathe: Same comment as above for the cast to HashMap, if this is working that's great just want to make sure it is verified.

src/org/labkey/remoteapi/assay/Protocol.java
```diff
@@ -0,0 +1,415 @@
...
+        result.put("name", _name);
+        result.put("description", _description);
+        result.put("providerName", _providerName);
+        JSONArray domains = new JSONArray();
+        result.put("domains", domains);
+        for (Domain domain : _domains)
+            domains.add(domain.toJSONObject(true));
+
+        if (_allowBackgroundUpload != null)
+            result.put("allowBackgroundUpload", _allowBackgroundUpload);
```
cnathe: I believe that this toJSONObject method is only called in the save case. If so, I think technically we could leave off the various "allow" related parameters since those are just informational coming back from the GetProtocolCommand and not something that can be set / saved. @labkey-klum thoughts?

src/org/labkey/remoteapi/assay/Protocol.java
```diff
@@ -0,0 +1,415 @@
...
+        if (_protocolParameters != null)
+            result.put("protocolParameters", _protocolParameters);
+
+        if (_protocolTransformScripts != null)
+        {
+            JSONArray transformScripts = new JSONArray();
+            result.put("protocolTransformScripts", transformScripts);
+            for (String script : _protocolTransformScripts)
+                transformScripts.add(script);
+        }
```
cnathe: This one fits with the comment above related to if we need to include the static parameters in the JSON object here (so I think the set would be the ones prefixed with "allow", prefixed with "available", and protocolTransformScripts. I don't know what protocolParameters are, might be worth checking with @labkey-klum

src/org/labkey/remoteapi/assay/Protocol.java
```diff
@@ -0,0 +1,415 @@
...
+    {
+        _allowBackgroundUpload = allowBackgroundUpload;
+        return this;
+    }
+    public Boolean getAllowBackgroundUpload()
+    {
+        return _allowBackgroundUpload;
+    }
+
+    public Protocol setAllowEditableResults(boolean allowEditableResults)
```
cnathe: Similar to the comments above, if some of these properties are not editable, we should probably remove the setters for them.

src/org/labkey/remoteapi/assay/Protocol.java
```diff
@@ -0,0 +1,415 @@
...
+    }
+
+    public String getDescription()
+    {
+        return _description;
+    }
+
+    public void setDescription(String description)
+    {
+        _description = description;
```
cnathe: Should this one be changed to `return this;` with a return type of `Protocol`?

src/org/labkey/remoteapi/assay/Protocol.java
```diff
@@ -0,0 +1,415 @@
...
+    }
+
+    public List<Domain> getDomains()
+    {
+        return _domains;
+    }
+
+    public void setDomains(List<Domain> domains)
+    {
+        _domains = domains;
```
cnathe: Similar to above, should this include `return this`?

src/org/labkey/test/AssayAPITest.java
```diff
@@ -106,6 +111,51 @@ protected void importAssayAndRun(File assayPath, int pipelineCount, String assay
...
+        String assayName = "testGpatAssay";
+
+        Connection connection = createDefaultConnection(true);
+        GetProtocolCommand getProtocolCommand = new GetProtocolCommand("General");                      // gets a template from the server
+        ProtocolResponse getProtocolResponse = getProtocolCommand.execute(connection, getCurrentContainerPath());
+
+        Protocol newAssayProtocol = getProtocolResponse.getProtocol();
+        newAssayProtocol.setName(assayName)
+            .allowQCStates(true)
+            .setAllowEditableResults(true)
```
cnathe: I don't think it makes sense to set the "allowQCStates" and "setAllowEditableResults", the assay provider either allows that or does not. For example, if you set those to false for a GPAT assay, I assume the response for the save protocol would still say true for those props.

src/org/labkey/remoteapi/assay/Protocol.java
labkey-jeckels: JSONObject subclasses HashMap, so this works. However, it might be a little clearer to call getJSONObject() instead of get(). Of course, JSONObject is a Map<String, Object>, not Map<String, String>, so it might need a little coercion since this code is blindly assuming that it will only contain Strings for values.

src/org/labkey/remoteapi/assay/Protocol.java
labkey-jeckels: Agreed. Chris, we talked about these "allow" properties last week. They're read-only.

src/org/labkey/test/AssayAPITest.java
```diff
@@ -106,6 +111,51 @@ protected void importAssayAndRun(File assayPath, int pipelineCount, String assay
...
+            .setEditableResults(true)
+            .setEditableRuns(true);
+        SaveProtocolCommand saveProtocolCommand = new SaveProtocolCommand(newAssayProtocol);
+        ProtocolResponse saveProtocolResponse = saveProtocolCommand.execute(connection, getCurrentContainerPath());
+
+        assertTrue(saveProtocolResponse.getProtocol().getAllowQCStates());
+        assertTrue(saveProtocolResponse.getProtocol().getAllowEditableResults());
+        assertTrue(saveProtocolResponse.getProtocol().getEditableResults());
```
labkey-jeckels: Consider also doing a separate call to GetProtocolCommand to make sure that the values stick

src/org/labkey/remoteapi/assay/Protocol.java
labkey-chrisj: Okay, I'll take those out but add a note on which ones are absent, and why

src/org/labkey/remoteapi/assay/Protocol.java
labkey-chrisj: That would be more consistent.  I'll do this for the rest of the setters like it to make chaining/builder syntax available

src/org/labkey/test/AssayAPITest.java
labkey-chrisj: good idea.
Doing so brought up that the GetProtocolCommand and Protocol classes disagree on whether the protocolID should be an int or a long- I've normalized them on 'Long'

src/org/labkey/test/AssayAPITest.java
labkey-chrisj: Thanks for this; I implemented most of these not knowing which ones would be ignored or non-settable.

src/org/labkey/remoteapi/assay/Protocol.java
labkey-chrisj: yes.

src/org/labkey/remoteapi/assay/GetProtocolCommand.java
```diff
@@ -9,7 +9,7 @@
 public class GetProtocolCommand extends Command<ProtocolResponse>
 {
     private String _providerName;
-    private Integer _protocolId;
+    private Long _protocolId;
```
labkey-jeckels: I'd make this a long instead of a Long. We need a value for the API to work, so we should prevent users from passing in null

labkey-chrisj: How does that square to the case where we get an instance using the string (providerName) constructor?  Also, the @Override getParameters method checks _protocolId for null to determine whether this instance uses protocolId or providerName

cnathe: @labkey-jeckels The case for having _protocolId be null is to get the template for an assay provider. The assay-getProtocol action when call with just a provider name and no protocolId will get you the template for that assay. I believe that is the reason for using "Long" here instead of "long", right?

labkey-chrisj: Keeping the member variable Long, updating constructors and setters that access it long

labkey-jeckels: Chris and I just talked over IM. I didn't realize it was either/or with the protocolId vs the name. I recommended keeping the member variable as a Long, but changing the constructor argument to a long, to signal if you're opting for the Id approach, you have to supply a value

cnathe: Yes, changing the constructor argument makes sense. Thanks.

src/org/labkey/remoteapi/assay/Protocol.java
```diff
@@ -0,0 +1,355 @@
...
+            result.put("backgroundUpload", _backgroundUpload);
+
+        if (_editableResults != null)
+            result.put("editableResults", _editableResults);
+
+        if (_qcEnabled != null)
+            result.put("qcEnabled", _qcEnabled);
+
+        if (_allowTransformationScript != null)
+            result.put("allowTransformationScript", _allowTransformationScript);
```
cnathe: I believe you want the `protocolTransformScripts` prop here in the result instead of the `allowTransformationScript `, right?

src/org/labkey/remoteapi/assay/Protocol.java
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    public Boolean getAllowEditableResults()
+    {
+        return _allowEditableResults;
+    }
+
+    public Protocol allowQCStates(boolean qcEnabled)
+    {
+        _allowQCStates = qcEnabled;
```
cnathe: This should be `setQcEnabled` and should be `_qcEnabled = qcEnabled` I believe.

src/org/labkey/remoteapi/assay/Protocol.java
```diff
@@ -0,0 +1,355 @@
...
+        return this;
+    }
+    public Map<String,String> getProtocolParameters()
+    {
+        return _protocolParameters;
+    }
+
+    public List<String> getProtocolTransformScripts()
+    {
+        return _protocolTransformScripts;
```
cnathe: I don't see a setter for _protocolTransformScripts, or maybe I'm just missing it in this PR.

src/org/labkey/remoteapi/assay/Protocol.java
cnathe: Oh nevermind, I see the setQcEnabled below, so this allowQcState setter should be removed.

src/org/labkey/test/AssayAPITest.java
```diff
@@ -106,6 +111,40 @@ protected void importAssayAndRun(File assayPath, int pipelineCount, String assay
...
+        String assayDescription = "generated for test purposes over remoteAPI";
+
+        Connection connection = createDefaultConnection(true);
+        GetProtocolCommand getProtocolCommand = new GetProtocolCommand("General");                      // gets a template from the server
+        ProtocolResponse getProtocolResponse = getProtocolCommand.execute(connection, getCurrentContainerPath());
+
+        Protocol newAssayProtocol = getProtocolResponse.getProtocol();
+        newAssayProtocol.setName(assayName)
+            .setDescription(assayDescription)
+            .allowQCStates(true)
```
cnathe: This should be calling the setter for qcEnabled right, `setQcEnabled(true)?`

src/org/labkey/test/AssayAPITest.java
```diff
@@ -106,6 +111,40 @@ protected void importAssayAndRun(File assayPath, int pipelineCount, String assay
...
+            .setEditableResults(true)
+            .setEditableRuns(true);
+        SaveProtocolCommand saveProtocolCommand = new SaveProtocolCommand(newAssayProtocol);
+        ProtocolResponse saveProtocolResponse = saveProtocolCommand.execute(connection, getCurrentContainerPath());
+        Long protocolId = saveProtocolResponse.getProtocol().getProtocolId();
+
+        assertEquals(assayDescription, saveProtocolResponse.getProtocol().getDescription());
+        assertTrue(saveProtocolResponse.getProtocol().getAllowQCStates());
```
cnathe: Shouldn't getAllowQCStates() be getQcEnabled()?

src/org/labkey/test/AssayAPITest.java
```diff
@@ -106,6 +111,40 @@ protected void importAssayAndRun(File assayPath, int pipelineCount, String assay
...
+        assertTrue(saveProtocolResponse.getProtocol().getEditableResults());
+        assertTrue(saveProtocolResponse.getProtocol().getEditableRuns());
+
+        GetProtocolCommand protocolCommand = new GetProtocolCommand(protocolId);
+        ProtocolResponse doubleCheckProtocolResponse = protocolCommand.execute(connection, getCurrentContainerPath());
+
+        assertEquals(assayDescription, doubleCheckProtocolResponse.getProtocol().getDescription());
+        assertTrue(doubleCheckProtocolResponse.getProtocol().getAllowQCStates());
```
cnathe: Shouldn't getAllowQCStates() be getQcEnabled()?

src/org/labkey/remoteapi/assay/Protocol.java
labkey-chrisj: good catch.  Removing it

src/org/labkey/remoteapi/assay/Protocol.java
labkey-chrisj: setter for _protocolTransformScripts was removed in  f14874bf9a3ba59c1d7df22f8508759a04259581

Should it be restored?

src/org/labkey/test/AssayAPITest.java
labkey-chrisj: doh, yes

src/org/labkey/test/AssayAPITest.java
labkey-chrisj: yep

src/org/labkey/remoteapi/assay/Protocol.java
labkey-chrisj: right

src/org/labkey/remoteapi/assay/Protocol.java
cnathe: Yes, that is how the transform scripts for an assay design would be set with this remoteapi.

labkey-chrisj: ok, bringing it back

src/org/labkey/test/AssayAPITest.java
labkey-chrisj: not unless allowing qc states is the same as setting them enabled/disabled

PR #198, opened 2020-01-10
Fb multi auth ui 6498

labkey-chrisj:
-updates to UI testing for multi-auth (particularly ldap) configuration
related PR: same branch name in ldap module


src/org/labkey/test/components/html/ToggleButton.java
```diff
@@ -0,0 +1,90 @@
...
+import org.labkey.test.components.WebDriverComponent;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+public class ToggleButton extends WebDriverComponent<ToggleButton.ElementCache>
+{
+    final WebElement _el;
+    final WebDriver _driver;
+
+    public ToggleButton(WebElement element, WebDriver driver)
```
labkey-tchad: Make the constructor `protected` so that only the `ToggleButtonFinder` can create it.

src/org/labkey/test/components/html/ToggleButton.java
```diff
@@ -0,0 +1,90 @@
...
+    {
+        return _driver;
+    }
+
+    public ToggleButton set(boolean enabled)
+    {
+        String desiredState = enabled ? "enabled" : "disabled";
+        if (get() != enabled)
+            getComponentElement().click();
+        getWrapper().waitFor(()-> get() == enabled,
```
labkey-tchad: Pay attention to IntelliJ warnings. `waitFor` is a static method:
```suggestion
        WebDriverWrapper.waitFor(()-> get() == enabled,
```

src/org/labkey/test/pages/core/login/AuthDialogBase.java
```diff
@@ -0,0 +1,97 @@
...
+import org.labkey.test.components.html.Input;
+import org.labkey.test.components.html.ToggleButton;
+import org.labkey.test.params.login.AuthenticationProvider;
+import org.openqa.selenium.WebDriver;
+
+public abstract class AuthDialogBase<T extends AuthDialogBase> extends ModalDialog
+{
+    private final LoginConfigRow _row;
+
+    protected AuthDialogBase(AuthenticationProvider<org.labkey.test.pages.ldap.LdapConfigureDialog> provider, WebDriver driver)
```
labkey-tchad: Remove reference to code in a different rep:
```suggestion
    protected AuthDialogBase(AuthenticationProvider provider, WebDriver driver)
```

src/org/labkey/test/pages/core/login/AuthDialogBase.java
```diff
@@ -0,0 +1,97 @@
...
+    protected ElementCache elementCache()
+    {
+        return  (ElementCache) super.elementCache();
+    }
+
+    protected class ElementCache extends ModalDialog.ElementCache
+    {
+//        WebElement enableToggle = Locator.tagWithClass("div", "toggle")
+//                //.withChild(Locator.tagWithClass("div", "toggle-group"))
+//                .findWhenNeeded(this).withTimeout(2000);
```
labkey-tchad: Cleanup old code

src/org/labkey/test/pages/core/login/AuthDialogBase.java
```diff
@@ -0,0 +1,97 @@
...
+        ToggleButton enableSlider = new ToggleButton.ToggleButtonFinder(getDriver())
```
labkey-tchad: Use consistent terminology. Name this variable `enableToggle` or name the class `SliderButton`. I could, honestly, go either way.

src/org/labkey/test/pages/core/login/LoginConfigRow.java
```diff
@@ -0,0 +1,115 @@
...
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.ui.ExpectedConditions;
+
+public class LoginConfigRow extends WebDriverComponent<LoginConfigRow.ElementCache>
+{
+    final WebElement _el;
+    final WebDriver _driver;
+
+    public LoginConfigRow(WebElement element, WebDriver driver)
```
labkey-tchad: Make the constructor `protected` to force usage of the `LoginConfigRowFinder`.

src/org/labkey/test/pages/core/login/LoginConfigRow.java
```diff
@@ -0,0 +1,115 @@
...
+        return new ElementCache();
+    }
+
+
+    protected class ElementCache extends WebDriverComponent.ElementCache
+    {
+        final WebElement baseFieldsElement = Locator.tagWithClass("div", "domain-row-base-fields").findWhenNeeded(this);
+        final WebElement description = Locator.tagWithClass("div", "description").findElement(baseFieldsElement);
+        final WebElement details = Locator.tagWithClass("div", "details").findElement(baseFieldsElement);
+        final WebElement provider = Locator.tagWithClass("div", "provider").findElement(baseFieldsElement);
```
labkey-tchad: Any reason to not use `findWhenNeeded` for these?

src/org/labkey/test/pages/core/login/LoginConfigurePage.java
```diff
@@ -20,61 +22,64 @@ public LoginConfigurePage(WebDriver driver)
...
-        {
-            throw new RuntimeException("Unable to instantiate page class: " + authType.getName(), e);
-        }
+        new LoginConfigRow.LoginConfigRowFinder(getDriver()).withDescription(description).waitFor()
+                .clickDelete();
 
-        WebElement row = Locator.xpath("//tbody/tr").containing(description).findElement(elementCache().tableElement);
-        clickAndWait(Locator.tagWithClass("a", "labkey-text-link").withText("edit").findElement(row));
+        // pause, because that works?
+        sleep(1000);
```
labkey-tchad: `LoginConfigRow.clickDelete` should handle this timing.

labkey-tchad (changes requested):

src/org/labkey/test/pages/core/login/LoginConfigRow.java
labkey-chrisj: nope.  it's a good idea

src/org/labkey/test/pages/core/login/LoginConfigurePage.java
labkey-chrisj: It just happens that the timing in clickDelete has changed; deleting a configuration now puts up a prompt (which clickDelete handles, then waits for the row to become stale)
...so this sleep (a holdover from the old UI) was probably unnecessary

src/org/labkey/test/pages/core/login/AuthDialogBase.java
```diff
@@ -0,0 +1,107 @@
...
+
+    public boolean isEnabled()
+    {
+        return elementCache().enableToggle.get();
+    }
+
+    public T clickButtonExpectingError(String buttonText)
+    {
+        Locators.dismissButton(buttonText).waitForElement(this, 2000).click();
+        return getThis();
```
labkey-tchad, line 68: `*ExpectingError` methods should wait for and return the error.

src/org/labkey/test/pages/core/login/LoginConfigurePage.java
```diff
@@ -86,25 +159,51 @@ protected ElementCache newElementCache()
...
+                    .child(Locator.tagWithClass("span", "clickable"));
+        }
+
+        WebElement selfSignupCheckBox = checkBoxLoc("Allow self sign up").findWhenNeeded(globalSettingsPanel())
+                .withTimeout(WAIT_FOR_JAVASCRIPT);
+        WebElement allowUserEmailEditCheckbox = checkBoxLoc("Allow users to edit their own email addresses")
+                .findWhenNeeded(globalSettingsPanel()).withTimeout(WAIT_FOR_JAVASCRIPT);
+        WebElement autoCreateCheckBox = checkBoxLoc("Auto-create authenticated users")
+                .findWhenNeeded(globalSettingsPanel())
+                .withTimeout(WAIT_FOR_JAVASCRIPT);
```
labkey-tchad: Does the `SvgCheckbox` component not work for these ones?

src/org/labkey/test/tests/SecondaryAuthenticationTest.java
```diff
@@ -1,169 +0,0 @@
...
-import java.util.Date;
-import java.util.List;
-import java.util.Map;
-
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-
-@Category({DailyA.class})
-@BaseWebDriverTest.ClassTimeout(minutes = 1)
-public class SecondaryAuthenticationTest extends BaseWebDriverTest
```
labkey-tchad, line 44: Why delete this test? It just needs an update for the new UI.

labkey-tchad (changes requested):

src/org/labkey/test/tests/SecondaryAuthenticationTest.java
labkey-adam: Moved to devtools (platform repo) I believe... which is where the provider lives now

src/org/labkey/test/pages/core/login/AuthDialogBase.java
labkey-chrisj: At this point, that's going to be a trick.
In the SamlConfigureDialog at least, some of the errors are at the field-level, some are text somewhere in the bottom of the dialog.  Not all of the errors are in a predictable or consistent place.

src/org/labkey/test/pages/core/login/LoginConfigurePage.java
labkey-chrisj: What happened here... is I didn't refactor to use them here.  Good catch

src/org/labkey/test/pages/core/login/AuthDialogBase.java
labkey-tchad: Oh, that makes sense.

src/org/labkey/test/tests/SecondaryAuthenticationTest.java
labkey-chrisj: it has been refactored to use the new UI, and moved into devtools, same branch name

src/org/labkey/test/components/html/ToggleButton.java
```diff
@@ -0,0 +1,91 @@
...
+    protected ElementCache newElementCache()
+    {
+        return new ElementCache();
+    }
+
+
+    protected class ElementCache extends WebDriverComponent.ElementCache
+    {
+
+    }
```
labkey-tchad: ElementCache stuff is just taking up space. There's nothing to cache and there likely never will be.

src/org/labkey/test/pages/core/admin/ShowAdminPage.java
```diff
@@ -197,7 +197,7 @@ protected ElementCache newElementCache()
         protected WebElement analyticsSettingsLink = Locator.linkWithText("analytics settings").findWhenNeeded(this);
         protected WebElement externalRedirectHostLink = Locator.linkWithText("External Redirect Hosts").findElement(this);
         protected WebElement auditLogLink = Locator.linkWithText("audit log").findWhenNeeded(this);
-        protected WebElement authenticationLink = Locator.linkWithText("authentication").findWhenNeeded(this);
+        protected WebElement authenticationLink = Locator.linkWithText("authentication (deprecated)").findWhenNeeded(this);
```
labkey-tchad: Remove this. Deprecated authentication page is no more.

src/org/labkey/test/pages/core/login/AuthDialogBase.java
```diff
@@ -0,0 +1,112 @@
...
+     * a new config vs. editing an existing one.
+     */
+    public LoginConfigRow clickApply()
+    {
+        String description= getDescription();
+        Locator.findAnyElement("Finish or Apply button", this,
+                Locators.dismissButton("Finish"), Locators.dismissButton("Apply")).click();
+        waitForClose(4);
+        return new LoginConfigRow.LoginConfigRowFinder(getDriver()).withDescription(description).waitFor();
+    }
```
labkey-tchad: We know whether this is a new config based on whether a `LoginConfigRow` was passed in. We could also use this knowledge to make `clickButtonExpectingError` not require button text.
```
dismiss(_row == null ? "finish" : "apply");
```

src/org/labkey/test/pages/core/login/DatabaseAuthConfigureDialog.java
```diff
@@ -0,0 +1,165 @@
...
+
+    public DatabaseAuthConfigureDialog setPasswordStrength(PasswordStrength desiredStrength)
+    {
+        if (getPasswordStrength().equals(desiredStrength))
+            return this;
+        else
+            if (desiredStrength.equals(PasswordStrength.Strong))
+                elementCache().strongButton.click();
+            else
+                elementCache().weakButton.click();
```
labkey-tchad: This `if/else` is hard to read without brackets. Removing the nesting makes it more readable to me.
```
if (getPasswordStrength().equals(desiredStrength))
    return this;
else if (desiredStrength.equals(PasswordStrength.Strong))
    elementCache().strongButton.click();
else
    elementCache().weakButton.click();
```

src/org/labkey/test/pages/core/login/DatabaseAuthConfigureDialog.java
```diff
@@ -0,0 +1,165 @@
...
+            if (oldStrength == null)
+                oldStrength = curStrength;
+            setPasswordStrength(newStrength);
+        }
+
+        if ( newExpiration != null && curExpiration != newExpiration)
+        {
+            if (oldExpiration == null)
+                oldExpiration = curExpiration;
+            setPasswordExpiration(newExpiration);
```
labkey-tchad: Remember the old settings in the individual setters or make them private. If we want to have the ability to reset we need to make it so people can't easily bypass this logic.

src/org/labkey/test/pages/core/login/DatabaseAuthenticationProvider.java
```diff
@@ -0,0 +1,31 @@
+package org.labkey.test.pages.core.login;
```
labkey-tchad: This isn't a page. It should go in `org.labkey.test.params.login`, next to the base AuthenticationProvider.

src/org/labkey/test/TestScrubber.java
```diff
@@ -85,16 +88,19 @@ public void cleanSiteSettings()
...
         }
 
         try
-        {
-            disableSecondaryAuthentication();
+        {   // disable any/all secondary auth configurations
+            LoginConfigurePage.beginAt(this).getSecondaryConfigurations()
+                .stream().filter(a -> a.getProvider().equals("TestSecondary"))
+                    .collect(Collectors.toList())
+                    .forEach(LoginConfigRow::clickDelete);
```
labkey-tchad: Adam mentioned deleting the secondary auth configurations via API. Did that not pan out?

src/org/labkey/test/pages/core/login/LoginConfigurePage.java
```diff
@@ -20,61 +23,130 @@ public LoginConfigurePage(WebDriver driver)
...
+    {
+        LoginConfigRow row = new LoginConfigRow.LoginConfigRowFinder(getDriver()).withDescription(description).waitFor(getDriver());
+        return row.clickEdit(authenticationProvider);
     }
 
-    public ShowAdminPage clickDone()
+    public ShowAdminPage clickSaveAndFinish()
     {
-        clickAndWait(elementCache().doneBtn());
+        shortWait().until(ExpectedConditions.elementToBeClickable(elementCache().saveAndFinishBtn()));
```
labkey-tchad, line 148: Does the save button get disabled in a way that this will notice?

src/org/labkey/test/tests/SecurityTest.java
```diff
@@ -223,7 +224,10 @@ public void assertUrlForbidden(String url)
     @LogMethod
     public void passwordResetTest()
     {
-        ConfigureDbLoginPage.beginAt(this).setDbLoginConfig(PasswordStrength.Weak, null);
+        LoginConfigurePage.beginAt(this)
+                .getPrimaryConfigurationRow("Standard database authentication")
+                .clickEdit(new DatabaseAuthenticationProvider())
+                .setDbLoginConfig(DatabaseAuthConfigureDialog.PasswordStrength.Weak.Weak, null);
```
labkey-tchad: Didn't even know you could do that.
```suggestion
                .setDbLoginConfig(DatabaseAuthConfigureDialog.PasswordStrength.Weak, null);
```

labkey-tchad (changes requested):

src/org/labkey/test/pages/core/admin/ShowAdminPage.java
labkey-chrisj: Good catch.  Changing it to no longer use the now-absent 'authentication (deprecated)' link, to use the one that does exist instead

src/org/labkey/test/pages/core/login/AuthDialogBase.java
labkey-chrisj: The only downside to this is that it means you don't know the button's text by just reading test code.  So, readability downside, but maintainability win

src/org/labkey/test/pages/core/login/DatabaseAuthConfigureDialog.java
labkey-chrisj: Okay.  I've split out the logic in that function and put the caching in the clickApply() override for this dialog

src/org/labkey/test/pages/core/login/LoginConfigurePage.java
labkey-chrisj: not 100% sure.
Added this on the chance that it might be helpful; we may want to come up with a consistent means of knowing if a button is in a disabled state

src/org/labkey/test/TestScrubber.java
labkey-chrisj: We did implement that via API for the particular case of deleting a secondary login config discovered during login; because that one occurs on a URL containing the config ID it was trivial

For the nonce this seemed easier to do via the UI

labkey-tchad (approved):

PR #201, opened 2020-01-15
Item 6630: test changes for the new designer

labkey-ankurj:


src/org/labkey/test/params/FieldDefinition.java
```diff
@@ -302,7 +302,8 @@ public String toString()
     {
         MultiLine("Multi-Line Text", "Multi-Line Text", "string"),
         Integer("Integer", "Integer", "int"),
-        String("Text", "Text (String)", "string"),
+        string("Text", "Text (String)", "string"),
+        String("Text", "Text (String)", "String"), // need upper case String for lookup tableType text
```
cnathe: I would rather not have the duplicate field type definitions here which just differ in text casing. Can we try editing the DomainFieldRow.setLookup so that we change line 383 to `tableType.toUpperCase()` instead?

cnathe (changes requested):

src/org/labkey/test/params/FieldDefinition.java
```diff
@@ -302,7 +302,7 @@ public String toString()
     {
         MultiLine("Multi-Line Text", "Multi-Line Text", "string"),
         Integer("Integer", "Integer", "int"),
-        String("Text", "Text (String)", "string"),
+        string("Text", "Text (String)", "string"),
```
cnathe: Can you put this back to the way it was now that we don't need both types?

cnathe (approved):

src/org/labkey/test/components/domain/DomainFieldRow.java
```diff
@@ -380,7 +380,7 @@ public DomainFieldRow setLookup(FieldDefinition.LookupInfo lookupInfo)
         String tableType = lookupInfo.getTableType();
         if (tableType == null)
             throw new IllegalArgumentException("No lookup type specified for " + lookupInfo.getTable());
-        setFromTargetTable(lookupInfo.getTable() + " (" + tableType + ")");
+        setFromTargetTable(lookupInfo.getTable() + " (" + tableType.toUpperCase() + ")");
```
cnathe: @labkey-ankurj on second thought I don't think toUpperCase() is what we want here. I just saw the TeamCity failure fo rFileContentUploadTest.testFileBrowser on this feature branch. We want to convert tableType from lower case to camel case (i.e. "string" -> "String") here.

cnathe (changes requested):

src/org/labkey/test/components/domain/DomainFieldRow.java
labkey-ankurj: done

cnathe (approved):

PR #204, opened 2020-01-22
Allow 'suite' parameter to specify a subset of an existing suite

labkey-tchad:
Got inspired to try this out last Friday. We might account for expected test execution time (based on timeout annotation) but that could be a future refinement.

With this change we could get rid of the DailyA/B/C suites and use `Daily[1/3]`, `Daily[2/3]`, `Daily[3/3]` or split up any suite to run segments in parallel.

### DIFF
--- a/src/org/labkey/test/SuiteBuilder.java
+++ b/src/org/labkey/test/SuiteBuilder.java
@@ -35,11 +35,14 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public class SuiteBuilder
 {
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/SuiteBuilder.java
```diff
@@ -176,14 +179,54 @@ public TestSet getTestSet(String suiteName)
...
+        List<Class> sorted = new ArrayList<>(tests);
+        sorted.sort(Comparator.comparing(Class::getName));
+
+        int size = sorted.size();
+        int index = subset - 1;
+        int subsetSize = size / subsetCount;
+        int remainder = size % subsetCount;
+        int isLargeSubset = index < remainder ? 1 : 0;
+        int fromIndex = subsetSize * index + Math.min(index, remainder);
+        int toIndex = Math.min(fromIndex + subsetSize + isLargeSubset, size);
```
labkey-danield, line 225: When I do this math with a size of 100 and subsetCount = 3, the first batch is 0 to 34, the second is 34 to 67 and the third is 67 to 100.
In this particular case two test cases would be run twice (if my math is right). 
I don't think this is a big deal just something we should be aware of.

Otherwise I think this is fine.

labkey-danield (approved):

src/org/labkey/test/SuiteBuilder.java
labkey-tchad: Your math is correct but `List.subList` doesn't include the element at `toIndex`.

PR #206, opened 2020-01-23
Sample Comparison Report Testing

labkey-tchad:
Add functionality to assay and sample set API helpers


src/org/labkey/test/Locators.java
```diff
@@ -21,7 +21,6 @@
     public static final Locator.IdLocator folderMenu = Locator.id("folderBar");
     public static final Locator.XPathLocator labkeyError = Locator.byClass("labkey-error");
     public static final Locator.XPathLocator labkeyMessage = Locator.byClass("labkey-message");
-    public static final Locator.XPathLocator alertWarning = Locator.byClass("alert").withClass("alert-warning");
```
labkey-danield, line 24: Just curious why this is being removed.

labkey-danield (commented):
Looks ok to me.

src/org/labkey/test/Locators.java
labkey-tchad: It was unused and is probably redundant with `BootstrapLocators.warningBanner`

labkey-chrisj (approved):
this looks good.  Good call using createSampleSetApi instead of createDomain for createEmptySampleSet()

labkey-danield (approved):
I'm ok with this, but it might be worth while to run this branch in the Sample Management Sample and SampleSet suites:
https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_LabkeyPremiumTrunk_GitModules_SampleManagementSamplesAndSampleSetsPostgres?branch=
https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_LabkeyPremiumTrunk_GitModules_SampleManagementSamplesAndSampleSetsSqlserver?branch=
The API helper is used in those tests.

PR #208, opened 2020-01-27
Item 6806: Domain designer test fixes for row deletion button/icon changes

cnathe:


src/org/labkey/test/components/domain/DomainFieldRow.java
```diff
@@ -138,24 +138,37 @@ public boolean isExpanded()
...
                 break;
-            }catch (NoSuchElementException notFound) {}
+            }
+            catch (NoSuchElementException notFound) {}
         }
 
         ModalDialog confirmDeletionDlg = new ModalDialog.ModalDialogFinder(getDriver())
-                .withTitle("Confirm Field Deletion").find();
+                .withTitle("Confirm Remove Field").find();
         return confirmDeletionDlg;
```
labkey-danield: I realize you didn't write this code, but I'm wondering if this function couldn't be simplified to this:

`
        elementCache().removeField.click();

        if (!confirmDialogExpected)
            return null;

       return  new ModalDialog.ModalDialogFinder(getDriver())
                    .withTitle("Confirm Remove Field").timeout(3_000).waitFor();

`
 The looping/retry when using a .waitFor just feel a little odd.
But maybe I don't understand the oddities of the ModalDialogFinder.

labkey-danield: Or, now that I have looked at how this is used,  what if the function was like this:

   public void clickRemoveField(String dlgButtonToClick)
    {
        getWrapper().mouseOver(elementCache().removeFieldBtn);

        elementCache().removeField.click();

        if (dlgButtonToClick != null && !dlgButtonToClick.isEmpty())
        {
            new ModalDialog.ModalDialogFinder(getDriver())
                    .withTitle("Confirm Remove Field").timeout(3_000).waitFor()
                    .dismiss(dlgButtonToClick);
        }
    }        

This way you don't have to pass the dialog back etc...
But using a string as a conditional parameter is a little weird as well, and isn't really intuitive.

And again I don't know what issues the original author hit that caused them to use a loop/retry model.

labkey-danield (approved):
Looks ok as is, just added a comment on a possible change to the RemoveField function.

src/org/labkey/test/components/domain/DomainFieldRow.java
cnathe: Thanks for the review Dan. I agree that it seems odd to have a retry loop over a waitFor. I'll try removing it and running the relevant tests on TeamCity to see what they think.

labkey-danield (approved):

PR #213, opened 2020-02-06
Item 6828: Test SM search UI

labkey-tchad:


src/org/labkey/test/util/BaseSearchHelper.java
```diff
@@ -0,0 +1,305 @@
...
+    @LogMethod
+    public void verifySearchResults(@LoggedParam String expectedResultsContainer, @LoggedParam @NotNull String baseScreenshotName)
+    {
+        // Note: adding this "waitForIndexer()" call should eliminate the need for sleep() and retry below.
+        SearchAdminAPIHelper.waitForIndexer();
+
+        verifySearchResults(expectedResultsContainer, baseScreenshotName, _searchResultsQueue, maxTries);
+    }
+
+    protected abstract void beforeVerify(String containerPath);
```
labkey-danield, line 104: Purely subjective comment:
This name, beforeVerify, gave me pause for a moment. 
Would a better name be doBeforeVerify?
Or am I misunderstanding what the intent is?

src/org/labkey/test/util/BaseSearchHelper.java
```diff
@@ -0,0 +1,305 @@
...
+
+            for (Locator loc : expectedResults)
+            {
+                if (!resultsPage.hasResultLocatedBy(loc))
+                {
+                    missingResults.add(loc);
+                    if (!failOnError)
+                    {
+                        // Stop checking for search results if we don't need them for a failure message
+                        break;
```
labkey-danield, line 168: Won't this stop on the first missing result?
Is there any value in seeing if there are any other missing results?

src/org/labkey/test/util/BaseSearchHelper.java
```diff
@@ -0,0 +1,305 @@
...
+            List<Locator> wrongResults = new ArrayList<>();
+            for (Locator loc : item.getUnwantedResults())
+            {
+                if (resultsPage.hasResultLocatedBy(loc))
+                {
+                    wrongResults.add(loc);
+                    if (!failOnError)
+                    {
+                        // Stop checking for search results if we don't need them for a failure message
+                        break;
```
labkey-danield, line 182: Same comment as above, is there any value in finding all of the incorrect results?

src/org/labkey/test/util/BaseSearchHelper.java
```diff
@@ -0,0 +1,305 @@
...
+
+    public void assertNoSearchResult(String searchTerm)
+    {
+        long startTime = System.currentTimeMillis();
+        List<WebElement> results;
+
+        do {
+            SearchResults searchResultsPage = searchFor(searchTerm);
+            results = searchResultsPage.getResults();
+        } while (System.currentTimeMillis() - startTime < BaseWebDriverTest.WAIT_FOR_PAGE && !results.isEmpty());
```
labkey-danield, line 260: I realize sleep is not ideal, but might this be a case for it? Just sleep for a period of time, then do the check, rather than have a tight loop like this.

labkey-danield (approved):

src/org/labkey/test/util/BaseSearchHelper.java
labkey-tchad: On the last retry (`failOnError=true`) it will collect all of the missing results.

PR #214, opened 2020-02-06
add support for setting experimental feature flags during test run

labkey-kevink:


src/org/labkey/test/BaseWebDriverTest.java
```diff
@@ -356,6 +356,8 @@ public URL getLastPageURL()
     private static long testCount;
     private static int currentTestNumber;
 
+    private static Map<String, Boolean> previousExperimentalFlags = Collections.emptyMap();
+
```
labkey-tchad: Have `ExperimentalFeaturesHelper` keep track of previous experimental feature settings. That way, one-off calls to enable/disable experimental features can be included in the reset.
See how we set and reset `PipelineToolsHelper._originalToolsDirs`.

labkey-tchad (changes requested):

src/org/labkey/test/util/ExperimentalFeaturesHelper.java
```diff
@@ -15,114 +15,103 @@
...
+            if (savePrevious && response.containsKey("previouslyEnabled"))
             {
-                Map<String, Boolean> data = (Map<String, Boolean>)response.get("data");
-                return data;
+                // Remember the previous setting for the feature flag if was different
+                Boolean previouslyEnabled = (Boolean)response.get("previouslyEnabled");
+                if (!Objects.equals(enable, previouslyEnabled.booleanValue()))
+                {
+                    TestLogger.log("Experimental feature will be reset back to " + (previouslyEnabled ? "enabled" : "disabled") + " after test is completed");
+                    _originalFeatureFlags.put(feature, previouslyEnabled.booleanValue());
```
labkey-tchad: Remove the `savePrevious` flag and always save the previous value if there isn't an existing saved value. Then clear out `_originalFeatureFlags` at the end of `resetFeatures`.
That way, enabling/disabling a feature multiple times won't clobber the initial setting.

labkey-tchad (changes requested):

labkey-tchad (approved):

PR #223, opened 2020-02-12
Allow duplicate sampledata

labkey-tchad:
This will ease Git migration as we move sampledata out of SVN

### DIFF
--- a/src/org/labkey/test/TestFileUtils.java
+++ b/src/org/labkey/test/TestFileUtils.java
@@ -212,7 +212,7 @@ public abstract class TestFileUtils
             File checkFile = new File(sampledataDir, relativePath);
             if (checkFile.exists())
             {
-                if (foundFile != null && !foundFile.equals(checkFile))
+                if (foundFile != null && foundFile.length() != checkFile.length()) // Allow duplicate files to ease migration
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/TestFileUtils.java
```diff
@@ -212,7 +212,7 @@ public static File getSampleData(String relativePath)
             File checkFile = new File(sampledataDir, relativePath);
             if (checkFile.exists())
             {
-                if (foundFile != null && !foundFile.equals(checkFile))
+                if (foundFile != null && foundFile.length() != checkFile.length()) // Allow duplicate files to ease migration
```
labkey-danield, line 215: Would a more appropriate check be to compare the hash of the file?
    if(foundFile != null && foundFile.hash() != checkFile.hash()))

labkey-tchad: `File` doesn't have any built in hashing functionality.
This isn't meant to be a rock-solid verification that they're the same file. Them probably being the same file is sufficient. We'll go back to being strict once we've gotten all of the sample data out of SVN.

labkey-danield (approved):

PR #232, opened 2020-02-17
Fb update admin console test new ui

labkey-chrisj:
This adds ability for AuthConfigRow and LoginConfigurePage to safely check to see if edit config links are shown, or if add configuration menus are present.
This updates the AppAdminTest to use the new UI to ensure that when logged in as AppAdmin, controls to edit authentication configs are correctly denied

...it also reveals that if the AppAdmin clicks 'Save and Finish' on the LoginConfigurePage, we'll get an error alert- possibly a product issue


src/org/labkey/test/pages/core/login/LoginConfigurePage.java
```diff
@@ -36,6 +36,12 @@ public static LoginConfigurePage beginAt(WebDriverWrapper webDriverWrapper)
         return authenticationProvider.getNewDialog(getDriver());
     }
 
+    public boolean canAddConfiguration()
+    {
+        togglePrimaryConfiguration();
+        return new MultiMenu.MultiMenuFinder(getDriver()).withText("Add New Primary Configuration").findOptional(getDriver()).isPresent();
```
labkey-tchad: Use the element cache version. With this technique, the button text could change and if the person doing the update didn't notice this method, the app admin check would become invalid because it would be validating the non-existence of a thing that doesn't ever exist.
```
try
{
    return elementCache().addPrimaryMenu.getComponentElement().isEnabled();
}
{
    return false;
}
```

src/org/labkey/test/pages/core/login/LoginConfigurePage.java
```diff
@@ -46,6 +52,12 @@ public static LoginConfigurePage beginAt(WebDriverWrapper webDriverWrapper)
         return authenticationProvider.getNewDialog(getDriver());
     }
 
+    public boolean canAddSecondaryConfiguration()
+    {
+        toggleSecondaryConfiguration();
+        return new MultiMenu.MultiMenuFinder(getDriver()).withText("Add New Secondary Configuration").findOptional(getDriver()).isPresent();
```
labkey-tchad: Use element cache. (See comment for `canAddConfiguration`)

src/org/labkey/test/tests/AdminConsoleTest.java
```diff
@@ -157,10 +160,11 @@ public void testAppAdminRole()
...
         //authentication
-        goToAdminConsole().clickAuthentication();
-        assertElementNotPresent("expect 'enable' links to be disabled for appAdmin", Locator.linkWithText("enable"));
-        assertElementNotPresent("expect 'configure' links to be disabled for appAdmin", Locator.linkWithText("configure"));
-        clickButton("Done");
+        LoginConfigurePage configurePage = goToAdminConsole().clickAuthentication();
+        List<LoginConfigRow> configRows = configurePage.getPrimaryConfigurations();
+        assertFalse("expect 'edit' links not to be available for auth configs", configRows.stream().anyMatch(a-> a.canEdit()));
+        assertFalse("expect 'add configuration' menu to be absent for AppAdmin", configurePage.canAddConfiguration());
+        configurePage.clickSaveAndFinish();
```
labkey-tchad: Wait until the product fix is in. This button might become disabled for app admins or something else that would require a followup test fix.

src/org/labkey/test/tests/AdminConsoleTest.java
```diff
@@ -157,10 +160,11 @@ public void testAppAdminRole()
         assertElementPresent("expect to return to admin console", siteAdminLoc, 1);
 
         //authentication
-        goToAdminConsole().clickAuthentication();
-        assertElementNotPresent("expect 'enable' links to be disabled for appAdmin", Locator.linkWithText("enable"));
-        assertElementNotPresent("expect 'configure' links to be disabled for appAdmin", Locator.linkWithText("configure"));
-        clickButton("Done");
+        LoginConfigurePage configurePage = goToAdminConsole().clickAuthentication();
+        List<LoginConfigRow> configRows = configurePage.getPrimaryConfigurations();
+        assertFalse("expect 'edit' links not to be available for auth configs", configRows.stream().anyMatch(a-> a.canEdit()));
```
labkey-tchad, line 165: We should perform the inverse of this check as a site admin to make sure it doesn't just always return false. Either because there aren't any configs to edit or because the Locator in `canEdit` becomes invalid.

labkey-tchad (changes requested):
Asserting the absence of things is fragile. Such a test can become invalid if the page design changes.

src/org/labkey/test/pages/core/login/LoginConfigurePage.java
labkey-chrisj: I've refactored these such that the canAddConfiguration... methods use the same MenuFinder as the cached menu objects; that way we find the menus in regular testing and check for their non-presence with the same thing

src/org/labkey/test/tests/AdminConsoleTest.java
labkey-chrisj: I've pulled the locator out into an elementCache member; now canEdit() references the same locator as the element used in the cache every time we edit a configuration

labkey-tchad (approved):

PR #233, opened 2020-02-19
Create jar using default configuration not compile configuration in o…

labkey-susanh:
…rder to pick up the dependent library classes as well.


labkey-tchad (approved):
The jars look like what I get in 19.3

labkey-susanh:
Personal run on TeamCity including ProgrammaticQCTest succeeded.

PR #234, opened 2020-02-20
Fb troubleshooter 39638

labkey-sweta:
Automation test

PS : Section of test is commented which verifies troubleshooter cannot save changes in authentication until the issue is fixed by rosaline.

### DIFF
--- a/src/org/labkey/test/tests/core/security/TroubleshooterRoleTest.java
+++ b/src/org/labkey/test/tests/core/security/TroubleshooterRoleTest.java
@@ -1,0 +1,94 @@
+package org.labkey.test.tests.core.security;
+
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.labkey.test.BaseWebDriverTest;
+import org.labkey.test.Locator;
+import org.labkey.test.TestTimeoutException;
+import org.labkey.test.categories.DailyC;
+import org.labkey.test.pages.core.admin.ShowAuditLogPage;
+import org.labkey.test.util.ApiPermissionsHelper;
+
+import java.io.File;
+import java.util.Arrays;
+import java.util.List;
+
+@Category({DailyC.class})
+public class TroubleshooterRoleTest extends BaseWebDriverTest
+{
+    private static final String TROUBLESHOOTER = "user@troubleshooter.test";
+
+    @BeforeClass
+    public static void setupProject()
+    {
+        TroubleshooterRoleTest init = (TroubleshooterRoleTest) getCurrentTest();
+        init.doSetup();
+    }
+
+    @Override
+    protected void doCleanup(boolean afterTest) throws TestTimeoutException
+    {
+        _userHelper.deleteUsers(false, TROUBLESHOOTER);
+    }
+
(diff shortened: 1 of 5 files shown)


labkey-adam:
Rosaline committed her auth save fix to platform yesterday and I verified it. I just merged develop into platform/fb_troubleshooter_39638, so should be able to uncomment that test.

src/org/labkey/test/tests/core/security/TroubleshooterRoleTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+        log("Verify External Redirect Hosts changes cannot be saved");
+        goToAdminConsole().goToSettingsSection().clickExternalRedirectHosts();
+        checker().verifyFalse("Save should not be present for troubleshooter under External Redirect Hosts",
+                isElementPresent(Locator.button("Save")));
+
+//        log("Verify authentication changes cannot be saved");
+//        goToAdminConsole().goToSettingsSection().clickAuthentication();
+//        checker().verifyFalse("Save should not be present for troubleshooter under authentication",
+//                isElementPresent(Locator.button("Save and Finish")));
```
labkey-tchad: Validating the absence of something can be dangerous. If the button text for any of these page changes, these checks would pass regardless of troubleshooter access.
Put these `verifyFalse` checks in a separate method (`verifySiteSettingsPermissions(boolean canSave)`) and call it once as a site admin to verify that we're checking the correct buttons, then once as troubleshooter to verify that they aren't present.

labkey-tchad (changes requested):
Looks like you branched off of a different feature branch instead of `develop`. Karl's assay test changes shouldn't be in here.

src/org/labkey/test/tests/core/security/TroubleshooterRoleTest.java
```diff
@@ -0,0 +1,104 @@
...
+    protected void doCleanup(boolean afterTest) throws TestTimeoutException
+    {
+        _userHelper.deleteUsers(false, TROUBLESHOOTER);
+    }
+
+    private void doSetup()
+    {
+        _userHelper.createUser(TROUBLESHOOTER);
+        ApiPermissionsHelper apiPermissionsHelper = new ApiPermissionsHelper(this);
+        goToSitePermissions();
```
labkey-tchad: Shouldn't need to navigate if you're using the API helper

src/org/labkey/test/tests/core/security/TroubleshooterRoleTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+import java.io.File;
+import java.util.Arrays;
+import java.util.List;
+
+@Category({DailyC.class})
+@BaseWebDriverTest.ClassTimeout(minutes = 4)
+public class TroubleshooterRoleTest extends BaseWebDriverTest
+{
+    private static final String TROUBLESHOOTER = "user@troubleshooter.test";
```
labkey-tchad: Avoid display name collision
```suggestion
    private static final String TROUBLESHOOTER = "troubleshooter@troubleshooter.test";
```

labkey-tchad (approved):

PR #236, opened 2020-02-21
Fb provenance runbuilder 6516

labkey-chrisj:
This PR adds components that were written to support provenance RunBuilder testing, but which belong in testAutomation


src/org/labkey/test/components/glassLibrary/components/RbtCombo.java
```diff
@@ -0,0 +1,133 @@
...
+import org.labkey.test.WebDriverWrapper;
+import org.labkey.test.components.Component;
+import org.labkey.test.components.WebDriverComponent;
+import org.labkey.test.selenium.EphemeralWebElement;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+import java.util.List;
+
+public class RbtCombo extends WebDriverComponent<RbtCombo.ElementCache>
```
labkey-tchad: I don't think this is a LabKey developed component. Also, this name doesn't help people find it or know what it is for. Looking at the product code, I think this is a [`react-bootstrap-typeahead`](http://ericgio.github.io/react-bootstrap-typeahead/) component.
Move/rename this to `org.labkey.test.components.react.Typeahead` and add a class javadoc indicating what it represents. (Here's where the component is imported in the run builder: https://github.com/LabKey/provenance/blob/0e9c1e3e8046e16eb600bc8079dc160ff2238b6b/src/client/RunBuilder/PropertyLookup.tsx#L3)

@labkey-martyp Can you confirm?

src/org/labkey/test/components/glassLibrary/components/RbtCombo.java
```diff
@@ -0,0 +1,133 @@
...
+    public boolean hasValue()
+    {
+        return !getValue().isEmpty();
+    }
+
+    // open
+    public boolean isOpened()
+    {
+        return elementCache().input.getAttribute("aria-expanded").equals("true");
+    }
```
labkey-tchad: Be mindful of the API surface of components, especially in core test components.
Make this `private` or `protected`

labkey-tchad (changes requested):

src/org/labkey/test/components/glassLibrary/components/RbtCombo.java
labkey-chrisj: I'd considered the cost/benefit of exposing it or not, then I figured it didn't hurt to let the test have access to it/seemed silly to have the functionality but not let a test access it

src/org/labkey/test/components/glassLibrary/CollapsiblePanel.java
```diff
@@ -0,0 +1,119 @@
...
+            elementCache().expandCollapseToggle.click();
+        WebDriverWrapper.waitFor(()-> !isExpanded(),
+                "the panel did not collapse", 2000);
+        return this;
+    }
+
+    public boolean isExpanded()
+    {
+        return elementCache().waitForBody()
+                .getAttribute("class").equals("panel-collapse collapse in");
```
labkey-danield: I don't think that is the best way to do this check. If the body is missing this will throw. Also It is dependent upon order of the attributes in the class.

I think it would be better to check for the existence of the  <div class="panel-body"> element in the panel. If it is not there the panel is not expanded.

src/org/labkey/test/components/glassLibrary/CollapsiblePanel.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+
+    protected class ElementCache extends Component<?>.ElementCache
+    {
+        WebElement header = Locator.tagWithClass("div", "collapsible-panel-header")
+                .findWhenNeeded(this).withTimeout(2000);
+        WebElement expandCollapseToggle = Locator.tagWithClass("span", "pull-right")
+                .findWhenNeeded(header).withTimeout(2000);
+        private Locator.XPathLocator body = Locator.tagWithClass("div", "panel-collapse");
```
labkey-danield, line 76: I don't think the body tag is correct the body is the <div class="panel-body"> tag contained in the domain-form-panel and the panel-collapse div:
<div class="domain-form-panel">
    <div class="panel-collapse">
        <div class="panel-body">
What you have will work, but you may want to avoid confusion in the future and reference the div tag that has the class value of panel-body.

src/org/labkey/test/components/glassLibrary/CollapsiblePanel.java
```diff
@@ -0,0 +1,119 @@
...
+
+        WebElement waitForBody()
+        {
+            return body.waitForElement(this, WAIT_FOR_JAVASCRIPT);
```
labkey-danield, line 82: I know @labkey-tchad suggested you wait for the element in the cache, I think we should talk as a team and come up with a standard on when to wait for elements.

src/org/labkey/test/components/glassLibrary/CollapsiblePanel.java
labkey-danield, line 76: I don't think the body tag is correct the body is the <div class="panel-body"> tag contained in the domain-form-panel and the panel-collapse div:
<div class="domain-form-panel">
    <div class="panel-collapse">
        <div class="panel-body">
What you have will work, but you may want to avoid confusion in the future and reference the div tag that has the class value of panel-body.

labkey-tchad (approved):
Provenance tests passed.
https://teamcity.labkey.org/viewLog.html?buildId=923770

src/org/labkey/test/components/glassLibrary/CollapsiblePanel.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
```
labkey-danield, line 84: A couple other elements that would be useful for a panel:
Panels can have an error message that appears right below them. Things like missing values in required fields can cause them. It would be useful if we could get the error message associated with the panel.
Second the the icon on the left hand side changes, it would be useful tog et that to validate that the expected icon is shown.
Third a panel can have text on the right hand side next to the expand collapse button. For example this gives the status of the fields in the panel. it would be useful to grab that text as well.

src/org/labkey/test/components/glassLibrary/components/FileUploadPanel.java
```diff
@@ -0,0 +1,128 @@
...
+
+    protected class ElementCache extends Component<?>.ElementCache
+    {
+        Input input()
+        {
+            return Input(Locator.id("fileUpload"), getDriver()).waitFor(this);
+        }
+
+        Locator.XPathLocator attachedFileContainer = Locator.tagWithClass("div", "attached-file--container")
+                .withChild(Locator.tagWithClass("span", "fa-times-circle"));
```
labkey-danield, line 91: Just curious. Is the model now to put locators in the ElementCache and not in a Locators class?

src/org/labkey/test/components/glassLibrary/components/react/typeahead/TypeAheadCombo.java
```diff
@@ -0,0 +1,136 @@
...
+
+    public TypeAheadCombo setValue(String value)
+    {
+        if (hasValue())
+            clear();
+        elementCache().input.sendKeys( value);
+
+        elementCache().option(value).click();
+        WebDriverWrapper.waitFor(()-> !isOpened() && getValue().equals(value),
+                "the value was not set, or the combo did not close", 2000);
```
labkey-danield, line 47: I like that you check that the value was set in the control.

src/org/labkey/test/components/glassLibrary/components/react/typeahead/TypeAheadCombo.java
```diff
@@ -0,0 +1,136 @@
...
+    public TypeAheadCombo clear()
+    {
+        if (hasValue())
+            elementCache().clearBtn.click();
+
+        WebDriverWrapper.waitFor(()-> !hasValue(), "the value was not cleared in time", 2000);
+        return this;
+    }
+
+
```
labkey-danield, line 76: Is there a way to get the values that are in the drop down list of the combo? A security test for example may need to check that only specific values are shown in a list/combo control to users with limited permissions.

labkey-danield (changes requested):
Unfortunately I only had time to review a couple of the files.
I would suggest that you put a try catch around the isExpanded check for the panel. It will throw if the body is not there.
All the other comments/suggestions can wait until later.

src/org/labkey/test/components/glassLibrary/components/FileUploadPanel.java
labkey-chrisj: I think yes- locators can go in either, which sorta makes the static Locators class redundant
But, I don't have strong feelings either way

src/org/labkey/test/components/glassLibrary/components/react/typeahead/TypeAheadCombo.java
labkey-chrisj: They will appear as rendered dom elements in the flyout; should be relatively easy to implement

src/org/labkey/test/components/glassLibrary/CollapsiblePanel.java
labkey-chrisj: I used the panel-collapse element (parent to the panel-body element) for 2 reasons: 
1) its class is used to signal its expand/collapse state
2) the use case I'm looking at puts provenance/runbuilder-specific classes in the panel-body div, I decided not to assume that it would exist in other usage

If we want to make that assumption, okay

src/org/labkey/test/components/glassLibrary/CollapsiblePanel.java
labkey-chrisj: If the body isn't there, the question of whether or not it is expanded is sorta meaningless

I will do this check by looking at the expand/collapse toggle (showing a minus-square means it's expanded, showing a plus-square means it's collapsed), but any way we look at it, we will throw if there's no body element

labkey-danield (approved):

PR #237, opened 2020-02-24
Epic 6798 : List Domain Designer Conversion

cnathe:


src/org/labkey/test/util/ListHelper.java
```diff
@@ -57,6 +59,25 @@ public ListHelper(WebDriver driver)
         this(() -> driver);
     }
 
+    private boolean NEW_LIST_DESIGNER_ENABLED = false;
+    private void toggleNewListDesigner(boolean enabled)
+    {
+        //if (enabled)
```
labkey-chrisj, line 65: out of curiosity, why comment the codepath to disable the new designer?

cnathe: Because I didn't want to affect anything on TeamCity yet and have any of the test run with the new UI until we are ready. I had those helpers in there and would uncomment them to test things out, but since there are a lot of tests that go through those helpers, leaving it on would break the ones I hadn't had a chance to look at yet.

cnathe:
@labkey-chrisj I guess another option if we aren't comfortable merging these changes yet would be to pull out the necessary fixes for the tests that will fail, because of the minor domain details text change from "New field" -> "New Field", into a separate branch and get that merged to develop.

labkey-chrisj (approved):

PR #242, opened 2020-02-26
Fb plate support 6715

labkey-sweta:


src/org/labkey/test/AssayAPITest.java
```diff
@@ -550,4 +504,30 @@ public void testGpatImportRunWithPlate() throws Exception
 
         verifyPlateMetadata(resp, expectedRows);
     }
+
+    public void createPlateTemplate()
```
labkey-klum, line 508: Did you mean to make this public and leave it in this class? It looks like it is in the APIAssayHelper class too.

src/org/labkey/test/tests/GpatPlateTemplateTest.java
```diff
@@ -0,0 +1,167 @@
...
+    private void doSetup()
+    {
+        _containerHelper.createProject(getProjectName(), "Assay");
+        setPipelineRoot(TestFileUtils.getSampleData("GPAT").getAbsolutePath(), false);
+
+        goToProjectHome();
+        APIAssayHelper assayHelper = new APIAssayHelper(this);
+        assayHelper.createAssayWithPlateSupport(ASSAY_NAME);
+        assayHelper.createPlateTemplate(templateName, "blank", "GPAT (General)");
+
```
labkey-klum: delete this line

labkey-klum: I meant there is an unnecessary blank line at the bottom of this method.

src/org/labkey/test/tests/GpatPlateTemplateTest.java
```diff
@@ -0,0 +1,167 @@
...
+        checker().verifyEquals("Well Location is incorrect", Arrays.asList("A1", "A2", "A3", "A4", "A5", "A6", "A7", "A8", "A9", "A10"),
+                table.getColumnDataAsText("WellLocation"));
+        checker().verifyEquals("Sample well group is incorrect", Arrays.asList("SA01", "SA01", "SA01", "SA02", "SA02", "SA02", "SA03", "SA03", "SA04", "SA04"),
+                table.getColumnDataAsText("PlateData/sample_well_groups"));
+        checker().verifyEquals("Barcode is incorrect", Arrays.asList("BC_111", "BC_111", "BC_111", "BC_222", "BC_222", "BC_222", "BC_333", "BC_333", "BC_444", "BC_444"),
+                table.getColumnDataAsText("PlateData/Barcode"));
+        checker().verifyEquals("Dilution is incorrect", Arrays.asList("1.0", "1.0", "1.0", "2.0", "2.0", "2.0", "3.0", "3.0", "4.0", "4.0"),
+                table.getColumnDataAsText(" PlateData/dilution"));
+        table.clearAllFilters();
+
```
labkey-klum: extra whitespace

src/org/labkey/test/tests/GpatPlateTemplateTest.java
```diff
@@ -0,0 +1,167 @@
...
+        importPlateData(assayId, TEST_PLATE_DATA, templateName, null);
+        checker().verifyEquals("Invalid error message for missing plate metadata", "No data file was uploaded. Please select a file.",
+                Locator.byClass("labkey-error").findElement(getDriver()).getText());
+
+        goToProjectHome();
+        clickAndWait(Locator.linkWithText(ASSAY_NAME));
+        importPlateData(assayId, TEST_PLATE_DATA, null, TEST_PLATE_METADATA);
+        checker().verifyEquals("Invalid error message for missing plate template", "PlateTemplate is required and must be of type Text (String).",
+                Locator.byClass("labkey-error").findElement(getDriver()).getText());
+
```
labkey-klum: extra whitespace

src/org/labkey/test/tests/GpatPlateTemplateTest.java
```diff
@@ -0,0 +1,167 @@
...
+        setFormElement(Locator.name("name"), name);
+        checkRadioButton(Locator.radioButtonById("Fileupload"));
+        setFormElement(Locator.input("__primaryFile__"), plateData);
+        if (templateName != null)
+            selectOptionByText(Locator.name("plateTemplate"), templateName);
+        if (plateMetadata != null)
+            setFormElement(Locator.name("__plateMetadataFile__"), TEST_PLATE_METADATA);
+
+        clickAndWait(Locator.tagWithText("span", "Save and Finish"));
+
```
labkey-klum, line 162: extra whitespace

src/org/labkey/test/tests/GpatPlateTemplateTest.java
```diff
@@ -0,0 +1,167 @@
...
+import java.io.File;
+import java.util.Arrays;
+import java.util.List;
+
+@Category({Assays.class, DailyB.class})
+@BaseWebDriverTest.ClassTimeout(minutes = 7)
+public class GpatPlateTemplateTest extends BaseWebDriverTest
+{
+    public static final File TEST_PLATE_DATA = TestFileUtils.getSampleData("GPAT/plateData.xlsx");
+    public static final File TEST_PLATE_METADATA = TestFileUtils.getSampleData("GPAT/plate-metadata-1.json");
```
labkey-klum: these variables can don't need to be referenced outside of the class so can be private

src/org/labkey/test/util/APIAssayHelper.java
```diff
@@ -295,4 +297,49 @@ public Protocol copyAssayDesign(Integer assayId, String containerPath, String as
...
+        SelectRowsCommand selectRowsCmd = new SelectRowsCommand("assay.General", "PlateTemplate");
+        selectRowsCmd.setColumns(List.of("Lsid"));
+
+        SelectRowsResponse resp = selectRowsCmd.execute(_test.createDefaultConnection(false), folderPath);
+
+        return String.valueOf(resp.getRows().get(0).get("Lsid"));
+    }
+
+    public void createPlateTemplate(String templateName, String templateType, String assayType)
+    {
```
labkey-klum: I don't think this method makes sense to move into this helper because it only creates a template with the same shape and has the weird behavior for GPAT template types. What would be better is to be able to parameterize the method so you could pass in size, well group and well location information. We could also consider whether it makes sense to push these into the PlateDesignerPage and have it create and return page objects like PlateWellGroup that in turn allows you to select wells for a particular well group.

labkey-klum (changes requested):
It would be useful to have the test load more than one run with 2 different plate metadata files that each introduce different properties and then make sure that all of the properties get shown in the run results view.

Also please address inline comments.

src/org/labkey/test/AssayAPITest.java
labkey-sweta: As per our discussion each test class is going to have its own createTemplateTest

labkey-klum (approved):

PR #246, opened 2020-02-28
Add Job Sample Automation

labkey-danield:
I'm still working on test cases but the support libs should be good to review.
Several of the changes in the repo are there just to get the story done, some of the modules I changed need to go away and be replaced with a shared component between Biologics and SampleManager and the other parts of the world.

### DIFF
--- a/src/org/labkey/test/components/glassLibrary/grids/GridBar.java
+++ b/src/org/labkey/test/components/glassLibrary/grids/GridBar.java
@@ -32,7 +32,7 @@ public class GridBar extends WebDriverComponent<GridBar.ElementCache>
 
     protected GridBar(WebDriver driver, WebElement container, ResponsiveGrid responsiveGrid)
     {
-        _gridBarElement = Locator.tagWithClass("div", "query-grid-bar").findElement(container);
+        _gridBarElement = Locators.gridBar().findWhenNeeded(container);
         _responsiveGrid = responsiveGrid;  // The responsive grid that is associated with this bar.
         _driver = driver;
     }
--- a/src/org/labkey/test/components/glassLibrary/grids/GridBar.java
+++ b/src/org/labkey/test/components/glassLibrary/grids/GridBar.java
@@ -30,9 +30,9 @@ public class GridBar extends WebDriverComponent<GridBar.ElementCache>
     final private WebDriver _driver;
     final private ResponsiveGrid _responsiveGrid;
 
-    protected GridBar(WebElement buttonBar, ResponsiveGrid responsiveGrid, WebDriver driver)
+    protected GridBar(WebDriver driver, WebElement container, ResponsiveGrid responsiveGrid)
     {
-        _gridBarElement = buttonBar;
+        _gridBarElement = Locator.tagWithClass("div", "query-grid-bar").findElement(container);
         _responsiveGrid = responsiveGrid;  // The responsive grid that is associated with this bar.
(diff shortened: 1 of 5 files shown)


src/org/labkey/test/components/glassLibrary/grids/GridBar.java
```diff
@@ -30,9 +30,9 @@
     final private WebDriver _driver;
     final private ResponsiveGrid _responsiveGrid;
 
-    protected GridBar(WebElement buttonBar, ResponsiveGrid responsiveGrid, WebDriver driver)
+    protected GridBar(WebDriver driver, WebElement container, ResponsiveGrid responsiveGrid)
     {
-        _gridBarElement = buttonBar;
+        _gridBarElement = Locator.tagWithClass("div", "query-grid-bar").findElement(container);
```
labkey-tchad: This Locator is defined below. Also, use `findWhenNeeded` so that this component can be initialized lazily:
```suggestion
        _gridBarElement = Locators.gridBar().findWhenNeeded(container);
```

src/org/labkey/test/components/glassLibrary/grids/QueryGrid.java
```diff
@@ -52,18 +50,8 @@ public WebDriver getDriver()
...
-    public boolean hasGridBar()
-    {
-        return _gridBar.isPresent();
-    }
-
-    public boolean hasGrid()
-    {
-        return _responsiveGrid.isPresent();
+        _responsiveGrid = new ResponsiveGrid.ResponsiveGridFinder(_driver).find(_queryGridPanel);
+        return _responsiveGrid;
```
labkey-tchad, line 57: `_responsiveGrid` was set in the constructor. We should be able to just return it here.
```suggestion
        return _responsiveGrid;
```

labkey-tchad (changes requested):
Where did we come down on renaming `QueryGrid` to `QueryGridPanel`? I think having panel in the name makes it a bit more clear what the component represents. Maybe our decision was to do that refactor on its own in a separate branch to make the code review for this a bit more straightforward?

I think we could simplify and clarify the hierarchy of these components. I would suggest having `GridBar` and `ResponsiveGrid` take only a `QueryGrid` in their constructor. That would be the `container` element, provide a `WebDriver` instance (need to make `QueryGrid.getDriver` public),  and provide the `GridBar` and `ResponsiveGrid` access to each-other via the getters in `QueryGrid`.
If we have a scenario in the future where there is a `GridBar` in a different context, we can pull the functionality into a base class that doesn't require a `QueryGrid`.

src/org/labkey/test/components/glassLibrary/grids/GridBar.java
labkey-danield: Updated.

src/org/labkey/test/components/glassLibrary/grids/QueryGrid.java
labkey-danield: The call to the finder was there originally (so to speak). The only thing that really changed here was to no longer using optionals. 
I remember we discussed this way back and landed on calling the finder when the grid is requested because it covers the scenario of the grid being updated between the constructor and when this function might be called.
An alternative could be to check for staleness and only call the finder then, otherwise return the grid found int he constructor.

labkey-tchad (approved):

PR #247, opened 2020-03-02
Item 6848: Sample Type Designer Update

labkey-tchad:


Sigmonia (approved):

src/org/labkey/test/components/labkey/ui/samples/SampleTypeDesigner.java
```diff
@@ -0,0 +1,225 @@
...
+        {
+            if (inputs.get(i).get().equals(parentAlias))
+            {
+                aliasIndex = i;
+                break;
+            }
+        }
+        if (aliasIndex < 0)
+        {
+            throw new NotFoundException("No such parent alias: " + parentAlias);
```
labkey-danield: Is it really a fatal issue if a user tries to remove a parentAlias that isn't there?

src/org/labkey/test/components/labkey/ui/samples/SampleTypeDesigner.java
```diff
@@ -0,0 +1,225 @@
...
+    public SampleTypeDesigner setParentAlias(int index, @Nullable String alias, @Nullable String optionDisplayText)
+    {
+        elementCache().parentAlias(index).setValue(alias);
+        if (optionDisplayText != null)
+        {
+            elementCache().parentAliasSelect(index).select(optionDisplayText);
+        }
+        return this;
+    }
+
```
labkey-danield, line 181: Would it be useful to have a setParetnAlias that uses the alias as the identifier? That is setParetnAlias(String alias, @Nullable optionDisplayText).

src/org/labkey/test/components/labkey/ui/samples/SampleTypeDesigner.java
```diff
@@ -0,0 +1,225 @@
...
+    public String getParentAliasSelectText(int index)
+    {
+        return elementCache().parentAliasSelect(index).getSelections().get(0);
+    }
+
+    protected class ElementCache extends Component<?>.ElementCache
+    {
+        protected final Input nameInput = Input.Input(Locator.id("entity-name"), getDriver()).findWhenNeeded(this);
+        protected final Input nameExpressionInput = Input.Input(Locator.id("entity-nameExpression"), getDriver()).waitFor(this);
```
labkey-danield, line 202: Why is this a waitFor and not a findWhenNeeded?

src/org/labkey/test/params/experiment/SampleSetDefinition.java
```diff
@@ -75,20 +79,70 @@ public SampleSetDefinition addField(@NotNull FieldDefinition field)
...
+    protected String getKind()
+    {
+        return "SampleSet";
+    }
+
+    @NotNull
+    @Override
+    protected Map<String, Object> getOptions()
+    {
+        Map<String, Object> json = new HashMap<>();
```
labkey-danield: Kind of an interesting variable name...

src/org/labkey/test/BootstrapLocators.java
```diff
@@ -2,10 +2,10 @@
...
 public abstract class BootstrapLocators
 {
-    public static final Locator infoBanner = Locator.tagWithClass("div", BannerType.INFO.getCss());
-    public static final Locator successBanner = Locator.tagWithClass("div", BannerType.SUCCESS.getCss());
-    public static final Locator errorBanner = Locator.tagWithClass("div", BannerType.ERROR.getCss());
-    public static final Locator warningBanner = Locator.tagWithClass("div", BannerType.WARNING.getCss());
+    public static final Locator.XPathLocator infoBanner = Locator.tagWithClass("div", BannerType.INFO.getCss());
+    public static final Locator.XPathLocator successBanner = Locator.tagWithClass("div", BannerType.SUCCESS.getCss());
+    public static final Locator.XPathLocator errorBanner = Locator.tagWithClass("div", BannerType.ERROR.getCss());
+    public static final Locator.XPathLocator warningBanner = Locator.tagWithClass("div", BannerType.WARNING.getCss());
```
labkey-danield: Not really sure what this change has to to do with this story.

src/org/labkey/test/tests/SampleSetTest.java
```diff
@@ -914,9 +886,7 @@ public void testDeleteSamplesSomeWithAssayData()
         portalHelper.addWebPart("Assay List");
         SampleSetHelper sampleHelper = new SampleSetHelper(this);
         log("Create a sample set");
-        sampleHelper.createSampleSet(SAMPLE_SET_NAME, null,
-                null,
-                sampleData);
+        sampleHelper.createSampleSet(new SampleSetDefinition(SAMPLE_SET_NAME), sampleData);
 
 //  Note that we currently will not find runs where the batch id references a sampleId.  See Issue 37918.
```
labkey-danield, line 891: This comment should be removed, issue 37918 has been closed.

labkey-danield (approved):
I'm going to follow dev's lead and approve even though I have comments.
Looks ok, just a slight concern about an assert for a situation that is not a catastrophic.

src/org/labkey/test/components/labkey/ui/samples/SampleTypeDesigner.java
labkey-tchad: I prefer to have things throw than to no-op silently.

src/org/labkey/test/components/labkey/ui/samples/SampleTypeDesigner.java
labkey-tchad: Not sure. Much of this class was extracted from the `SampleSetDesignPage` in Sample Manager. I'm guessing this acts as insurance that the form has loaded so I'm hesitant to change it.

src/org/labkey/test/params/experiment/SampleSetDefinition.java
labkey-tchad: Quite. This was a `JSONObject` in an earlier iteration. Changing it to `options`

src/org/labkey/test/BootstrapLocators.java
labkey-tchad: I think an earlier iteration needed access to `XPathLocator`-specific methods. Reverted.

labkey-tchad:
A little background in case others want to review (@labkey-sweta @labkey-chrisj).

Creating/editing sample sets used to be two separate pages; one for the name, name expression, and import aliases; and one for defining the fields. APIs, and test helpers were similarly divided. This merges tests' handling of sample set designs and sample set domains.

UI Related changes:
 - I refactored those methods to be more atomic and made the various `createSampleSet` methods take a `SampleSetDefinition` rather than having numerous overloads for various use cases. 
 - Created new `SampleTypeDesigner` component based on existing Sample Manager test helper.

API Related changes:
Similar to the UI, creating a Sample Set via API went from two calls ('createSampleSet' and 'saveDomain') to a single call ('createDomain').
 - `SampleSetAPIHelper` was updated to use `CreateDomainCommand` and utilize the improved functionality of `SampleSetDefinition` (see below)
 - Improvements to `FieldDefinition`
    - Extend `PropertyDescriptor` from remoteapi and renamed some methods to match. This allows you to use `FieldDefinition` directly with various remote API methods that take `PropertyDescriptor`.
    - Extend properties currently supported by the Java API (See `FieldDefinition.toJSONObject`).
    - Added test API support for special `User`, `Sample`, and `Flag` column types.
 - Add general interface for creating domain definitions via API (`DomainProps`)
    - `SampleSetDefinition` extends it and is widely used.
    - `ListDefinition` is just a proof of concept. Needs to be iterated on and integrated into `ListHelper` and a future `ListApiHelper`.

src/org/labkey/test/components/labkey/ui/samples/SampleTypeDesigner.java
labkey-tchad: Probably. Added.

labkey-chrisj (approved):

PR #248, opened 2020-03-05
Item 6640: Update tests for new list designer UI

labkey-chrisj:
Enables tests that use the new list designer UI to come out from behind their exp switch


src/org/labkey/test/components/list/AdvancedListSettingsDialog.java
```diff
@@ -0,0 +1,169 @@
...
+    public AdvancedListSettingsDialog allowMultipleDiscussionsPerItem()
+    {
+        elementCache().radio(this, "Allow multiple discussions per item").check();
+        return this;
+    }
+
+    public AdvancedListSettingsDialog indexEntireListAsASingleDocument(boolean checked, String docTitle, String includeRadioLabel, String indexRadioLabel)
+    {
+        String labelText = "Index entire list as a single document";
+        elementCache().checkbox(labelText).set(checked);
```
labkey-tchad, line 55: Create `enum`s for these radio button labels. They are known. There's no reason to force individual tests to know the label text.
Also, create separate `enable` and `disable` methods. The code paths are fairly distinct and half of the parameters become pointless when disabling.

src/org/labkey/test/components/list/AdvancedListSettingsDialog.java
```diff
@@ -0,0 +1,169 @@
...
+            WebElement expandCollapsePane = elementCache().collapsibleField(labelText);
+            expandPane(expandCollapsePane);
+            Input.Input(Locator.id("entireListTitleTemplate"), getDriver()).find().set(docTitle);
+            elementCache().radio(expandContainer, includeRadioLabel).check();
+            elementCache().radio(expandContainer, indexRadioLabel).check();
+        }
+        return this;
+    }
+
+    public AdvancedListSettingsDialog indexEachItemAsASeparateDocument(boolean checked, String docTitle, String radioLabel)
```
labkey-tchad: See my comment about `indexEntireListAsASingleDocument`

src/org/labkey/test/components/list/AdvancedListSettingsDialog.java
```diff
@@ -0,0 +1,169 @@
...
+    private boolean isPaneExpanded(WebElement expandCollapsePane)
+    {
+        return Locator.tagWithClass("svg", "fa-angle-down").existsIn(expandCollapsePane);
+    }
+
+    public AdvancedListSettingsDialog setIndexFileAttachments(boolean checked)
+    {
+        Locator loc = Locator.tagWithClass("span", "list__advanced-settings-model__index-checkbox")
+                .withChild(Locator.tagWithText("span", "Index file attachments"))
+                .child(Locator.tagWithClass("span", "list__properties__checkbox--no-highlight"));
```
labkey-tchad, line 115: I presume this predates the Locator in the element cache.
```suggestion
        Locator loc = elementCache().checkBoxLoc("Index file attachments");
```

src/org/labkey/test/pages/list/EditListDefinitionPage.java
```diff
@@ -69,6 +62,54 @@ public static EditListDefinitionPage beginAt(WebDriverWrapper driver, String con
...
+    }
+
+    public EditListDefinitionPage setDescription(String description)
+    {
+        expandPropertiesPane();
+        elementCache().descriptionInput().set(description);
+        return this;
+    }
+
+    public AdvancedListSettingsDialog getAdvancedListSettings()
```
labkey-tchad: I generally prefer to name methods `getSomething` only if they don't change the page state. They are also, usually, terminal methods (e.g. `getDescription`).
```suggestion
    public AdvancedListSettingsDialog openAdvancedListSettings()
```

src/org/labkey/test/pages/list/EditListDefinitionPage.java
```diff
@@ -69,6 +62,54 @@ public static EditListDefinitionPage beginAt(WebDriverWrapper driver, String con
...
+
+    public EditListDefinitionPage checkIndexFileAttachements(boolean checked)
+    {
+        getAdvancedListSettings()
+                .setIndexFileAttachments(checked)
+                .clickApply();
+        return this;
+    }
+
     public DomainFormPanel getFieldsPanel()
```
labkey-tchad: Remove `expandFieldsPanel` and just have `getFieldsPanel` expand it. Anybody that wants to interact with the fields panel would reasonably assume that `getFieldsPanel` would return something in a usable state.

src/org/labkey/test/pages/list/EditListDefinitionPage.java
```diff
@@ -96,112 +138,88 @@ public DomainFormPanel setKeyField(ListHelper.ListColumnType listKeyType, String
...
+    public void clickCancel()
+    {
+        clickAndWait(Locator.button("Cancel").waitForElement(getDriver(), WAIT_FOR_JAVASCRIPT));
+    }
+
+    // domain properties
+    public DomainFormPanel listFieldsPane()
+    {
+        return elementCache().fieldsPanel;
+    }
```
labkey-tchad: Redundant with `getFieldsPanel`. Pick one and toss the other.

src/org/labkey/test/params/FieldDefinition.java
```diff
@@ -323,7 +323,9 @@ public String toString()
         Flag("Flag", "Flag (String)", null),
         Attachment("Attachment", "Attachment", "attachment"),
         User("User", "User", "int"),
-        Lookup("Lookup", "Lookup", null),
+        Lookup("Lookup", "Lookup", null),   // Todo: convert usage to lookupTo<type>
+        LookupToInteger("Lookup", "Lookup", "Integer"),
+        LookupToString("Lookup", "Lookup", "String"),
```
labkey-tchad, line 328: These seem redundant. The uses I see already have the lookup column type specified in the `ListColumn` constructor:
`new ListHelper.ListColumn("ListLookup", "ListLookup", `**`ListHelper.ListColumnType.Integer`**`, "List Lookup", new ListHelper.LookupInfo(getProjectName() + "/" + SUB_FOLDER_A, "lists", lookupTargetListName).setTableType(FieldDefinition.ColumnType.LookupToInteger));`
or
`new ListHelper.ListColumn("StudyLookup", "StudyLookup", `**`ListHelper.ListColumnType.String`**`, "", new ListHelper.LookupInfo(null, "viscstudies", "studies").setTableType(FieldDefinition.ColumnType.LookupToString))`

src/org/labkey/test/tests/SimpleModuleTest.java
```diff
@@ -1069,29 +1070,44 @@ protected void createList()
...
+                        TestDataGenerator.simpleFieldDef("Age", FieldDefinition.ColumnType.Integer).setDescription("Age"),
+                        TestDataGenerator.simpleFieldDef("Crazy", FieldDefinition.ColumnType.Boolean).setDescription("Crazy?")
+                ));
+        dgen.createList(createDefaultConnection(true), "Key");
+        goToManageLists();
+        _listHelper.goToList(LIST_NAME);
     }
 
-    private void createPeopleListInTab(String tabLabel)
+    private void createPeopleListInTab(String tabLabel, String containerPath) throws Exception
```
labkey-tchad: Add a TODO or something to switch this back to using the UI.
This is a fine workaround to get the test working but it isn't testing folder tab functionality.

labkey-tchad (changes requested):

src/org/labkey/test/components/list/AdvancedListSettingsDialog.java
labkey-chrisj: When I was testing this, it turned out that the span element on the checkboxes adjacent to expand-fields has a different class than standalone checkboxes- "model__index" vs "modal__index"
I just worked around this here, used the very-similar locator inline rather than have two side-by-side in elementCache

src/org/labkey/test/tests/SimpleModuleTest.java
labkey-chrisj: okay

src/org/labkey/test/params/FieldDefinition.java
labkey-chrisj: Unfortunately we have some usages where we use the columnType fields to do double-duty and that's why we still allow columnType to take a string.  Changing this will require time to shake out the failures and fix them.  I have a todo in ColumnType now, to coalesce these usages- I suggest we shake this out after 20.3

cnathe (changes requested):
Looks like develop was inadvertently merged into this 20.3 branch.

labkey-tchad (approved):
Assuming tests pass. This looks good to me.

cnathe (approved):

cnathe:
Latest TeamCity run complete. Looks like we are all set: https://teamcity.labkey.org/project.html?projectId=LabKey_203Release&branch_LabKey_203Release_Premium_SampleManager=listDesignerTestConversion

PR #250, opened 2020-03-09
enables regression coverage for issue 38341

labkey-chrisj:
Found the issue when writing automation; the test method for regressing it has been marked @Ignore since it was checked in
This just removes the @Ignore annotation

(side note: this fix is also in 20.3)

### DIFF
--- a/src/org/labkey/test/tests/DomainDesignerTest.java
+++ b/src/org/labkey/test/tests/DomainDesignerTest.java
@@ -587,7 +587,6 @@ public class DomainDesignerTest extends BaseWebDriverTest
      * @throws Exception
      */
     @Test
-    @Ignore("ignore this test until issue 38341 is resolved")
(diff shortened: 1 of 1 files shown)


labkey-tchad (commented):
Why not update the test in 20.3?

labkey-chrisj:
> Why not update the test in 20.3?
I wanted to do that, but thought we were in stabilization there.  If that's the better course of action, this PR could target release20.3-SNAPSHOT ?

labkey-tchad:
Watch the contents of your branch. This would merge a bunch of incompatible develop changes into 20.3.
This would be a pretty low-impact regression so let's just merge to develop.

labkey-chrisj:
Okay, retargeted to develop.

labkey-tchad (approved):

PR #251, opened 2020-03-09
Extract parseUrlQueryString from parseUrlQuery

labkey-nicka:
#### Changes

- parseUrlQuery safe against NPE
- expose query string parsing via parseUrlQueryString
- used as a part of fix for [37293](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=37293)

#### Related Pull Requests
* https://github.com/LabKey/biologics/pull/561

### DIFF
--- a/src/org/labkey/test/WebTestHelper.java
+++ b/src/org/labkey/test/WebTestHelper.java
@@ -359,20 +359,32 @@ public class WebTestHelper
 
     public static Map<String, String> parseUrlQuery(URL url)
     {
-        String query = url.getQuery();
-        if (query.startsWith("?"))
-            query = query.substring(1);
-        String[] queryArgs = query.split("&");
+        if (url != null)
+            return parseUrlQueryString(url.getQuery());
 
-        Map<String, String> parsedQuery = new HashMap<>(queryArgs.length);
+        return Collections.emptyMap();
+    }
 
-        for (String arg : queryArgs)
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/WebTestHelper.java
```diff
@@ -359,20 +359,32 @@ public static String buildRelativeUrl(String controller, @Nullable String contai
...
     {
-        String query = url.getQuery();
-        if (query.startsWith("?"))
-            query = query.substring(1);
-        String[] queryArgs = query.split("&");
+        if (url != null)
+            return parseUrlQueryString(url.getQuery());
 
-        Map<String, String> parsedQuery = new HashMap<>(queryArgs.length);
+        return Collections.emptyMap();
```
labkey-tchad, line 365: A `null` URL and a URL with no query would return the same thing. I'd distinguish them by returning `null` here.
```suggestion
        return null;
```

labkey-tchad (changes requested):

src/org/labkey/test/WebTestHelper.java
labkey-tchad: Actually. Now that I look at the usage, that technique would just cause NPEs in the test. Why do we need to protect against NPE here? A `null` URL seems like it would indicate that something went very wrong.

labkey-tchad (approved):

PR #256, opened 2020-03-13
Issue 39945: match related text in change summary in tests

labkey-ankurj:


src/org/labkey/test/WebDriverWrapper.java
```diff
@@ -1458,7 +1458,7 @@ public void assertTextPresent(TextSearcher searcher, String... texts)
 
         if (!missingTexts.isEmpty())
         {
-            String failMsg = (missingTexts.size() == 1 ? "Text '" : "Texts ['") + String.join("', '", missingTexts) + "'";
+            String failMsg = (missingTexts.size() == 1 ? "Text '" : "Texts ['") + String.join("','", missingTexts) + "'";
```
labkey-jeckels: Is this change necessary?

labkey-ankurj: It's not.. Just keeping the fail msg consistent with the text.

labkey-ankurj: I see, this is used by many places. I will update.

labkey-jeckels (approved):
Just saw Ankur's update - I'm approving this change since it's now just targeted to the issues test, but @labkey-sweta  and @labkey-tchad please let us know if you have concerns

labkey-tchad (approved):

labkey-tchad (changes requested):
This is just causing the test to fail: https://teamcity.labkey.org/viewLog.html?buildId=939541&tab=buildResultsDiv&buildTypeId=LabKey_203Release_Community_ModuleSuites_IssuesSqlserver

labkey-ankurj:
@labkey-tchad Yes, I reversed the order of related matching text.

labkey-tchad (approved):
Didn't realize this was verifying the comment text. Maybe add a comment to that effect.

labkey-ankurj:
@labkey-tchad added

PR #257, opened 2020-03-13
revert to setting description the old way

labkey-chrisj:


src/org/labkey/test/components/domain/DomainFieldRow.java
```diff
@@ -234,7 +235,12 @@ public DomainFieldRow setDescription(String description)
     {
         expand();
         getWrapper().shortWait().until(ExpectedConditions.elementToBeClickable(elementCache().descriptionTextArea));
-        getWrapper().setFormElementJS(elementCache().descriptionTextArea, description);
+        try{
+            getWrapper().setFormElement(elementCache().descriptionTextArea, description);
+        }catch (ElementNotInteractableException retry)
+        {
+            getWrapper().setFormElementJS(elementCache().descriptionTextArea, description);
```
labkey-tchad: I don't think setting via JS is a reliable solution here. If the input remains not-interactable, throwing is preferable to just failing silently.
```suggestion
            WebDriverWrapper.sleep(500);
            getWrapper().setFormElement(elementCache().descriptionTextArea, description);
```

labkey-tchad (changes requested):

labkey-tchad (approved):

PR #264, opened 2020-03-19
Update dependencies for labkey-client-api to use proper group and vesion number

labkey-susanh:


labkey-susanh:
Re-targeting to 20.3-SNAPSHOT

PR #266, opened 2020-03-19
Issue 40017: Fix group name and version for labkey-client-api dependency

labkey-susanh:


labkey-susanh:
Wrong name for 20.3 branch

PR #272, opened 2020-03-23
Item 6693: query metadata editor test migration

labkey-ankurj:


src/org/labkey/test/tests/ExpTest.java
```diff
@@ -132,15 +135,15 @@ public void testSteps()
...
-        PropertiesEditor.FieldRow row = editor.selectField("Created");
-        row.setLabel("editedCreated");
-        PropertiesEditor.FieldPropertyDock.FormatTabPane tabPane = row.properties().selectFormatTab();
-        tabPane.setPropertyFormat("ddd MMM dd yyyy");
-        clickButton("Save", 0);
+        DomainFieldRow domainRow = designerPage.fieldsPanel().getField("Created");
+        domainRow.setLabel("editedCreated");
+        domainRow.setDateFormat("ddd MMM dd yyyy");
+        designerPage.clickFinish();
         waitForText(WAIT_FOR_JAVASCRIPT, "Save successful.");
```
cnathe: Does this Save Successful text/message still happen for the new UI designer?

src/org/labkey/test/tests/ExpTest.java
```diff
@@ -163,17 +166,17 @@ public void testSteps()
         waitForElement(Locator.xpath("//span[contains(text(), 'Reset to Default')]"), defaultWaitForPage);
         int lastFieldIndex = getElementCount(Locator.xpath("//input[starts-with(@name, 'ff_label')]")) - 1;
         Locator lastField = Locator.xpath("//input[@name='ff_label" + lastFieldIndex + "']");
-        click(lastField);
-        click(Locator.xpath("//span").append(Locator.lkButton("Alias Field")));
-        selectOptionByText(Locator.name("sourceColumn"), "RowId");
-        click(Locator.xpath("//span").append(Locator.lkButton("OK")));
+        domainRow = designerPage.fieldsPanel().getField(lastFieldIndex);
```
cnathe: is this lastFieldIndex valid for the new UI designer? Looks like it is based off of looking for the count of "ff_label" as an input name. I don't think we use those type of names anymore. Should that line above changes as well? Or should we instead be using the domainDesigner test helper to get the count of the fields (I believe it has that helper on the DomainFormPanel)?

src/org/labkey/test/tests/ExpTest.java
```diff
@@ -163,17 +166,17 @@ public void testSteps()
...
+        clickButton("Alias Field");
+        SelectWrapper.Select(Locator.tagWithAttribute("select", "name", "aliasField")).findWhenNeeded(getDriver()).selectByVisibleText("RowId");
+        clickButton("OK");
 
         // Make it a lookup into our custom query
         editor = PropertiesEditor.PropertiesEditor(getDriver()).withTitleContaining("Metadata Properties").find();
         int fieldCount = getElementCount(Locator.xpath("//input[contains(@name, 'ff_type')]"));
```
cnathe: Same comment as above related to the "ff_type" input name. I don't think that will exist for the new designer UI. Also, shouldn't the use of the PropertiesEditor be removed completely?

cnathe (changes requested):

src/org/labkey/test/tests/ExpTest.java
labkey-ankurj: Yes

src/org/labkey/test/tests/ExpTest.java
```diff
@@ -24,7 +24,10 @@
 import org.labkey.test.TestTimeoutException;
 import org.labkey.test.categories.DailyB;
 import org.labkey.test.categories.FileBrowser;
+import org.labkey.test.components.DomainDesignerPage;
 import org.labkey.test.components.PropertiesEditor;
```
cnathe: One good indicator that the test has been fully converted is when this PropertiesEditor import can be removed. Is that the case here?

src/org/labkey/test/tests/ExpTest.java
```diff
@@ -131,20 +134,16 @@ public void testSteps()
...
-        tabPane.setPropertyFormat("ddd MMM dd yyyy");
-        clickButton("Save", 0);
-        waitForText(WAIT_FOR_JAVASCRIPT, "Save successful.");
+        clickButton("Edit Metadata", 6000);
+        DomainDesignerPage designerPage = new DomainDesignerPage(getDriver());
+
+        DomainFieldRow domainRow = designerPage.fieldsPanel().getField("Created");
+        domainRow.setLabel("editedCreated");
+        domainRow.setDateFormat("ddd MMM dd yyyy");
+        designerPage.click(Locator.button("Save"));
```
cnathe: Doesn't DomainDesignerPage have a clickSave() or clickFinish() method for this?

src/org/labkey/test/tests/ExpTest.java
```diff
@@ -160,29 +159,25 @@ public void testSteps()
         selectQuery("exp", "Data"); // Select the one we want to edit
         waitForElement(Locator.linkWithText("edit metadata"), WAIT_FOR_JAVASCRIPT); //on Ext panel
         clickAndWait(Locator.linkWithText("edit metadata"));
-        waitForElement(Locator.xpath("//span[contains(text(), 'Reset to Default')]"), defaultWaitForPage);
-        int lastFieldIndex = getElementCount(Locator.xpath("//input[starts-with(@name, 'ff_label')]")) - 1;
-        Locator lastField = Locator.xpath("//input[@name='ff_label" + lastFieldIndex + "']");
-        click(lastField);
```
cnathe, line 166: Does the new designer just put the newly added alias field on the end of the list? If so, that matches this previous behavior. Looks like in the old GWT designer it was clicking the last field so that the alias field was added at the end.

src/org/labkey/test/tests/ExpTest.java
```diff
@@ -160,29 +159,25 @@ public void testSteps()
...
-        click(Locator.xpath("//span").append(Locator.lkButton("Alias Field")));
-        selectOptionByText(Locator.name("sourceColumn"), "RowId");
```
cnathe, line 168: The test used to be selecting the "RowId" for the alias field. I don't see that being done explicitly in your updated test, is that because it is the first field and already selected?

src/org/labkey/test/tests/ExpTest.java
```diff
@@ -160,29 +159,25 @@ public void testSteps()
...
-        row = editor.selectField(fieldCount-1);
-        row.setType(new FieldDefinition.LookupInfo(null, "exp", "dataCustomQuery"));
+        domainRow = designerPage.fieldsPanel().getField(fieldCount-1);
+        domainRow.setType(FieldDefinition.ColumnType.Lookup).setFromSchema("exp").setFromTargetTable("dataCustomQuery" + " (Integer)");
 
         // Save it
-        clickButton("Save", 0);
-        waitForText(WAIT_FOR_JAVASCRIPT, "Save successful.");
-        clickButton("View Data");
+        designerPage.click(Locator.button("Save"));
```
cnathe: Same as above. I believe the DomainDesignerPage has a clickFinish() for this.

src/org/labkey/test/tests/ExpTest.java
```diff
@@ -194,9 +189,7 @@ public void testSteps()
         // Wait for query to load
         waitForText("edit metadata");
         clickAndWait(Locator.linkWithText("edit metadata"));
-        waitForElement(Locator.xpath("//span[contains(text(), 'Reset to Default')]"), defaultWaitForPage);
-        click(Locator.xpath("//span").append(Locator.lkButton("Reset to Default")));
-        click(Locator.xpath("//span").append(Locator.lkButton("OK")));
-        waitForText(WAIT_FOR_JAVASCRIPT, "Reset successful");
+        designerPage = new DomainDesignerPage(getDriver());
+        designerPage.click(Locator.button("Reset To Default"));
```
cnathe: does the updated page button click for "Reset To Default" not have a confirmation dialog? looks like the old test was clicking that button and then clicking OK to confirm it.

cnathe (changes requested):

src/org/labkey/test/tests/ExpTest.java
labkey-ankurj: It expects the page to navigate to next page and that doesn't happen here.

src/org/labkey/test/tests/ExpTest.java
labkey-ankurj: I switched to this call from clickFinish() due to expectations of navigation to a new page.

src/org/labkey/test/tests/ExpTest.java
cnathe: Ah ok, got it.

src/org/labkey/test/tests/ExpTest.java
cnathe: Got it. Thanks.

src/org/labkey/test/tests/ExpTest.java
labkey-ankurj: I think the confirmation for Reset to Default comes if there are any changes to the domain form in the current state. Do we always want the confirmation?

src/org/labkey/test/tests/ExpTest.java
labkey-ankurj: Correct.

src/org/labkey/test/tests/ExpTest.java
cnathe: I would say that, yes we do want the confirmation always. It could be the case that they had saved some metadata changes, then come back to this page and chose to reset at that point.

src/org/labkey/test/tests/ExpTest.java
labkey-ankurj: Yes, the alias field is added at the end of the list. Do we want to put the alias field next to the source field?

src/org/labkey/test/tests/ExpTest.java
cnathe: Cool. Maybe add a quick comment to that affect then (i.e. that the selected options is already RowId so no need to change).

src/org/labkey/test/tests/ExpTest.java
cnathe: Nope. I was just curious. I think that putting the alias field at the end makes the most sense. That is what the old GWT UI does after you save anyway, so putting it at the end right away sounds good to me.

src/org/labkey/test/tests/ExpTest.java
labkey-ankurj: removed the import.

src/org/labkey/test/tests/ExpTest.java
labkey-ankurj: added

src/org/labkey/test/tests/ExpTest.java
labkey-ankurj: changes related to this in the commit here: https://github.com/LabKey/platform/pull/949/commits/b9ce7eec0f41baaa13d9e8142334c0cc8d70a366

cnathe (approved):

src/org/labkey/test/components/QueryMetadataEditorPage.java
```diff
@@ -0,0 +1,91 @@
...
+        shortWait().until(ExpectedConditions.elementToBeClickable(newElementCache().aliasFieldButton()));
+        click(Locator.button(newElementCache().aliasFieldButton().getText()));
+        click(Locator.button("OK")); //the selected option is the first field
+    }
+
+    @Override
+    protected DomainDesignerPage.ElementCache elementCache()
+    {
+        return super.elementCache();
+    }
```
labkey-tchad: If you cast this then you won't need to call `newElementCache` in other methods.
```suggestion
    protected QueryMetadataEditorPage.ElementCache elementCache()
    {
        return (QueryMetadataEditorPage.ElementCache) super.elementCache();
    }
```

src/org/labkey/test/components/QueryMetadataEditorPage.java
```diff
@@ -0,0 +1,91 @@
...
+        click(Locator.button(elementCache().finishButton().getText()));
+    }
+
+    public void reset()
+    {
+        scrollIntoView(newElementCache().resetButton());
+        shortWait().until(ExpectedConditions.elementToBeClickable(newElementCache().resetButton()));
+        click(Locator.button(newElementCache().resetButton().getText()));
+        click(Locator.button("Reset")); // Reset confirmation on the confirm modal
+    }
```
labkey-tchad: This pattern will find the same button four times and unnecessarily get its text once (`WebElement.getText` is a fairly time consuming operation and should be avoided).
 - You can eliminate the redundant element finding by refactoring the element cache away from using methods (See corresponding feedback below).
 - The `getText` is just unnecessary; you can click a `WebElement` directly.
 - `newElementCache` shouldn't need to be called directly. Doing so prevents the cache from working correctly. (See `elementCache` feedback)
```suggestion
    public void reset()
    {
        scrollIntoView(elementCache().resetButton);
        shortWait().until(ExpectedConditions.elementToBeClickable(elementCache().resetButton));
        elementCache().resetButton.click();
        click(Locator.button("Reset")); // Reset confirmation on the confirm modal
    }
```

src/org/labkey/test/components/QueryMetadataEditorPage.java
```diff
@@ -0,0 +1,91 @@
...
+        return new QueryMetadataEditorPage.ElementCache();
+    }
+
+    public class ElementCache extends DomainDesignerPage.ElementCache
+    {
+        WebElement resetButton()
+        {
+            return Locator.button("Reset To Default")
+                    .waitForElement(getDriver(), WAIT_FOR_JAVASCRIPT);
+        }
```
labkey-tchad: Using a method prevents the element from being cached; `findWhenNeeded` will find and save the element the first time you try to inspect or interact with it.
```suggestion
        WebElement resetButton = Locator.button("Reset To Default")
                .findWhenNeeded(this).timeout(WAIT_FOR_JAVASCRIPT);
```

src/org/labkey/test/components/QueryMetadataEditorPage.java
```diff
@@ -0,0 +1,91 @@
...
+    {
+        super(driver);
+    }
+
+    @Override
+    public void clickFinish()
+    {
+        scrollIntoView(elementCache().finishButton());
+        shortWait().until(ExpectedConditions.elementToBeClickable(elementCache().finishButton()));
+        click(Locator.button(elementCache().finishButton().getText()));
```
labkey-tchad, line 20: Is there any sort of success indicator to wait for? Just clicking and continuing can cause tests to ignore errors.

src/org/labkey/test/components/QueryMetadataEditorPage.java
```diff
@@ -0,0 +1,91 @@
...
+        shortWait().until(ExpectedConditions.elementToBeClickable(newElementCache().resetButton()));
+        click(Locator.button(newElementCache().resetButton().getText()));
+        click(Locator.button("Reset")); // Reset confirmation on the confirm modal
+    }
+
+    public void viewData()
+    {
+        scrollIntoView(newElementCache().viewDataButton());
+        shortWait().until(ExpectedConditions.elementToBeClickable(newElementCache().viewDataButton()));
+        click(Locator.button(newElementCache().viewDataButton().getText()));
```
labkey-tchad: Wait for navigation.
```suggestion
        clickAndWait(elementCache().viewDataButton);
```

src/org/labkey/test/components/QueryMetadataEditorPage.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    public void editSource()
+    {
+        scrollIntoView(newElementCache().editSourceButton());
+        shortWait().until(ExpectedConditions.elementToBeClickable(newElementCache().editSourceButton()));
+        click(Locator.button(newElementCache().editSourceButton().getText()));
```
labkey-tchad: Wait for navigation.
```suggestion
        clickAndWait(elementCache().editSourceButton);
```

src/org/labkey/test/components/QueryMetadataEditorPage.java
```diff
@@ -0,0 +1,91 @@
...
+        shortWait().until(ExpectedConditions.elementToBeClickable(elementCache().finishButton()));
+        click(Locator.button(elementCache().finishButton().getText()));
+    }
+
+    public void reset()
+    {
+        scrollIntoView(newElementCache().resetButton());
+        shortWait().until(ExpectedConditions.elementToBeClickable(newElementCache().resetButton()));
+        click(Locator.button(newElementCache().resetButton().getText()));
+        click(Locator.button("Reset")); // Reset confirmation on the confirm modal
```
labkey-tchad, line 29: Is there any navigation or other indication that the reset is complete?

src/org/labkey/test/components/QueryMetadataEditorPage.java
```diff
@@ -0,0 +1,91 @@
...
+        shortWait().until(ExpectedConditions.elementToBeClickable(newElementCache().editSourceButton()));
+        click(Locator.button(newElementCache().editSourceButton().getText()));
+    }
+
+    public void aliasField()
+    {
+        scrollIntoView(newElementCache().aliasFieldButton());
+        shortWait().until(ExpectedConditions.elementToBeClickable(newElementCache().aliasFieldButton()));
+        click(Locator.button(newElementCache().aliasFieldButton().getText()));
+        click(Locator.button("OK")); //the selected option is the first field
```
labkey-tchad, line 53: Is there any success indicator?

labkey-tchad (changes requested):
These button clicking methods are doing a lot of redundant work. I added feedback for the reset button but it's applicable to all of the buttons here.
A lot of these methods also don't wait for any indication that operations have completed. This can lead to race conditions and random failures.

src/org/labkey/test/components/QueryMetadataEditorPage.java
labkey-ankurj: the success indicator is the field that gets added in the domain form.

src/org/labkey/test/components/QueryMetadataEditorPage.java
labkey-ankurj: I tried to make the buttons similar to DomainDesginerPage, is there something different in those buttons?
Screen Shot 2020-03-25 at 10 34 22 PM

src/org/labkey/test/components/QueryMetadataEditorPage.java
labkey-ankurj: 'Save successful' is a text response after clicking Save on Query Metadata Editor.

src/org/labkey/test/components/QueryMetadataEditorPage.java
labkey-ankurj: there is no indication saying Reset successful.

labkey-tchad (approved):
Looks good for now. I just added some TODOs for the next round of testing

PR #280, opened 2020-03-26
Item 7002: Automated test fixes related to shared component domain designer refactors for shared code

cnathe:
#### Rationale
Changes were made in the shared components package to consolidate shared pieces of the various domain designer components. Those changes were pulled into platform and SM with the related PRs and these test fixes were needed as a result.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/195
* https://github.com/LabKey/platform/pull/997
* https://github.com/LabKey/sampleManagement/pull/218

#### Changes
* Issue 39938: Fix for domain designer text option handling of the "max text length" input and radio options
* Re-enable the test cases in ColumnResizeTest (related to fix for 38838)
* Add SampleSetTest test case for verifying error messages for reserved field check on create sample set
* Update SampleSetParentColumnTest test case for parent alias validation against domain fields to check on both create and update of sample set

### DIFF
--- a/src/org/labkey/test/components/domain/DomainFormPanel.java
+++ b/src/org/labkey/test/components/domain/DomainFormPanel.java
@@ -249,7 +249,7 @@ public class DomainFormPanel extends WebDriverComponent<DomainFormPanel.ElementC
     public WebElement getPanelAlertWebElement()
     {
         getWrapper().waitFor(()-> BootstrapLocators.infoBanner.existsIn(getDriver()),
-                "the error alert did not appear as expected", 1000);
+                "the info alert did not appear as expected", 1000);
 
         // It would be better to not return a raw WebElement but who knows what the future holds, different alerts
         // may show different controls.
--- a/src/org/labkey/test/components/domain/DomainFormPanel.java
+++ b/src/org/labkey/test/components/domain/DomainFormPanel.java
@@ -215,6 +215,21 @@ public class DomainFormPanel extends WebDriverComponent<DomainFormPanel.ElementC
         return hasPanelTitle() ? elementCache().panelTitle.getText() : null;
     }
 
+    public String getPanelErrorText()
+    {
+        return getPanelErrorWebElement().getText();
+    }
+
+    public WebElement getPanelErrorWebElement()
+    {
+        getWrapper().waitFor(()-> BootstrapLocators.errorBanner.existsIn(getDriver()),
+                "the error alert did not appear as expected", 1000);
+
+        // It would be better to not return a raw WebElement but who knows what the future holds, different alerts
+        // may show different controls.
+        return BootstrapLocators.errorBanner.existsIn(getDriver()) ? BootstrapLocators.errorBanner.findElement(getDriver()) : null;
+    }
+
     /**
      * Get the alert message that is shown only in the alert panel. An example of this is the Results Field in
      * Sample Manager requires a field that is a sample look-up, if it missing an alert is shown.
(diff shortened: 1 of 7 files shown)


src/org/labkey/test/components/domain/DomainFormPanel.java
```diff
@@ -217,6 +215,21 @@ public String getPanelTitle()
...
+
+    public WebElement getPanelErrorWebElement()
+    {
+        getWrapper().waitFor(()-> BootstrapLocators.errorBanner.existsIn(getDriver()),
+                "the error alert did not appear as expected", 1000);
+
+        // It would be better to not return a raw WebElement but who knows what the future holds, different alerts
+        // may show different controls.
+        return BootstrapLocators.errorBanner.existsIn(getDriver()) ? BootstrapLocators.errorBanner.findElement(getDriver()) : null;
+    }
```
labkey-danield, line 231: I think this should be ok for now, but a possible trip up is that the scope of the existsIn is the Driver which, I believe, is the entire page. So if there is an error anywhere on the page this will find the first one it comes across.

A backlog item for the test team is to revisit the shared test component for panels. When we revisit that we should look more closely on how we find the error banner associated with a panel.

src/org/labkey/test/components/domain/DomainFormPanel.java
```diff
@@ -238,12 +248,12 @@ public String getPanelAlertText()
      */
     public WebElement getPanelAlertWebElement()
     {
+        getWrapper().waitFor(()-> BootstrapLocators.infoBanner.existsIn(getDriver()),
+                "the error alert did not appear as expected", 1000);
```
labkey-danield: Something to keep in mind about the waitFor call, when the wait expires it will throw an exception if there is a message parameter ("the error alert did not appear as expected.").
Adding this waitFor and message now means this function could throw when previously it did not.

src/org/labkey/test/components/labkey/ui/samples/SampleTypeDesigner.java
```diff
@@ -27,13 +27,20 @@
...
         _driver = driver;
         _el = Locator.id("app").findElement(_driver); // Full page component
     }
 
+    public SampleTypeDesigner(WebDriver driver, boolean useFinishButton)
+    {
+        this(driver);
+        _useFinishButton = useFinishButton;
+    }
+
```
labkey-danield: Did you want to change this one as well to not pass a parameter but put the smarts/logic in the getSaveButton?

labkey-danield (approved):

src/org/labkey/test/components/domain/DomainFormPanel.java
cnathe: Good to know. Thanks.

src/org/labkey/test/components/domain/DomainFormPanel.java
cnathe: Good to know. Thanks. Looks like the only usage to date was in the SMAssayCreateTest, which I'll verify passes.

src/org/labkey/test/components/labkey/ui/samples/SampleTypeDesigner.java
cnathe: Bad timing. I checked in this change right after you completed your review. I'll re-request review now that this change is in. Thanks.

src/org/labkey/test/components/labkey/ui/samples/SampleTypeDesigner.java
```diff
@@ -208,6 +209,24 @@ public String getParentAliasSelectText(int index)
         protected final WebElement cancelButton = Locator.button("Cancel").findWhenNeeded(this);
         protected final WebElement saveButton = Locator.button("Save").findWhenNeeded(this);
 
+        // the SM app uses alternate text for the sample type designer save buttons
+        protected final WebElement finishButton = Locator.buttonContainingText("Finish").findWhenNeeded(this);
```
labkey-tchad: You could make a Locator that would match either button.
```suggestion
        protected final WebElement saveButton = Locator.union(Locator.button("Save"),
                Locator.buttonContainingText("Finish")).findWhenNeeded(this);
```

cnathe: Awesome, thanks @labkey-tchad. I don't think I've ever used the Locator.union. I'll give it a try.

labkey-danield (approved):

PR #287, opened 2020-03-30
test updates for list automation

labkey-chrisj:
related to branches in premium, platform, provenance

### DIFF
--- a/src/org/labkey/test/params/FieldDefinition.java
+++ b/src/org/labkey/test/params/FieldDefinition.java
@@ -289,12 +289,14 @@ public class FieldDefinition extends PropertyDescriptor
         MultiLine("Multi-Line Text", "Multi-Line Text", "string"),
         Integer("Integer", "Integer", "int"),
         String("Text", "Text (String)", "string"),
-        Subject("Subject/Participant", "Subject/Participant (String)", "string"),
+        Subject("Subject/Participant", "Subject/Participant (String)",
+                "string", "http://www.labkey.org/exp/xml#string", null),
         DateTime("DateTime", "DateTime", "date"), // TODO remove this after GWT designer removed
         DateAndTime("Date Time", "Date Time", "date"),
         Boolean("Boolean", "Boolean", "boolean"),
         Double("Number (Double)", "Number (Double)", "float"), // TODO remove this after GWT designer removed
-        Decimal("Decimal", "Decimal", "float"),
+        Decimal("Decimal", "Decimal", "float",
+            "http://www.labkey.org/exp/xml#float", null),
         File("File", "File", "fileLink"),
         AutoInteger("Auto-Increment Integer", "Auto-Increment Integer", "int"),
         Flag("Flag", "Flag (String)", "string",
@@ -339,7 +341,7 @@ public class FieldDefinition extends PropertyDescriptor
 
         public String getRangeURI() { return _rangeURI; }
 
-        protected String getConceptURI()
+        public String getConceptURI()
         {
             return _conceptURI;
         }
--- a/src/org/labkey/test/params/FieldDefinition.java
+++ b/src/org/labkey/test/params/FieldDefinition.java
@@ -295,8 +295,7 @@ public class FieldDefinition extends PropertyDescriptor
         DateAndTime("Date Time", "Date Time", "date"),
         Boolean("Boolean", "Boolean", "boolean"),
         Double("Number (Double)", "Number (Double)", "float"), // TODO remove this after GWT designer removed
-        Decimal("Decimal", "Decimal", "float",
-            "http://www.labkey.org/exp/xml#float", null),
+        Decimal("Decimal", "Decimal", "float"),
         File("File", "File", "fileLink"),
         AutoInteger("Auto-Increment Integer", "Auto-Increment Integer", "int"),
         Flag("Flag", "Flag (String)", "string",
--- a/src/org/labkey/test/params/FieldDefinition.java
+++ b/src/org/labkey/test/params/FieldDefinition.java
@@ -340,7 +340,7 @@ public class FieldDefinition extends PropertyDescriptor
 
         public String getRangeURI() { return _rangeURI; }
 
-        public String getConceptURI()
+        protected String getConceptURI()
         {
             return _conceptURI;
         }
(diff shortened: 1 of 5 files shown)


src/org/labkey/test/tests/list/ListTest.java
```diff
@@ -751,6 +751,29 @@ public void testRemoveColumnOverAPI() throws Exception
...
+                .withColumns(cols);
+        DomainResponse createResponse = dgen.createList(createDefaultConnection(true), "key");
+        Domain listDomain = createResponse.getDomain();
+        listDomain.setName("remoteAPIAfterRename");
+
+        SaveDomainCommand saveCmd = new SaveDomainCommand(listDomain.getDomainId());
+        saveCmd.setDomainDesign(listDomain);
+        DomainResponse saveResponse = saveCmd.execute(createDefaultConnection(true), info.getFolder());
+
+        assertEquals("remoteAPIAfterRename", saveResponse.getDomain().getName());
```
labkey-chrisj, line 775: The test expects the rename to take here- you can rename a list over the UI- but the save domain doesn't seem to save it

labkey-tchad: I think you need to set the name in the `options` object of the command. For some reason, it is only exposed when creating the domain (`CreateDomainCommand.setOptions()`). Might need to `@Ignore` this test until the API is expanded.

src/org/labkey/test/tests/list/ListTest.java
```diff
@@ -683,6 +688,91 @@ public void testBatchInsertErrors() throws IOException, CommandException
...
+        listDomain.setFields(listFields);
+
+        SaveDomainCommand saveCmd = new SaveDomainCommand(info.getSchema(), info.getTable());
+        saveCmd.setDomainDesign(listDomain);
+        DomainResponse saveResponse = saveCmd.execute(createDefaultConnection(true), info.getFolder());
+
+        checker().verifyFalse("expect field [title] with type [" +FieldDefinition.ColumnType.String.getConceptURI()+ "]",
+                saveResponse.getDomain().getFields().stream()
+                        .anyMatch(a -> a.getName().equals("removeMe") &&
+                                a.getRangeURI().endsWith(FieldDefinition.ColumnType.String.getConceptURI())));
```
labkey-tchad: Why are you using `conceptURI` here? From my experimentation, it is just to provide UI styling to particular types. e.g. A 'Sample' column is really just a lookup to `exp.Materials` but the 'sample' concept makes the domain designer look different.

That being said; still having a "removeMe" column, just with a different type seems like a failure.

labkey-tchad (changes requested):

src/org/labkey/test/tests/list/ListTest.java
labkey-chrisj: ignoring for now

src/org/labkey/test/tests/list/ListTest.java
labkey-chrisj: it was the difference between "float" and "http://www.labkey.org/exp/xml#float", and I think the API is now regarding the full URL as the canonical way of describing the rangeURI even though it will accept "float" as an acceptable input
I just thought 'conceptURI' was odd nomenclature for a field that holds the value for a canonical rangeURI

src/org/labkey/test/params/FieldDefinition.java
```diff
@@ -289,12 +289,14 @@ public String toString()
         MultiLine("Multi-Line Text", "Multi-Line Text", "string"),
         Integer("Integer", "Integer", "int"),
         String("Text", "Text (String)", "string"),
-        Subject("Subject/Participant", "Subject/Participant (String)", "string"),
+        Subject("Subject/Participant", "Subject/Participant (String)",
+                "string", "http://www.labkey.org/exp/xml#string", null),
```
labkey-tchad: Wrong `conceptURI`:
```suggestion
        Subject("Subject/Participant", "Subject/Participant (String)",
                "string", "http://cpas.labkey.com/Study#ParticipantId", null),
```

src/org/labkey/test/params/FieldDefinition.java
```diff
@@ -289,12 +289,14 @@ public String toString()
...
         DateTime("DateTime", "DateTime", "date"), // TODO remove this after GWT designer removed
         DateAndTime("Date Time", "Date Time", "date"),
         Boolean("Boolean", "Boolean", "boolean"),
         Double("Number (Double)", "Number (Double)", "float"), // TODO remove this after GWT designer removed
-        Decimal("Decimal", "Decimal", "float"),
+        Decimal("Decimal", "Decimal", "float",
+            "http://www.labkey.org/exp/xml#float", null),
```
labkey-tchad: Decimal fields don't have a `conceptURI`
```suggestion
        Decimal("Decimal", "Decimal", "float"),
```

src/org/labkey/test/params/FieldDefinition.java
```diff
@@ -339,7 +341,7 @@ public String getLabel()
 
         public String getRangeURI() { return _rangeURI; }
 
-        protected String getConceptURI()
+        public String getConceptURI()
```
labkey-tchad: `conceptURI` isn't useful outside this class, it just muddies the interface.
```suggestion
        protected String getConceptURI()
```

src/org/labkey/test/tests/list/ListTest.java
```diff
@@ -683,6 +689,91 @@ public void testBatchInsertErrors() throws IOException, CommandException
...
+        Domain listDomain = createResponse.getDomain();
+        List<PropertyDescriptor> listFields = createResponse.getDomain().getFields();
+        listFields.removeIf(a-> a.getName().equals("removeMe"));
+        listDomain.setFields(listFields);
+
+        SaveDomainCommand saveCmd = new SaveDomainCommand(info.getSchema(), info.getTable());
+        saveCmd.setDomainDesign(listDomain);
+        DomainResponse saveResponse = saveCmd.execute(createDefaultConnection(true), info.getFolder());
+
+        checker().verifyFalse("expect field [title] with type [" +FieldDefinition.ColumnType.String.getConceptURI()+ "]",
```
labkey-tchad: What is `title`? Copy/paste error?
```suggestion
        checker().verifyFalse("'removeMe' field was not deleted.",
```

labkey-tchad (changes requested):
Just need to clean up some of the `conceptURI` stuff.

labkey-tchad (approved):

PR #288, opened 2020-03-30
Create panorama suite

labkey-tchad:


labkey-jeckels:
What tests will be included here? It seems like all of the TargetedMS module's tests are not, at least not yet? That's likely to cause confusion, at least for me, because I think of TargetedMS and Panorama as synonyms.

labkey-tchad:
Picking a less misleading suite name

PR #293, opened 2020-04-02
Use more intuitive method names in 'AdvancedSettingsDialog'

labkey-tchad:


src/org/labkey/test/components/domain/AdvancedFieldSetting.java
```diff
@@ -0,0 +1,41 @@
+package org.labkey.test.components.domain;
+
+import org.labkey.test.components.PropertiesEditor;
+
+import java.util.function.BiConsumer;
+
+public class AdvancedFieldSetting<T>
```
labkey-chrisj, line 7: This approach seems interesting; would definitely like to talk a bit about how it works

src/org/labkey/test/components/domain/AdvancedSettingsDialog.java
```diff
@@ -28,43 +29,56 @@ public AdvancedSettingsDialog(DomainFieldRow row)
...
     }
+
     public AdvancedSettingsDialog showInDefaultView(boolean checked)
     {
         elementCache().showInDefaultView.set(checked);
         return this;
     }
 
-    public boolean showOnInsertView()
+    public boolean isShownOnInsertView()
```
labkey-chrisj, line 52: general note: the method names here are pretty intuitive; I like this

labkey-chrisj (approved):

src/org/labkey/test/components/domain/DomainFieldRow.java
```diff
@@ -74,9 +74,6 @@ public Input nameInput()
 
     /**
      * selects the field data type.  Note: after the field is initially created, the select will be disabled
-     *
-     * @param columnType
-     * @return
      */
```
labkey-danield, line 77: The signature didn't change, the @param and @return values in the javadocs comment should be filled in.

src/org/labkey/test/components/domain/DomainFieldRow.java
```diff
@@ -214,8 +217,6 @@ public boolean isNewField()
 
     /**
      * indicates that the field has been edited (but is not new). The user will need to save changes to persist
-     *
-     * @return
```
labkey-danield, line 218: Same comment as above.

labkey-danield (approved):

PR #294, opened 2020-04-03
fix query parameters for query editor

labkey-ankurj:


src/org/labkey/test/tests/LinkedSchemaTest.java
```diff
@@ -563,11 +564,10 @@ private void checkLinkedSchema(String updatedMetaData, String studyName, int exp
...
     {
         navigateToMetadataQuery(schema, query);
-        clickButton("Alias Field", 0);
-        selectOptionByValue(Locator.gwtListBoxByName("sourceColumn"), fieldToWrap);
-        clickButton("OK", 0);
-        setFormElement(Locator.tagWithAttribute("input", "value", "Wrapped" + fieldToWrap), aliasFieldName);
-        clickButton("Save", 0);
+        QueryMetadataEditorPage queryMetadataEditorPage = new QueryMetadataEditorPage(getDriver());
+        queryMetadataEditorPage.aliasField().selectAliasField(fieldToWrap).clickApply();
+        queryMetadataEditorPage.fieldsPanel().getField("WrappedParticipantId").setName(aliasFieldName);
```
cnathe, line 569: I don't know if this wrapField() method is used for anything other than the ParticipantId field, but using that string `WrappedParticipantId` here makes it specific to that field. Since we know that the alias field is always added to the end of the list, you could probably more generally get the field based on the last field index.

cnathe (approved):

src/org/labkey/test/tests/LinkedSchemaTest.java
labkey-ankurj: I checked it earlier, only being used at one place.

PR #301, opened 2020-04-09
Moving some of the tests from BVT to DailyB.

labkey-sweta:
Moving some of the tests from BVT to DailyB.


labkey-danield (approved):

labkey-tchad (changes requested):
BVTs are down to 33 minutes with these tests removed. Let's see how long it is with `DomainDesignerTest` and `GpatAssayTest`

labkey-tchad (approved):

PR #307, opened 2020-04-13
Fix URL save/recall in AbstractExportTest

labkey-tchad:
Cleaning up some old feature branches.

I noticed this misuse of `saveLocation`/`recallLocation` a while back. The saved location for those is stored in a non-static variable and won't survive between different `@Test` methods.


src/org/labkey/test/tests/AbstractExportTest.java
```diff
@@ -483,7 +492,7 @@ protected BrowserType bestBrowser()
         return BrowserType.CHROME;
     }
 
-    public ColumnHeaderType[] getExportHeaderTypes()
+    protected ColumnHeaderType[] getExportHeaderTypes()
```
labkey-danield, line 495: Just curious why this is changed.

labkey-danield (approved):

src/org/labkey/test/tests/AbstractExportTest.java
labkey-tchad: I think I was just making `getExportHeaderTypes` and `goToDataRegionPage` consistent.

PR #311, opened 2020-04-13
Unified domain test components

labkey-tchad:
This is an attempt to unify some of the test pages and components used by domain designers (List, Data Class, Sample Type, etc.) and pseudo-domain designers (Query metadata)  
Primary components here are:
```
BaseDomainDesigner (Handles save/cancel buttons and finding errors)
 ├─DomainDesigner (For designers with a properties panel and a single domain)
 │ ├─SampleTypeDesigner
 │ ├─EditListDefinitionPage
 │ ├─QueryMetadataEditorPage
 │ ├─InsertDataClassPage
 │ └─*DatasetDesigner
 └─*AssayDesigner

DomainPanel (Handles expand/collapse and defines panel finder)
 └DomainPanel.DomainPanelImpl (Used to expand/collapse properties panels or to construct other panels)
   ├DomainFormPanel (Field designer)
   └EntityDesignFieldsPanel (Panel for inferring fields from a file, currently in SM)

*-Not yet implemented
```

I'm still working on a better pattern for integrating these into different products. Currently, for the sample type designer, we wrap the `SampleTypeDesigner` component in a LKS or SM page and just pass most of the functionality through.


src/org/labkey/test/components/domain/BaseDomainDesigner.java
```diff
@@ -0,0 +1,100 @@
...
+            elementCache().cancelButton.click();
+        }
+    }
+
+    public boolean isSaveButtonEnabled()
+    {
+        return elementCache().saveButton.isEnabled();
+    }
+
+    public Object clickSave()
```
labkey-danield, line 66: Not sure I understand why this is not a void if it will always return a null. Is the idea that a derived class would override it?
Also since this is itself an overload of the function below, (clickSave(boolean expectPageLoad)) shouldn't that function also return?

src/org/labkey/test/components/QueryMetadataEditorPage.java
```diff
@@ -52,9 +42,10 @@ public AliasFieldDialog aliasField()
     }
 
     @Override
-    protected QueryMetadataEditorPage.ElementCache elementCache()
+    public Void clickSave()
```
labkey-danield: Ummm.... did you really want **V**oid and not _void_?
Void is a class which is why no error is shown with the return statement.

src/org/labkey/test/components/domain/BaseDomainDesigner.java
```diff
@@ -0,0 +1,100 @@
...
+    {
+        return driver;
+    }
+
+    public boolean isCancelButtonEnabled()
+    {
+        return elementCache().cancelButton.isEnabled();
+    }
+
+    public Object clickCancel()
```
labkey-danield, line 50: Same comment as for clickSave. Why return anything if you are always returning null? And why have this method have a return value but the overloaded function clickCancel(boolean expectPageLoad) is void?

src/org/labkey/test/components/domain/DomainPanel.java
```diff
@@ -0,0 +1,154 @@
...
+import org.labkey.test.components.WebDriverComponent;
+import org.labkey.test.util.LabKeyExpectedConditions;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+/**
+ * Wraps the functionality of the LabKey ui component defined in domainproperties/CollapsiblePanelHeader.tsx
+ * Subclasses should add panel-specific functionality.
+ */
+public abstract class DomainPanel<EC extends DomainPanel<EC, T>.ElementCache, T extends DomainPanel<?, T>> extends WebDriverComponent<EC>
```
labkey-danield: Is it possible to make this class a little more generic? If you look at Sample Manager Workflow create a new job, the "Define Job Tasks" is a domain panel but has a few more html elements in the title bar. For example it has text that appears on the right hand side of the heard (ex "1 Task Defined"). It also has an icon next to the panel title.

src/org/labkey/test/components/domain/BaseDomainDesigner.java
```diff
@@ -0,0 +1,100 @@
...
+        else
+        {
+            elementCache().saveButton.click();
+        }
+    }
+
+    public List<String> clickSaveExpectingErrors()
+    {
+        elementCache().saveButton.click();
+        return getWrapper().getTexts(BootstrapLocators.errorBanner.waitForElements(getWrapper().shortWait()));
```
labkey-danield: My experience with the .getTexts call is that if there is any kind of a mark-up in the message it can add spaces at the front and end of the text. This becomes a tricky for text expecting specific text in the error message.
Maybe this code should do some clean up of the text before adding them to the list.

labkey-danield (changes requested):
I'm good with the approach, just a couple of comments on return types an how generic the DomainPanel class can be.

src/org/labkey/test/components/domain/BaseDomainDesigner.java
labkey-tchad: Type `Object` is so that subclasses can return a specific page (see `QueryMetadataEditor` or `SampleSetDesignPage`).

The `clickSave(boolean)` and `clickCancel(boolean)` methods were to allow subclasses to specify whether a page load is expected. I wasn't happy with the solution and have deleted them in favor of having subclasses interact directly with the buttons in the element cache.

src/org/labkey/test/components/QueryMetadataEditorPage.java
labkey-tchad: I did mean that but it should have been `QueryMetadataEditorPage` since this page saves in place.
I've updated it accordingly.

src/org/labkey/test/components/domain/DomainPanel.java
labkey-tchad: Are you thinking of having methods to fetch the icon class and task count?
I'll look at adding them. Regular domain editors have those elements as well, so that seems reasonable.

src/org/labkey/test/components/domain/BaseDomainDesigner.java
labkey-tchad: I wasn't aware of that. I've updated it to `.trim()` the error texts.

src/org/labkey/test/components/domain/DomainPanel.java
labkey-tchad: I also added a class to the title element of the workflow job panes so that these helpers should work with them. I'm not sure how reliable they will be though. It looks like the workflow jobs form doesn't actually use any of the domain components, it just styles itself to look like them.

labkey-danield (approved):

PR #312, opened 2020-04-14
Item 6961: DataClassDesigner updates for LKS

cnathe:
#### Rationale
A DataClassDesigner component was created for a recent SM story to give consistency between creating sample types and creating data classes (i.e. sources in SM app). As a follow-up, we have pulled that same DataClassDesigner component into the LKS side to make that experience consistent with our other domain designers as well.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/222
* https://github.com/LabKey/platform/pull/1064
* https://github.com/LabKey/sampleManagement/pull/238
* https://github.com/LabKey/biologics/pull/577

#### Changes
* domain designer helper component changes to support SampleTypeDesigner change to defaulting to "infer from file" UI in empty fields state
* factor out EntityTypeDesigner, add DataClassDesigner which extends it, and update SampleTypeDesigner to extend it as well
* refactor InsertDataClassPage as CreateDataClassPage to work with experiment-editDataClass action
* misc sample type test updates for changes to experiment-editSampleSet action


src/org/labkey/test/components/domain/DomainFormPanel.java
```diff
@@ -108,6 +108,11 @@ public DomainFieldRow addField(String name)
         return newFieldRow;
     }
 
+    public boolean isStartNewDesignPresent()
+    {
+        return getWrapper().isElementPresent(elementCache().startNewDesignLoc);
+    }
+
```
labkey-tchad, line 115: I've never been a fan of the `startNewDesign` terminology in these helpers. I was very confused the first time I saw it in a test because you're already on the designer page.
Could we make this make this more clear about what it is checking? Something like `isInferFieldsFromFilePresent`.

src/org/labkey/test/components/labkey/ui/samples/DataClassDesigner.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public void expandPropertiesPanel()
+    {
+        elementCache().propertiesPanelHeader.click();
+    }
+
+    protected class ElementCache extends EntityTypeDesigner.ElementCache
+    {
+        protected final WebElement propertiesPanelHeader = Locator.id("dataclass-properties-hdr").findWhenNeeded(this);
```
labkey-tchad: Use `DomainFormPanel` to expand/collapse. I've pulled that functionality into a base domain panel class in another branch but this will make it easier to rafactor this when I'm ready to merge. You can also pull this up into the base EntityTypeDesigner since you won't be tied to this header ID.

src/org/labkey/test/components/labkey/ui/samples/EntityTypeDesigner.java
```diff
@@ -0,0 +1,156 @@
...
+import org.labkey.test.components.html.Input;
+import org.labkey.test.params.FieldDefinition;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+import java.util.List;
+
+import static org.labkey.test.WebDriverWrapper.sleep;
+
+public abstract class EntityTypeDesigner extends WebDriverComponent<EntityTypeDesigner.ElementCache>
```
labkey-tchad: Add a type parameter to this so that the subclasses can chain properly.
`sampleTypeDesigner.setName(name).addParentAlias("", "")`  wouldn't work because `getName` doesn't return a SampleTypeDesigner.
```suggestion
public abstract class EntityTypeDesigner<T extends EntityTypeDesigner> extends WebDriverComponent<EntityTypeDesigner.ElementCache>
```
Then add:
```
protected abstract T getThis();
```
And change setters to look like:
```
    public T setName(String name)
    {
        elementCache().nameInput.set(name);
        return getThis();
    }
```

src/org/labkey/test/tests/TriggerScriptTest.java
```diff
@@ -802,13 +802,11 @@ private void setupDataClass()
...
-        DomainFormPanel domainFormPanel = domainDesignerPage.fieldsPanel();
-        domainFormPanel.addField(COMMENTS_FIELD).setLabel(COMMENTS_FIELD);
-        domainFormPanel.addField(COUNTRY_FIELD).setLabel(COUNTRY_FIELD);
-        domainDesignerPage.clickFinish();
+        drt.clickHeaderMenu("New Data Class", "Design Manually");
+
+        CreateDataClassPage createDataClassPage = new CreateDataClassPage(getDriver());
+        createDataClassPage.setName(DATA_CLASSES_NAME)
+            .addFields(List.of(new FieldDefinition(COMMENTS_FIELD), new FieldDefinition(COUNTRY_FIELD)))
+            .clickSave();
```
labkey-tchad: Consider creating this data class via API:
```suggestion

        DataClassDefinition dataClass = new DataClassDefinition(DATA_CLASSES_NAME)
                .setFields(List.of(new FieldDefinition(COMMENTS_FIELD), new FieldDefinition(COUNTRY_FIELD)));
        TestDataGenerator.createDomain(getProjectName(), dataClass);
```

labkey-tchad (changes requested):
I'm working on a broader refactor of the domain designers (https://github.com/LabKey/testAutomation/pull/311) but this should be pretty easy to merge in.

src/org/labkey/test/components/domain/DomainFormPanel.java
cnathe: Yes, good point. The text on that link in the domain fields panel originally said something like "or Start a New Design" which is likely why these helpers got that name. I can change that.

src/org/labkey/test/components/labkey/ui/samples/DataClassDesigner.java
cnathe: Perfect. Thanks, I'll make that change.

src/org/labkey/test/components/labkey/ui/samples/EntityTypeDesigner.java
cnathe: Great. Thanks, I'll make that change.

src/org/labkey/test/tests/TriggerScriptTest.java
cnathe: Yeah, I can do that. I think this might actually be one of the only places in the selenium tests that use the UI for data class creation, so we should definitely add a new set of test cases for that at some point. @labkey-tchad is there a test backlog place that makes sense to add that item (i.e. more selenium test coverage of the Data Class create/edit via UI)?

labkey-tchad: Dan added some UI automation in SM but I aggree, we should have some LKS coverage as well. I'd say add a new test story to the existing epic (6961)

src/org/labkey/test/components/labkey/ui/samples/DataClassDesigner.java
cnathe: When looking into this, the DomainFormPanel expand/collapse is for the fields panel. This expand/collapse I am trying to do is for the top properties panel of these designers. Are you saying there is a helper somewhere for that already?

labkey-tchad: Assuming the properties panel uses `CollapsiblePanelHeader` from labkey-ui-components, the `DomainFormPanel` expand/collapse should work correctly. If not, I have some rethinking to do in my branch.

cnathe: Ah, I getcha. Yes, the properties panel does use CollapsiblePanelHeader. I'll give it a try. Thanks.

src/org/labkey/test/tests/DataClassTest.java
```diff
@@ -0,0 +1,173 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+import static org.labkey.test.util.DataRegionTable.DataRegion;
+
+@Category({DailyC.class})
+@BaseWebDriverTest.ClassTimeout(minutes = 5)
+public class DataClassTest extends BaseWebDriverTest
```
cnathe, line 38: @labkey-tchad @labkey-danield FYI, I added an initial DataClassTest here with some test cases to error validation checks on the new data class designer UI. There is a TODO in here and story added to the backlog for adding more UI cases for this designer.

labkey-tchad (approved):
Fixed up the type parameter but looks good otherwise

cnathe:
> Fixed up the type parameter but looks good otherwise

Thank you!

PR #314, opened 2020-04-15
create component/subcomponent wrappers for lineage graph component

labkey-chrisj:
adds components to interact with lineage graph details panel

related PRs: provenance/fb_lineageGraphTesting


src/org/labkey/test/components/experiment/LineageGraph.java
```diff
@@ -0,0 +1,105 @@
...
+
+import org.labkey.test.Locator;
+import org.labkey.test.components.Component;
+import org.labkey.test.components.WebDriverComponent;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+import java.util.List;
+
+public class LineageGraph extends WebDriverComponent<LineageGraph.ElementCache>
```
labkey-tchad, line 16: Add a comment indicating what product component is controlled by this class. (Similar to [SampleTypeDesigner.java](https://github.com/LabKey/testAutomation/blob/develop/src/org/labkey/test/components/labkey/ui/samples/SampleTypeDesigner.java#L20))

src/org/labkey/test/components/experiment/LineageGraph.java
```diff
@@ -0,0 +1,105 @@
...
+        List<NodeDetails> summaryLists()
+        {
+            return new NodeDetails.NodeDetailsFinder(getDriver()).findAll(nodeDetailContainer);
+        }
+    }
+
+
+    public static class LineageGraphFinder extends WebDriverComponentFinder<LineageGraph, LineageGraphFinder>
+    {
+        private final Locator.XPathLocator _baseLocator = Locator.tagWithAttributeContaining("div", "id", "run-graph-app");
```
labkey-tchad: This Locator is specific to LabKey Server. Try to find one that will work in Sample Manager as well.

src/org/labkey/test/components/experiment/LineageGraph.java
```diff
@@ -0,0 +1,105 @@
...
+import org.openqa.selenium.WebElement;
+
+import java.util.List;
+
+public class LineageGraph extends WebDriverComponent<LineageGraph.ElementCache>
+{
+    final WebElement _el;
+    final WebDriver _driver;
+
+    public LineageGraph(WebElement element, WebDriver driver)
```
labkey-tchad: Make constructor protected since only the `Finder` knows how to find it.
```suggestion
    protected LineageGraph(WebElement element, WebDriver driver)
```

src/org/labkey/test/components/experiment/NodeDetailItem.java
```diff
@@ -0,0 +1,113 @@
...
+    {
+        final WebElement image = Locator.tagWithClass("img", "lineage-sm-icon")
+                .findWhenNeeded(this);
+
+        final WebElement overviewLink = Locator.linkWithSpan("Overview")
+                .findWhenNeeded(this).withTimeout(2000);
+        final WebElement lineageGraphLink = Locator.linkWithSpan("Lineage")
+                .findWhenNeeded(this).withTimeout(2000);
+        final Locator nameLink = Locator.tagWithClass("a", "pointer");
+        final Locator nameSpan = Locator.tag("span");
```
labkey-tchad: You can combine Locators rather than checking for one and falling back to the other.
```suggestion
        final WebElement nameEl = Locator.union(
                Locator.tagWithClass("a", "pointer"), 
                Locator.tag("span"))
            .findWhenNeeded(this);
```

src/org/labkey/test/components/experiment/NodeDetailItem.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public static class NodeDetailItemFinder extends WebDriverComponentFinder<NodeDetailItem, NodeDetailItemFinder>
+    {
+        private final Locator.XPathLocator _baseLocator = Locator.tagWithClass("li", "lineage-name");
+        private String _name = null;
+        private String _linkedName = null;
```
labkey-tchad: Remove unused variables.
```suggestion
```

src/org/labkey/test/components/experiment/NodeDetailItem.java
```diff
@@ -0,0 +1,113 @@
+package org.labkey.test.components.experiment;
+
+import org.labkey.test.Locator;
+import org.labkey.test.components.Component;
+import org.labkey.test.components.WebDriverComponent;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+public class NodeDetailItem extends WebDriverComponent<NodeDetailItem.ElementCache>
```
labkey-tchad: Add a comment indicating what product component is controlled by this class.

src/org/labkey/test/components/experiment/NodeDetails.java
```diff
@@ -0,0 +1,121 @@
...
+import org.labkey.test.Locator;
+import org.labkey.test.components.Component;
+import org.labkey.test.components.WebDriverComponent;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class NodeDetails extends WebDriverComponent<NodeDetails.ElementCache>
```
labkey-tchad: Add a comment indicating what product component is controlled by this class.

labkey-tchad (changes requested):
Looking good. Needs some documentation indicating what product components are being instrumented and description of the relationship and behavior of the test components.

src/org/labkey/test/components/experiment/LineageGraph.java
```diff
@@ -0,0 +1,105 @@
...
+    {
+        final WebElement visGraphContainer = Locator.tagWithClass("div", "lineage-visgraph-ct")
+                .findWhenNeeded(this);
+        final WebElement nodeDetailContainer = Locator.tagWithClass("div", "lineage-node-detail-container")
+                .findWhenNeeded(this).withTimeout(4000);
+
+        WebElement nodeDetails = Locator.tagWithClass("div", "lineage-node-detail")
+                .findWhenNeeded(nodeDetailContainer).withTimeout(3000);
+        WebElement nodeDetailsName = Locator.tagWithClass("div", "lineage-name-data")
+                .findWhenNeeded(nodeDetails);
```
labkey-danield, line 139: Doesn't look like this element is used and I cannot find a div with that class in the lineage graph in Biologics or SampleManager. You should remove this.

src/org/labkey/test/components/experiment/NodeDetailItem.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public void clickOverViewLink()
+    {
+        getWrapper().mouseOver(getComponentElement());
+        getWrapper().waitFor(()-> elementCache().overviewLink.isEnabled(), 1000);
+        getWrapper().clickAndWait(elementCache().overviewLink);
+    }
+
+    public void getLineageGraphLink()
```
labkey-danield: This function name should be changed to "click" not get, it's not returning anything.

src/org/labkey/test/components/experiment/NodeDetails.java
```diff
@@ -0,0 +1,121 @@
...
+            return elementCache().summary.getText();
+    }
+
+    public NodeDetailItem getItem(String itemName)
+    {
+        return  elementCache().items().stream().filter(a-> a.getName().equals(itemName))
+                .findFirst().orElseThrow();
+    }
+
+    public NodeDetailItem getItemByTitle(String title)
```
labkey-danield: How does this function differ from getItem(String itemName)? From the consumer perspective they both return the same thing the node detail based on the text it is showing.

src/org/labkey/test/components/experiment/NodeDetails.java
```diff
@@ -0,0 +1,121 @@
...
+    final String _groupName;
+
+    public NodeDetails(WebElement element, String groupName, WebDriver driver)
+    {
+        _el = element;
+        _driver = driver;
+        _groupName = groupName;
+    }
+
+    public String getGroupName()
```
labkey-danield: Using groupName in a class named NodeDetails is a little confusing. This is really the text contained in the summary.lineage-name tag, which is the node. So shouldn't this be called getNodeDetailName?
Or alternatively since the finder calls it "Title" (find.withTitle) maybe it should be called getTitle. 
At any rate I think you could be a little more consistent with the name of the class (the tag it wraps), this member attribute and the finder.

src/org/labkey/test/components/experiment/NodeDetails.java
labkey-danield: I don't think the name should be plural, it is a single node detail. It has a member function which return a collection of items of a different class, but this represents a single object.

src/org/labkey/test/components/experiment/NodeDetails.java
```diff
@@ -0,0 +1,121 @@
...
+    public List<NodeDetailItem> getItems()
+    {
+        return elementCache().items();
+    }
+
+    public List<String> getItemNames()
+    {
+        return elementCache().items().stream().map(NodeDetailItem::getName).collect(Collectors.toList());
+    }
+
```
labkey-danield: I don't think it is needed, but this component (NodeDetails) can collapse and expand, any value in adding that functionality?

src/org/labkey/test/components/experiment/NodeDetailItem.java
```diff
@@ -0,0 +1,113 @@
...
+    {
+        getWrapper().mouseOver(getComponentElement());
+        getWrapper().waitFor(()-> elementCache().lineageGraphLink.isEnabled(), 1000);
+        getWrapper().clickAndWait(elementCache().lineageGraphLink);
+    }
+
```
labkey-danield: Consider adding the icon element to the element cache. Different types of elements will have different types of icons. The getter for it could return the src or alt attribute of the img tag.

src/org/labkey/test/components/experiment/LineageGraph.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * finds the list of edges/nodes to the one currently focused, by name
+     * @param listTitle
+     * @return
+     */
+    public NodeDetails getDetails(String listTitle)
+    {
+        return elementCache().summaryList(listTitle);
```
labkey-danield, line 49: This doesn't return a list, it returns a single NodeDetails item. The NodeDetails may contain a list of NodeDetailItems .

src/org/labkey/test/components/experiment/LineageGraph.java
```diff
@@ -0,0 +1,105 @@
...
+    protected ElementCache newElementCache()
+    {
+        return new ElementCache();
+    }
+
+
+    protected class ElementCache extends Component<?>.ElementCache
+    {
+        final WebElement visGraphContainer = Locator.tagWithClass("div", "lineage-visgraph-ct")
+                .findWhenNeeded(this);
```
labkey-danield, line 115: Doesn't look like this element is used.

src/org/labkey/test/components/experiment/LineageGraph.java
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * finds all details lists in the details panels
+     * @return
+     */
+    public List<NodeDetails> getDetails()
+    {
+        return elementCache().summaryLists();
+    }
+
```
labkey-danield, line 66: The attributes of the sample, or whatever has the focus, are also shown in the right hand panel next to the canvas/visgraph. I don't see how I get those attributes in this control. There is also an "Overview" and "Lineage" link at the top of the panel as well (next to the items name).

src/org/labkey/test/components/experiment/LineageGraph.java
labkey-danield: This won't work for the lineage page for a sample in Biologics either (it is the same page as in SampleManager).

labkey-danield (changes requested):
LineageGraph_PR_Feedback
I found the name NodeDetails and NodeDetailItems a little confusing, especially when comparing them to the UI.

src/org/labkey/test/components/experiment/NodeDetailItem.java
labkey-chrisj: I can't resolve Locator.union from anywhere; suggestions?
I considered using Locator.findAnyElement(description, context, Locator... locators) but didn't like the idea of not really knowing which it is
I ended up with what's there because it worked

src/org/labkey/test/components/experiment/NodeDetails.java
labkey-chrisj: Title is an attribute on the component element, its text is a prefix like Data: + [the text shown in the name].  It's the preferable way to search; the alternative is to find them all and pick the one with a matching name

src/org/labkey/test/components/experiment/NodeDetails.java
labkey-chrisj: I'm not especially happy with the nomenclature I've started with either.  
Essentially the NodeDetails class is a collection of inputs or outputs (or edges, in graph parlance?) where the members of that collection are NodeDetailItems.
Let's talk about better naming

src/org/labkey/test/components/experiment/LineageGraph.java
labkey-chrisj: I've added a function to the top-level component that will return the data in the table as a map.

Worth noting that it's in a perfectly good responsiveGrid that we use in lots of places to show key/value pairs, but which doesn't have this utility (of extrcting columnName/value).  Think we should extend responsiveGrid with a special subType that will pull keys/values out of this kind of table

src/org/labkey/test/components/experiment/LineageGraph.java
labkey-chrisj: For now, it isn't.

src/org/labkey/test/components/experiment/NodeDetails.java
labkey-chrisj: I've re-done the taxonomy now- the top-level component is called the LineageGraph
The details panel has one or many NodeDetailGroups (which can be a run, or a sampleset)
each NodeDetailGroup has one or many NodeDetail instances under it.

Each NodeDetail has a title, a text, an icon, might or might not link to someplace else, has an overview link, maybe a lineage link

src/org/labkey/test/components/experiment/LineageGraph.java
labkey-chrisj: I have one that works in both labkey and in biologics.
I've asked Nick to put a class on the top-level component so we aren't reliant on a class that's style-related.
He will put that into his current branch, we'll need to update/align once his commit is in

src/org/labkey/test/components/core/DetailComponentTable.java
```diff
@@ -0,0 +1,106 @@
...
+    public String getTableValue(String fieldKey)
+    {
+        return elementCache().getTableValue(fieldKey);
+    }
+
+    public List<String> fieldKeys()
+    {
+        return elementCache().fieldKeys();
+    }
+
```
labkey-danield, line 58: Should add a getter for the Component name. The h4 tag with class attribute of lineage-name-data.

src/org/labkey/test/components/experiment/NodeDetail.java
```diff
@@ -0,0 +1,120 @@
...
+            getWrapper().clickAndWait(elementCache().overviewLink);
+        else
+            elementCache().overviewLink.click();
+    }
+
+    public void clickLineageGraphLink()
+    {
+        getWrapper().mouseOver(getComponentElement());
+        getWrapper().waitFor(()-> elementCache().lineageGraphLink.isEnabled(), 1000);
+        getWrapper().clickAndWait(elementCache().lineageGraphLink);
```
labkey-danield: I think you will need to add a wait parameter to this function call as well. You cannot depend upon clicking lineage to open/load a new page.

src/org/labkey/test/components/experiment/NodeDetailGroup.java
```diff
@@ -0,0 +1,121 @@
...
+        else
+            return elementCache().summary.getText();
+    }
+
+    public NodeDetail getItem(String itemName)
+    {
+        return  elementCache().items().stream().filter(a-> a.getName().equals(itemName))
+                .findFirst().orElseThrow();
+    }
+
```
labkey-danield, line 38: I'm still unclear what the difference is between these two functions (getItem & getItemByTitle). They both take a string value and return a NodeDetail object. What is the difference between an nodes name and its title?

labkey-danield (changes requested):

src/org/labkey/test/components/experiment/NodeDetail.java
labkey-chrisj: good call.

src/org/labkey/test/components/core/DetailComponentTable.java
labkey-chrisj: This is implemented in a function on the LineageGraph class; the name h4 (along with the overview and lineage links and its icon) is outside of the DetailComponentTable's componentElement.

LineageGraph now has the following getters for the current/selected node:
getFocusedNodeData() - gets a map of the contents of the detailComponentTable
getFocusedNodeImage() - returns the icon/image
getFocusedNodeText() - returns the text of the h4.lineage-name-data element

src/org/labkey/test/components/experiment/NodeDetailGroup.java
labkey-chrisj: image
Title is an attribute on the li element of the NodeDetail, Name is the contents of the link/span that are shown in the body.  Title has a prefix like "Data: " + Name.  This prefix seems to be related to the grouping (like in the image).
We can take away name and force callers to go in via title if that's preferable?  I've been supporting both ways with the thought that it might be easier to round-trip some data if it comes out as name but not title

labkey-tchad (approved):
Requested documentation has been added.

labkey-danield (approved):

PR #316, opened 2020-04-20
Fb parent sample editing automation

labkey-danield:
#### Rationale
Added some support controls for sample detail pages.

#### Related Pull Requests
https://github.com/LabKey/sampleManagement/pull/232

#### Changes
- Added ParentEntityEditPanel.java. Panel used to edit lineage for a sample.
- Added comment about a bug in the react select control.
- Added two functions to the DataGenerator. 
- - A static method to delete a domain (matches the static create method).
- - A static function to see if a domain exists.


src/org/labkey/test/components/glassLibrary/ParentEntityEditPanel.java
```diff
@@ -0,0 +1,475 @@
...
+    {
+        return getAllTypeCombo().get(index);
+    }
+
+    /**
+     * Get the type combos by it's index (zero based). This would include any 'add new' combos.
+     *
+     * @return A combo at the given position in the colection of combos..
+     */
+    protected ReactSelect getTypeCombo(String labelText)
```
labkey-tchad: Update copy/pasted javadoc to match method.

src/org/labkey/test/components/glassLibrary/ParentEntityEditPanel.java
```diff
@@ -0,0 +1,475 @@
...
+                .followingLabelWithSpan(labelText)
+                .find(this);
+    }
+
+    /**
+     * Get the type combos by it's index (zero based). This would include any 'add new' combos.
+     *
+     * @return A combo at the given position in the colection of combos..
+     */
+    protected ReactSelect getTypeComboWithSelection(String selection)
```
labkey-tchad: Update copy/pasted javadoc to match method.

src/org/labkey/test/components/glassLibrary/ParentEntityEditPanel.java
```diff
@@ -0,0 +1,475 @@
...
+     * @return A collection of all of the combos that select the type.
+     */
+    protected List<ReactSelect> getAllTypeCombo()
+    {
+        // Calling ReactSelect.finder(getDriver()).findAll(this) is not specific enough, it returns all divs with
+        // 'select' in the class. That can return multiple controls for one react select (div.Select, div.Select-control,
+        // div.Select-value,div.Select-input). That is if there is only one element selected. If it is a multi-select
+        // control each selected value will have div.Select-value.
+
+        List<ReactSelect> allControls = ReactSelect.finder(getDriver()).findAll(this);
```
labkey-tchad: This sounds like a bug in the `ReactSelect` finder. You might be able to fix this problem somewhat by changing [`BaseReactSelect.Locators#selectContainer`](https://github.com/LabKey/testAutomation/blob/develop/src/org/labkey/test/components/glassLibrary/components/BaseReactSelect.java#L352) to use `tagWithClass` instead of `tagWithClassContaining`. Though that would get the multi-select id combos so it doesn't really help here.
You might be able to avoid all this rigamarole though by using your own `Locator`.
```
List<ReactSelect> typeCombos = ReactSelect.finder(getDriver())
        .locatedBy(Locator.tag("div").withClasses("Select", "Select--single"))
        .findAll(this);
```

src/org/labkey/test/components/glassLibrary/ParentEntityEditPanel.java
```diff
@@ -0,0 +1,475 @@
...
+     * @return A collection of all of the combos that select the sample or source ids.
+     */
+    protected List<FilteringReactSelect> getAllIdCombo()
+    {
+        // Calling ReactSelect.finder(getDriver()).findAll(this) is not specific enough, it returns all divs with
+        // 'select' in the class. That can return multiple controls for one react select (div.Select, div.Select-control,
+        // div.Select-value,div.Select-input). That is if there is only one element selected. If it is a multi-select
+        // control each selected value will have div.Select-value.
+
+        List<FilteringReactSelect> allControls = FilteringReactSelect.finder(getDriver()).findAll(this);
```
labkey-tchad: Might be able to use your own Locator to avoid sifting through to find what you want.
```
List<ReactSelect> idCombos = FilteringReactSelect.finder(getDriver())
        .locatedBy(Locator.tag("div").withClasses("Select", "Select--multi"))
        .findAll(this);
```

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -411,4 +413,22 @@ public static TestDataGenerator createDomain(String containerPath, DomainProps d
...
         return def.getTestDataGenerator(containerPath);
     }
+
+    static public CommandResponse deleteDomain(final String containerPath, final String schema, final String domainName)
+            throws IOException, CommandException
+    {
+        Connection connection = WebTestHelper.getRemoteApiConnection();
+        DropDomainCommand delCmd = new DropDomainCommand(schema, domainName);
+        return delCmd.execute(connection, containerPath);
+    }
```
labkey-tchad: I'd use the same terminology as the `DropDomainCommand` for these parameters; `query` or `queryName` instead of `domainName`

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -411,4 +413,22 @@ public static TestDataGenerator createDomain(String containerPath, DomainProps d
 
         return def.getTestDataGenerator(containerPath);
     }
+
+    static public CommandResponse deleteDomain(final String containerPath, final String schema, final String domainName)
+            throws IOException, CommandException
```
labkey-tchad: I usually catch and rethrow `IOException` wrapped in a `RuntimeException`. It generally means something catastrophic has happened (like the server isn't responding) and there isn't much a caller would/could do about it.

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -411,4 +413,22 @@ public static TestDataGenerator createDomain(String containerPath, DomainProps d
...
+    }
+
+    static public boolean doesDomainExists(final String containerPath, final String schema, final String domainName)
+            throws IOException, CommandException
+    {
+        GetQueriesCommand qcmd = new GetQueriesCommand(schema);
+        GetQueriesResponse resp = qcmd.execute(WebTestHelper.getRemoteApiConnection(), containerPath);
+
+        return resp.getQueryNames().contains(domainName);
+    }
```
labkey-tchad: `GetDomainCommand` would be more appropriate here. As is, this could return true for a query that doesn't actually have a domain.
You'd need to catch `CommandException` and check `getStatusCode()` for a 404 for your `false` case, otherwise return true.

labkey-tchad (changes requested):
This class is gorgeous! Loving the JavaDoc.

src/org/labkey/test/components/glassLibrary/ParentEntityEditPanel.java
labkey-danield: Fixed.

src/org/labkey/test/components/glassLibrary/ParentEntityEditPanel.java
labkey-danield: Fixed.

src/org/labkey/test/components/glassLibrary/ParentEntityEditPanel.java
labkey-danield: Good call! I changed the function to return this collection.

src/org/labkey/test/components/glassLibrary/ParentEntityEditPanel.java
labkey-danield: Good call! I changed the function to return this collection.
I did move the comment explaining why I did this to the select finder in the shared ui components. I think I should open a test bug on this issue..

src/org/labkey/test/util/TestDataGenerator.java
labkey-danield: Changed to queryName.

src/org/labkey/test/util/TestDataGenerator.java
labkey-danield: Modified.  Copied the same try/catch from createDomain.

src/org/labkey/test/util/TestDataGenerator.java
labkey-danield: Ok, I think I updated it the way you suggested. At least the test passed with the change.

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -404,31 +404,74 @@ public static TestDataGenerator createDomain(String containerPath, DomainProps d
...
+            exist = response.getStatusCode() != 404;
+        }
+        catch (CommandException ce)
+        {
+            exist = false;
+        }
+        catch (IOException ioe)
+        {
+            throw new RuntimeException("IO exception while looking for the domain.", ioe);
+        }
```
labkey-tchad: If the status code is 400 or higher, the `execute` will throw a `CommandException`. If the command succeeds, you're good 👍. The `CommandException` is where you should check for the 404 since it could represent other problems.
```suggestion
        try
        {
            DomainResponse response = cmd.execute(connection, containerPath);
            return true;
        }
        catch (CommandException ce)
        {
            if (ce.getStatusCode() == 404)
            {
                return false;
            }
            throw new RuntimeException("Exception while looking for the domain.", ce);
        }
        catch (IOException ioe)
        {
            throw new RuntimeException("IO exception while looking for the domain.", ioe);
        }
```

labkey-tchad (approved):
Just a small correction for remoteapi behavior. `doesDomainExist` will mostly work as written but it could mask some non-404 errors.

src/org/labkey/test/util/TestDataGenerator.java
labkey-danield: Thanks for the clarification. Updated.

PR #317, opened 2020-04-21
Update audit message

labkey-susanh:
#### Rationale
Audit message changed to be more consistent with other messages.  Need to update tests accordingly.


labkey-danield (approved):
I don't know, this looks like a very risky change. But I'll approve it anyway.

PR #322, opened 2020-04-29
7066: Invalid Dataset API call update

RosalineP:
#### Rationale
Creating a dataset that is demographics data and also has a third key column is not valid.

#### Changes
Toggle 'demographics' to false. Since the test isn't testing any demographics or third key behavior, this change shouldn't effect test coverage.


cnathe (changes requested):
We will need to change this feature branch name in order to get TeamCity to run this change with the fb for the platform changes (i.e. fb_datasetDesignerStory3_7066).

PR #324, opened 2020-04-30
Create package derived module suite subsets

labkey-tchad:
Use test packages to define more targeted suites:
Consider the test `org.labkey.test.tests.moduleName.featureArea.subArea.TestClass
Currently, such a test would be added to the following suites:
 - moduleName
 - featureArea
 - subArea
This change would, additionally, create the following suites:
 - moduleName.featureArea
 - moduleName.featureArea.subArea

#### Rationale
We want to trigger subsets of module-specific tests without hard-coding test lists on TeamCity. Sample Manager tests already have a package structure that indicates what product sub-area they cover. This will leverage that pattern to define new suites made up of particular sub-areas.

_Note: this also cleans up numerous IntelliJ warnings._

### DIFF
--- a/src/org/labkey/junit/runner/WebTestProperties.java
+++ b/src/org/labkey/junit/runner/WebTestProperties.java
@@ -34,7 +34,7 @@ public abstract class WebTestProperties
     private static ModuleMap associatedModules = new ModuleMap();
     private static final List<String> installedModules = getInstalledModules();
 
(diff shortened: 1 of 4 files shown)


labkey-tchad:
I updated the paused Sample Manager Sample Set suites to verify this feature: https://teamcity.labkey.org/viewQueued.html?itemId=984027

labkey-danield (approved):

PR #329, opened 2020-05-06
RlabkeyTest fixes related to labkey.domain.create changes with study datasets

cnathe:
#### Rationale
The RlabkeyTest has a few test cases that make use of the labkey.domain.create API helper from Rlabkey to test cases related to study dataset creation and issue list def creation. Recent server side API changes for both study dataset and issue list def look to have caused failures in this test. This PR fixes those issues (see changes mentioned below).

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1146

#### Changes
* RlabkeyTest.testRlabkeyStudyApi remove an invalid category test now that we do allow for new category creation during labkey.domain.create

### DIFF
--- a/data/api/rlabkey-api-study.xml
+++ b/data/api/rlabkey-api-study.xml
@@ -27,23 +27,7 @@
         </url>
         <response>
             <![CDATA[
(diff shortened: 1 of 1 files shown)


RosalineP (approved):
Looks good!

PR #330, opened 2020-05-11
Fb dataset designer test conversion

labkey-chrisj:
#### Rationale
This branch provides automation to activate experimental switches for tests to interact with the new DatasetDesigner UI, and it refactors tests to do that.

#### Related Pull Requests
https://github.com/LabKey/compliance/pull/84
https://github.com/LabKey/premium/pull/188
https://github.com/LabKey/platform/pull/1164

#### Changes

-  Add components and pages to wrap the new UI
- Insert calls to enable Experimental switch to show the new UI in helper code/main entry points into this UI
- refactor tests to support this


src/org/labkey/test/components/domain/DomainFormPanel.java
```diff
@@ -173,6 +180,9 @@ public DomainFormPanel setInferFieldFile(File file)
         return this;
     }
 
+    // TODO: add ability to select "import data"/"don't import" here, after inferring fields from file
+    // for datasets, the ability to map key columns is exposed when 'import data' is selected
```
cnathe, line 184: Where are the helpers for that "Import Data" toggle for the list designer case? Or did the list tests not need to implement that at this time?

src/org/labkey/test/pages/EditDatasetDefinitionPage.java
```diff
@@ -15,267 +15,258 @@
...
+//        WebElement additionalKeyDataFieldParentSpan = Locator.xpath("//span[descendant::input[@id='button_dataField']]").findWhenNeeded(this);
+//        WebElement additionalKeyNoneParentSpan = Locator.xpath("//span[descendant::input[@id='button_none']]").findWhenNeeded(this);
+//
+//        Locator.XPathLocator additionalKeyDisabledNone = Locator.xpath("//span[contains(@class,'gwt-RadioButton-disabled')]/input[@id='button_none']");
+//        Locator.XPathLocator additionalKeyDataFieldDisabled = Locator.xpath("//span[contains(@class,'gwt-RadioButton-disabled')]/input[@id='button_none']");
+//        Locator.XPathLocator additionalKeyMangedFieldDisabled = Locator.xpath("//span[contains(@class,'gwt-RadioButton-disabled')]/input[@id='button_none']");
+//
+//        PropertiesEditor fieldsEditor = PropertiesEditor.PropertiesEditor(getDriver()).withTitleContaining("Dataset Fields").findWhenNeeded();
+//    }
+//}
```
cnathe: Looks like this whole file has been commented out, should it be deleted?

src/org/labkey/test/pages/dataset/EditDatasetDefinitionPage.java
```diff
@@ -0,0 +1,282 @@
...
+    {
+        return elementCache().descriptionInput.get();
+    }
+
+    // properties panel interactions
+    // get/set category
+    public EditDatasetDefinitionPage setCategory(String category)
+    {
+        expandPropertiesPanel();
+        elementCache().categorySelect.filterSelect(category);
```
cnathe: Note that this "Category" input is the only dropdown in this panel which also allows for a text input that does not match a dropdown option. This is for the case of allowing a new category to be created for this dataset.

src/org/labkey/test/pages/dataset/EditDatasetDefinitionPage.java
```diff
@@ -0,0 +1,282 @@
...
+        new RadioButton(elementCache().dataRowRadioBtn(type.getIndex()).findElement(getDriver())).check();
+        return this;
+    }
+
+    public boolean isAdditionalKeyManagedEnabled()
+    {
+        expandPropertiesPanel();
+        return elementCache().additionalKeyFieldRadioBtn.isChecked() &&
+                elementCache().keyFieldSelect.hasValue() &&
+                elementCache().keyPropertyManagedBox.isChecked();
```
cnathe: the method name indicates that this is a check for if the checkbox is enabled or not but this last part of the statement would only return true if it was also checked

src/org/labkey/test/pages/dataset/EditDatasetDefinitionPage.java
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    public boolean isAdditionalKeyDataFieldEnabled()
+    {
+        return  elementCache().keyFieldSelect.isEnabled();
```
cnathe: The method name and the check here don't seem to match. If this is just checking the enabled state of the keyFieldSelect, then the "Data" part of the method name should be dropped.

src/org/labkey/test/pages/dataset/EditDatasetDefinitionPage.java
```diff
@@ -0,0 +1,282 @@
...
+    {
+        getWrapper().doAndWaitForPageToLoad(()-> elementCache().saveButton.click());
+        return new DatasetPropertiesPage(getDriver());
+    }
+
+    public enum LookupAdditionalKeyColType
+    {
+        NONE("Participants only (demographic data)", 0),
+        DATAFIELD("Participants and visits", 1),
+        MANAGEDFIELD("Participants, visits, and additional key field", 2);
```
cnathe: The name of this enum and these values is very misleading at this point given the way this UI was refactored. I would suggest that we change this enum to be DataRowUniquenessType and the values to be DEMOGRPHIC, PTID_TIMEPOINT, ADDITIONAL_KEY

src/org/labkey/test/pages/DatasetPropertiesPage.java
```diff
@@ -32,13 +34,22 @@ public DatasetPropertiesPage(WebDriver driver)
...
+
     protected DatasetPropertiesPage.ElementCache newElementCache()
     {
         return new DatasetPropertiesPage.ElementCache();
     }
 
     public EditDatasetDefinitionPage clickEditDefinition()
     {
+        ExperimentalFeaturesHelper.enableExperimentalFeature(
+                createDefaultConnection(true),"experimental-reactdesigner");
```
labkey-tchad, line 52: There are numerous failures in tests that don't take this route to the dataset domain designer and fail because they get the old designer.
Just set this flag for all the tests from the jump. A good spot would be in `BWDT.doPreamble`, immediately after signing in.

src/org/labkey/test/pages/EditDatasetDefinitionPage.java
```diff
@@ -15,267 +15,258 @@
...
-        WebElement additionalKeyNoneParentSpan = Locator.xpath("//span[descendant::input[@id='button_none']]").findWhenNeeded(this);
-        
-        Locator.XPathLocator additionalKeyDisabledNone = Locator.xpath("//span[contains(@class,'gwt-RadioButton-disabled')]/input[@id='button_none']");
-        Locator.XPathLocator additionalKeyDataFieldDisabled = Locator.xpath("//span[contains(@class,'gwt-RadioButton-disabled')]/input[@id='button_none']");
-        Locator.XPathLocator additionalKeyMangedFieldDisabled = Locator.xpath("//span[contains(@class,'gwt-RadioButton-disabled')]/input[@id='button_none']");
-
-        PropertiesEditor fieldsEditor = PropertiesEditor.PropertiesEditor(getDriver()).withTitleContaining("Dataset Fields").findWhenNeeded();
-    }
-}
+//public class EditDatasetDefinitionPage extends BaseDesignerPage<EditDatasetDefinitionPage.ElementCache>
```
labkey-tchad: Delete this file.

src/org/labkey/test/components/dataset/AdvancedDatasetSettingsDialog.java
```diff
@@ -0,0 +1,142 @@
+package org.labkey.test.components.dataset;
```
labkey-tchad: Packages should be named based on the module they are related to or the JavaScript component they wrap. Something like:
```suggestion
package org.labkey.test.components.study;
```
_(but don't put it in the platform repository)_

src/org/labkey/test/pages/dataset/EditDatasetDefinitionPage.java
```diff
@@ -0,0 +1,282 @@
...
+import org.openqa.selenium.WebElement;
+
+import java.util.List;
+
+import static org.hamcrest.CoreMatchers.hasItem;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.labkey.test.WebDriverWrapper.WAIT_FOR_PAGE;
+import static org.labkey.test.WebDriverWrapper.waitFor;
+
+public class EditDatasetDefinitionPage extends DomainDesigner<EditDatasetDefinitionPage.ElementCache>
```
labkey-tchad: Add a reference to the product code wrapped by this page. I presume it is "`study/src/client/DatasetDesigner/DatasetDesigner.tsx`".
Also, rename the class to match: `DatasetDesignerPage`

src/org/labkey/test/pages/dataset/EditDatasetDefinitionPage.java
```diff
@@ -0,0 +1,282 @@
+package org.labkey.test.pages.dataset;
```
labkey-tchad:
```suggestion
package org.labkey.test.pages.study;
```

src/org/labkey/test/pages/DefineDatasetPage.java
```diff
@@ -16,6 +16,7 @@
 package org.labkey.test.pages;
 
 import org.labkey.test.Locator;
+import org.labkey.test.pages.dataset.EditDatasetDefinitionPage;
```
labkey-tchad: Even on develop, `DefineDatasetPage` class is unused. Delete it.

src/org/labkey/test/pages/DatasetPropertiesPage.java
```diff
@@ -18,6 +18,8 @@
 import org.labkey.test.Locator;
```
labkey-tchad, line 18: Move to appropriate package:
`org.labkey.test.pages.study`

src/org/labkey/test/pages/ManageDatasetsPage.java
```diff
@@ -16,7 +16,8 @@
 package org.labkey.test.pages;
```
labkey-tchad, line 16: Move to appropriate package:
```suggestion
package org.labkey.test.pages.study;
```

src/org/labkey/test/pages/ManageDatasetsPage.java
```diff
@@ -48,10 +49,12 @@ public void clickManageDatasetSecurity()
         clickAndWait(elementCache().manageDatasetSecurityLink);
     }
 
-    public CreateDatasetPage clickCreateNewDataset()
+    public EditDatasetDefinitionPage clickCreateNewDataset()
     {
+        ExperimentalFeaturesHelper.enableExperimentalFeature(
+                createDefaultConnection(true),"experimental-reactdesigner");
```
labkey-tchad, line 55: As mentioned earlier, just set this globally.

src/org/labkey/test/pages/dataset/EditDatasetDefinitionPage.java
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    public enum LookupAdditionalKeyColType
+    {
+        NONE("Participants only (demographic data)", 0),
+        DATAFIELD("Participants and visits", 1),
+        MANAGEDFIELD("Participants, visits, and additional key field", 2);
+
+        private String _label;
+        private Integer _index;
```
labkey-tchad: Make final:
```suggestion
        private final String _label;
        private final Integer _index;
```

src/org/labkey/test/pages/dataset/EditDatasetDefinitionPage.java
```diff
@@ -0,0 +1,282 @@
...
+            this._index = index;
+        }
+    }
+
+    public enum ShareDemographicsBy
+    {
+        NONE("No"),
+        PTID("Share by ParticipantId");
+
+        private String _option;
```
labkey-tchad: Make final

src/org/labkey/test/pages/dataset/EditDatasetDefinitionPage.java
cnathe: @labkey-chrisj @labkey-sweta FYI, I am making this change now and will have it checked in shortly.

src/org/labkey/test/pages/dataset/EditDatasetDefinitionPage.java
```diff
@@ -0,0 +1,282 @@
...
+                .findWhenNeeded(rowUniquenessContainer));
+
+        private WebElement keyFieldRow = Locator.XPathLocator.union(
+                Locator.tagWithClass("div", "dataset_data_row_element_show"),
+                Locator.tagWithClass("div", "dataset_data_row_element_hide"))
+                .containing("Additional Key Field").refindWhenNeeded(propertiesPanel);
+        protected ReactSelect keyFieldSelect = ReactSelect.finder(getDriver()).findWhenNeeded(keyFieldRow);
+
+        Checkbox keyPropertyManagedBox = new Checkbox(Locator.inputById("keyPropertyManaged")
+                .findWhenNeeded(propertiesPanel));
```
labkey-tchad: Element cache members should, generally, be final.

src/org/labkey/test/pages/study/CreateDatasetPage.java
```diff
@@ -17,8 +17,8 @@
 
 import org.labkey.test.Locator;
 import org.labkey.test.components.html.Input;
-import org.labkey.test.pages.EditDatasetDefinitionPage;
 import org.labkey.test.pages.LabKeyPage;
+import org.labkey.test.pages.dataset.EditDatasetDefinitionPage;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
```
labkey-tchad: This class is (almost) unused. Remove last usage and delete it.

src/org/labkey/test/pages/study/ManageDatasetsPage.java
```diff
@@ -4,6 +4,8 @@
 import org.labkey.test.WebDriverWrapper;
```
labkey-tchad: There are two `ManageDatasetsPage`s. Get rid of one of them.

labkey-tchad (changes requested):

src/org/labkey/test/pages/dataset/EditDatasetDefinitionPage.java
labkey-chrisj: Yes, the page puts a FilteringReactSelect there because its 'filterSelect' function supports that scenario.  At least one test sets a category that doesn't exist yet, and it works.

src/org/labkey/test/pages/dataset/EditDatasetDefinitionPage.java
labkey-chrisj: Thanks.  It was confusing before and maybe became more confusing after the refactor.

src/org/labkey/test/pages/EditDatasetDefinitionPage.java
labkey-chrisj: Yes, that was the intention.  I'd only left it in to provide a reference to the old one.

src/org/labkey/test/components/domain/DomainFormPanel.java
labkey-chrisj: I'm positive that I implemented support for that slider.  I don't know where to find it, tho.

src/org/labkey/test/pages/EditDatasetDefinitionPage.java
labkey-chrisj: Yes.  I left it in at first so I could refer to the old one while doing conversion.

src/org/labkey/test/components/dataset/AdvancedDatasetSettingsDialog.java
labkey-chrisj: moved it to the existing study package there

src/org/labkey/test/pages/dataset/EditDatasetDefinitionPage.java
labkey-chrisj: cool.

src/org/labkey/test/pages/study/ManageDatasetsPage.java
labkey-chrisj: I saw that there were different classes, one under test/pages/study, the other under test/pages.  I didn't know if they were there and separate for a reason, wanted to avoid biting off too much in this change.  But OK, let's do this

src/org/labkey/test/tests/visualization/ColumnChartTest.java
```diff
@@ -86,42 +87,41 @@ private void doSetup()
...
+        domainFormPanel.getField(PREGNANCY_COLUMN_NAME)
+                .setDimension(true)
+                .setMeasure(false);
         DATA_SOURCE_1_DIMENSIONS.add(PREGNANCY_COLUMN_NAME);
 
-        editor.selectField(LANGUAGE_COLUMN_NAME);
-        pane.setDimension(true);
-        pane.setMeasure(false);
+        domainFormPanel.getField(LANGUAGE_COLUMN_NAME)
+            .setMeasure(true)
```
cnathe: @labkey-chrisj this should be `setDimension(true)`

src/org/labkey/test/pages/DefineDatasetPage.java
labkey-chrisj: okay, gone.

src/org/labkey/test/tests/visualization/ColumnChartTest.java
labkey-chrisj: lol whoops

src/org/labkey/test/pages/study/CreateDatasetPage.java
```diff
@@ -17,7 +17,6 @@
 
 import org.labkey.test.Locator;
 import org.labkey.test.components.html.Input;
-import org.labkey.test.pages.EditDatasetDefinitionPage;
 import org.labkey.test.pages.LabKeyPage;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
```
labkey-tchad: I think this `CreateDatasetPage` is for the old designer. Delete it?

labkey-tchad (changes requested):
Please move `DatasetPropertiesPage` and `ManageDatasetsPage` into the `...pages.study` package

labkey-tchad (approved):

src/org/labkey/test/pages/study/CreateDatasetPage.java
labkey-chrisj: Looks not to be used anywhere- I think it was there to wrap the create-domain view which the old UI went through before showing an editor created thereby.  I'll remove it.

PR #334, opened 2020-05-18
Fb dataset from file test migration

labkey-chrisj:
#### Rationale
updates tests that upload file information into dataset at create-time.  Also addresses product issues found while converting these tests.

#### Related Pull Requests
https://github.com/LabKey/platform/pull/1185

#### Changes
pull out GWT dataset designer code, navigate via new UI by default
pull out calls to activate the exp switch to cause callers to go via new UI

### DIFF
--- a/src/org/labkey/test/pages/study/DatasetDesignerPage.java
+++ b/src/org/labkey/test/pages/study/DatasetDesignerPage.java
@@ -9,6 +9,7 @@ import org.labkey.test.components.glassLibrary.components.ReactSelect;
 import org.labkey.test.components.html.Checkbox;
 import org.labkey.test.components.html.Input;
 import org.labkey.test.components.html.RadioButton;
+import org.labkey.test.components.html.ToggleButton;
 import org.labkey.test.components.study.AdvancedDatasetSettingsDialog;
 import org.labkey.test.pages.DatasetPropertiesPage;
 import org.openqa.selenium.WebDriver;
@@ -168,8 +169,6 @@ public class DatasetDesignerPage extends DomainDesigner<DatasetDesignerPage.Elem
         return  elementCache().keyFieldSelect.isEnabled();
     }
 
-    // get/select additional key field
-    // get/set 'let server manage fields to make entries unique' checkbox
 
     public DatasetDesignerPage setShowInOverview(boolean checked)
     {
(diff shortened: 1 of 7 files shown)


src/org/labkey/test/pages/study/DatasetDesignerPage.java
```diff
@@ -272,5 +300,19 @@ protected Locator dataRowRadioBtn(Integer index)
 
         protected final Checkbox keyPropertyManagedBox = new Checkbox(Locator.inputById("keyPropertyManaged")
                 .findWhenNeeded(propertiesPanel));
+
+        // this is only shown when inferring fields from a file
+        protected ToggleButton autoImportSlider()
```
labkey-tchad: Use consistent naming.
```suggestion
        protected ToggleButton autoImportToggle()
```

src/org/labkey/test/tests/visualization/ColumnChartTest.java
```diff
@@ -83,12 +83,8 @@ private void doSetup()
         selectQuery("study", DATA_SOURCE_1);
         click(Locator.linkWithText("edit definition"));
 
-        waitForText("Edit Dataset Definition");
-
         log("Set the '" + PREGNANCY_COLUMN_NAME + "', '" + LANGUAGE_COLUMN_NAME + "', and '" + SIGNATURE_COLUMN_NAME + "' fields to be dimensions but not measures.");
-
         DatasetDesignerPage datasetDesignerPage = new DatasetDesignerPage(getDriver());
```
labkey-tchad: This should probably wait for a page load. The `waitForText` was compensating for its absence and this might cause intermittent race conditions (depending how much the designer page waits for things). Might as well use the page class while we're at it.
```suggestion
        DatasetDesignerPage datasetDesignerPage = new DatasetPropertiesPage(getDriver()).clickEditDefinition();
```

labkey-tchad (changes requested):
Fix for a possible race condition and a naming nit.

src/org/labkey/test/tests/visualization/ColumnChartTest.java
labkey-chrisj: good call, I like it.
Unfortunately, the suggested edit didn't work when I tested locally.  DatasetDesignerPage does wait until the fields panel element is rendered, so we are likely okay there- but, I've modified the click(edit definition) to clickAndWait (which does work)

labkey-chrisj: For context, we're clicking 'Edit Definition' from a query-begin.view page in schema browser- do we have a page class for that?  Would be nice to have pages for those

labkey-tchad: Ahh, my bad. I just saw the "edit definition" and assumed it was on the dataset designer.
As for a schema browser page class, agreed. Currently, most of the functionality is sitting in `BaseWebDriverTest` (`selectSchema`, `selectQuery`, etc.).

labkey-tchad (approved):

PR #336, opened 2020-05-20
Fb sample type test automation

labkey-danield:
#### Rationale
Need to fill some test gaps in the Sample Type automation. 

#### Changes
* Moved the lineage tests to their own test class.
* Moved 'stress' tests to their own test class.
* Renamed test cases to be more meaningful.
* Added tests for lineage update test gaps.
* General clean up of other test cases.


labkey-danield:
For the most part a lot of this is just moving existing tests to new tests classes. All of the lineage tests were moved to SampleTypeLineageTest. Two new tests were added:
testUpdateLineageUsingFileImport
testDeriveSampleByUI

I also added comments to testSampleTypeAndLineageInSubfolder to help clarify what the test was doing.

I moved tests that are 'stress' tests to the SampleTypeLimitsTests. If they fail they tend to impact other tests that follow and are difficult to run on a local machine.

src/org/labkey/test/tests/SampleTypeLineageTest.java
```diff
@@ -0,0 +1,1088 @@
...
+        checker().verifyEquals("Number of entries in the column '" + columnName + "' is not as expected.",
+                expectedValues.size(), dataInTable.size());
+
+        for(String expectedValue : expectedValues)
+        {
+            checker().verifyTrue(String.format("Value '%s' was not shown in column '%s' in data region '%s'.",
+                    expectedValue, columnName, dataRegionName),
+                    dataInTable.contains(expectedValue));
+
+        }
```
labkey-tchad: Consider using `verifyEquals`. It could catch some errors that this would miss (wrong numbers of duplicated values).
```suggestion
        Collections.sort(expectedValues);
        Collections.sort(dataInTable);
        checker().verifyEquals("Entries in the column '" + columnName + "' are not as expected.",
                expectedValues, dataInTable);
```

labkey-tchad (approved):
We should move these tests to an appropriate package (`org.labkey.test.tests.experiment`). I'd also consider moving them to `platform/experiment/test/...`; next to `VocabularyViewSupportTest`.
If they stay in `testAutomation`, the new sample data should live here as well; in `testAutomation/data`

labkey-susanh (commented):
Test cases look good to me.

labkey-danield:
@labkey-tchad Moving these test to experiment sounds like a good idea, I would like to do that in a separate story. I would also like to change the name of the other older test classes to SampleType...

I will add a new story to the back log for these two tasks.

PR #337, opened 2020-05-20
Fix for ListTest.doRenameFieldsTest failure timing issue by removing the assertTextPresent check and using the fields panel helpers

cnathe:
#### Rationale
The ListTest.doRenameFieldsTest test intermittently fails because of the assertTextPresent() call when it goes to the list designer and tries to verify that a column name is present before changing it. This PR changes that test to use the fields panel helpers instead, which will do the same validation that the column name is present, but should be more reliable.


cnathe:
TeamCity run of ListTest was successful:
https://teamcity.labkey.org/project.html?projectId=LabKey_Trunk&branch_LabKey_Trunk=listTestRenameFieldsFix

labkey-tchad (approved):

PR #340, opened 2020-05-21
Item 7350: GpatAssayTest fix for updating assay results domain properties via domain designer UI

cnathe:
#### Rationale
With the refactor of the "Create New General Assay Design" from file scenario so that it no longer uses the GWT PropertiesEditor (see related PR), the GpatAssayTest needs an update so that it goes through the assay designer for the results property updates that it was previously making on the TsvImportAction page.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1204
* https://github.com/LabKey/provenance/pull/23

#### Changes
* GpatAssayTest fix for updating assay results domain properties via domain designer UI

### DIFF
--- a/src/org/labkey/test/Locator.java
+++ b/src/org/labkey/test/Locator.java
@@ -841,11 +841,6 @@ public abstract class Locator extends By
         return tagWithAttribute("input", "type", "checkbox").withPredicate(xpath("../../following-sibling::td/span").containing(label));
     }
 
-    public static XPathLocator gwtNextButtonOnImportGridByColLabel(String label)
-    {
-        return tagWithClass("div", "x-tbar-page-next").withPredicate(xpath("../preceding-sibling::td/span").containing(label));
-    }
-
     public static Locator permissionRendered()
     {
         return Locators.pageSignal("policyRendered");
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/Locator.java
```diff
@@ -841,11 +841,6 @@ public static XPathLocator gwtCheckBoxOnImportGridByColLabel(String label)
         return tagWithAttribute("input", "type", "checkbox").withPredicate(xpath("../../following-sibling::td/span").containing(label));
     }
 
-    public static XPathLocator gwtNextButtonOnImportGridByColLabel(String label)
-    {
-        return tagWithClass("div", "x-tbar-page-next").withPredicate(xpath("../preceding-sibling::td/span").containing(label));
-    }
-
     public static Locator permissionRendered()
```
labkey-tchad, line 839: Can we get rid of some of these other gwt locators?

labkey-tchad (approved):
Maybe get rid of more of those GWT locators but looks good.

src/org/labkey/test/Locator.java
cnathe: Done. I got rid of any used gwt locators in this file.

labkey-danield (approved):

PR #351, opened 2020-05-31
Fb grid row

labkey-chrisj:
#### Rationale
Refactors the shared test Wrapper components related to grids to support Biologics tests, and updates Biologics tests to use the shared components instead of the biologics-only ones.

#### Related Pull Requests
https://github.com/LabKey/sampleManagement/pull/281
https://github.com/LabKey/provenance/pull/26
https://github.com/LabKey/biologics/pull/598

#### Changes
Refactor shared components to support scenarios in biologics as well


src/org/labkey/test/components/glassLibrary/grids/GridRow.java
```diff
@@ -0,0 +1,203 @@
...
+    final ResponsiveGrid _grid;
+    private Map<String, String> _rowMap = null;
+
+    public GridRow(ResponsiveGrid grid, WebElement element, WebDriver driver)
+    {
+        _el = element;
+        _driver = driver;
+        _grid = grid;
+    }
+
```
labkey-danield, line 34: It would be useful to add JavaDoc comments above the public functions.

src/org/labkey/test/components/glassLibrary/grids/GridRow.java
```diff
@@ -0,0 +1,203 @@
...
+    public boolean isSelected()
+    {
+        return hasSelectColumn() && elementCache().selectCheckbox.isSelected();
+    }
+
+    public GridRow select(boolean checked)
+    {
+        elementCache().selectCheckbox.set(checked);
```
labkey-danield, line 62: Would it be more robust to put a check for hasSelectColumn before trying to click the check box?

src/org/labkey/test/components/glassLibrary/grids/GridRow.java
```diff
@@ -0,0 +1,203 @@
...
+    {
+        WebElement link = getLink(text);
+        String href = link.getAttribute("href");
+        link.click();
+        log("waiting for url to be: [" + href + "]");
+        WebDriverWrapper.waitFor(()-> getWrapper().getURL().toString().endsWith(href), 1000);
+        getWrapper().shortWait().until(ExpectedConditions.stalenessOf(link));
+    }
+
+    public String getValue(String columnText)
```
labkey-danield: I personally don't think getValue is a good choice of a name for this and the following function, I think it is a little misleading. You are not getting the value of the cell you are getting the text.
What if in the future a cell contains something like an img, a check-box, or just a color to convey information, getValue implies that this will return the appropriate value for those type of cells.
Also calling getWrapper().**getTexts** is also a bit of a giveaway what is being returned.

src/org/labkey/test/components/glassLibrary/grids/GridRow.java
```diff
@@ -0,0 +1,203 @@
...
+    {
+        return Locator.tag("td").index(colIndex).findElement(this);
+    }
+
+    public WebElement getCell(String columnHeader)
+    {
+        return getCell(_grid.getColumnIndex(columnHeader));
+    }
+
+    public boolean contains(Locator loc)
```
labkey-danield: Why not let the caller just do the loc.existsIn(localReferenceToGridRow)?
Alternatively why not also add an overloaded function that takes a WebElement?

src/org/labkey/test/components/glassLibrary/grids/GridRow.java
```diff
@@ -0,0 +1,203 @@
...
+        List<String> columnValues = getWrapper().getTexts(Locator.css("td")
+                .findElements(getComponentElement()));
+        if (hasSelectColumn())
+            columnValues.remove(0);
+        return columnValues;
+    }
+
+    public Map<String, String> getRowMap()
+    {
+        if (_rowMap == null)
```
labkey-danield, line 157: Isn't there a danger with this check that it assumes the data in the row has not changed since it was last retrieved?
Which, I think, begs the larger question of our UI component libraries... should they be stateless or stateful?

src/org/labkey/test/components/glassLibrary/grids/GridRow.java
```diff
@@ -0,0 +1,203 @@
...
+
+    public void clickLink(String text)
+    {
+        WebElement link = getLink(text);
+        String href = link.getAttribute("href");
+        link.click();
+        log("waiting for url to be: [" + href + "]");
+        WebDriverWrapper.waitFor(()-> getWrapper().getURL().toString().endsWith(href), 1000);
+        getWrapper().shortWait().until(ExpectedConditions.stalenessOf(link));
+    }
```
labkey-danield, line 126: This feels like a lot of work that is better left to the test code (i.e. clickAndWait(Locator.linkWithText(...). Where the link has a search scope of the row, or better yet a cell.

src/org/labkey/test/components/glassLibrary/grids/GridRow.java
```diff
@@ -0,0 +1,203 @@
...
+            return this;
+        }
+
+        public GridRowFinder withDescendant(Locator.XPathLocator descendant)
+        {
+            _locator = _xPathLocator.withDescendant(descendant);
+            return this;
+        }
+
+        public GridRowFinder withValue(String value)
```
labkey-danield: I would rename this to withCellContaining or withContainingText.

src/org/labkey/test/components/glassLibrary/grids/GridRow.java
```diff
@@ -0,0 +1,203 @@
...
+        {
+            _locator = _xPathLocator.withChild(Locator.tagWithText("td", value));
+            return this;
+        }
+
+        public GridRowFinder withValueAtColumnIndex(String value, int columnIndex)
+        {
+            _locator = _xPathLocator.withChild(Locator.tag("td").index(columnIndex).withText(value));
+            return this;
+        }
```
labkey-danield, line 245: For completeness I think if you are going to have a finder function taking a columnIndex you should also have one taking a columnText

src/org/labkey/test/components/glassLibrary/grids/GridRow.java
```diff
@@ -0,0 +1,203 @@
...
+    public GridRow select(boolean checked)
+    {
+        elementCache().selectCheckbox.set(checked);
+        return this;
+    }
+
+    public WebElement getCell(int colIndex)
+    {
+        return Locator.tag("td").index(colIndex).findElement(this);
+    }
```
labkey-danield, line 82: This function creates an inconsistency with what is and is not a column.
In this function the checkbox column is treated the same as every other column, it would be at index 0. But in later code we ignore the checkbox column. That just feels a little inconsistent to me.

src/org/labkey/test/components/glassLibrary/grids/QueryGrid.java
```diff
@@ -95,11 +158,58 @@ public GridTabBar getGridTabBar()
...
+
     public int getRecordCount()
     {
         return _gridBar.getRecordCount();
     }
 
+    public QueryGrid waitForRecordCount(int expectedCount)
+    {
+        WebDriverWrapper.waitFor(()-> getRecordCount() == expectedCount,
+                "did not get to the expected record count ["+expectedCount+"] in time",  WAIT_FOR_JAVASCRIPT);
```
labkey-danield:
```suggestion
                "Did not get to the expected record count ["+expectedCount+"] in time.",  WAIT_FOR_JAVASCRIPT);
```
I think it would be good to add an interface where the user is allowed to set the timeout limit, especially since the name of the interface is **wait**ForRecordCount.

src/org/labkey/test/components/glassLibrary/components/OmniBox.java
```diff
@@ -175,9 +176,9 @@ public OmniBox setSearch(String searchTerm)
         return this.setText("search \"" + searchTerm + "\"");
     }
 
-    public OmniBox setSort(String columnName, boolean descending)
+    public OmniBox setSort(String columnName, SortDirection direction)
     {
-        return this.setText("sort \"" + columnName + "\"" + (descending ? " desc" : ""));
+        return this.setText("sort \"" + columnName + "\"" + (direction.equals(SortDirection.DESC) ? " desc" : ""));
```
labkey-danield, line 181: Not really sure what this has to do with the grid row change.

src/org/labkey/test/components/glassLibrary/grids/GridRow.java
```diff
@@ -0,0 +1,203 @@
...
+    {
+        return getRowMap().get(columnText);
+    }
+
+    public List<String> getValues()
+    {
+        List<String> columnValues = getWrapper().getTexts(Locator.css("td")
+                .findElements(getComponentElement()));
+        if (hasSelectColumn())
+            columnValues.remove(0);
```
labkey-danield, line 147: This is the part of the code I commented on above. Here the select column is not treated the same as the other columns.

src/org/labkey/test/components/glassLibrary/grids/QueryGrid.java
```diff
@@ -51,10 +58,58 @@ public WebDriver getDriver()
...
+    // get rowMaps
+
+    public Map<String, String> getRowMap(int rowIndex)
     {
-        _responsiveGrid = new ResponsiveGrid.ResponsiveGridFinder(_driver).find(_queryGridPanel);
-        return _responsiveGrid;
+        return getGrid().getRowMap(rowIndex);
+    }
+
+    public Map<String, String> getRowMap(String containsText)
```
labkey-danield: I think naming the parameter **contains**Text can be a little misleading. What is ultimately called is withText which does a "=" comparison. Using the word contains implies you can give a partial string.
For example look at our locators (withClassContaining), you can give a partial string for it to resolve. In that case the contains here means something different than the contains elsewhere in our code.

src/org/labkey/test/components/glassLibrary/grids/QueryGrid.java
```diff
@@ -51,10 +58,58 @@ public WebDriver getDriver()
...
+    public List<Map<String, String>> getRowMaps()
+    {
+        return getGrid().getRowMaps();
+    }
+
+    // row selection
+
+    public QueryGrid selectRow(int index, boolean checked)
+    {
+        getGrid().getRows().get(index).select(checked);
```
labkey-danield: Another place where a check could be done that the grid has a select column before trying to click it.

src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
```diff
@@ -87,36 +87,33 @@ public Boolean hasData()
         return elementCache().emptyGrid.isEmpty();
     }
 
-    public List<WebElement> getRows()
-    {
-        return elementCache().getRows();
-    }
-
-    // TODO Responsive grids cannot be sorted, this should be moved to QueryGrid (or what ever it will be called).
```
labkey-danield, line 95: Are you absolutely sure that all grids can be sorted?

src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
```diff
@@ -159,18 +154,31 @@ public ResponsiveGrid selectRow(String columnName, List<String> columnValues, bo
         return this;
     }
 
-    // TODO I don't think Responsive grids have a checkbox need to verify with dev. If not this should be moved to QueryGrid (or what ever it will be called).
     public boolean isRowSelected(int index)
     {
-        WebElement checkBox = elementCache().getRowCheckbox(index);
-        return checkBox.isSelected();
+        GridRow row = new GridRow.GridRowFinder(this).index(index).find(this);
+        return row.isSelected();
```
labkey-danield:
```suggestion
        return new GridRow.GridRowFinder(this).index(index).find(this).row.isSelected();
```

src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
```diff
@@ -159,18 +154,31 @@ public ResponsiveGrid selectRow(String columnName, List<String> columnValues, bo
...
-    public ResponsiveGrid selectAllOnPage(boolean checked)
+    protected Checkbox selectAllBox()
     {
         Checkbox box = elementCache().selectAllCheckbox;
         getWrapper().shortWait().until(ExpectedConditions.elementToBeClickable(box.getComponentElement()));
+        return box;
+    }
+
+    // TODO I don't think Responsive grids have a checkbox need to verify with dev. If not this should be moved to QueryGrid (or what ever it will be called).
+    public boolean areElementsSelected()
```
labkey-danield: Maybe change to areAllRowsOnPageSelected. Why call them elements and not rows?

src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
```diff
@@ -186,100 +194,52 @@ public ResponsiveGrid selectAllOnPage(boolean checked)
...
+    public List<GridRow> getRows()
     {
-        int colIndex = getColumnIndex(columnHeader);
-        return Locator.xpath("//tbody/tr[./td[" + (colIndex + 1) + "]/span[text()=normalize-space('" + exactColumnText + "')]]")
-                .findOptionalElement(this).orElseThrow(() -> new NoSuchElementException("Could not find a row where column '" + columnHeader + "' had a value of '" + exactColumnText + "'."));
+        return new GridRow.GridRowFinder(this).findAll(getComponentElement());
     }
 
-    public Integer getRowIndex(Locator containing)
+    protected WebElement getCell(int rowIndex, String columnHeader)
```
labkey-danield: Kind of a big change from getRowIndex to getCell.

src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
```diff
@@ -391,29 +338,33 @@ public void initGridData()
...
 
-    private WebElement getLink(String text)
+    public void clickLink(String text)
     {
-        log("seeking link with text [" + text + "]");
-        WebElement link = getComponentElement().findElement(By.partialLinkText(text));
-        getWrapper().scrollIntoView(link);
-        log("found element with text [" + link.getText() + "]");
-        return link;
+        getRow(text).orElseThrow(()-> new NotFoundException("did not find a row with a link with text ["+text+"]"))
```
labkey-danield:
```suggestion
        getRow(text).orElseThrow(()-> new NotFoundException("Did not find a row with a link with text ["+text+"]"))
```

src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
```diff
@@ -291,71 +251,58 @@ public boolean hasSelectColumn()
...
+        return elementCache().columnNames;
     }
 
+    /**
+     * there are ways to get rowMaps without exposing indexes to outside use
+     * @param rowIndex
+     * @return
+     */
+    @Deprecated
     public List<String> getRowValues(int rowIndex)
```
labkey-danield: Not getting values getting the text.

src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
```diff
@@ -291,71 +251,58 @@ public boolean hasSelectColumn()
...
-        Map<String, String> rowMap = new HashMap<>();
-        for (int i=0; i < tableColumns.size(); i++)
-        {
-            rowMap.put(tableColumns.get(i), rowValues.get(i));
-        }
-        return  rowMap;
+        return new GridRow(this, row, getDriver()).getRowMap();
     }
 
     public String getCellValue(int rowIndex, String columnHeader)
```
labkey-danield:
```suggestion
    public String getCellText(int rowIndex, String columnHeader)
```

src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
```diff
@@ -291,71 +251,58 @@ public boolean hasSelectColumn()
...
     {
-        return getRowMap(rowIndex).get(columnHeader);
+        return getRow(rowIndex).getValue(columnHeader);
     }
 
     public String getCellValue(WebElement row, String columnHeader)
```
labkey-danield:
```suggestion
    public String getCellText(WebElement row, String columnHeader)
```

labkey-danield (changes requested):

src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
labkey-chrisj: Yes, it was a big change to eliminate getRowIndex- fortunately, there weren't many usages of it and I refactored them out to have tests rely on object references rather than having to know how indexes work in the row.

Also, by the time I was done refactoring, there were also no usages of getCell

src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
labkey-chrisj: changing it to getRowTexts.

src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
labkey-chrisj: If possible, I'd like to eliminate usage of webElements as input parameters like the one at line 308.  Likewise, on 303 the external reliance on rowIndex.  Because line 308 isn't being used from anywhere, I'm cutting it

src/org/labkey/test/components/glassLibrary/grids/GridRow.java
labkey-chrisj: I initially wrote the synchronization/wait-for-url-to-update code to work around the fact that clicking a link doesn't get a page refresh in a single-page app. 
Previously the clickLink() function was on the grid, I wanted to push the implementation down into the row as a way to avoid multiple implementations of the same thing.

Speaking of which, there are several use cases that call for more than just text to get an element in a row or cell- I'd like to add better support for these cases like an overload allowing the test code to pass in a locator.  Tests can still get a row or cell and use it as a search scope that way, so win-win?

src/org/labkey/test/components/glassLibrary/grids/GridRow.java
labkey-chrisj: Yes, we have to be careful about caching anything in React.  Well worth discussing.
Part of what I'd like this component to handle is the test code assumptions that (for example) rowIndices won't change. Where possible I'd like test code to depend on object references and let the objects handle the indexing work.  The other major objective of this component is to consolidate implementation/supporting needed cases so that if product implementation changes, we don't have to go and touch so much test code

src/org/labkey/test/components/glassLibrary/grids/GridRow.java
labkey-chrisj: Let's keep the text!=value idea, but avoid the implication that 'containing' means partial text matching

src/org/labkey/test/components/glassLibrary/grids/GridRow.java
labkey-chrisj: Actually, the public-facing functions taking a columnText do the work of getting an index and calling this one.  I'd like to make it protected and discourage test code from using indexes as much as possible.

src/org/labkey/test/components/glassLibrary/grids/GridRow.java
labkey-chrisj: having this contains() method on the object supports streaming over a collection of GridRows, and it doesn't prevent test code from doing its own loc.existsIn(localRefToRow).

I like the idea of supporting as many use cases as possible; when we consolidate implementation into the component, future product changes can require that much less need to touch test code

src/org/labkey/test/components/glassLibrary/grids/GridRow.java
labkey-chrisj: Yes. This is the right call and it gets to the heart of why I want to handle indexes internally.  Internally, the grid maintains separate indices for raw and normalized column index (mapped to column header) for those cases where a selector row being present shifts them sideways.
The remaining use case for it is in heatmap usage; the columns change when the month does. I've made the tests using it figure out the correct current and previous months, and made this function private

src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
labkey-chrisj: Not all can.  The ones that don't have drop-menus in the header cells and the ones that don't have omniBoxes can't be sorted.  Until we have a separate class for un-sortableGrid, or one for a SortableGrid, it doesn't hurt to support those functions here and if someone tries to call these functions on grids that can't be sorted, they'll find out quickly enough

labkey-danield (approved):

src/org/labkey/test/components/glassLibrary/grids/GridRow.java
labkey-chrisj: Adding the ones I missed now

src/org/labkey/test/components/glassLibrary/components/ReactCheckBox.java
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright (c) 2019 LabKey Corporation. All rights reserved. No portion of this work may be reproduced in
+ * any form or by any electronic or mechanical means without written permission from LabKey Corporation.
+ */
```
labkey-tchad: Delete copyright block. This file should be `Apache 2.0` and we have scripts to handle it anyway.

src/org/labkey/test/components/glassLibrary/components/ReactCheckBox.java
```diff
@@ -0,0 +1,86 @@
...
+package org.labkey.test.components.glassLibrary.components;
+
```
labkey-tchad: We should probably discontinue the `glassLibrary` package since the UI components repository isn't called that anymore.
Depending on where the checkbox is defined, this class probably belongs somewhere like `org.labkey.test.components.bootstrap` or `org.labkey.test.components.react`.

src/org/labkey/test/components/glassLibrary/components/ReactCheckBox.java
```diff
@@ -0,0 +1,86 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.labkey.test.WebDriverWrapper;
+import org.labkey.test.components.html.Checkbox;
+import org.openqa.selenium.WebElement;
+
+/* this component wraps the ternary checkbox that uses hidden attributes 'checked' and 'indeterminate' to signal its state.
+* if the box is 'checked', the hidden attribute called 'checked' equals 'true'- otherwise it is null
+* if the box is 'indeterminate', that attribute is true - otherwise it is null
+* if both are null, the box is unchecked */
+public class ReactCheckBox extends Checkbox
```
labkey-tchad: Please indicate the specific product component that is instrumented by this class.

src/org/labkey/test/components/glassLibrary/grids/EditableGrid.java
```diff
@@ -271,6 +288,57 @@ public void setCellValue(int row, String columnName, Object value)
...
+        Clipboard c = Toolkit.getDefaultToolkit().getSystemClipboard();
+        StringSelection sel = new StringSelection(pasteText);
+        c.setContents(sel, sel);
+
+        new Actions(getDriver())
+                .keyDown(cmdKey)
+                .sendKeys("v")       // paste the contents into whatever has focus now
+                .keyUp(cmdKey)
+                .perform();
+        return this;
```
labkey-tchad: Could you use `getWrapper().actionPaste()` for this? Maybe update it to paste without a target if a WebElement isn't provided.

src/org/labkey/test/components/glassLibrary/grids/GridRow.java
labkey-tchad: I agree with Dan on this. I'm not a fan of duplicating basic functionality all over the place.

src/org/labkey/test/components/glassLibrary/grids/GridRow.java
```diff
@@ -0,0 +1,269 @@
...
+    final WebDriver _driver;
+    final ResponsiveGrid _grid;
+    private Map<String, String> _rowMap = null;
+
+    public GridRow(ResponsiveGrid grid, WebElement element, WebDriver driver)
+    {
+        _el = element;
+        _driver = driver;
+        _grid = grid;
+    }
```
labkey-tchad: Make `GridRow` constructor private. Only `GridRowFinder` knows how to find one properly.

src/org/labkey/test/components/glassLibrary/grids/GridRow.java
```diff
@@ -0,0 +1,269 @@
...
+
+    protected class ElementCache extends Component<?>.ElementCache
+    {
+        public Optional<WebElement> selectColumn = Locator.xpath("//td/input[@type='checkbox']")
+                .findOptionalElement(getComponentElement());
+        public ReactCheckBox selectCheckbox = new ReactCheckBox(Locator.tagWithAttribute("input", "type", "checkbox")
+            .findWhenNeeded(this));
+        public WebElement column = Locator.tag("tr").findWhenNeeded(getComponentElement());
+    }
```
labkey-tchad: Remove unused `column` element.

src/org/labkey/test/components/glassLibrary/grids/QueryGrid.java
```diff
@@ -85,6 +195,14 @@ public String getGridErrorMessage()
         return errorMsg;
     }
 
+    // subcomponent getters
+
+    public ResponsiveGrid getGrid()
+    {
+        _responsiveGrid = new ResponsiveGrid.ResponsiveGridFinder(_driver).find(_queryGridPanel);
+        return _responsiveGrid;
+    }
```
labkey-tchad: Could we have `QueryGrid` just extend `ResponsiveGrid`? Having to call `getGrid` from a grid is not intuitive to me.

src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
```diff
@@ -128,49 +124,125 @@ private void sortColumn(String columnName, boolean descending)
...
+     *      unchecked (meaning no rows on the page are selected)
+     * @return
+     */
+    public boolean areAllRowsOnPageSelected()
+    {
+        Checkbox box = selectAllBox();
+
+        String isIndeterminate = box.getComponentElement().getAttribute("indeterminate");
+        return box.isChecked() || (isIndeterminate != null && isIndeterminate.equals("true"));
+    }
```
labkey-tchad: The logic here seems incorrect. If an indeterminate state indicates that some rows are NOT selected, I would expect this to return `false` in that case.

labkey-tchad (changes requested):
There is a pattern in a lot of these helpers matching text in a column to put the text first in the parameter list (`getRow(text, column)`). This feels backward to me. I would expect a `key, value` ordering (`getRow(column, text)`).

src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
labkey-chrisj: It's also redundant.  ReactCheckBox also handles this logic internally.

src/org/labkey/test/components/glassLibrary/grids/QueryGrid.java
labkey-chrisj: I'll try.  We may have to re-think the way we're using ResponsiveGrid as a subcomponent; there may be timing/initialization/lifecycle errors that arise out of the change.

src/org/labkey/test/components/glassLibrary/grids/EditableGrid.java
labkey-chrisj: Yes, I'll update it accordingly.

src/org/labkey/test/components/glassLibrary/grids/GridRow.java
labkey-chrisj: Okay.

src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
```diff
@@ -447,44 +491,39 @@ protected final WebElement getColumnHeaderCell(String headerText)
...
                 {
                     headerCellElements.remove(0);
+                    offset = 1;
                 }
                 columnNames = getWrapper().getTexts(headerCellElements);
+                indexes = new HashMap<>();
                 for (int i = 0; i < headerCellElements.size(); i++)
                 {
                     headerCells.put(columnNames.get(i), headerCellElements.get(i)); // Fill out the headerCells Map since we have them all
+                    indexes.put(columnNames.get(i), Map.of("normalizedIndex", i, "rawIndex", i + offset));
```
labkey-tchad: Make a little [POJO](https://en.wikipedia.org/wiki/Plain_old_Java_object) for these indices instead of using magical map keys.

labkey-chrisj: done

src/org/labkey/test/components/glassLibrary/components/ReactCheckBox.java
labkey-chrisj: I'm not aware of any specific product components for this; it is most-useful for the select-all checkboxes on grids that have the magic 'indeterminate' state

src/org/labkey/test/components/glassLibrary/components/ReactCheckBox.java
labkey-chrisj: Okay.  I'll start moving classes out of that package, into other locations.
I'll ping you if I have questions about where they should be

labkey-chrisj: Okay, I've moved the reactCheckBox class into `org.labkey.test.components.react`.  I'm not sure where we want the rest of the classes in the `glassLibrary `package moved, so let's do that work in a different story

labkey-tchad (approved):

PR #352, opened 2020-06-01
Fb fix ModeratorReviewTest

labkey-danield:
#### Rationale
It looks like messages are taking a little longer to show up in dumpster. Modified the test to allow for this delay.

#### Related Pull Requests
* none

#### Changes
* Added code to refresh the page and check again if the message isn't there. Fail after 5 retries.

### DIFF
--- a/src/org/labkey/test/tests/announcements/ModeratorReviewTest.java
+++ b/src/org/labkey/test/tests/announcements/ModeratorReviewTest.java
@@ -159,12 +159,15 @@ public class ModeratorReviewTest extends BaseWebDriverTest
         goToProjectHome();
         impersonate(user);
         log("Insert message");
+        log("Currently impersonating user:" + user);
         // make title unique in test scope, to check email notifications later
         title = title + " " + System.currentTimeMillis();
+        log("Inserting message with title " + title);
         InsertPage.beginAt(this)
                 .setTitle(title)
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/announcements/ModeratorReviewTest.java
```diff
@@ -159,12 +159,15 @@ private String insertMessage(String user, String title, boolean expectAutoApprov
...
         log("Insert message");
+        log("Currently impersonating user:" + user);
         // make title unique in test scope, to check email notifications later
         title = title + " " + System.currentTimeMillis();
+        log("Inserting message with title " + title);
         InsertPage.beginAt(this)
                 .setTitle(title)
                 .setBody(title)
                 .submit();
+        log("Submission of message (supposedly) complete");
```
RosalineP, line 170: This was a for-my-own debugging message, maybe it should be removed?

RosalineP (approved):
Looks good! Left one comment.

src/org/labkey/test/tests/announcements/ModeratorReviewTest.java
labkey-danield: Extra logging in a test case is not a bad thing, and it may turn out to be useful at some future date.

PR #362, opened 2020-06-09
Move sampledata files from SVN to appropriate repositories

Sigmonia:
#### Rationale
As part of git migration we need to move the SVN sampledata folder's contents to appropriate git repos


labkey-susanh (approved):
I will admit that I have not read all of these files, but folder organization seems reasonable.  It seems there may be a good amount of test data we don't actually need and there are folder zip files that should be expanded to be not zip files, but that's something for a different day.

labkey-tchad (approved):

PR #369, opened 2020-06-16
Make tests work with CAS-only server

labkey-tchad:
#### Rationale
Our automated tests should be more flexible about the configuration of the server they wish to test. Many tests currently require database authentication against a server running locally with a full dev enlistment and a test user with site admin permission. Most if not all of these restrictions are not actually necessary for most of our automated tests.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1351
* LabKey/ldap#14
* LabKey/nlp#166
* LabKey/compliance#88
* LabKey/cas#26
* LabKey/biologics#627 
* https://github.com/LabKey/labkey-api-java/pull/7

#### Changes
* Add `webtest.primary.app.admin` parameter to allow tests to alter their behavior when running as a non site-admin.
* Add `webtest.without.test.modules` parameter to allow tests to alter their behavior when test modules aren't installed.
* Update remoteapi `Connection` helpers to copy browser session more aggressively
* Update `WebDriverWrapper` helpers to use Java API instead of JavaScript to get current user info
* Adjust some tests to not use admin-only APIs while impersonating


src/org/labkey/test/WebTestHelper.java
```diff
@@ -753,47 +757,4 @@ public int hashCode() {
...
-        public String getShape()
-        {
-            return _shape;
-        }
-
-        public String getTitle()
-        {
-            return _title;
-        }
-    }
```
labkey-danield, line 798: I'm assuming this was removed because it wasn't used (basic clean up work).

src/org/labkey/test/WebTestHelper.java
```diff
@@ -389,27 +394,26 @@ public static String buildRelativeUrl(String controller, @Nullable String contai
 
     public static Connection getRemoteApiConnection()
     {
-        return getRemoteApiConnection(false);
+        return getRemoteApiConnection(true);
```
labkey-danield, line 397: Any particular reason why this was changed from false to true?

src/org/labkey/test/tests/GpatAssayTest.java
```diff
@@ -240,20 +214,41 @@ private void setAssayResultsProperties()
...
-        setFormElement(Locator.name("AssayDesignerName"), assayName);
-        fireEvent(Locator.xpath("//input[@id='AssayDesignerName']"), SeleniumEvent.blur);
+        startCreateGpatAssay(fnaFile, assayName);
 
         clickButton("Begin import");
         clickButton("Next", defaultWaitForPage);
         clickButton("Save and Finish", defaultWaitForPage);
     }
 
+    @LogMethod
```
labkey-danield, line 227: Do you really want to log the calls to this function when the first line in the function is a log?

src/org/labkey/test/tests/GpatAssayTest.java
```diff
@@ -240,20 +214,41 @@ private void setAssayResultsProperties()
...
+    private void startCreateGpatAssay(File dataFile, @LoggedParam String assayName)
+    {
+        log("Create GPAT assay from " + dataFile.getName());
```
labkey-danield, line 230: Why not add the assayName to this log?

labkey-danield (approved):
Looks ok to me, just some trivial comments/questions.

src/org/labkey/test/WebTestHelper.java
labkey-tchad: I want to encourage tests to use a `Connection` tied to the current browser session unless they have specific reasons not to. Most scenarios don't need a totally fresh `Connection` and this will make tests more likely to work against CAS-only servers.

PR #378, opened 2020-06-24
Add tests for HTML encoding of report names

labkey-sweta:
#### Rationale
Automated coverage for report name HTML encoding

Issue : https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=40558

#### Related Pull Requests
Only this one.

#### Changes
1. Added new test class to DailyB
2. New method in ManageViewPage.


src/org/labkey/test/pages/reports/ManageViewsPage.java
```diff
@@ -63,4 +63,12 @@ public void deleteReport(String reportName)
                 "Failed to delete report: " + reportName, WAIT_FOR_JAVASCRIPT);
     }
 
+    @LogMethod
+    public void editReport(String reportName)
+    {
+        //td[1] is the pencil
+        final Locator report = Locator.tag("tr").withClass("x4-grid-row").containing(reportName).childTag("td").position(1);
+        click(report);
```
labkey-tchad, line 71: If this navigates, we should wait for the page load.
```suggestion
        clickAndWait(report);
```

src/org/labkey/test/tests/CrossSiteScriptingForDeleteTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @Test
+    public void verifyReportNameIsEncoded()
+    {
+        goToProjectHome();
+        goToManageViews();
+
+        log("Adding the link report");
+        new BootstrapMenu.BootstrapMenuFinder(getDriver()).withButtonTextContaining("Add Report").find()
+                .clickSubMenu(true, "Link Report");
```
labkey-tchad: Use the page returned by `goToManageViews`.
```suggestion
        log("Adding the link report");
        goToManageViews().clickAddReport("Link Report");
```

src/org/labkey/test/tests/CrossSiteScriptingForDeleteTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+import java.util.Collections;
+import java.util.List;
+
+@Category({DailyB.class})
+@BaseWebDriverTest.ClassTimeout(minutes = 3)
+public class CrossSiteScriptingForDeleteTest extends BaseWebDriverTest
+{
+    protected static final String PROJECT_NAME = "CrossSiteScriptingForDeleteTest";
+    protected static final String REPORT_NAME = "<span onclick='alert(\"fooled you!\")'>mouseover me, it's totally safe</span>";
```
labkey-tchad: Use an injection string that doesn't require direct interaction. It gives us a chance to catch something we aren't specifically checking for.
```suggestion
    protected static final String REPORT_NAME = BaseWebDriverTest.INJECT_CHARS_1;
```

src/org/labkey/test/tests/CrossSiteScriptingForDeleteTest.java
```diff
@@ -0,0 +1,104 @@
...
+                .clickSubMenu(true, "Link Report");
+
+        setFormElement(Locator.name("viewName"), REPORT_NAME);
+        setFormElement(Locator.name("linkUrl"), WebTestHelper.getContextPath() + LINK_REPORT_URL);
+        clickButton("Save");
+        waitForText("Manage Views");
+
+        log("Clicking on the report - No XSS");
+        click(Locator.linkWithText(REPORT_NAME));
+        switchToMainWindow();
```
labkey-tchad, line 68: Add a `selectReport` method to `ManageViewsPage`. Check `ManageViewsPage.deleteReport` for a Locator that won't open the report.

labkey-tchad (changes requested):

src/org/labkey/test/pages/reports/ManageViewsPage.java
labkey-sweta: It does not navigate, that's why only click.

src/org/labkey/test/tests/CrossSiteScriptingForDeleteTest.java
labkey-sweta: Done

src/org/labkey/test/tests/CrossSiteScriptingForDeleteTest.java
labkey-sweta: Done

src/org/labkey/test/tests/CrossSiteScriptingForDeleteTest.java
labkey-sweta: Done.

labkey-jeckels (changes requested):
This looks like it should prevent regressions on the two delete dialogs that had problems, but do nothing to help us detect any other XSS with report names. There are plenty of other places in the code that could have problems, like we saw with the user DisplayName property. It was only when we started setting the display name for the user our tests were running as that we discovered those other places.

I don't expect the automation changes here to protect against any and all possible report-related XSS, but can we please go more general than this? Lots of other tests create reports. We could update some of them to set report names, descriptions, and other values that verify that we don't have XSS issues.

I think that @labkey-tchad  and @cnathe  both worked on the user display name testing and fixes. Hopefully they will have some suggestions here too.

labkey-tchad (approved):

labkey-tchad:
I think a good target would be to add injection characters to all of the report names, titles, and descriptions in `DataRerportsTest`, `NonStudyReportsTest`, and `ScatterPlotTest` (or any other `GenericChartsTest` subclasses).

src/org/labkey/test/tests/CrossSiteScriptingForDeleteTest.java
```diff
@@ -0,0 +1,98 @@
...
+                .waitFor()
+                .clickButton("Delete", false);
+        deleteMsg = "Are you sure you want to delete \"" + REPORT_NAME + "\"";
+
+        log("Verifying the error message");
+        confirmReportNameInMessage("Delete?", deleteMsg, true);
+
+    }
+
+    public void confirmReportNameInMessage(String windowTitle, String msg, boolean action)
```
labkey-danield, line 88: Personally I would change the name of this to something like confirmReportNameAndDoAction. The current name of the function implies it is just a check. Not a big deal.

You may also want to change the access to private, unless you expect it to be used outside of the test.

labkey-danield (approved):

labkey-jeckels (changes requested):
NonStudyReportsTest isn't passing in the feature branch. ExtHelper.Locators.extDialog() isn't properly escaping the report name when it builds the XPath, which causes the test to fail.

   org.openqa.selenium.InvalidSelectorException: invalid selector: Unable to locate an element with the xpath expression 

I can repro this locally.

labkey-jeckels (changes requested):
Instead of making the validation less rigorous by removing some characters, I'd really prefer that we properly encode the XPath. We must do this for at least some other Locators.

https://stackoverflow.com/questions/642125/encoding-xpath-expressions-with-both-single-and-double-quotes

labkey-tchad:
> Instead of making the validation less rigorous by removing some characters, I'd really prefer that we properly encode the XPath. We must do this for at least some other Locators.
> 

@labkey-sweta `Locator.containing` and `Locator.withText` will escape text properly.

labkey-jeckels (approved):
DailyC is still running, but the test of interest, NonStudyReportsTest, has passed, so it should be safe to merge.

src/org/labkey/test/util/ExtHelper.java
```diff
@@ -565,8 +565,9 @@ private static String jsString(String s)
          */
         public static Locator.XPathLocator extDialog(String title)
         {
-            return Locator.xpath("//span[" + Locator.NOT_HIDDEN + " and contains(@class, 'window-header-text') and contains(string(), '" + title + "')]");
-        }
+            return Locator.xpath("//div").withClass("x4-window").notHidden().
+                    withDescendant(Locator.xpath("//span").withClass("x4-window-header-text").withText(title));
+     }
```
labkey-tchad: This Locator isn't intended to find the window, just the title bar. It also needs to work with ext3 and we shouldn't impose a strict text match where there wasn't one before.
```suggestion
            return Locator.tag("span").notHidden().withAttributeContaining("class", "window-header-text").containing(title);
         }
```

labkey-tchad (changes requested):
`ExtHelper` refactor is a bit too aggressive.

labkey-tchad (approved):

PR #380, opened 2020-06-25
Test Automation for Sample Manager Timeline Tests

labkey-danield:
#### Rationale
Cleaning up some code that was needed for the timeline automation in Sample Manager.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/306

#### Changes
* Updating EditParentsPanel and ParentEntityEditPanel to allow adding a parent to an already existing parent type.
* Added a better check for removing a type.
* Moved shared code to base class.


src/org/labkey/test/components/glassLibrary/ParentEntityEditPanel.java
```diff
@@ -353,14 +353,44 @@ public ParentEntityEditPanel removeType(String typeName)
...
         {
            elementCache().removeButton(index + 1).click();
-           getWrapper().waitFor(()-> getAllTypeCombo().size() < typeCombos.size(),
-                   "The type '" + typeName + "' was not successfully removed.",
-                   1_000);
+
+           // Need to check if this is removing the last/only type.
+           if(typeCombos.size() > 1)
+           {
+               // If it is not removing the last one can simple check that the count of combos is as expected.
```
labkey-susanh:
```suggestion
               // If it is not removing the last one can simply check that the count of combos is as expected.
```

labkey-susanh (approved):

src/org/labkey/test/components/glassLibrary/ParentEntityEditPanel.java
```diff
@@ -353,14 +353,44 @@ public ParentEntityEditPanel removeType(String typeName)
...
+               getWrapper().waitFor(() -> getAllTypeCombo().size() < typeCombos.size(),
+                       "The type '" + typeName + "' was not successfully removed.",
+                       1_000);
+           }
+           else
+           {
+               // If this is removing the last/only one the count of combos will still be 1, so need a different check.
+               ReactSelect rs = getAllTypeCombo().get(0);
```
labkey-chrisj, line 368: This might be out of scope for this work, but if there is at all any way other than index to identify these controls, we ought to consider doing that.

labkey-chrisj (approved):
Approved, with comment about the practice of referencing typeCombos by their index

src/org/labkey/test/components/glassLibrary/ParentEntityEditPanel.java
labkey-danield: There might be, but I think that would be a challenge to address at a later time.

PR #383, opened 2020-06-29
Fb sample delete 7179

labkey-chrisj:
#### Rationale
As we merge SampleManager features and components to Biologics, there is call to pull test components from SM and into shared (testAutomation) location.  These changes support new Biologics tests, in class BiologicsCRUDTest

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/621
https://github.com/LabKey/sampleManagement/pull/311

#### Changes
adds default support for random data generation for columnTypes of date, boolean.


src/org/labkey/test/components/glassLibrary/grids/QueryGrid.java
```diff
@@ -218,6 +218,25 @@ public QueryGrid clearSortsAndFilters()
...
+    /**
+     *
+     * @return
+     */
+    public QueryGrid selectAllInSet()
+    {
+        doAndWaitForUpdate(()->
+                getGridBar().selectAllInSet());
+        return this;
+    }
```
labkey-tchad: Not sure what "set" means here. Does it just select rows on the current page or all rows in the query? Something like `selectAllRows`, `selectVisibleRows`, or `selectRowsOnPage` would be more consistent with terminology elsewhere.

src/org/labkey/test/components/glassLibrary/grids/QueryGrid.java
```diff
@@ -218,6 +218,25 @@ public QueryGrid clearSortsAndFilters()
         return this;
     }
 
+    /**
+     *
+     * @return
+     */
+    public QueryGrid selectAllInSet()
```
labkey-tchad: Fill out this JavaDoc.

src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
```diff
@@ -448,10 +448,10 @@ public void clickLink(String text)
 
     /**
      * locates the first link in the specified column, clicks it, and waits for the URL to update
-     * @param text  text for link to match
      * @param column    column in which to search
+     * @param text  text for link to match
      */
-    public void clickLink(String text, String column)
```
labkey-chrisj, line 454: for consistency.  Elsewhere we follow a key/value ordering

src/org/labkey/test/components/react/dialogs/SampleDeleteDialog.java
```diff
@@ -0,0 +1,18 @@
+package org.labkey.test.components.react.dialogs;
+
+import org.labkey.test.components.experiment.DeleteConfirmationDialog;
+import org.labkey.test.pages.LabKeyPage;
+import org.labkey.test.pages.samplemanagement.samples.SampleTypePage;
+
+public class SampleDeleteDialog<SourcePage extends LabKeyPage> extends DeleteConfirmationDialog<SourcePage, SampleTypePage>
```
labkey-tchad: The build is failing because this references code in `samplemanagment`. `DeleteConfirmationDialog` is the shared component. This is the sample manager specific implementation.

labkey-tchad (changes requested):
Don't use the sample manager confirmation dialog.

src/org/labkey/test/components/glassLibrary/grids/QueryGrid.java
labkey-chrisj: In this context, 'set' refers to the domain.
Agree, we could use better terminology

src/org/labkey/test/components/react/dialogs/SampleDeleteDialog.java
labkey-chrisj: Ha, mildly infuriating to learn that there's a shared component for this after all.

labkey-chrisj: Just sent it back to SM, refactored SM tests to point back to the old location.
Hopefully this doesn't break Biologics/require me to implement a similar pattern over there

src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
labkey-tchad: Please do the same for `selectRows` and `isRowSelected`.
Could you also refactor all of the parameter names. Various methods in `ResponsiveGrid` use 'column', 'columnName', 'columnHeader', or 'columnText'. I don't have a particular preference but we should pick one. I prefer `columnLabel` or `columnText` to indicate we are matching the display text on the column, not the database column name.

labkey-tchad (changes requested):

src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
labkey-chrisj: columnLabel it is, my good man

labkey-tchad (approved):

PR #393, opened 2020-07-08
move test components to shared library to support SM integration to biologics

labkey-chrisj:
#### Rationale
This moves InsertRecord.java and related dialog(s) from SM, to the shared library in support of integrating SM into biologics.
Proposed rename -> SampleCRUDPanel

#### Related Pull Requests
https://github.com/LabKey/sampleManagement/pull/320
https://github.com/LabKey/biologics/pull/628

#### Changes
Update bulk edit lookup method to use FilteringReactSelect, helpful for cases where there are many entities to resolve in the lookup

### DIFF
--- a/src/org/labkey/test/components/glassLibrary/components/BaseReactSelect.java
+++ b/src/org/labkey/test/components/glassLibrary/components/BaseReactSelect.java
@@ -45,9 +45,8 @@ public abstract class BaseReactSelect<T extends BaseReactSelect> extends WebDriv
         {
             WebElement selectMenuElement = Locators.selectMenu.findElementOrNull(getComponentElement());
 
-            if((selectMenuElement != null && selectMenuElement.isDisplayed()) && selectMenuElement.getText().toLowerCase().contains("loading"))
-                waitFor(()-> !selectMenuElement.getText().toLowerCase().contains("loading"),
-                        "React Select is stuck loading.", WAIT_FOR_JAVASCRIPT);
+            if((selectMenuElement != null && selectMenuElement.isDisplayed()) && isLoading())
+                waitForLoaded();
 
             return (selectMenuElement != null && selectMenuElement.isDisplayed()) || hasClass("is-open");
         }
@@ -86,31 +85,27 @@ public abstract class BaseReactSelect<T extends BaseReactSelect> extends WebDriv
 
     public boolean isLoading()
     {
-        // if it's present, we're loading dropdown options
-        return Locators.loadingSpinner.findElementOrNull(getComponentElement()) != null;
+        // if either are present, we're loading options
+        return Locators.loadingSpinner.existsIn(getComponentElement()) ||
+                getComponentElement().getText().equals("Loading...");
     }
 
-    public boolean isLoaded()
+    protected T waitForLoaded()
     {
-        return !isLoading() && !getOptions().contains("Loading...");
-    }
-
-    public T waitForLoaded(int waitMsec)
-    {
-        _wrapper.waitFor(()-> isLoaded(),
-                "Took too long for to become loaded", waitMsec);
+        _wrapper.waitFor(()-> !isLoading(),
+                "Took too long for to become loaded", WAIT_FOR_JAVASCRIPT);
         return (T) this;
     }
 
     public String getValue()
     {
-        return getWrapper().shortWait().withMessage(() -> "Select stuck loading. [" + getComponentElement().getText() + "]").until(ignored -> {
-            String val = getComponentElement().getText();
-            if (val.equals("Loading..."))
-                return null;
-            else
-                return val;
-        });
+        waitForLoaded();
+
+        String val = getComponentElement().getText();
+        if (val.equals("Loading..."))
+            return null;
+        else
+            return val;
     }
 
     public boolean hasOption(String value)
@@ -243,7 +238,7 @@ public abstract class BaseReactSelect<T extends BaseReactSelect> extends WebDriv
 
     public List<String> getSelections()
     {
-
+        waitForLoaded();
         try
         {
             // Wait for at least one of the elements to be visible.
--- a/src/org/labkey/test/components/glassLibrary/components/BaseReactSelect.java
+++ b/src/org/labkey/test/components/glassLibrary/components/BaseReactSelect.java
@@ -30,6 +30,7 @@ public abstract class BaseReactSelect<T extends BaseReactSelect> extends WebDriv
     final WebElement _componentElement;
     final WebDriver _driver;
     final WebDriverWrapper _wrapper;
+    private final String LOADING_TEXT = "loading...";
 
     public BaseReactSelect(WebElement selectOrParent, WebDriver driver)
     {
@@ -87,7 +88,7 @@ public abstract class BaseReactSelect<T extends BaseReactSelect> extends WebDriv
     {
         // if either are present, we're loading options
         return Locators.loadingSpinner.existsIn(getComponentElement()) ||
-                getComponentElement().getText().equals("Loading...");
+                getComponentElement().getText().toLowerCase().equals(LOADING_TEXT);
     }
 
     protected T waitForLoaded()
@@ -102,7 +103,7 @@ public abstract class BaseReactSelect<T extends BaseReactSelect> extends WebDriv
         waitForLoaded();
 
         String val = getComponentElement().getText();
-        if (val.equals("Loading..."))
+        if (val.toLowerCase().equals(LOADING_TEXT))
             return null;
         else
             return val;
--- a/src/org/labkey/test/components/glassLibrary/components/BaseReactSelect.java
+++ b/src/org/labkey/test/components/glassLibrary/components/BaseReactSelect.java
@@ -90,6 +90,18 @@ public abstract class BaseReactSelect<T extends BaseReactSelect> extends WebDriv
         return Locators.loadingSpinner.findElementOrNull(getComponentElement()) != null;
     }
 
+    public boolean isLoaded()
+    {
+        return !isLoading() && !getOptions().contains("Loading...");
+    }
+
+    public T waitForLoaded(int waitMsec)
+    {
+        _wrapper.waitFor(()-> isLoaded(),
+                "Took too long for to become loaded", waitMsec);
+        return (T) this;
+    }
+
     public String getValue()
     {
         return getWrapper().shortWait().withMessage(() -> "Select stuck loading. [" + getComponentElement().getText() + "]").until(ignored -> {
--- a/src/org/labkey/test/components/glassLibrary/components/InsertEntitiesPanel.java
+++ b/src/org/labkey/test/components/glassLibrary/components/InsertEntitiesPanel.java
@@ -14,13 +14,13 @@ import java.util.Map;
 
 // This is the same component (collection of atomic elements) used in insertAssay and createSamples.
 
-public class SampleCRUDPanel extends WebDriverComponent<SampleCRUDPanel.ElementCache>
+public class InsertEntitiesPanel extends WebDriverComponent<InsertEntitiesPanel.ElementCache>
 {
 
     private final WebDriver _driver;
     private final WebElement _editingDiv;
 
-    public SampleCRUDPanel(WebElement element, WebDriver driver)
+    public InsertEntitiesPanel(WebElement element, WebDriver driver)
     {
         _driver = driver;
         _editingDiv = element;
@@ -38,7 +38,7 @@ public class SampleCRUDPanel extends WebDriverComponent<SampleCRUDPanel.ElementC
         return _driver;
     }
 
-    public SampleCRUDPanel addRecords(List<Map<String, Object>> records)
+    public InsertEntitiesPanel addRecords(List<Map<String, Object>> records)
     {
         getWrapper().setFormElement(elementCache().addRowsTxtBox, Integer.toString(records.size()));
         elementCache().addRowsButton.click();
@@ -61,13 +61,13 @@ public class SampleCRUDPanel extends WebDriverComponent<SampleCRUDPanel.ElementC
         return this;
     }
 
-    public SampleCRUDPanel setRecordValues(Map<String, Object> columnValues)
+    public InsertEntitiesPanel setRecordValues(Map<String, Object> columnValues)
     {
         int insertRowIndex = elementCache().grid.listOfnotPopulatedRows().get(0);
         return setRecordValues(columnValues, insertRowIndex);
     }
 
-    public SampleCRUDPanel setRecordValues(Map<String, Object> columnValues, int row)
+    public InsertEntitiesPanel setRecordValues(Map<String, Object> columnValues, int row)
     {
         for(String columnName : columnValues.keySet())
         {
@@ -97,13 +97,13 @@ public class SampleCRUDPanel extends WebDriverComponent<SampleCRUDPanel.ElementC
         return elementCache().grid.isDisplayed();
     }
 
-    public SampleCRUDPanel setAddRows(int numOfRows)
+    public InsertEntitiesPanel setAddRows(int numOfRows)
     {
         getWrapper().setFormElement(elementCache().addRowsTxtBox, Integer.toString(numOfRows));
         return this;
     }
 
-    public SampleCRUDPanel clickAddRows()
+    public InsertEntitiesPanel clickAddRows()
     {
         elementCache().addRowsButton.click();
         return this;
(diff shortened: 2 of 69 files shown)


src/org/labkey/test/components/glassLibrary/components/InsertEntitiesPanel.java
```diff
@@ -0,0 +1,193 @@
...
+import org.openqa.selenium.NoSuchElementException;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+import java.util.List;
+import java.util.Map;
+
+// This is the same component (collection of atomic elements) used in insertAssay and createSamples.
+
+public class InsertEntitiesPanel extends WebDriverComponent<InsertEntitiesPanel.ElementCache>
```
labkey-tchad: Reference the specific product component that this class controls. Also rename to match product component (EntityInsertPanel.tsx).
```suggestion
/**
 * This class automates the UI component defined in <a href="https://github.com/LabKey/labkey-ui-components/blob/master/packages/components/src/components/entities/EntityInsertPanel.tsx">components/entities/EntityInsertPanel.tsx</a>
 * This is the same component (collection of atomic elements) used in insertAssay and createSamples.
 */
public class EntityInsertPanel extends WebDriverComponent<EntityInsertPanel.ElementCache>
```

src/org/labkey/test/components/react/dialogs/BulkCreateSamplesDialog.java
```diff
@@ -0,0 +1,159 @@
...
+    private boolean waitForReactSelectReady(ReactSelect select)
+    {
+        getWrapper().waitFor(()->!select.getSelections().contains("Loading..."), "The reactSelect field did not populate in time.", 1_000);
+        return true;
+    }
+
+    public BulkCreateSamplesDialog setSelectionField(String fieldCaption, List<String> selectValues)
+    {
+        FilteringReactSelect reactSelect = FilteringReactSelect.finder(getDriver()).followingLabelWithSpan(fieldCaption).find();
+        getWrapper().waitFor(()-> reactSelect.isInteractive(), 5000);
```
labkey-danield: Are you sure this will work?
My experience was that the arrow on the reactSelect was showing up enabled when data was still loading. That is why the waitForReactSelectReady was added.

labkey-danield (changes requested):

src/org/labkey/test/components/react/dialogs/BulkCreateSamplesDialog.java
labkey-chrisj: I'm willing to be wrong on this.  But, if waitForReactSelectReady works, I'd be happy to implement a real waitForLoaded method on BaseReactSelect.  

After a bit of investigation, it's clear that `.isInteractive()` was the wrong condition to check. I'm implementing a public 'waitForLoaded' method on BaseReactSelect that uses both the absence of a loading spinner and of the "Loading..." value placeholder to determine whether the control is still loading options.

src/org/labkey/test/components/glassLibrary/components/InsertEntitiesPanel.java
```diff
@@ -0,0 +1,193 @@
+package org.labkey.test.components.glassLibrary.components;
```
labkey-tchad: Avoid using the `glassLibrary` package. Use `org.labkey.test.components.labkey.ui`
```suggestion
package org.labkey.test.components.labkey.ui;
```

src/org/labkey/test/components/glassLibrary/components/BaseReactSelect.java
```diff
@@ -90,6 +90,18 @@ public boolean isLoading()
...
+    public T waitForLoaded(int waitMsec)
+    {
+        _wrapper.waitFor(()-> isLoaded(),
+                "Took too long for to become loaded", waitMsec);
+        return (T) this;
+    }
+
     public String getValue()
     {
         return getWrapper().shortWait().withMessage(() -> "Select stuck loading. [" + getComponentElement().getText() + "]").until(ignored -> {
```
labkey-tchad: There are too many different methods for determining if the select has loaded. `isLoading` and `isLoaded` are redundant; then `getValue` and `isExpanded` do their own thing.

src/org/labkey/test/components/glassLibrary/components/BaseReactSelect.java
```diff
@@ -90,6 +90,18 @@ public boolean isLoading()
...
+    {
+        return !isLoading() && !getOptions().contains("Loading...");
+    }
+
+    public T waitForLoaded(int waitMsec)
+    {
+        _wrapper.waitFor(()-> isLoaded(),
+                "Took too long for to become loaded", waitMsec);
+        return (T) this;
+    }
```
labkey-tchad: I don't see a reason for any of the loaded/loading methods to be public. Any interaction methods should handle waiting internally.
```suggestion
    protected void waitForLoaded(int waitMsec)
    {
        _wrapper.waitFor(()-> isLoaded(),
                "Took too long for select to load: " + getComponentElement().toString(), waitMsec);
    }
```

src/org/labkey/test/components/react/dialogs/BulkCreateSamplesDialog.java
```diff
@@ -0,0 +1,160 @@
+package org.labkey.test.components.react.dialogs;
```
labkey-tchad: This is a LabKey developed component
```suggestion
org.labkey.test.components.labkey.ui.samples;
```

src/org/labkey/test/components/react/dialogs/BulkCreateSamplesDialog.java
```diff
@@ -0,0 +1,160 @@
...
+    {
+        String forField = Locator.xpath("//div[@class='modal-body']//label[./span[contains(text(), '" + fieldCaption + "')]]").findElement(getComponentElement()).getAttribute("for");
+        return getWrapper().getFormElement(Locator.tagWithId("input", forField));
+    }
+
+    private boolean waitForReactSelectReady(FilteringReactSelect select)
+    {
+        getWrapper().waitFor(()->!select.getSelections().contains("Loading..."), "The reactSelect field did not populate in time.", 1_000);
+        return true;
+    }
```
labkey-tchad: This component shouldn't need to determine whether the select is ready and it certainly shouldn't need to know how to do so.

src/org/labkey/test/components/react/dialogs/BulkCreateSamplesDialog.java
```diff
@@ -0,0 +1,160 @@
...
+    protected ElementCache newElementCache()
+    {
+        return (ElementCache) super.elementCache();
+    }
+
+    @Override
+    protected ElementCache elementCache()
+    {
+        return new ElementCache();
+    }
```
labkey-tchad: You have these backward.
```suggestion
    @Override
    protected ElementCache newElementCache()
    {
        return new ElementCache();
    }

    @Override
    protected ElementCache elementCache()
    {
        return (ElementCache) super.elementCache();
    }
```

src/org/labkey/test/components/react/dialogs/BulkCreateSamplesDialog.java
```diff
@@ -0,0 +1,160 @@
...
+
+    public BulkCreateSamplesDialog(WebDriver driver)
+    {
+        this(new ModalDialogFinder(driver).withTitle("Bulk Creation of Samples"));
+    }
+
+    private BulkCreateSamplesDialog(ModalDialogFinder finder)
+    {
+        super(finder);
+    }
```
labkey-tchad: Take an `InsertEntitiesPanel` in the constructor. That way this component could wait for the list to update in `clickAddRows`.
```suggestion
    private final InsertEntitiesPanel _panel;

    public BulkCreateSamplesDialog(InsertEntitiesPanel panel)
    {
        super(new ModalDialogFinder(panel.getDriver()).withTitle("Bulk Creation of Samples"));
        _panel = panel;
    }
```
`InsertEntitiesPanel.getDriver()` needs to be public for this to work (I really wish I'd made `WebDriverComponent.getDriver` public in the first place 🤦)

src/org/labkey/test/components/react/dialogs/BulkCreateSamplesDialog.java
```diff
@@ -0,0 +1,160 @@
...
+        try
+        {
+            if (BootstrapLocators.errorBanner.findElement(getComponentElement()).isDisplayed())
+                throw new IllegalStateException("Error message on dialog: '" + BootstrapLocators.errorBanner.findElement(getComponentElement()).getText() + "'.");
+        }
+        catch(StaleElementReferenceException stale)
+        {
+            // Do nothing if stale.
+        }
+    }
```
labkey-tchad: Wait for dialog to disappear. Maybe wait for the insert page to update somehow as well.
```suggestion
        waitForClose();
        /* Wait for insert panel to update */
    }
```

src/org/labkey/test/components/react/dialogs/BulkCreateSamplesDialog.java
```diff
@@ -0,0 +1,160 @@
...
+        {
+            if (BootstrapLocators.errorBanner.findElement(getComponentElement()).isDisplayed())
+                throw new IllegalStateException("Error message on dialog: '" + BootstrapLocators.errorBanner.findElement(getComponentElement()).getText() + "'.");
+
+        }
+        catch (StaleElementReferenceException stale)
+        {
+            // Do nothing if stale.
+        }
+    }
```
labkey-tchad: Wait for dialog to disappear.
```suggestion
        waitForClose();
    }
```

labkey-tchad (changes requested):

src/org/labkey/test/components/react/dialogs/BulkCreateSamplesDialog.java
labkey-danield: Ok, that sounds like a good approach. The call to .isInteractive() is not a reliable check if the dropdown control is a look up field. This is why the [SMSampleCreateTest.testBulkInsert](https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_SampleManager_SampleManagerAPostgres/1055312?buildTab=tests&pager.currentPage=1) is failing in the branch. A call to isInteractive is returning true but there is a split second between that and the list int he drop down being populated, which is enough time for the test code to type the text which is not found.

src/org/labkey/test/components/react/dialogs/BulkCreateSamplesDialog.java
labkey-danield: I agree. This was a hack/work-around for an issue with the ReactSelect. I'll start an email thread to discuss best solutions.

src/org/labkey/test/components/glassLibrary/components/BaseReactSelect.java
labkey-chrisj: Okay.  I'll make waitForLoaded() protected and call it from interaction/entry points
If the logic is different in derived classes we can override there

src/org/labkey/test/components/react/dialogs/BulkCreateSamplesDialog.java
labkey-chrisj: I've added a call to waitForClose() to the cancel and clickInsertRows methods (right after clicking the buttons, which assumes that the checks for errors are looking for errors in the EntityInsertPane (and not in the dialog)

src/org/labkey/test/components/react/dialogs/BulkCreateSamplesDialog.java
labkey-chrisj: At the moment, SMSampleCreateTest.testBulkInsert is failing for a different reason.
By switching to use FilteringReactSelect.filterSelect(), the test filters the input, and looks for a value that contains the filter clause- but values here don't contain the full filter being supplied.  Easy workaround: omit the Color: prefix, just search for the color ("orange")
I've sent you a patch with this workaround, wanted to give you the input on whether it's the right fix

src/org/labkey/test/components/react/dialogs/BulkCreateSamplesDialog.java
labkey-chrisj: done

src/org/labkey/test/components/react/dialogs/BulkCreateSamplesDialog.java
labkey-chrisj: done

src/org/labkey/test/components/react/dialogs/BulkCreateSamplesDialog.java
labkey-chrisj: BCSD now takes an InsertEntitesPanel in its constructor, getDriver() is now public on InsertEntitiesPanel

src/org/labkey/test/components/glassLibrary/components/BaseReactSelect.java
```diff
@@ -86,19 +85,27 @@ public boolean hasValue()
 
     public boolean isLoading()
     {
-        // if it's present, we're loading dropdown options
-        return Locators.loadingSpinner.findElementOrNull(getComponentElement()) != null;
+        // if either are present, we're loading options
+        return Locators.loadingSpinner.existsIn(getComponentElement()) ||
+                getComponentElement().getText().equals("Loading...");
```
labkey-danield: Sure you don't want to make this check case insensitive, and not dependent upon the ellipse?
`getComponentElement().getText().toLowerCase().contains("loading")`

src/org/labkey/test/components/glassLibrary/components/BaseReactSelect.java
```diff
@@ -86,19 +85,27 @@ public boolean hasValue()
...
-            String val = getComponentElement().getText();
-            if (val.equals("Loading..."))
-                return null;
-            else
-                return val;
-        });
+        waitForLoaded();
+
+        String val = getComponentElement().getText();
+        if (val.equals("Loading..."))
```
labkey-danield: Same question as above for case insensitive and the ellipse.
However if you do want to use `equals("Loading...")` you may want to consider making "Loading..." a constant.

labkey-danield (approved):

src/org/labkey/test/components/glassLibrary/components/BaseReactSelect.java
labkey-tchad: This `if/else` shouldn't even be necessary now that we call `waitForLoaded` at the start of this method.

src/org/labkey/test/components/glassLibrary/components/BaseReactSelect.java
labkey-chrisj: A quick survey of dev code shows that yes, there are cases where lower-case is being used, so I'm down with lower-case.
The ellipsis seems to be universal, however, and I'm worried that `.toLowerCase().contains("loading")` could match actual values.  I propose to do `getComponentElement().getText().toLowerCase().equals("loading...")`.

src/org/labkey/test/components/ui/samples/BulkCreateSamplesDialog.java
```diff
@@ -0,0 +1,155 @@
...
+import org.labkey.test.Locator;
+import org.labkey.test.components.bootstrap.ModalDialog;
+import org.labkey.test.components.glassLibrary.components.FilteringReactSelect;
+import org.labkey.test.components.ui.EntityInsertPanel;
+import org.openqa.selenium.StaleElementReferenceException;
+import org.openqa.selenium.WebElement;
+
+import java.util.List;
+
+public class BulkCreateSamplesDialog extends ModalDialog
```
labkey-tchad: Sorry I didn't notice this earlier. This should be more generic. The panel that pops this dialog is for any kind of entity. I don't think it needs any functional changes, just a rename and a move to a different package:
`org.labkey.test.components.ui.EntityBulkInsertDialog`
Should also update the title passed to the `ModalDialogFinder` in the constructor. Based on the [product code](https://github.com/LabKey/labkey-ui-components/blob/master/packages/components/src/components/entities/EntityInsertPanel.tsx#L779), `"Bulk Creation of"` should do the trick.

labkey-tchad (changes requested):

src/org/labkey/test/components/ui/samples/BulkCreateSamplesDialog.java
labkey-chrisj: I think this is a good catch.  I'll do it.

labkey-tchad (approved):

PR #409, opened 2020-07-24
Fb biologics sample update test

labkey-chrisj:
#### Rationale
This supports regression testing for sample Update in Biologics

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/640

#### Changes
moves DropdownButtonGroup to shared location,
adds ability to check if a cell in EditableGrid is 'read-only'

### DIFF
--- a/src/org/labkey/test/components/glassLibrary/components/BaseReactSelect.java
+++ b/src/org/labkey/test/components/glassLibrary/components/BaseReactSelect.java
@@ -79,9 +79,9 @@ public abstract class BaseReactSelect<T extends BaseReactSelect> extends WebDriv
         return hasClass("is-clearable");
     }
 
-    public boolean isDisabled()
+    public boolean isEnabled()
     {
-        return hasClass("is-disabled");
+        return !hasClass("is-disabled");
     }
 
     public boolean hasValue()
@@ -285,12 +285,6 @@ public abstract class BaseReactSelect<T extends BaseReactSelect> extends WebDriv
     public String getName()
     {
         return elementCache().input.getAttribute("name");
-    }
-
-    public boolean isEnabled()
-    {
-        // If the class attribute does not contain 'is-disabled' the component is enabled.
-        return !getComponentElement().getAttribute("class").toLowerCase().contains("is-disabled");
     }
 
     protected T scrollIntoView()
--- a/src/org/labkey/test/components/glassLibrary/components/BaseReactSelect.java
+++ b/src/org/labkey/test/components/glassLibrary/components/BaseReactSelect.java
@@ -79,6 +79,11 @@ public abstract class BaseReactSelect<T extends BaseReactSelect> extends WebDriv
         return hasClass("is-clearable");
     }
 
+    public boolean isDisabled()
+    {
+        return hasClass("is-disabled");
+    }
+
     public boolean hasValue()
     {
         return hasClass("has-value");
--- a/src/org/labkey/test/components/react/DropdownButtonGroup.java
+++ b/src/org/labkey/test/components/react/DropdownButtonGroup.java
@@ -221,7 +221,7 @@ public class DropdownButtonGroup extends WebDriverComponent<DropdownButtonGroup.
         }
         public static Locator.XPathLocator menuAnchor(String text)
         {
-            return Locator.tagWithAttribute("a", "role", "menuitem").withText();
+            return Locator.tagWithAttribute("a", "role", "menuitem").withText(text);
         }
 
         public static Locator.XPathLocator subMenuItem()
--- a/src/org/labkey/test/components/react/DropdownButtonGroup.java
+++ b/src/org/labkey/test/components/react/DropdownButtonGroup.java
@@ -217,7 +217,7 @@ public class DropdownButtonGroup extends WebDriverComponent<DropdownButtonGroup.
 
         public static Locator.XPathLocator menuItem(String text)
         {
-            return menuItem().withChild(Locator.linkWithText(text)).notHidden();
+            return menuItem().withChild(menuAnchor(text)).notHidden();
         }
         public static Locator.XPathLocator menuAnchor(String text)
         {
--- a/src/org/labkey/test/components/react/DropdownButtonGroup.java
+++ b/src/org/labkey/test/components/react/DropdownButtonGroup.java
@@ -1,7 +1,4 @@
-/*
- * Copyright (c) 2018-2019 LabKey Corporation. All rights reserved. No portion of this work may be reproduced in
- * any form or by any electronic or mechanical means without written permission from LabKey Corporation.
- */
+
 package org.labkey.test.components.react;
 
 import org.labkey.test.Locator;
--- a/src/org/labkey/test/components/react/DropdownButtonGroup.java
+++ b/src/org/labkey/test/components/react/DropdownButtonGroup.java
@@ -1,0 +1,297 @@
+/*
+ * Copyright (c) 2018-2019 LabKey Corporation. All rights reserved. No portion of this work may be reproduced in
+ * any form or by any electronic or mechanical means without written permission from LabKey Corporation.
+ */
+package org.labkey.test.components.react;
+
+import org.labkey.test.Locator;
+import org.labkey.test.WebDriverWrapper;
+import org.labkey.test.WebDriverWrapperImpl;
+import org.labkey.test.components.WebDriverComponent;
+import org.labkey.test.util.LogMethod;
+import org.labkey.test.util.LoggedParam;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+import java.util.List;
+
+import static org.labkey.test.WebDriverWrapper.sleep;
+
+
+public class DropdownButtonGroup extends WebDriverComponent<DropdownButtonGroup.ElementCache>
+{
+    final WebElement _el;
+    final WebDriver _driver;
+    private Integer _expandRetries = 1;
+
+    /* componentElement should contain the toggle anchor *and* the UL containing list items */
+    public DropdownButtonGroup(WebDriver driver, WebElement componentElement)
+    {
+        this(componentElement, new WebDriverWrapperImpl(driver));
+    }
+
+    public DropdownButtonGroup(WebElement element, WebDriverWrapper wrapper)
+    {
+        _el = element;
+        _driver = wrapper.getDriver();
+    }
+
+    @Override
+    public WebElement getComponentElement()
+    {
+        return _el;
+    }
+
+    @Override
+    public WebDriver getDriver()
+    {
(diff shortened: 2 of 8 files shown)


src/org/labkey/test/components/glassLibrary/components/BaseReactSelect.java
```diff
@@ -79,6 +79,11 @@ public boolean isClearable()
         return hasClass("is-clearable");
     }
 
+    public boolean isDisabled()
+    {
+        return hasClass("is-disabled");
+    }
```
labkey-tchad: Consider inverting this to be consistent with the `WebElement` interface.
```suggestion
    public boolean isEnabled()
    {
        return !hasClass("is-disabled");
    }
```

src/org/labkey/test/components/ui/EntityBulkUpdateDialog.java
```diff
@@ -0,0 +1,205 @@
...
+import org.labkey.test.components.html.Checkbox;
+import org.labkey.test.components.html.Input;
+import org.labkey.test.components.html.ToggleButton;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+import java.util.List;
+
+
+public class EntityBulkUpdateDialog extends ModalDialog
```
labkey-tchad: Indicate the specific product component that is instrumented by this class.
Also, this seems very similar to the `EntityBulkInsertDialog`. I suspect they are the same product component, with a flag modifying their behavior slightly. If so, this class should extend `EntityBulkInsertDialog` (or vice-versa if that makes more sense).

src/org/labkey/test/components/react/DropdownButtonGroup.java
```diff
@@ -0,0 +1,297 @@
+/*
+ * Copyright (c) 2018-2019 LabKey Corporation. All rights reserved. No portion of this work may be reproduced in
+ * any form or by any electronic or mechanical means without written permission from LabKey Corporation.
+ */
```
labkey-tchad: Remove copyright. This is no longer in a closed-source repository.

labkey-tchad (changes requested):

src/org/labkey/test/components/react/DropdownButtonGroup.java
labkey-chrisj: thanks

src/org/labkey/test/components/ui/EntityBulkUpdateDialog.java
labkey-chrisj: I thought the dialog having enable/disable sliders for editing fields made it different enough from the outside to handle them separately.  If it's necessary to derive one from the other I expect a lot of overrides or forking (to handle the enable/disable field edit sliders) in a lot of places; not sure that will be helpful.

As I look, it seems there are several versions of BulkUpdateForm.d.ts out there, in multiple modules, as well as BulkAddUpdateForm.d.ts as well.  They appear to both use https://github.com/LabKey/labkey-ui-components/blob/master/packages/components/src/components/forms/QueryInfoForm.tsx in common to generate the modal dialog, but use different forms.

src/org/labkey/test/components/react/DropdownButtonGroup.java
```diff
@@ -0,0 +1,297 @@
...
+import java.util.List;
+
+import static org.labkey.test.WebDriverWrapper.sleep;
+
+
+public class DropdownButtonGroup extends WebDriverComponent<DropdownButtonGroup.ElementCache>
+{
+    final WebElement _el;
+    final WebDriver _driver;
+    private Integer _expandRetries = 1;
```
labkey-danield, line 22: Is this used anyplace? I see where it can be assigned but it doesn't look like it is used.

src/org/labkey/test/components/react/DropdownButtonGroup.java
```diff
@@ -0,0 +1,297 @@
...
+    public DropdownButtonGroup closeVisibleSubmenus()
+    {
+        getWrapper().waitFor(()-> {
+            WebElement openToggle = Locators.openMenuCaret.findElementOrNull(getComponentElement());
+            if (openToggle == null)
+                return true;
+            else
+                openToggle.click();
+            return false;
+        }, 4000);
```
labkey-danield: Four seconds feels like a long wait to fail if a submenu didn't close.

src/org/labkey/test/components/react/DropdownButtonGroup.java
```diff
@@ -0,0 +1,297 @@
...
+        sleep(500);
+        WebElement context = getRootMenu();
+
+        for (int i = 0; i < subMenuLabels.length; i++)
+        {
+            context = openSubmenu(subMenuLabels[i], context);
+        }
+
+        return Locators.menuItem().findElements(context);
+    }
```
labkey-danield, line 130: I think there is value in providing an similar function that returns a list containing the text of menu items. It makes it easier to write tests that validate menu content (for dynamic menus). Example
```
List<String> expectedMenu = Arrays.asList(...);
checker().verifyEquals(Msg, expectedMenu, dbg.getMenuItemsText("foo menu"));
```

src/org/labkey/test/components/react/DropdownButtonGroup.java
```diff
@@ -0,0 +1,297 @@
...
+    // finds and opens the next submenu item.  Returns the child UL of the specified
+    private WebElement openSubmenu(@LoggedParam String itemText, WebElement context)
+    {
+        WebElement item = Locators.subMenuItem(itemText).waitForElement(context, 4000);
+
+        if (!menuItemIsExpanded(item))
+        {
+            getWrapper().log("attempting to expand menu item [" + itemText + "]");
+            item.click();
+            sleep(1000);
```
labkey-danield: Not sure why there is a sleep here and not a wait for menuItemIsExpanded.

src/org/labkey/test/components/react/DropdownButtonGroup.java
```diff
@@ -0,0 +1,297 @@
...
+        }
+
+        getWrapper().waitFor(()-> menuItemIsExpanded(item), "menu item ["+itemText+"] failed to open", 2000);
+
+        return Locators.childMenuLoc().waitForElement(item, 2000);  // It's the child <ul> of the current submenu item
```
labkey-danield, line 165: Can there ever be more than one child sub-menu?

src/org/labkey/test/components/ui/EntityBulkUpdateDialog.java
```diff
@@ -0,0 +1,205 @@
...
+        selectValues.forEach(s -> {reactSelect.filterSelect(s);});
+        return this;
+    }
+
+    public List<String> getSelectionFieldValues(String fieldKey)
+    {
+        return elementCache().getSelect(fieldKey).getSelections();
+    }
+
+    public EntityBulkUpdateDialog setAliases(List<String> aliases)
```
labkey-danield: Why is there a specific field for Aliases when it is a selection field? Same can be asked about description. This should be a more generalized control and not assume that there are any pre-defined fields.

labkey-danield (changes requested):

src/org/labkey/test/components/react/DropdownButtonGroup.java
labkey-chrisj: I don't think it is ever accessed.

src/org/labkey/test/components/react/DropdownButtonGroup.java
labkey-chrisj: Something along the lines of
```
public List<String> getMenuItemTexts(@LoggedParam String ... subMenuLabels)
    {
        return getWrapper().getTexts(getMenuItemsAt(subMenuLabels));
    }
```

src/org/labkey/test/components/react/DropdownButtonGroup.java
labkey-chrisj: No.  The child UL here will contain all of the children of the current menuItem

src/org/labkey/test/components/ui/EntityBulkUpdateDialog.java
labkey-chrisj: I've assumed that these come standard with whatever set of fields you specify, since that's what I've been doing- creating a set of fields and getting them + these two.
I got rid of them and made tests call into the more generic functions

src/org/labkey/test/components/glassLibrary/components/BaseReactSelect.java
labkey-chrisj: Haha, it turns out I'd already implemented 'isEnabled' on it

labkey-tchad (approved):

src/org/labkey/test/components/react/DropdownButtonGroup.java
labkey-chrisj: It's gone

src/org/labkey/test/components/react/DropdownButtonGroup.java
labkey-chrisj: Are you requesting its removal?

src/org/labkey/test/components/react/DropdownButtonGroup.java
labkey-chrisj: This code anticipates that there might be many nested submenus; I figured plan for 2 or 3 max and give a second each plus a bit of buffer.  Buuuut, 2 seconds it is and I hope that doesn't cause failures or race conditions

labkey-tchad (approved):

src/org/labkey/test/components/react/DropdownButtonGroup.java
labkey-danield: Having an used private variable is not a big deal, but in this case there is a public interface that sets it (see withExpandRetries(Integer expandRetries) below) and never uses it. Because of that it should be cleaned up.

This is where a linter would help because it can call out unused variables.

src/org/labkey/test/components/react/DropdownButtonGroup.java
```diff
@@ -0,0 +1,298 @@
...
+        return _el;
+    }
+
+    @Override
+    public WebDriver getDriver()
+    {
+        return _driver;
+    }
+
+    public DropdownButtonGroup withExpandRetries(Integer expandRetries)
```
labkey-danield, line 48: Remove this public function.

labkey-danield (approved):

src/org/labkey/test/components/react/DropdownButtonGroup.java
labkey-chrisj: Okay, I'll do this when I can

PR #413, opened 2020-07-28
add pipeline status details page test helper

labkey-kevink:
#### Rationale
- fix S3FileRootMigrationTest: support s3 cloud status files in StatusDetailsBean
- fix RedCapTest: find expected errors in log text

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1431
* https://github.com/LabKey/redcap/pull/29
* https://github.com/LabKey/cloud/pull/74
* https://github.com/LabKey/dataintegration/pull/58

### DIFF
--- a/src/org/labkey/test/pages/pipeline/PipelineStatusDetailsPage.java
+++ b/src/org/labkey/test/pages/pipeline/PipelineStatusDetailsPage.java
@@ -230,6 +230,13 @@ public class PipelineStatusDetailsPage extends LabKeyPage<PipelineStatusDetailsP
         return "ERROR".equalsIgnoreCase(getStatus());
     }
 
+    /** Return true if the log status file was found (the log data element is present.) */
+    public PipelineStatusDetailsPage waitForLogPresent()
+    {
+        waitForElement(elementCache().logDataId);
+        return this;
+    }
+
     public boolean isShowingLogSummary()
     {
         String details = elementCache().showFullLogLink.getAttribute("data-details");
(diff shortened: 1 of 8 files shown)


src/org/labkey/test/pages/pipeline/PipelineStatusDetailsPage.java
```diff
@@ -0,0 +1,433 @@
...
+        protected WebElement showGridButton = Locator.lkButton("Show Grid").findWhenNeeded(this);
+        protected WebElement showDataButton = Locator.id("show-data-btn").findWhenNeeded(this);
+        protected WebElement browseFilesButton = Locator.lkButton("Browse Files").findWhenNeeded(this);
+        protected WebElement cancelButton = Locator.id("cancel-btn").findWhenNeeded(this);
+        protected WebElement retryButton = Locator.id("retry-btn").findWhenNeeded(this);
+
+        protected WebElement showFullLogLink = Locator.id("show-full-log").findWhenNeeded(this);
+        protected WebElement logContainer = Locator.id("log-container").findWhenNeeded(this);
+        protected Locator logDataId = Locator.id("log-data");
+        protected WebElement logData = logDataId.findWhenNeeded(this);
```
labkey-tchad: Make element cache members `final`

src/org/labkey/test/pages/pipeline/PipelineStatusDetailsPage.java
```diff
@@ -0,0 +1,433 @@
+package org.labkey.test.pages.pipeline;
+
+import org.apache.commons.lang3.StringUtils;
+import org.jetbrains.annotations.Nullable;
+import org.labkey.api.collections.CaseInsensitiveHashSet;
+import org.labkey.api.pipeline.PipelineJob;
```
labkey-tchad: Just an FYI. Avoid using server APIs. On TeamCity, we pull the API jar from Artifactory to avoid building it before every test suite (some tests need access to the XML schema definitions). The version used by tests lags behind the actual server version to avoid breaking tests every time we bump versions for monthly releases so using server APIs too extensively could lead to test noise on TeamCity.
That being said, these are probably fine because they're unlikely to change in any way the tests would care about.

labkey-tchad (approved):

PR #414, opened 2020-07-28
API bootstrap test

labkey-tchad:
#### Rationale
Some suites shouldn't need to do the full suite of initial user checks done by our standard `signIn` method. This stub test allows us to bootstrap via API (mostly)

#### Related Pull Requests
* https://github.com/LabKey/biologics/pull/643

#### Changes
* Add a stub test that overrides the standard `signIn` to avoid initial user checks.


labkey-nicka (approved):

labkey-danield (commented):
Looks good to me.

PR #422, opened 2020-07-31
Updates the test to no longer specify a public name on schema query l…

labkey-chrisj:
#### Rationale
This changes the test to no longer expect 'ctq' as the public name of the query.  This test was failing because the text of the link appears to have changed.

#### Related Pull Requests
n/a

#### Changes
don't expect the schema query link text to contain schema, query, and public name- schema and query are enough.


labkey-martyp (approved):
This should back out all my changes that broke the expression matrix and Nab tests

PR #430, opened 2020-08-16
Fb shared component test 7220

labkey-chrisj:
#### Rationale
This provides support for testing components via the core-Components view, and creates tests specificially for grid
pulls the Pager component out of Biologics and adapts it for common use, updates QueryGrid to use it for paging

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/652

#### Changes
Creates a page class that wraps the core-components.view action
Creates forking logic within QueryGrid test component to support transition from old QueryGridPanel toward new QueryGrid


src/org/labkey/test/components/glassLibrary/grids/GridBar.java
```diff
@@ -28,12 +27,14 @@
 public class GridBar extends WebDriverComponent<GridBar.ElementCache>
 {
     final private WebElement _gridBarElement;
+    final private WebElement _containerElement;
```
labkey-chrisj, line 31: Just a general note on this 'containerElement'- because the newer QueryGrid component moves things like the OmniBox and selection info controls out of the _gridBarElement, we may want to consider whether to expand that element to contain them, or re-think whether or not GridBar ought to be replaced

src/org/labkey/test/components/glassLibrary/components/OmniBox.java
```diff
@@ -144,20 +166,42 @@ private void sendClearValue()
...
+        this.setText(commandExpression.toString());
+        if (WebDriverWrapper.waitFor(()->  getValue(expectedFilterText.toString()) != null, 2500))
             return this;
 
         // try again if necessary and fail if it doesn't work
         this.setText("filter \"" + columnName + "\" " + operator + val);
-        WebDriverWrapper.waitFor(()->  getValue(expectedValue) != null,
-                "Expect a new value to be present with [" + expectedValue + "]", 1500);
+        WebDriverWrapper.waitFor(()->  getValue(expectedFilterText.toString()) != null,
+                "Expect a new value to be present with [" + expectedFilterText.toString() + "]", 1500);
```
labkey-tchad, line 191: Log the retry so that somebody could investigate whether it is having to do so too frequently.

src/org/labkey/test/components/glassLibrary/grids/GridBar.java
```diff
@@ -68,11 +69,46 @@ public File exportData(ExportType exportType)
...
+    /**
+     * A word of caution for callers; the Pager is not aware of the grid state.
+     * To navigate (next page, previous, first/last) use methods on GridBar to do it-
+     * they will synchronize against grid updates
+     * @return
+     */
+    public Pager pager()
+    {
+        return new Pager.PagerFinder(getDriver()).waitFor(this);
+    }
```
labkey-tchad: Why not have the `Pager` handle synchronization? Pass the `_responsiveGrid` in its constructor and have the `Pager` make the `doAndWaitForUpdate` calls to the grid.

src/org/labkey/test/components/glassLibrary/grids/GridBar.java
```diff
@@ -83,51 +119,25 @@ public int getRecordCount()
...
     public ResponsiveGrid getPreviousPage()
     {
-        WebElement prevButton = Locators.pgLeftButton.waitForElement(this, 2500);
-        if (prevButton.isEnabled())
-        {
-            _responsiveGrid.doAndWaitForUpdate(()->prevButton.click());
-        }
+        _responsiveGrid.doAndWaitForUpdate(()-> pager().clickPrevious());
         return _responsiveGrid;
     }
```
labkey-tchad: `get*` implies a non-mutating operation. I'd go with `clickNext`/`clickPrevious` to match the `Pager`

src/org/labkey/test/components/glassLibrary/grids/Pager.java
```diff
@@ -0,0 +1,194 @@
...
+import org.labkey.test.components.WebDriverComponent;
+import org.labkey.test.components.react.DropdownButtonGroup;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+public class Pager extends WebDriverComponent<Pager.ElementCache>
+{
+    private final WebElement _el;
+    private final WebDriver _driver;
+    //private final ResponsiveGrid _grid;
```
labkey-tchad: I presume the `ResponsiveGrid` is omitted so that the biologics report list can use the pager. You could make a simple interface for components that can update:
```
public interface UpdatingComponent
{
    public static final UpdatingComponent DEFAULT = Runnable::run;

    void doAndWaitForUpdate(Runnable func);
}
```
`Pager` would take an `UpdatingComponent` in its constructor. Classes would either implement it and pass themselves to the `Pager` or just pass `UpdatingComponent.DEFAULT` to it. Alternately, you could leave the existing `Pager` constructor signature and have it pass through `UpdatingComponent.DEFAULT`.

src/org/labkey/test/components/glassLibrary/grids/Pager.java
```diff
@@ -0,0 +1,194 @@
...
+        elementCache().prevButton.click();
+        return this;
+    }
+
+    public boolean isPreviousEnabled()
+    {
+        return !elementCache().prevButton.getAttribute("disabled").equals("true");
+    }
+
+    public Pager clickNextButton()
```
labkey-tchad: Use consistent method names
```suggestion
    public Pager clickNext()
```

src/org/labkey/test/components/glassLibrary/grids/Pager.java
```diff
@@ -0,0 +1,194 @@
+/*
+ * Copyright (c) 2019 LabKey Corporation. All rights reserved. No portion of this work may be reproduced in
+ * any form or by any electronic or mechanical means without written permission from LabKey Corporation.
+ */
```
labkey-tchad: Remove inappropriate copyright.

src/org/labkey/test/components/glassLibrary/grids/Pager.java
```diff
@@ -0,0 +1,194 @@
...
+package org.labkey.test.components.glassLibrary.grids;
```
labkey-tchad: Don't use the `glassLibrary` package. It'll just be one more thing to move once we decide to refactor. I've been using `org.labkey.test.components.labkey.ui`.
```suggestion
package org.labkey.test.components.labkey.ui.grids;
```

src/org/labkey/test/components/glassLibrary/grids/Pager.java
```diff
@@ -0,0 +1,194 @@
...
+        WebElement oldTotal = Locator.tagWithClass("span", "pagination-info__total")
+                .findElementOrNull(this);
+        if (oldTotal != null)
+            return Integer.parseInt(oldTotal.getText());
+        else
+        {
+            String value = elementCache().counts.getAttribute("data-total");
+            return Integer.parseInt(value);
+        }
+    }
```
labkey-tchad: I suggest combining `start()`, `end()`, and `total()` into a single `getPaginationInfo()` method that gathers all three and wraps them in a little class like:
```
public static class PaginationInfo
{
    public final int _start, _end, _total;
    PaginationInfo(int start, int end, int total)
    {
        _start = start;
        _end = end;
        _start = total;
    }
}
```

src/org/labkey/test/components/react/ReactCheckBox.java
```diff
@@ -78,6 +77,10 @@ public void set(@NotNull Boolean checked)
 
         if (checked != isChecked())
             toggle();
-        WebDriverWrapper.waitFor(() -> checked == isChecked(), "Failed to " + (checked ? "check" : "uncheck") + " checkbox", 1000);
+
+        // Sometimes (like when the box has a ternary state, as the 'select all' box in grids)
+        // checking the box won't result in a checked state- it will instead land on 'indeterminate'.
+        WebDriverWrapper.waitFor(() -> checked == isChecked() || checked == isIndeterminate(),
+                "Failed to " + (checked ? "check" : "uncheck") + " checkbox", 1000);
```
labkey-tchad: As discussed. Create a `set` overload that takes an expected state.

src/org/labkey/test/tests/component/QueryGridTest.java
```diff
@@ -0,0 +1,443 @@
...
+    {
+        _containerHelper.createProject(getProjectName(), null);
+    }
+
+    @Before
+    public void preTest() throws Exception
+    {
+        goToProjectHome();
+    }
+
```
labkey-tchad: This seems unnecessary (I should probably remove it from the template).
All of the tests start with a `CoreComponentsTestPage.beginAt`

src/org/labkey/test/tests/component/QueryGridTest.java
```diff
@@ -0,0 +1,443 @@
...
+        grid.filterOn("Description", "contains", "off-view");
+        assertThat(grid.getGridBar().pager().summary(), is ("1 - 5"));
+        grid.clearSortsAndFilters();
+
+        sampleSetDataGenerator.deleteDomain(createDefaultConnection());
+    }
+
+    @Ignore // until https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41169 is resolved
+    @Test
+    public void testEmptyNotEmptyFilter() throws Exception
```
labkey-tchad: Why `@Ignore` this test? It looks like the "has any value" check will pass with the current (incorrect) behavior which seems like the right approach.

labkey-tchad (dismissed):

src/org/labkey/test/components/glassLibrary/grids/GridBar.java
labkey-chrisj: There are cases where the pager interacts with not-a-grid (specifically, in Biologics, the reportList).
I initially did that and passed in the _responsiveGrid to synchronize internally, but opted to make it relatively dumb and have the methods calling it (from QueryGrid) handle the synchronizing.
I'm willing to pass in the grid and locate the synchronization inside this component, but then I'd have to fix the BiologicsReportTest to not use it.  The upside to doing that would mean only having to support grid usage, and that would simplify a lot of code

src/org/labkey/test/components/glassLibrary/grids/GridBar.java
labkey-chrisj: Good call.  I'd been trying not to mess with established naming unless necessary but I like it.

src/org/labkey/test/tests/component/QueryGridTest.java
labkey-chrisj: If it was behaving correctly, it would say 1- 20 of 30 but it instead current behavior has it saying 1-20 of 35 because the custom rows (which don't have values in Int Column) aren't filtered out.  
`@Ignore` is there so it's not noisy until the bug is addressed.  If that's not right, I'm happy to change it

src/org/labkey/test/components/glassLibrary/grids/GridBar.java
labkey-chrisj: ...later in this review you suggested using an interface to deal with the coupling issues of wanting to pass in the grid.  I like that as a concept and am going to make a brave attempt, but may need help implementing that.

src/org/labkey/test/components/glassLibrary/grids/Pager.java
labkey-chrisj: Good guess.  Thanks for providing enough guidance here to help me create the interface and wire it all up

src/org/labkey/test/components/glassLibrary/grids/Pager.java
labkey-chrisj: I've been putting off making a new package/migrating everything into it, largely because I don't have a clear vision for what the new destination should look like and I didn't want to make the changes and then be told I did it wrong.  I'll just do my best for now.

src/org/labkey/test/components/glassLibrary/grids/Pager.java
labkey-chrisj: I confess; I'm not 100% sure what that will get us.  I agree that the current implementation is pretty verbose and would like to clean it up.  Let's talk about how this can be used

labkey-danield (approved):

src/org/labkey/test/components/glassLibrary/grids/GridBar.java
labkey-chrisj: Actually, I did it and it works

src/org/labkey/test/components/glassLibrary/components/OmniBox.java
labkey-chrisj: done, not sure why github doesn't mark this as outdated

src/org/labkey/test/components/glassLibrary/grids/Pager.java
labkey-chrisj: done

src/org/labkey/test/components/glassLibrary/grids/GridBar.java
labkey-chrisj: done

src/org/labkey/test/components/glassLibrary/grids/Pager.java
labkey-chrisj: Right now, ` summary()` (line 138) consumes the text from the 'counts' element, which only exists in the new GridPanel.

src/org/labkey/test/components/react/ReactCheckBox.java
labkey-chrisj: done

PR #431, opened 2020-08-17
Run automated tests against trial servers

labkey-tchad:
#### Rationale
Trial servers have numerous limitations that conflict with assumptions made by our automated tests. There are workarounds for many such limitations.

#### Changes
* Add property for targeting a trial server: `webtest.server.trial=true` (`TestProperties`)
* Allow suites to be defined by text file (`SuiteBuilder`, `TestFileUtils`)
* Better handling of errors logged during server startup (`BaseWebDriverTest`)
* Make WebDav helpers use a session key on trial instances (`WebDavUtils`, `WebDavUploadHelper`)
* Update some tests to work against trial instances


src/org/labkey/test/BaseWebDriverTest.java
```diff
@@ -577,13 +576,24 @@ private static boolean canConnectWithPrimaryUser()
     private void doPreamble()
     {
         signIn();
+
+        // Only do this as part of test startup if we haven't already checked. Since we do this as the last
+        // step in the test, there's no reason to bother doing it again at the beginning of the next test
+        if (!_checkedLeaksAndErrors && !"DRT".equals(System.getProperty("suite")))
```
labkey-chrisj, line 582: I'm curious to know why we plan to behave differently for DRT than for other suites.

labkey-chrisj (approved):

src/org/labkey/test/BaseWebDriverTest.java
labkey-tchad: This is just maintaining existing behavior. The thought is that we want DRTs to run as fast a possible so we only do the error and leak checks once; at the end of the suite.

src/org/labkey/test/TestFileUtils.java
```diff
@@ -186,46 +186,81 @@ public static String getDefaultWebAppRoot()
...
 
-        File sampledataDirsFile = new File(getTestBuildDir(), "sampledata.dirs");
-        if (sampledataDirsFile.exists())
+        if (sampleDatas.isEmpty())
         {
-            String path = getFileContents(sampledataDirsFile);
-            sampledataDirs.addAll(Arrays.stream(path.split(";")).map(File::new).collect(Collectors.toList()));
+            throw new NotFoundException("Sample data not found: " + relativePath + "\n" +
+                    "Run `./gradlew :server:test:build :server:test:writeSampleDataFile` once to locate all sampledata" + "\n" +
+                    "Currently known sample data locations: " + getSampleDataDirs().stream().map(File::getAbsolutePath).collect(Collectors.joining("\n")));
```
labkey-danield, line 205: For a string like this it might be better to use String.format().
`String.format("Sample data not found: %s%nRun './gradlew :server:test:build :server:test:writeSampleDataFile' once to locate all sampledata%nCurrently known sample data locations: %s", relativePath, getSampleDataDirs().stream().map(File::getAbsolutePath).collect(Collectors.joining("\n"))`

src/org/labkey/test/TestFileUtils.java
```diff
@@ -186,46 +186,81 @@ public static String getDefaultWebAppRoot()
...
         }
-        else
+        if (sampleDatas.size() > 1)
         {
-            sampledataDirs.add(new File(getTestRoot(), "data"));
+            throw new IllegalArgumentException(
+                "Ambiguous file specified: " + relativePath + "\n" +
+                    "Found:\n" +
+                    sampleDatas.stream().map(File::getAbsolutePath).collect(Collectors.joining("\n")));
```
labkey-danield, line 212: Same style comment as above, String.format might be useful here.

labkey-danield (approved):

PR #433, opened 2020-08-19
Create and publish a labkey-api-selenium jar with just the classes from this module

labkey-susanh:
#### Rationale
Currently the default jar file for this package is empty because the source code is not in the default location for the Gradle java plugin.  This causes a problem when trying to run the `moduleUiTests` task that declares a dependency on the default output of the test project.  We had been producing a `testJar` artifact from the Gradle plugins containing all of the test classes from all modules as well as this module.  This is more than is needed for running individual module tests, so we scale back on the classes included in the jar and publish it to Artifactory to get us closer to being able to have developers use this test api jar without a need for an enlistment in this repository. (We're not quite there yet because of some need for the `test.properties` file that lives in this repository, but this gets us pretty close.)

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1498

#### Changes
* Specify the custom `src` directory for the default java source set so the `jar` (or `build`) command produces a jar file with classes for use in writing Selenium tests
* Set up publishing for the default jar file to `org.labkey.api:labkey-api-selinim`
* Move configuration and dependency declarations out of the Gradle plugins and into here for easier maintenance

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,12 @@
 apply plugin: 'java'
+apply plugin: 'java-library'
 apply plugin: 'org.labkey.testRunner'
+apply plugin: 'maven-publish'
 
 import org.labkey.gradle.util.BuildUtils
+import org.labkey.gradle.util.PomFileHelper
+
+project.group = "org.labkey.test"
 
 project.dependencies {
     implementation "org.slf4j:slf4j-log4j12:${slf4jLog4j12Version}" // Suppresses unwanted logging from remoteapi and webdriver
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -62,3 +78,86 @@ if (project.hasProperty("teamcity"))
...
+                        name = "LabKey Test Automation Classes"
+                        description = "Library of classes and utilities for Selenium testing of LabKey Server components"
+                        url = PomFileHelper.LABKEY_ORG_URL
+                        developers PomFileHelper.getLabKeyTeamDevelopers()
+                        licenses PomFileHelper.getApacheLicense()
+                        organization PomFileHelper.getLabKeyOrganization()
+                        scm {
+                            connection = 'scm:git:https://github.com/LabKey/testAutomation'
+                            developerConnection = 'scm:git:https://github.com/LabKey/testAutomation'
+                            url = 'scm:git:https://github.com/LabKey/testAutomation'
```
labkey-tchad: I think the scm `url` property is supposed to be the plain repository url: http://maven.apache.org/pom.html#SCM
```suggestion
                            url = 'https://github.com/LabKey/testAutomation'
```

labkey-tchad (approved):

PR #432, opened 2020-08-17
Adding test coverage for "Unhappy rendering in wiki "Show Page Tree" …

labkey-sweta:
…widget"

#### Rationale
Adding test coverage for 
1.  Stored XSS in wiki titles and names 
2.  Unhappy rendering in wiki "Show Page Tree" widget 

#### Changes
1. Edited the WikiTest and WikiLongTest

PR #436, opened 2020-08-21
Fix `WikiLongTest`

labkey-tchad:
#### Rationale
Encoding/decoding the wiki name or the link's href is complicated. Let's just verify the link's destination instead.

#### Related Pull Requests
* #432 

#### Changes
* Validate destination instead of link `href`


labkey-tchad:
Fixed in #437

labkey-sweta (approved):

PR #437, opened 2020-08-21
WikiLongTest fix

labkey-sweta:
#### Rationale
Removing the special character from the wiki name which is part of wiki body.

#### Related Pull Requests
#432

#### Changes
wikiLongTest.java


labkey-tchad (dismissed):

labkey-sweta:
Trey fixed the test in different fb.

PR #445, opened 2020-08-25
Update ListArchiveImportTest

labkey-tchad:
#### Rationale
Validators are now applied _after_ data is imported from a list archive, so can't cause the import to fail. This test needs a different way of triggering an import error. This does so by using an archive where a list record references a missing file.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1410

#### Changes
* Update `ListArchiveImportTest` to work for new import behavior
* Add functionality to some test helpers
* Some SonarLint fixes


labkey-tchad:
Verifying fix: https://teamcity.labkey.org/viewQueued.html?itemId=1097946

labkey-chrisj (approved):

PR #454, opened 2020-09-01
Update Pagination Selectors, rename QueryGridTest.java to GridPanelTest.java

labkey-alan:
#### Rationale
The QueryGridTest was not testing the QueryGridPanel at all, so I renamed it to GridPanelTest to better reflect what it is testing. I also added some new selectors to the pagination related components and updated the test code to use them.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/332
* https://github.com/LabKey/platform/pull/1538

#### Changes
* Rename QueryGridTest.java to GridPanelTest.java
* Update selectors for pagination components

### DIFF
--- a/src/org/labkey/test/components/react/DropdownButtonGroup.java
+++ b/src/org/labkey/test/components/react/DropdownButtonGroup.java
@@ -250,6 +250,11 @@ public class DropdownButtonGroup extends WebDriverComponent<DropdownButtonGroup.
         {
             return Locator.tagWithClass("div", "dropdown").withChild(Locators.toggleAnchor().withAttribute("id", Id));
         }
+
+        static public Locator.XPathLocator buttonGroupWithClass(String cls)
+        {
+            return Locator.tagWithClass("div", "dropdown").withChild(Locators.toggleAnchor().withClass(cls));
+        }
     }
 
     public static class DropdownButtonGroupFinder extends WebDriverComponentFinder<DropdownButtonGroup, DropdownButtonGroupFinder>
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/components/react/DropdownButtonGroup.java
```diff
@@ -281,6 +286,12 @@ public DropdownButtonGroupFinder withButtonId(String Id)
             return this;
         }
 
+        public DropdownButtonGroupFinder withButtonClass(String Id)
+        {
+            _baseLocator = Locators.buttonGroupWithClass(Id);
```
labkey-chrisj: seems a little confusing to pass an Id to a method that advertises taking a class

src/org/labkey/test/tests/component/GridPanelTest.java
```diff
@@ -53,7 +52,7 @@ public void testJumpToLastPage() throws Exception
         sampleSetDataGenerator.insertRows();
 
         CoreComponentsTestPage testPage = CoreComponentsTestPage.beginAt(this, getProjectName());
-        QueryGrid grid = testPage.getQueryGrid("samples", "grid_paging_samples");
+        QueryGrid grid = testPage.getGridPanel("samples", "grid_paging_samples");
```
labkey-chrisj, line 55: Eventually, once we consolidate our grid use down to GridPanels, we may want to rename QueryGrid (the test wrapper class) to 'GridPanel'.  For now tho, because it has to support both GridPanel and QueryGridPanel, we can leave it as it is

labkey-chrisj (approved):

src/org/labkey/test/components/react/DropdownButtonGroup.java
labkey-alan: I'll update the be variable to be cls

src/org/labkey/test/tests/component/GridPanelTest.java
labkey-alan: Agreed.

PR #460, opened 2020-09-03
Fb fix specimen import test

labkey-chrisj:
#### Rationale
This attempts to fix a test failing to find a control by its label by finding it by Id

#### Related Pull Requests
n/a


src/org/labkey/test/tests/SpecimenImportTest.java
```diff
@@ -116,7 +117,7 @@ protected void changeTimepointType()
         clickTab("Manage");
         waitAndClick(Locator.linkContainingText("Change Study Properties"));
         waitForText("Timepoint Type:");
```
labkey-tchad: Your fix seems more effective than this `waitForText`. There's also a page load here, I think.
```suggestion
        waitAndClickAndWait(Locator.linkContainingText("Change Study Properties"));
```

labkey-tchad (approved):

PR #461, opened 2020-09-03
Remove unnecessary navigation from BaseFlowTest

labkey-tchad:
#### Rationale
This method is called by `doCleanup`. If the crawler ends on a particular action, this method will attempt to open a project menu that doesn't exist.
The test does a bunch of navigation just to check if a folder exists. If we use the container helper to check for the folder we don't need the project menu to be present.
_Note: This code predates our current project menu and might not function as expected at all._

#### Changes
* Use `APIContainerHelper` to check for folder

### DIFF
--- a/src/org/labkey/test/tests/flow/BaseFlowTest.java
+++ b/src/org/labkey/test/tests/flow/BaseFlowTest.java
@@ -174,16 +174,10 @@ abstract public class BaseFlowTest extends BaseWebDriverTest
     //Issue 12597: Need to delete exp.data objects when deleting a flow run
     protected void deleteAllRuns()
     {
-        if (!isElementPresent(Locator.linkWithText(getProjectName())))
-            goToHome();
-        if (!isElementPresent(Locator.linkWithText(getProjectName())))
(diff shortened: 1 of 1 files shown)


labkey-danield (approved):

labkey-tchad:
Looks like this is causing trouble in the flow suites. They don't report results to this repository, so I'm adding links for easier checking in the future.
[Postgres](https://teamcity.labkey.org/viewType.html?buildTypeId=LabkeyTrunk_ModuleSuites_FlowPostgres&branch_LabkeyTrunk_ModuleSuites=flowDeleteAllRuns&tab=buildTypeStatusDiv)
[SQL Server](https://teamcity.labkey.org/viewType.html?buildTypeId=LabkeyTrunk_ModuleSuites_FlowSqlserver&branch_LabkeyTrunk_ModuleSuites=flowDeleteAllRuns&tab=buildTypeStatusDiv)

labkey-sweta (approved):

labkey-chrisj (approved):

PR #468, opened 2020-09-11
Have 'isDisplayed' return false for stale or unfound WebElements

labkey-tchad:
#### Rationale
We often check whether a component or page is ready by checking for the presence and visibility of particular element. e.g.:
```
try
{
    return elementCache().element.isDisplayed();
}
{
    return false;
}
```
This update would allow us to remove the try/catch from such statements:
```
return elementCache().element.isDisplayed();
```

#### Changes
* Have 'isDisplayed' return false for stale or unfound `LazyWebElement`s

#### Alternate solutions
* Move this override up to `WebElementWrapper`.
  * More likely to introduce unexpected behavior to other tests but would make behavior more consistent.
* Create a new `WebElementWrapper` or `LazyWebElement` subclass to allow components to opt in to this behavior.
  * Won't risk breaking any existing tests but might be too cumbersome for people to bother with.

### DIFF
--- a/src/org/labkey/test/selenium/LazyWebElement.java
+++ b/src/org/labkey/test/selenium/LazyWebElement.java
@@ -17,7 +17,9 @@ package org.labkey.test.selenium;
 
 import org.jetbrains.annotations.NotNull;
 import org.labkey.test.Locator;
+import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.SearchContext;
+import org.openqa.selenium.StaleElementReferenceException;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.ui.FluentWait;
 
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/selenium/LazyWebElement.java
```diff
@@ -39,6 +41,19 @@ public LazyWebElement(@NotNull Locator locator, @NotNull SearchContext searchCon
...
 
+    @Override
+    public boolean isDisplayed()
+    {
+        try
+        {
+            return super.isDisplayed();
+        }
+        catch (StaleElementReferenceException | NoSuchElementException ex)
+        {
```
labkey-chrisj, line 52: Would it be helpful to note in a log that this happened?

...or possibly, put a _staleReferenceCount member on the class and increment it, provide an isStale() method that returns whether or not the count is > 0?  Maybe start logging when it exceeds a threshold?

labkey-chrisj (approved):
I like the idea behind this.
Have you considered whether other webElement methods (like `isEnabled(), isSelected()`) might be candidates for similar treatment?

labkey-tchad:
@labkey-chrisj Definitely don't want to modify `isEnabled`. That the standard method used to trigger a staleness check (see `ExpectedConditions.stalenessOf`). Modifying `isSelected` would likely cause tests to blow past missing elements (1. Attempt to uncheck checkbox; 2. Already looks unchecked (actually doesn't exist); 3. Renamed checkbox on page stays checked and waves as the test passes by).
`isDisplayed` seems reasonable to override because "stale", "not present", "hidden" should be effectively identical states as far as tests are concerned.
I could see an argument for making `getText` return `null` but that would break numerous places that are already catching exceptions from `getText`.

labkey-danield (approved):
Excellent!

PR #469, opened 2020-09-11
Fb test sample create sm integration

labkey-chrisj:
#### Rationale
This adds support for SM integration to Biologics

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/678

#### Changes
updates to EntityInsertPanel, EditableGrid

### DIFF
--- a/src/org/labkey/test/components/glassLibrary/grids/EditableGrid.java
+++ b/src/org/labkey/test/components/glassLibrary/grids/EditableGrid.java
@@ -513,6 +513,24 @@ public class EditableGrid extends WebDriverComponent<EditableGrid.ElementCache>
     }
 
     @Override
+    protected void clearElementCache()
+    {
+        super.clearElementCache();
+    }
+
+    public void doAndWaitForUpdate(Runnable func)
+    {
+        func.run();
+        waitForUpdate();
+    }
+
+    private void waitForUpdate()
+    {
+        waitForLoaded();
+        clearElementCache();
+    }
+
+    @Override
     protected ElementCache newElementCache()
     {
         return new ElementCache();
--- a/src/org/labkey/test/components/glassLibrary/grids/EditableGrid.java
+++ b/src/org/labkey/test/components/glassLibrary/grids/EditableGrid.java
@@ -512,12 +512,6 @@ public class EditableGrid extends WebDriverComponent<EditableGrid.ElementCache>
         }
     }
 
-    @Override
-    protected void clearElementCache()
-    {
-        super.clearElementCache();
-    }
-
     public void doAndWaitForUpdate(Runnable func)
     {
         func.run();
--- a/src/org/labkey/test/components/ui/EntityBulkInsertDialog.java
+++ b/src/org/labkey/test/components/ui/EntityBulkInsertDialog.java
@@ -4,6 +4,9 @@ import org.labkey.test.BootstrapLocators;
 import org.labkey.test.Locator;
 import org.labkey.test.components.bootstrap.ModalDialog;
 import org.labkey.test.components.glassLibrary.components.FilteringReactSelect;
+import org.labkey.test.components.html.Checkbox;
+import org.labkey.test.components.html.Input;
+import org.openqa.selenium.Keys;
 import org.openqa.selenium.StaleElementReferenceException;
 import org.openqa.selenium.WebElement;
 
@@ -88,6 +91,43 @@ public class EntityBulkInsertDialog extends ModalDialog
         return getWrapper().getFormElement(Locator.tagWithId("input", id));
     }
 
+    public EntityBulkInsertDialog setDateField(String fieldKey, String dateString)
+    {
+        Input input = elementCache().textInput(fieldKey);
+        input.set(dateString);
+        input.getComponentElement().sendKeys(Keys.ENTER);               // enter to collapse the datepicker
+        getWrapper().waitFor(()-> !isDatePickerExpanded(), 1500);
+        return this;
+    }
+
+    private boolean isDatePickerExpanded()
+    {
+        return Locator.tagWithClass("div", "react-datepicker-popper").existsIn(this);
+    }
+
+    public String getDateField(String fieldKey)
+    {
+        return elementCache().textInput(fieldKey).get();
+    }
+
+    public EntityBulkInsertDialog setBooleanField(String fieldKey, boolean checked)
+    {
+        Checkbox box = getCheckBox(fieldKey);
+        box.set(checked);
+        return this;
+    }
+
+    public boolean getBooleanField(String fieldKey)
+    {
+        return getCheckBox(fieldKey).get();
+    }
+
+    private Checkbox getCheckBox(String fieldKey)
+    {
+        WebElement row = elementCache().formRow(fieldKey);
+        return new Checkbox(elementCache().checkBoxLoc.findElement(row));
+    }
+
     public void clickAddRows()
     {
         elementCache().addRowsButton.click();
@@ -136,6 +176,19 @@ public class EntityBulkInsertDialog extends ModalDialog
 
     protected class ElementCache extends ModalDialog.ElementCache
     {
+        public WebElement formRow(String fieldKey)
+        {
+            return Locator.tagWithClass("div", "row")
+                    .withChild(Locator.tagWithAttribute("label", "for", fieldKey))
+                    .findElement(this);
+        }
+
+        public Input textInput(String fieldKey)
+        {
+            WebElement inputEl = textInputLoc.findElement(elementCache().formRow(fieldKey));
+            return new Input(inputEl, getDriver());
+        }
+
         WebElement cancelButton = Locator.tagWithClass("button", "test-loc-cancel-button")
                 .findWhenNeeded(getComponentElement());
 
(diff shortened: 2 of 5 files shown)


src/org/labkey/test/components/glassLibrary/grids/EditableGrid.java
```diff
@@ -512,6 +512,24 @@ public boolean isDisplayed()
         }
     }
 
+    @Override
+    protected void clearElementCache()
+    {
+        super.clearElementCache();
+    }
+
```
labkey-tchad: Why override `clearElementCache`? This doesn't change the functionality or access level.
```suggestion
```

src/org/labkey/test/components/ui/EntityBulkInsertDialog.java
```diff
@@ -88,6 +91,43 @@ public String getFieldWithId(String id)
...
+    {
+        return getCheckBox(fieldKey).get();
+    }
+
+    private Checkbox getCheckBox(String fieldKey)
+    {
+        WebElement row = elementCache().formRow(fieldKey);
+        return new Checkbox(elementCache().checkBoxLoc.findElement(row));
+    }
+
```
labkey-tchad: Add `setCheckbox`?

src/org/labkey/test/components/ui/EntityBulkUpdateDialog.java
```diff
@@ -104,6 +104,25 @@ public String getNumericField(String fieldKey)
...
+        elementCache().title.click();
+        getWrapper().waitFor(()-> !isDatePickerExpanded(), 1500);
+        input.set(dateString);
+        return this;
+    }
+
+    private boolean isDatePickerExpanded()
+    {
+        return Locator.tagWithClass("div", "react-datepicker-popper").existsIn(this);
+    }
```
labkey-tchad: Consider adding a `ReactDateInput` that handles waiting for the date picker popup to disappear after setting.

labkey-tchad: Actually, is this a different field type than in `EntityBulkInsertDialog.setDateField`? They do different things so I'm wondering if one or the other is an earlier implementation or if they are actually different components.

src/org/labkey/test/components/ui/EntityInsertPanel.java
```diff
@@ -190,8 +204,13 @@ protected ElementCache newElementCache()
...
     protected class ElementCache extends Component.ElementCache
     {
+        Locator deleteRowsBtnLoc = Locator.XPathLocator.union(
+                Locator.button("Delete rows"),
+                Locator.buttonContainingText("Remove"));
+
         WebElement bulkInsert = Locator.button("Bulk Insert").findWhenNeeded(_driver);
-        WebElement deleteRows = Locator.button("Delete rows").findWhenNeeded(_driver);
+        WebElement bulkUpdate = Locator.button("Bulk update").findWhenNeeded(_driver);
+        WebElement deleteRows = deleteRowsBtnLoc.findWhenNeeded(_driver);
```
labkey-tchad: Should the elements in this cache be scoped to the component instead of the entire page? e.g.
```suggestion
        WebElement deleteRows = deleteRowsBtnLoc.findWhenNeeded(this);
```

labkey-tchad (changes requested):

src/org/labkey/test/components/glassLibrary/grids/EditableGrid.java
labkey-chrisj: I was following the pattern, looks like I got it wrong.  I'll take this change

src/org/labkey/test/components/ui/EntityBulkUpdateDialog.java
labkey-chrisj: Functionally, the datepicker here appears to be the same in both dialogs.

I've implemented a component class to wrap it- right now it exposes get/set/clear methods that seem to work in this context.
My initial (in-dialog) means of closing the datePicker popup was to click on the dialog title banner; that's a safe spot that won't dismiss the dialog.  Right now, the component isn't dialog-aware, so it clicks half the height of the edit box above or below, depending on whether the datepicker popup is below vs. above.

labkey-chrisj: I'll assume for now you meant that I should create a ReactDateInput that extends input, and handles the popup

labkey-chrisj: Unfortunately, it's hard to close the datepicker if you aren't either clicking in a safe spot outside of it, or selecting a time from the time-list.  I've got an attempt at a basic component that works so far in the use cases I have for it.

src/org/labkey/test/components/ui/EntityBulkInsertDialog.java
labkey-danield: I understand the getCheckBox method but I think the name is confusing. Why not change it to getCheckboxElement?

src/org/labkey/test/components/ui/EntityBulkInsertDialog.java
```diff
@@ -88,6 +91,43 @@ public String getFieldWithId(String id)
...
     }
 
+    public EntityBulkInsertDialog setDateField(String fieldKey, String dateString)
+    {
+        Input input = elementCache().textInput(fieldKey);
+        input.set(dateString);
+        input.getComponentElement().sendKeys(Keys.ENTER);               // enter to collapse the datepicker
+        getWrapper().waitFor(()-> !isDatePickerExpanded(), 1500);
+        return this;
+    }
```
labkey-danield, line 107: Should add a getDateField as well, other wise this is a "write only" field? Even if the getDateField calls getTextField (which it might) it makes the overall interface more consistent/coherent.

src/org/labkey/test/components/ui/EntityBulkInsertDialog.java
```diff
@@ -88,6 +91,43 @@ public String getFieldWithId(String id)
         return getWrapper().getFormElement(Locator.tagWithId("input", id));
     }
 
+    public EntityBulkInsertDialog setDateField(String fieldKey, String dateString)
```
labkey-danield, line 102: We should be more consistent on parameter names for public functions. Examples of other public functions in this class are:
`public EntityBulkInsertDialog setTextField(String fieldCaption, String value)`
and
`public EntityBulkInsertDialog setFieldWithId(String id, String value)`
I'm not sure fieldCaption is a great option, it could be misleading. I find "fieldKey" a little confusing as well. I think we should use parameters name that indicates hints at what attribute the locator will use to find the field.

src/org/labkey/test/components/ui/EntityBulkInsertDialog.java
```diff
@@ -136,6 +176,19 @@ protected ElementCache elementCache()
...
+            return Locator.tagWithClass("div", "row")
+                    .withChild(Locator.tagWithAttribute("label", "for", fieldKey))
+                    .findElement(this);
+        }
+
+        public Input textInput(String fieldKey)
+        {
+            WebElement inputEl = textInputLoc.findElement(elementCache().formRow(fieldKey));
+            return new Input(inputEl, getDriver());
+        }
```
labkey-danield, line 191: Why not replace the locator in the getTextField and setTextField with this locator.

labkey-danield (dismissed):

src/org/labkey/test/components/ui/EntityBulkInsertDialog.java
labkey-chrisj: good call

labkey-chrisj: TBH I'd avoided doing that because doing it would require me to crack open SM tests.
It turns out that getTextField as implemented took the column label text in the dialog, not the fieldKey/column name.
Buuuuut, I'll do it

src/org/labkey/test/components/ui/EntityBulkInsertDialog.java
labkey-chrisj: It's already handled in `setBooleanField , getBooleanField`

src/org/labkey/test/components/ui/EntityBulkInsertDialog.java
labkey-danield: I'm still not a fan of fieldKey as a parameter name because it is not an attribute in the tag, id and name are.
But this is a discussion better had in the Test Team Working Session meeting.

src/org/labkey/test/components/ui/EntityBulkInsertDialog.java
labkey-danield: Thanks!

src/org/labkey/test/components/ui/EntityBulkInsertDialog.java
labkey-chrisj: FWIW, I like the term here because it is also literally the column name, in addition to being an attribute of the field row, and the ID of the inputs that aren't reactSelects.  For them, we're kinda stuck using label text.
Either way, I'm cool with calling the parameters whatever but yeah It's worth talking about

src/org/labkey/test/components/glassLibrary/grids/EditableGrid.java
labkey-chrisj: Done

src/org/labkey/test/components/ui/EntityInsertPanel.java
labkey-chrisj: I've scoped these finders down to the insertPanel

labkey-tchad (approved):

PR #471, opened 2020-09-16
Make test URL helper more flexible

labkey-tchad:
#### Rationale
Adding non-String URL parameters requires converting them to strings before passing them to one of the `buildURL` variants. This change allows one to pass them in directly. This is most useful for `Integer` and `Boolean` parameters.

#### Changes
* Allow non-String URL parameters
* Fix synchronization in `WebTestHelper` (SonarLint suggestion)


labkey-danield (approved):
It looks ok to me, but I'm not sure I understand the "why" of the change.

labkey-tchad:
> It looks ok to me, but I'm not sure I understand the "why" of the change.

I was writing a [`beginAt` method](https://github.com/LabKey/ontology/commit/6c1925f4004a6cc3aa6067b298197d2697e3034b#diff-6de84eadc6476adcfdccef2d3aafa434R28-R30) that takes an integer id and didn't want to call `String.valueOf`.

PR #476, opened 2020-09-18
Fb study cohort export investigation

labkey-chrisj:
#### Rationale
This fixes a purely test issue by using an Html checkbox in a case where the Ext4 checkbox incorrectly returns false on get(), thus doesn't attempt to toggle state if it is checked, leaving it checked.
It doesn't fix the test, however- the issue at hand is that after un-enrolling a cohort and exporting/importing the test, the re-imported cohort shows as enrolled.  It's possible the test imports the wrong file (not the exported one?), or that export is broken/doesn't capture cohort enrollment state correctly

#### Related Pull Requests
https://github.com/LabKey/platform/pull/1579  (updates the base test class to sort archives before selecting for import)

#### Changes
use the right kind of checkbox
use dataregion helper to get rowIndexes in it
add column sort to fileBrowserHelper

### DIFF
--- a/src/org/labkey/test/BaseWebDriverTest.java
+++ b/src/org/labkey/test/BaseWebDriverTest.java
@@ -1719,26 +1719,8 @@ public abstract class BaseWebDriverTest extends LabKeySiteWrapper implements Cle
      */
     public void verifyCohortStatus(DataRegionTable table, String cohort, boolean enrolled)
     {
-        int row = getCohortRow(table, cohort);
+        int row = table.getRowIndex("Label", cohort);
         assertEquals("Enrollment state for cohort " + cohort, String.valueOf(enrolled).toLowerCase(), table.getDataAsText(row, "Enrolled").toLowerCase());
-    }
-
-    /**
-     * Used by CohortTest and StudyCohortExportTest
-     * Retrieves the row for the cohort matching the label passed in
-     */
-    public int getCohortRow(DataRegionTable cohortTable, String cohort)
-    {
-        int row;
-        for (row = 0; row < cohortTable.getDataRowCount(); row++)
-        {
-            String s = cohortTable.getDataAsText(row, "Label");
-            if (0 == s.compareToIgnoreCase(cohort))
-            {
-                break;
-            }
-        }
-        return row;
     }
 
     /**
@@ -1748,13 +1730,11 @@ public abstract class BaseWebDriverTest extends LabKeySiteWrapper implements Cle
     public void changeCohortStatus(DataRegionTable cohortTable, String cohort, boolean enroll)
     {
         // if the row does not exist then most likely the cohort passed in is incorrect
-        int rowIndex = getCohortRow(cohortTable, cohort);
+        int rowIndex = cohortTable.getRowIndex("Label", cohort);
         cohortTable.clickEditRow(rowIndex);
+        WebElement checkboxElement = Locator.name("quf_enrolled").waitForElement(getDriver(), 2000);
+        new org.labkey.test.components.html.Checkbox(checkboxElement).set(enroll);   // the ext4 checkbox doesn't correctly parse checked state of this html checkbox
 
-        if (!enroll)
-            uncheckCheckbox(Locator.name("quf_enrolled"));
-        else
-            checkCheckbox(Locator.name("quf_enrolled"));
         clickButton("Submit");
     }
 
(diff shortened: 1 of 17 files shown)


labkey-tchad (changes requested):
I don't understand the point of this. The old code wasn't doing anything with Ext4. `checkCheckbox` and `uncheckCheckbox` are for html checkboxes. `getCohortRow` is janky and unnecessary but I don't think any of this warrants a separate change until we have an actual fix (though it sounds like there might just be an actual product but).

Also, the `FileBrowserHelper` changes appear to be something intended for a different branch?

labkey-chrisj:
I tried using DataRegionTable's helper (the ListHelper, really) to pass in a map of keys and values, but it also failed to set the checkbox. I assumed from there that, like in BaseWebDriverTest, that an ext4 import was the first Checkbox.
Explicitly using an htmlcheckbox worked here.

Buuut, we can toss this if it makes more sense to you

src/org/labkey/test/util/FileBrowserHelper.java
```diff
@@ -145,6 +148,41 @@ public void selectFileBrowserItem(@LoggedParam String path)
...
+
+    private SortDirection getSortDirection(Locator headerElementLoc)
+    {
+        String elementClass = headerElementLoc.findElement(getDriver()).getAttribute("class");
+        if (elementClass.contains("column-header-sort-DESC"))
+            return SortDirection.DESC;
+        else if (elementClass.contains("column-header-sort-ASC"))
+            return SortDirection.ASC;
+        else
+            return SortDirection.ASC; // if it doesn't have an explicit sort direction, default behavior is ASC
```
labkey-tchad: Default sort behavior isn't anything in particular. Should return null.
```suggestion
            return null; // No sort defined for column
```

src/org/labkey/test/util/FileBrowserHelper.java
```diff
@@ -145,6 +148,41 @@ public void selectFileBrowserItem(@LoggedParam String path)
...
+        // often initial state is unsorted, first click sets it to ASC
+        if (!getSortDirection(headerLoc).equals(sortDirection))
+        {
+            doAndWaitForFileListRefresh(()-> headerElement.click());
+        }
+        // if we intended DESC, it may require another click
+        if (!getSortDirection(headerLoc).equals(sortDirection))
+        {
+            doAndWaitForFileListRefresh(()-> headerElement.click());
+        }
```
labkey-tchad: Flip `equals` checks to prevent NPE from unsorted columns (see my comment on `getSortDirection`).

src/org/labkey/test/BaseWebDriverTest.java
```diff
@@ -1719,42 +1719,22 @@ public DataRegionTable getCohortDataRegionTable()
...
      * Changes the enrolled status of the passed in cohort name
      */
     public void changeCohortStatus(DataRegionTable cohortTable, String cohort, boolean enroll)
     {
         // if the row does not exist then most likely the cohort passed in is incorrect
-        int rowIndex = getCohortRow(cohortTable, cohort);
+        int rowIndex = cohortTable.getRowIndex("Label", cohort);
         cohortTable.clickEditRow(rowIndex);
+        WebElement checkboxElement = Locator.name("quf_enrolled").waitForElement(getDriver(), 2000);
+        new org.labkey.test.components.html.Checkbox(checkboxElement).set(enroll);   // the ext4 checkbox doesn't correctly parse checked state of this html checkbox
```
labkey-tchad: This is more likely to hinder than it is to help. The html Checkbox component works with the same page elements as the `checkCheckbox` and `uncheckCheckbox` methods but it doesn't validate the checkbox state after clicking.

labkey-tchad (changes requested):
There's no import of html checkbox in BaseWebDriverTest because the `un/checkCheckbox` methods predate our current component pattern.

src/org/labkey/test/util/FileBrowserHelper.java
labkey-chrisj: I've figured out a way to fly out the header sort menu; clicking that seems better

src/org/labkey/test/BaseWebDriverTest.java
labkey-chrisj: OK, so it turns out that using DataRegionTable.updateRow(index, map, validate) also does this

labkey-tchad (approved):

PR #477, opened 2020-09-22
Fix intermittent test issues

labkey-danield:
#### Rationale
Trying to fix long running intermittent test issues with [ReportThumbnailTest.testSteps](https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailySqlserverb/1121712?buildTab=tests&name=ReportThumbnail&pager.currentPage=1) and [VocabularyViewSupportTest.testSampleSetViewSupport](https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyCPostgres/1121786?buildTab=tests&name=Vocabulary&pager.currentPage=1).

#### Related Pull Requests
* [Change in platform module](https://github.com/LabKey/platform/pull/1584).

#### Changes
* Added a better check to the vocabulary test to validate a table as a map object and not as a string.
* Added a sleep to wait for the thumbnail graph to get generated

Tried to add a validation to the checkbox control to assert that it is set as expected. However this proved to be unreliable for radio buttons which inherit from the checkbox. So I removed that change but added comments to various other test libraries calling out some of the issues I ran into.

### DIFF
--- a/src/org/labkey/test/WebDriverWrapper.java
+++ b/src/org/labkey/test/WebDriverWrapper.java
@@ -3457,10 +3457,11 @@ public abstract class WebDriverWrapper implements WrapsDriver
         RadioButton radioButton = RadioButton(radioButtonLocator).find(getDriver());
         radioButton.check();
 
+        // TODO: Should review this code, and understand why it was written.
         // I'm not sure this code is doing what was intended, or is the correct thing to do.
         // I think this may be exposing a bug in the RadioButton finder. Look at baseWebDriverTest.prepareForFolderExport (line 1448).
         // The locator sent into this function is a table, which will find the input element, but it makes the base
-        // element table which never returns true for checked.
+        // element the table as well which will never returns true for checked.
         if (!waitFor(()-> radioButton.isChecked(), 250))
             radioButton.check();
         return radioButton;
--- a/src/org/labkey/test/WebDriverWrapper.java
+++ b/src/org/labkey/test/WebDriverWrapper.java
@@ -3456,6 +3456,11 @@ public abstract class WebDriverWrapper implements WrapsDriver
     {
         RadioButton radioButton = RadioButton(radioButtonLocator).find(getDriver());
         radioButton.check();
+
+        // I'm not sure this code is doing what was intended, or is the correct thing to do.
+        // I think this may be exposing a bug in the RadioButton finder. Look at baseWebDriverTest.prepareForFolderExport (line 1448).
+        // The locator sent into this function is a table, which will find the input element, but it makes the base
+        // element table which never returns true for checked.
         if (!waitFor(()-> radioButton.isChecked(), 250))
             radioButton.check();
         return radioButton;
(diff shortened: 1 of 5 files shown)


src/org/labkey/test/components/html/Table.java
```diff
@@ -79,6 +79,19 @@ public int getRowCount()
...
     }
 
+    /**
+     * For a well formed html table get the text from the th elements in the thead.
+     *
+     * @return A list of the text values contained in the th tags in the thead.
+     */
+    public List<String> getTableHeaderTexts()
+    {
+        List<WebElement> headerEls = Locator.xpath("//thead//th").findElements(this);
```
labkey-tchad, line 89: That Locator would find headers in nested tables. We can make it a bit more strict and only find a specific table's header.
```suggestion
        List<WebElement> headerEls = Locator.xpath("./thead/th").findElements(this);
```

src/org/labkey/test/WebDriverWrapper.java
```diff
@@ -3456,6 +3456,11 @@ public RadioButton checkRadioButton(Locator radioButtonLocator)
     {
         RadioButton radioButton = RadioButton(radioButtonLocator).find(getDriver());
         radioButton.check();
+
+        // I'm not sure this code is doing what was intended, or is the correct thing to do.
+        // I think this may be exposing a bug in the RadioButton finder. Look at baseWebDriverTest.prepareForFolderExport (line 1448).
+        // The locator sent into this function is a table, which will find the input element, but it makes the base
+        // element table which never returns true for checked.
```
labkey-tchad: `RadioButton` and `Checkbox` have an `assertElementType` to make sure they're the correct type of element. However, I never enabled it because I knew in my heart that there would be places like this that are using them as a glorified click helper.

src/org/labkey/test/components/html/RadioButton.java
```diff
@@ -20,6 +20,9 @@
 import org.openqa.selenium.SearchContext;
 import org.openqa.selenium.WebElement;
 
+// TODO: We may want to revisit this inheritance. Specifically I have run into problems being able to check if a
+//  radio button is checked or not. It doesn't look like their status (checked/unchecked) is always stored as an
+//  attribute of the input control. Sometimes it is stored as an attribute in one of it's ancestor tags.
 public class RadioButton extends Checkbox
```
labkey-tchad: `WebElement.isSelected` should be pretty reliable for standard HTML checkboxes and radio buttons. The problems you mention might be more due to wrapping the wrong element (like the `table` passed in from `prepareForFolderExport`)
I don't disagree about revisiting the inheritance though. Not sure that we get a ton of utility . `check` isn't a great verb for radio buttons (doubly true for `uncheck`)

labkey-tchad (approved):
Just a note about the table header locator

src/org/labkey/test/WebDriverWrapper.java
labkey-danield: I'm afraid that there is code like this in multiple places.

src/org/labkey/test/components/html/RadioButton.java
labkey-danield: I have trimmed the comment to be less verbose.

src/org/labkey/test/components/html/Table.java
labkey-danield: Changed.

PR #478, opened 2020-09-22
Update staleness checks and other fixes

labkey-tchad:
#### Rationale
* Don't check server errors after tests when running locally
  * Doing so might need to stop impersonating, which will navigate away from the test failure.
* Don't double-log impersonation
  * `LKSW.impersonate` calls `SiteNavBar.UserMenu.impersonate`, and they would both would get logged
* Include partial server stack traces in test log
  * TeamCity discards artifacts (such as server logs) after a couple of weeks but the test logs stick around for much longer.
* Use 'isEnabled' to trigger staleness checks
  * `LazyWebElement.isDisplayed` won't throw `StaleElementReferenceException` or `NoSuchElementException` (#468). Need to call `isEnabled` instead.

### DIFF
--- a/src/org/labkey/test/aspects/ImpersonationLoggingAspect.java
+++ b/src/org/labkey/test/aspects/ImpersonationLoggingAspect.java
@@ -43,7 +43,7 @@ public class ImpersonationLoggingAspect
             _impersonating = impersonating;
             _startTime = System.currentTimeMillis();
         }
-        else
+        else if (!_impersonating.equals(impersonating))
         {
             TestLogger.decreaseIndent();
             TestLogger.log("><Switch Impersonation : " + _impersonating + TestLogger.formatElapsedTime(System.currentTimeMillis() - _startTime) + " -> " + impersonating);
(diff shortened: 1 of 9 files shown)


src/org/labkey/test/aspects/ImpersonationLoggingAspect.java
```diff
@@ -43,7 +43,7 @@ public void beforeImpersonation(String impersonating)
             _impersonating = impersonating;
             _startTime = System.currentTimeMillis();
         }
-        else
+        else if (!_impersonating.equals(impersonating))
```
labkey-danield, line 46: Just curious why you have to do this check. Would there be any value in adding an else clause that basically says _impersonating equals impersonating so not going to do anything?

labkey-danield (approved):

src/org/labkey/test/aspects/ImpersonationLoggingAspect.java
labkey-tchad: After we pushed impersonation methods into the `UserMenu` component, we ended up having two methods in the call stack that matched this Aspect and they would both get logged, resulting in something like:
```
03:03:42,909   >>Impersonate - abstractor1ncitestproject@nlpui.test
03:03:42,910   ><Switch Impersonation : abstractor1ncitestproject@nlpui.test <1ms> -> abstractor1ncitestproject@nlpui.test
```

PR #479, opened 2020-09-23
Create a component test class for EditableGrid

labkey-chrisj:
#### Rationale
This adds basic test coverage for component behavior of EditableGrid in core-components view.
To date, the only ways to test components like this has been to bloat scenario tests in Biologics or SM.
The test in BiologicsComponentTest.checkErrorCasesInEditableGrid sometimes fails because testing the second case in it collides with the (expected) prior error- creating this should replace that coverage, but more reliably.

For the nonce, the core-components.view does not support getting an EditableGrid for an arbitrary schema- [issue 41459](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41459) should address this.  Once this is possible we can add test coverage for cases like lookup field interactions

#### Related Pull Requests
n/a
(possibly, remove checkErrorCasesInEditableGrid from the Biologics suite)

#### Changes
add ability to get cell error text from its related popover

### DIFF
--- a/src/org/labkey/test/tests/component/EditableGridTest.java
+++ b/src/org/labkey/test/tests/component/EditableGridTest.java
@@ -17,7 +17,7 @@ import static org.hamcrest.CoreMatchers.hasItems;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
 
-@Category({DailyB.class})   // until there's a component suite
+@Category({DailyB.class})
 public class EditableGridTest extends BaseWebDriverTest
 {
     @Override
--- a/src/org/labkey/test/tests/component/EditableGridTest.java
+++ b/src/org/labkey/test/tests/component/EditableGridTest.java
@@ -13,9 +13,10 @@ import org.labkey.test.pages.test.CoreComponentsTestPage;
 import java.util.Arrays;
 import java.util.List;
 
-import static org.hamcrest.CoreMatchers.hasItems;
+import static org.hamcrest.CoreMatchers.everyItem;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.Assert.assertEquals;
 
 @Category({DailyB.class})
 public class EditableGridTest extends BaseWebDriverTest
@@ -56,6 +57,8 @@ public class EditableGridTest extends BaseWebDriverTest
         assertThat("Expect cell error to explain that paste cannot add columns",
                 testGrid.getCellPopoverText(0, "Description"),
                 is("Unable to paste. Cannot paste columns beyond the columns found in the grid."));
+        assertThat("Expect failed paste to leave data unchanged",
+                testGrid.getColumnData("LSID *"), everyItem(is("")));
     }
 
     @Test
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/tests/component/EditableGridTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @Test
+    public void testTooWideErrorCase() throws Exception
+    {
+        CoreComponentsTestPage testPage = CoreComponentsTestPage.beginAt(this, getProjectName());
+        EditableGrid testGrid = testPage.getEditableGrid();
+        String wideShape = "Too wide\tACME\tthing\tanother\televen\toff the map\tMoar columns\n";
+
+        testGrid.pasteFromCell(0, "Description", wideShape);
+        assertThat("Expect cell error to explain that paste cannot add columns",
```
cnathe, line 57: might also be worth an assert that the column to the right of description is still empty (or hasn't change from value prior to paste)

cnathe (dismissed):
These initial 2 test cases look good to me. I'll leave the test code review to the Test team. I assume there is a backlog list of other scenarios to add to this editable grid test in the future (likely after I get that issue fixed which allows the core-components page to select which schema/query to use for the editable grid).

cnathe (commented):
Oops, meant to submit my review as "Comment" not "Request changes"

labkey-kevink (approved):

src/org/labkey/test/tests/component/EditableGridTest.java
```diff
@@ -0,0 +1,99 @@
...
+                "38";
+
+        int initialRowCount = testGrid.getRowCount();
+        assertThat(initialRowCount, is(2));
+
+        testGrid.pasteFromCell(0, "Description", tallShape);
+        List<String> pastedColData = testGrid.getColumnData("Description");
+
+        assertThat(pastedColData, hasItems("42", "41", "40", "39", "38"));
+        assertThat(testGrid.getRowCount(), is(5));
```
labkey-tchad: You can assert the size and contents in one line:
```suggestion
        assertEquals("Didn't get correct values", List.of("42", "41", "40", "39", "38"), pastedColData);
```
This would also verify that the values appear in the correct order.

src/org/labkey/test/tests/component/EditableGridTest.java
```diff
@@ -0,0 +1,99 @@
...
+    }
```
labkey-tchad, line 85: Maybe verify that other columns are empty? Are there other columns?

src/org/labkey/test/tests/component/EditableGridTest.java
```diff
@@ -0,0 +1,99 @@
...
+import org.labkey.test.pages.test.CoreComponentsTestPage;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static org.hamcrest.CoreMatchers.hasItems;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+@Category({DailyB.class})   // until there's a component suite
```
labkey-tchad: There _is_ a component suite!
Our test runner creates one based on the package name after `tests`, e.g. `org.labkey.test.tests.`**`component`**
That being said. We want this to be in a daily suite as well.
```suggestion
@Category({DailyB.class})
```

labkey-tchad (approved):
Some minor suggestions

src/org/labkey/test/tests/component/EditableGridTest.java
labkey-chrisj: done, tested locally

src/org/labkey/test/tests/component/EditableGridTest.java
labkey-chrisj: Yes, I've added checks to see that adjacent cells aren't affected in both tests

src/org/labkey/test/tests/component/EditableGridTest.java
labkey-chrisj: done

src/org/labkey/test/tests/component/EditableGridTest.java
labkey-chrisj: Nice!

PR #482, opened 2020-09-25
Support selecting a grid row by partial map of column, text values

labkey-chrisj:
#### Rationale
This adds support for selecting a row in a QueryGrid based on values it might have in more than one column.
The grid already supports finding a row by passing in a partialMap (of column, text).

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/697

#### Changes
New overload for `selectRow()`

### DIFF
--- a/src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
+++ b/src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
@@ -147,9 +147,12 @@ public class ResponsiveGrid<T extends ResponsiveGrid> extends WebDriverComponent
      * @param checked       the desired checkbox state
      * @return
      */
-    public ResponsiveGrid selectRow(Map partialMap, boolean checked)
+    public ResponsiveGrid selectRow(Map<String, String> partialMap, boolean checked)
     {
-        getRow(partialMap).select(checked);
+        GridRow row = getRow(partialMap);
+        selectRowAndVerifyCheckedCounts(row, checked);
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
```diff
@@ -141,6 +141,18 @@ public ResponsiveGrid selectRow(int index, boolean checked)
         return getThis();
     }
 
+    /**
+     * Finds the first row with the specified texts in the specified columns, and sets its checkbox
+     * @param partialMap    key-column, value-text in that column
+     * @param checked       the desired checkbox state
+     * @return
+     */
+    public ResponsiveGrid selectRow(Map partialMap, boolean checked)
```
labkey-kevink: You should include the generic types for the Map.

```suggestion
    public ResponsiveGrid selectRow(Map<String, String> partialMap, boolean checked)
```

Also the `selectRow(String columnLabel, String text, boolean checked)` method validates the checkbox count after selecting.  Do you need to do that here too?

labkey-kevink (changes requested):

src/org/labkey/test/components/glassLibrary/grids/ResponsiveGrid.java
labkey-chrisj: It's probably a good idea so I will

labkey-kevink (approved):

labkey-tchad (approved):

PR #485, opened 2020-09-29
SchemaHelper cleanup

labkey-tchad:
#### Rationale
Taking separate project and folder parameters is limiting and inconsistent with other parameters in the same method even. I'd made these changes as part of the new `ContainerFilterQueryTest`. I'd pulled them out to simplify that pull request but forgot the leading slash fix.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1587

#### Changes
* Make create/update linked schema take container paths instead of separate project and folder parameters
* Add leading slash ('/') to container paths to work with combo box

### DIFF
--- a/src/org/labkey/test/tests/LinkedSchemaTest.java
+++ b/src/org/labkey/test/tests/LinkedSchemaTest.java
@@ -424,7 +424,7 @@ public class LinkedSchemaTest extends BaseWebDriverTest
     public void lookupTest()
     {
         String sourceContainerPath = "/" + getProjectName() + "/" + SOURCE_FOLDER;
-        _schemaHelper.createLinkedSchema(getProjectName(), TARGET_FOLDER, "BasicLinkedSchema", sourceContainerPath, null, "lists", "NIMHDemographics,NIMHPortions", null);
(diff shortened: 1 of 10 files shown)


src/org/labkey/test/tests/LinkedSchemaTest.java
```diff
@@ -424,7 +424,7 @@ void deleteLinkedSchema()
     public void lookupTest()
     {
         String sourceContainerPath = "/" + getProjectName() + "/" + SOURCE_FOLDER;
-        _schemaHelper.createLinkedSchema(getProjectName(), TARGET_FOLDER, "BasicLinkedSchema", sourceContainerPath, null, "lists", "NIMHDemographics,NIMHPortions", null);
+        _schemaHelper.createLinkedSchema(getProjectName() + "/" + TARGET_FOLDER, "BasicLinkedSchema", sourceContainerPath, null, "lists", "NIMHDemographics,NIMHPortions", null);
```
labkey-chrisj, line 427: At some point it might be helpful to have a paths or urls util do the work of assembling strings of <container name><delimiter char> etc.   Not asking that we do that now, but we should give some thought to how to approach it

labkey-chrisj (approved):

labkey-danield (approved):

PR #487, opened 2020-09-30
Create a test class for EntityInsertPanel component

labkey-chrisj:
#### Rationale
This doesn't yet provide full coverage for all the important behaviors of the EntityInsertPanel.  It is a starting point.
Currently, most of the functional testing EntityInsertPanel sees is done as a part of automted testing in SampleManager and Biologics.  Covering functional behavior here may mean we don't have to do so much in the product suites

### DIFF
--- a/src/org/labkey/test/tests/component/EntityInsertPanelTest.java
+++ b/src/org/labkey/test/tests/component/EntityInsertPanelTest.java
@@ -71,6 +71,21 @@ public class EntityInsertPanelTest extends BaseWebDriverTest
         assertThat(success.getText(), is("Created 3 samples in sample type 'insert_from_edit_grid'."));
         List<Map<String, Object>> insertedRows = dgen.getRowsFromServer(createDefaultConnection()).getRows();
         assertThat(insertedRows.size(), is(3));
+        Map<String, Object> ed_row = insertedRows.stream().filter(a-> a.get("name").equals("ed")).findFirst().get();
+        assertThat(ed_row.get("intColumn"), is(2));
+        // eventually, parse the date out of dateColumn
+        assertThat(ed_row.get("stringColumn"), is("stringfellow"));
+        assertThat(ed_row.get("boolColumn"), is(true));
+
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/tests/component/EntityInsertPanelTest.java
```diff
@@ -0,0 +1,123 @@
...
+        String sampleTypeName = "insert_from_edit_grid";
+        SampleTypeDefinition props = new SampleTypeDefinition(sampleTypeName).setFields(standardTestSampleFields());
+        TestDataGenerator dgen = SampleTypeAPIHelper.createEmptySampleType(getProjectName(), props);
+        CoreComponentsTestPage testPage = CoreComponentsTestPage.beginAt(this, getProjectName());
+        EntityInsertPanel testPanel = testPage.getEntityInsertPanel();
+        testPanel.getEntityTypeSelect("Sample Type").select(sampleTypeName);
+
+        String pasteText = "ed\tbrother\tthe quiet one\t\t2\tstringfellow\t11/11/2020\ttrue\n" +
+                "jed\tother brother\tthe squinty one\t\t3\tstrongfellow\t11/11/2020\tfalse\n" +
+                "ted\tother other brother\tisn't sure about the others\t\t4\tstrangefellow\t11/11/2020\ttrue";
```
labkey-danield, line 66: I guess I'm still a little confused about this test data. 
It looks like test validates the number of rows added but not the content. If test does not care about the content and only the count, this data is a little bit of overkill.
If there is a concern that the form may not accurately set a field (because this is validating the test form and not LKS), then the validation code below should also validate that the fields have the values that were set.

labkey-danield (approved):

src/org/labkey/test/tests/component/EntityInsertPanelTest.java
labkey-chrisj: OK, I will validate the complete dataset here before pushing

PR #490, opened 2020-10-05
Fix timeout in 'testInsertLargeLineageGraph'

labkey-tchad:
#### Rationale
It regularly causes timeout failures and isn't providing interesting test coverage. Hopefully deleting the entire container at the end of the test is more reliable.

#### Changes
* Don't delete domain at end of test.


labkey-danield (approved):
Let's give it a shot and see.

PR #491, opened 2020-10-05
Item 7767: Automated test updates for the new error page

labkey-ankurj:
#### Rationale
Automated test updates for the new error page

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1589

### DIFF
--- a/src/org/labkey/test/BaseWebDriverTest.java
+++ b/src/org/labkey/test/BaseWebDriverTest.java
@@ -187,9 +187,6 @@ public abstract class BaseWebDriverTest extends LabKeySiteWrapper implements Cle
     protected static boolean _checkedLeaksAndErrors = false;
     private static final String ACTION_SUMMARY_TABLE_NAME = "actions";
 
-    protected static final String PERMISSION_ERROR = "User does not have permission to perform this operation.";
-    protected static final String NOT_FOUND_ERROR = "notFound";
-
     static final Set<String> urlsSeen = new HashSet<>();
 
     static
@@ -1541,12 +1538,6 @@ public abstract class BaseWebDriverTest extends LabKeySiteWrapper implements Cle
         {
             log("properties were already set, no changed needed");
         }
-    }
-
-    public void assertAtUserUserLacksPermissionPage()
-    {
-        assertTextPresent(PERMISSION_ERROR);
-        assertTitleEquals("403: Error Page -- User does not have permission to perform this operation.");
     }
 
     public void assertNavTrail(String... links)
--- a/src/org/labkey/test/BaseWebDriverTest.java
+++ b/src/org/labkey/test/BaseWebDriverTest.java
@@ -188,6 +188,7 @@ public abstract class BaseWebDriverTest extends LabKeySiteWrapper implements Cle
     private static final String ACTION_SUMMARY_TABLE_NAME = "actions";
 
     protected static final String PERMISSION_ERROR = "User does not have permission to perform this operation.";
+    protected static final String NOT_FOUND_ERROR = "notFound";
 
     static final Set<String> urlsSeen = new HashSet<>();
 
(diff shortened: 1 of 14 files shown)


src/org/labkey/test/components/html/SiteNavBar.java
```diff
@@ -285,10 +286,8 @@ public void impersonate(String fakeUser)
 
             AbstractUserHelper.saveCurrentDisplayName(getWrapper());
 
-            if (getWrapper().isElementPresent(Locator.lkButton("Home")))
-            {
-                getWrapper().clickAndWait(Locator.lkButton("Home"));
-            }
+            getWrapper().goToHome();
+
```
labkey-tchad: Don't change the behavior of this method (It only navigates if ends up on a permission denied page after impersonating). It is fairly widely used and some probably expect to be on the same page after impersonating.
Does the 403 page (or whatever page we end up on) not have a link to go to home anymore?

src/org/labkey/test/tests/SecurityTest.java
```diff
@@ -858,7 +861,8 @@ public void loginSelfRegistrationDisabledTest()
         assertFalse("Self-registration button is visible", link != null && link.isDisplayed());
 
         beginAt(buildURL("login", "register"));
-        assertElementPresent(Locators.labkeyError.withText("Registration is not enabled."));
+        waitForAnyElement(Locators.labkeyError, Locators.labkeyErrorSubHeading);
+        assertElementPresent(Locators.labkeyErrorSubHeading.withText("Registration is not enabled."));
```
labkey-tchad: This check should be specific. We know what page we should end up on. If timing is a problem, we can wait for the error.
```suggestion
        waitForElement(Locators.labkeyErrorSubHeading.withText("Registration is not enabled."));
```

src/org/labkey/test/BaseWebDriverTest.java
```diff
@@ -188,6 +188,7 @@
     private static final String ACTION_SUMMARY_TABLE_NAME = "actions";
 
     protected static final String PERMISSION_ERROR = "User does not have permission to perform this operation.";
+    protected static final String NOT_FOUND_ERROR = "notFound";
```
labkey-tchad: Move `PERMISSION_ERROR` and `NOT_FOUND_ERROR` to `SecurityTest`. The only other place that uses `PERMISSION_ERROR` is an unused method, which can be removed: `BaseWebDriverTest.assertAtUserUserLacksPermissionPage`.

labkey-tchad (changes requested):

src/org/labkey/test/components/html/SiteNavBar.java
labkey-ankurj: We do not have home button on the error page anymore. Should I check for the response code from the title of the page and then route to home?

labkey-tchad: Yeah. And just click the logo in the upper-left. I'd rather not move `goToHome` into `WebDriverWrapper`.

labkey-tchad (commented):
A couple of questions.
 - In `Crawler.isIgnoredError`, we ignore 404s from modules that aren't enabled in the current container. Should we update that check or do we think 404 is an inappropriate response for those actions? It probably isn't appropriate since we have links to all modules in the admin menu, regardless of whether they are enabled.
 - Is `Locator.css("table.server-error")` still valid? Its used in `Crawler.checkForServerError` and `Crawler.crawlLink`.

labkey-jeckels:
@labkey-tchad An action from an inactive module won't automatically return a 404. Most should return 200 and work OK. For example, neither Study nor Flow is enabled in this folder:

https://www.labkey.org/home/study-begin.view?

https://www.labkey.org/home/flow-begin.view?

I see exactly one spot in the code that returns a 404 with the message about a module not being enabled, in a MS2 protein search action.

labkey-ankurj:
@labkey-tchad I think `Locator.css("table.server-error")` is still valid because that is still used in WebPartErrorRenderer.

labkey-ankurj:
@labkey-tchad all the tests passed. Let me know if there are changes to be made.

labkey-tchad (approved):

PR #492, opened 2020-10-07
Back-port test fixes from develop to 20.7 branch.

labkey-danield:
#### Rationale
CDS development is done in the 20.7 branch. There were some changes to shared test code in the develop branch that would be useful in fixing /updating the CDS tests. I'd like to back port these two changes to the 20.7 branch.

#### Related Pull Requests
* None at this time.

#### Changes
* The change to Crawler.getDefaultExcludedActions that ignores two actions in CDS.
* The change to LazyWebElement.isDisplayed to return false if the element is not present on the page.

### DIFF
--- a/src/org/labkey/test/selenium/LazyWebElement.java
+++ b/src/org/labkey/test/selenium/LazyWebElement.java
@@ -17,7 +17,9 @@ package org.labkey.test.selenium;
 
 import org.jetbrains.annotations.NotNull;
 import org.labkey.test.Locator;
+import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.SearchContext;
+import org.openqa.selenium.StaleElementReferenceException;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.ui.FluentWait;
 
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/selenium/LazyWebElement.java
```diff
@@ -39,6 +41,19 @@ public LazyWebElement(@NotNull Locator locator, @NotNull SearchContext searchCon
...
+        try
+        {
+            return super.isDisplayed();
+        }
+        catch (StaleElementReferenceException | NoSuchElementException ex)
+        {
+            return false;
+        }
+    }
+
```
labkey-tchad: There are a handful of places that use 'isDisplayed' to trigger a staleness check (which this breaks). Some are in [testAutomation](https://github.com/LabKey/testAutomation/pull/478), [adjudication](https://github.com/LabKey/adjudication/pull/44), and [sample manager](https://github.com/LabKey/sampleManagement/pull/371).
I checked the CDS changes that go along with this and you could tweak the existing assertions and not need the `isDisplayed` change. They would look something like:
```
checker().wrapAssertion(() -> assertElementPresent(Locator.tagWithClass("div", "detail-gray-text").containing("HIV")));
```

labkey-tchad (changes requested):

src/org/labkey/test/selenium/LazyWebElement.java
labkey-danield: I was unaware of that.
I've reverted this change.

labkey-tchad (approved):

PR #496, opened 2020-10-09
Make toggleLocator settable

labkey-chrisj:
#### Rationale
A previous change modified the BootstrapMenu's toggleAnchor locator to be the union of 2 possible locators, but this broke some tests.  This restores the old locator as the default, but makes that locator settable in situations where that will make the BootstrapMenu component wrapper work for menus that define their toggle element differently

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/709

#### Changes
revert the original change to leave the default locator how it was
add a setter for an optional locator which if defined will be used instead of the default

### DIFF
--- a/src/org/labkey/test/components/html/BootstrapMenu.java
+++ b/src/org/labkey/test/components/html/BootstrapMenu.java
@@ -26,6 +26,8 @@ import java.util.List;
 
 public class BootstrapMenu extends BaseBootstrapMenu
 {
+    private Locator _toggleLocator;
+
     /* componentElement should contain the toggle anchor *and* the UL containing list items */
     public BootstrapMenu(WebDriver driver, WebElement componentElement)
     {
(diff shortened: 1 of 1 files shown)


labkey-nicka:
I presume this is intending to revert #494, however, I don't see that revert in these changes (as is mentioned in the PR summary).

labkey-chrisj:
> I presume this is intending to revert #494, however, I don't see that revert in these changes (as is mentioned in the PR summary).

Yes, this doesn't revert #494, it just restores the same default locator as it had before.
The difference here is that it adds a setter, so the component can work with a toggle that looks different.
If we want, we could just revert #494 but that will require a re-write of test code in biologics that uses the new setter

labkey-nicka (approved):
Want to unblock broken premium tests.

PR #499, opened 2020-10-13
Have tests wait for app content

labkey-tchad:
#### Rationale
Now that error pages are a special single-page app, tests need to know how to wait for them to load. Several tests have been failing because an expected error wasn't present soon enough. Waiting for such errors on a case-by-case basis isn't a viable solution because we need to notice them when they aren't expected (especially in the Crawler).

#### Changes
* If a single page app container is present after a page load, wait for it to contain something.

### DIFF
--- a/src/org/labkey/test/Locator.java
+++ b/src/org/labkey/test/Locator.java
@@ -939,7 +939,7 @@ public abstract class Locator extends By
 
     public static String cq(String value)
     {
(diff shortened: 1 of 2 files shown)


labkey-ankurj (approved):
This looks good, I was adding waits in fbs for the cases.
Thanks,
Ankur

labkey-chrisj (approved):
It may be worth wrapping an error page up into its own page class eventually, but this is good

labkey-danield (approved):

PR #501, opened 2020-10-14
7711: Surface 'add user' button on project users page

RosalineP:
#### Rationale
Functionality has been added to ensure a Folder Admin can add a new user.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1631

#### Changes
* Add test projectUsersTest()


src/org/labkey/test/tests/SecurityTest.java
```diff
@@ -553,6 +555,23 @@ protected void clonePermissionsTest()
...
+        usersPage
+                .clickAddUsers()
+                .setNewUsers(Arrays.asList(userEmail))
+                .setSendNotification(true)
+                .clickAddUsers();
+
+        assertTextPresent(userEmail);
+        stopImpersonating();
+    }
+
```
RosalineP, line 586: Could also add a check, if it seems worthwhile, to impersonate folder admin and verify the 'Add Users' button is not present.

src/org/labkey/test/tests/SecurityTest.java
```diff
@@ -553,6 +555,23 @@ protected void clonePermissionsTest()
         assertNavTrail("Site Users", "User Details", "Permissions");
     }
 
+    protected void projectUsersTest()
+    {
+        beginAt("/project/SecurityVerifyProject/begin.view?");
+        impersonateRoles(PROJECT_ADMIN_ROLE);
+        ShowUsersPage usersPage = goToProjectUsers();
+        String userEmail = "fromprojectusers@gmail.com";
```
labkey-tchad: Don't use real domains for test email addresses. We don't want to end up on a spam list from TeamCity spitting nonsense emails into the world. We usually use "`.test`" as it is a reserved top level domain, safe for testing.
```suggestion
        String userEmail = "fromprojectusers@security.test";
```
Also, we should delete this user during cleanup, otherwise this test would fail if you tried to run it twice. Define the email in a class constant and add it to the list of users to delete in `doCleanup` like we do with the other users (e.g. `NORMAL_USER`).

src/org/labkey/test/tests/SecurityTest.java
```diff
@@ -553,6 +555,23 @@ protected void clonePermissionsTest()
         assertNavTrail("Site Users", "User Details", "Permissions");
     }
 
+    protected void projectUsersTest()
+    {
+        beginAt("/project/SecurityVerifyProject/begin.view?");
```
labkey-tchad: Use helper methods when able. It makes future refactors easier.
```suggestion
        beginAt(WebTestHelper.buildUrl("project", getProjectName(), "begin"));
```

src/org/labkey/test/tests/SecurityTest.java
```diff
@@ -128,6 +129,7 @@ public void testSteps() throws IOException
         }
 
         clonePermissionsTest();
+        projectUsersTest();
```
labkey-tchad: `projectUsersTest` should go into its own `@Test` method, if able. It doesn't appear to depend on or impact any of the other tests.

src/org/labkey/test/tests/SecurityTest.java
labkey-tchad: I think that's a worthwhile test. The distinction between project and folder admins can be unclear at times and would be worth verifying here.

labkey-tchad (changes requested):

src/org/labkey/test/tests/SecurityTest.java
RosalineP: Ah, gotcha! Done.

src/org/labkey/test/tests/SecurityTest.java
RosalineP: Done!

src/org/labkey/test/tests/SecurityTest.java
RosalineP: When I pulled it out, it flopped because `clonePermissionsTest()` in `testSteps()` creates the project in the first place. In this case, is it alright to piggyback off of that created project, or should I be setting up my own project/factoring out the project creation into a setup fn?

labkey-tchad (approved):

labkey-tchad (changes requested):
This test is failing on TeamCity. It looks like the corresponding branch in platform has different casing.

labkey-tchad (approved):

PR #502, opened 2020-10-15
Fix intermittent error in 'SpecimenTest'

labkey-tchad:
#### Rationale
`SpecimenTest` regularly fails on TeamCity because it can't find the expected email attachments. My presumption is that the radio button to set the attachment type is sometimes right at the bottom of the browser viewport; in that sweet spot that sometimes doesn’t scroll automatically but doesn’t register the click.

#### Changes
* Make sure radio buttons are in view
* Separate Excel and TSV attachment checks so we don't have to worry about order
* Add debugging screenshots for apparently incomplete 'verifyInactiveUsersInRequests'


labkey-chrisj (approved):
Thanks for doing this

src/org/labkey/test/pages/study/specimen/ManageNotificationsPage.java
```diff
@@ -0,0 +1,175 @@
...
+    {
+        webDriverWrapper.beginAt(WebTestHelper.buildURL("study-samples", containerPath, "manageNotifications"));
+        return new ManageNotificationsPage(webDriverWrapper.getDriver());
+    }
+
+    public ManageNotificationsPage setReplyTo(@Nullable String email)
+    {
+        if (email == null)
+        {
+            elementCache().replyToAdminRadio.check();
```
labkey-danield, line 43: Should this be uncheck?

src/org/labkey/test/pages/study/specimen/ManageNotificationsPage.java
```diff
@@ -0,0 +1,175 @@
...
+            shortWait().until(ExpectedConditions.visibilityOf(elementCache().ccInput.getComponentElement()));
+            elementCache().ccInput.set(String.join("\n", users));
+        }
+        return this;
+    }
+
+    public ManageNotificationsPage setDefaultEmailNotification(DefaultEmailNotify option)
+    {
+        scrollIntoView(elementCache().saveButton); // Make sure radio buttons are fully in view
+        elementCache().defaultEmailNotifyRadio(option).check();
```
labkey-danield, line 92: Why not scroll the option into view? I'm assuming that the button is scrolled into view because it is next to the options.

src/org/labkey/test/pages/study/specimen/ManageNotificationsPage.java
```diff
@@ -0,0 +1,175 @@
...
+import org.labkey.test.components.html.Input;
+import org.labkey.test.components.html.RadioButton;
+import org.labkey.test.pages.LabKeyPage;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.ui.ExpectedConditions;
+
+import java.util.List;
+
+public class ManageNotificationsPage extends LabKeyPage<ManageNotificationsPage.ElementCache>
```
labkey-danield, line 21: Aren't we adding a comment at the top identifying the dev code / page object this represents?

labkey-danield (approved):
I think there might be one issue (check vs uncheck).
Screen shot would have helped.

src/org/labkey/test/pages/study/specimen/ManageNotificationsPage.java
labkey-tchad: There are two radio buttons. This one says to use a use a "reply-to" address inherited from global site setting. The other one makes an input appear where you can specify an arbitrary email address. I decided to give `null` a special meaning instead of making 2 or 3 methods for setting the reply-to address.

PR #505, opened 2020-10-21
Announcements: API endpoint tests

labkey-nicka:
#### Rationale
Adds API integration testing for announcements.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1644

#### Changes
* Introduce `AnnouncementAPITest` for coverage of all announcement APIs.
* Create wrappers for command requests and responses for all announcement APIs.


src/org/labkey/remoteapi/announcements/AbstractMessageThreadCommand.java
```diff
@@ -0,0 +1,20 @@
+package org.labkey.remoteapi.announcements;
+
+import org.json.simple.JSONObject;
+import org.labkey.remoteapi.PostCommand;
+
+public class AbstractMessageThreadCommand extends PostCommand<MessageThreadResponse>
```
labkey-klum: should this be abstract, or is it useful on it's own?

src/org/labkey/remoteapi/announcements/TestAnnouncementModel.java
```diff
@@ -0,0 +1,263 @@
...
+import org.json.simple.JSONArray;
+import org.json.simple.JSONObject;
+import org.labkey.test.util.WikiHelper;
+
+import java.util.ArrayList;
+import java.util.List;
+
+// POJO for working with responses from Announcements APIs.
+// This is modeled after AnnouncementModel on the server.
+public class TestAnnouncementModel
```
labkey-klum: I was a little confused by the name of this class when I saw it, especially since it is exposed in the remote API. Would it be too confusing to call it AnnouncementModel as well?

labkey-klum (approved):

src/org/labkey/remoteapi/announcements/TestAnnouncementModel.java
labkey-nicka: Renamed.

src/org/labkey/remoteapi/announcements/AbstractMessageThreadCommand.java
labkey-nicka: Good catch. I missed this when reviewing.

PR #511, opened 2020-10-27
Remove ignored test case that provided coverage for issue 38436

labkey-chrisj:
#### Rationale
This removes test coverage for [issue 38436](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=38436), which we have decided not to fix.  The test has been marked 'ignore', this culls it

#### Changes
removes an ignored test


labkey-tchad (commented):
I'm not sure what the conclusion of the related issue was. If we don't plan on ever supporting this, remove the test. If we expect to support it sometime down the road, keeping it ignored will save us rewriting this test.

PR #513, opened 2020-10-28
Fb sm test fix for units field

labkey-danield:
#### Rationale
Ran across an issue with the grid component when writing Sample Manager test. If the grid is long, and has been scrolled,  the previous and next buttons can be scrolled out of view. Need to scroll them into view before clicking them.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/393

#### Changes
* Added a scrollIntoView in clickPrevious and clickNext.


src/org/labkey/test/components/react/Pager.java
```diff
@@ -63,8 +63,11 @@ public int getPageSize()                // only works on GridPanel
     public Pager clickPrevious()
     {
         if (elementCache().prevButton().isPresent())
-            _pagedComponent.doAndWaitForUpdate(()->
+        {
+            getWrapper().scrollIntoView(elementCache().prevButton().get());
+            _pagedComponent.doAndWaitForUpdate(() ->
                     elementCache().prevButton().get().click());
+        }
```
labkey-tchad: We're finding the button three times here. We should avoid that.
```suggestion
        elementCache().prevButton().ifPresent(button -> {
            _pagedComponent.doAndWaitForUpdate(()-> {
                getWrapper().scrollIntoView(button);
                button.click();
            });
        });
```

src/org/labkey/test/components/react/Pager.java
```diff
@@ -83,8 +86,11 @@ public boolean isPreviousEnabled()
     public Pager clickNext()
     {
         if (elementCache().prevButton().isPresent())
-            _pagedComponent.doAndWaitForUpdate(()->
+        {
+            getWrapper().scrollIntoView(elementCache().nextButton().get());
+            _pagedComponent.doAndWaitForUpdate(() ->
                     elementCache().nextButton().get().click());
+        }
```
labkey-tchad: Find button once.
```suggestion
        elementCache().nextButton().ifPresent(button -> {
            _pagedComponent.doAndWaitForUpdate(()-> {
                getWrapper().scrollIntoView(button);
                button.click();
            });
        });
```

labkey-tchad (approved):
Looks ok.
The suggestion here is for a pre-existing problem since you're modifying this file already. Perhaps apply similar changes to `isNext/PreviousEnabled`.

src/org/labkey/test/components/react/Pager.java
labkey-danield: I was unaware of the ifPresent option.
I have made the update, but it does touch on an issue we discussed in our coding standards meeting. If the previous, or next button, is not there this code will do nothing. Does that run counter to what we had talked about?

src/org/labkey/test/components/react/Pager.java
labkey-danield: Implemented the change.

labkey-chrisj (approved):
This seems sensible.  Thanks!

src/org/labkey/test/params/experiment/MetricUnit.java
```diff
@@ -0,0 +1,7 @@
+package org.labkey.test.params.experiment;
+
+public interface MetricUnit<T extends MetricUnit<T>>
```
labkey-tchad: This sort of self-referential type parameter is just a hack we do to allow chaining in some pages and components. It shouldn't be necessary here.
```suggestion
public interface MetricUnit
```

src/org/labkey/test/params/experiment/SampleTypeDefinition.java
```diff
@@ -30,6 +30,9 @@
     // Indicates which parent aliases reference 'exp.dataInputs' instead of 'exp.materialInputs'
     private Set<String> _dataParentAliases = new HashSet<>();
 
+    private MetricUnit _inventoryMetricUnit;
+    private String _labelColor;
```
labkey-tchad, line 35: Maybe add a note indicating that these are used only by Freezer Manager and Sample Manager

src/org/labkey/test/params/experiment/SampleTypeDefinition.java
```diff
@@ -68,6 +71,28 @@ public SampleTypeDefinition setDescription(String description)
...
+    {
+        return _labelColor;
+    }
+
+    protected SampleTypeDefinition setLabelColor(String color)
+    {
+        _labelColor = color;
+        return this;
+    }
+
```
labkey-tchad, line 96: To make these work with API helpers, add `metricUnit` and `labelColor` to the map produced by `getOptions`:
```
if (getInventoryMetricUnit() != null)
{
    options.put("metricUnit", getInventoryMetricUnit());
}
if (getLabelColor() != null)
{
    options.put("labelColor", getLabelColor());
}
```

labkey-tchad (approved):
Suggestion to make the new functionality work over API

src/org/labkey/test/params/experiment/SampleTypeDefinition.java
labkey-danield: Updated. I limited the comment to the SampleManager module because that is the only module that takes advantage of them.

src/org/labkey/test/params/experiment/SampleTypeDefinition.java
labkey-danield: Updated.

src/org/labkey/test/params/experiment/MetricUnit.java
labkey-danield: Removed. I was trying to work around an issue and forgot I had added this.

PR #515, opened 2020-10-29
Issue 40442: LinkedSchema should throw handled exception

labkey-kevink:
#### Rationale
Slightly better handling of a bad LinkedSchema configuration.  Instead of showing an error page, throw QueryParseException which is more gracefully handled in the schema browser and query-executeQuery.view action.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1659

#### Changes
- update expected error

### DIFF
--- a/src/org/labkey/test/params/list/IntListDefinition.java
+++ b/src/org/labkey/test/params/list/IntListDefinition.java
@@ -1,10 +1,32 @@
 package org.labkey.test.params.list;
 
+import org.jetbrains.annotations.NotNull;
+import org.labkey.test.params.FieldDefinition;
+
+import java.util.List;
+
 public class IntListDefinition extends ListDefinition
 {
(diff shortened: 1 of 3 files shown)


labkey-kevink:
LinkedSchemaTest passed on TC: https://teamcity.labkey.org/viewLog.html?buildId=1156210&buildTypeId=bt20

labkey-tchad (approved):
I'm happy that `IntListDefinition` worked as well as it did! I don't think I ever actually tried it. I made some tweaks to clean it up a bit.
 - Made `ListDefinition` not handle any special `IntList` behavior.
 - Made it unnecessary to declare the key name twice for an `IntList`

labkey-tchad:
Re-verified with my additional change: https://teamcity.labkey.org/buildConfiguration/bt20/1156412?buildTab=tests

labkey-kevink (commented):
Thanks, that's much nicer.

PR #517, opened 2020-10-29
attempt to wait after pasting to grid

labkey-chrisj:
#### Rationale
Some of our tests paste data into an editable grid component, and submit the contents immediately.
Sometimes, tests can fail if the submit occurs before the grid component can bind the contents to the form state.
This attempts to prevent that from occurring, by waiting briefly after pasting

#### Changes
wait, after paste into the grid, for value change/selection/more rows - or else, for 1.5 seconds


labkey-tchad (approved):
Seems like a reasonable condition

PR #521, opened 2020-11-05
20.11 fb field export import automated test

labkey-chrisj:
#### Rationale
this adds test coverage for the new field export/import functionality in the domain designer

#### Related Pull Requests
n/a

#### Changes
update domainFieldRow page class to download fields exported as json
update webDriverWrapper to cause firefox to automatically download application/json content

### DIFF
--- a/src/org/labkey/test/components/domain/DomainFormPanel.java
+++ b/src/org/labkey/test/components/domain/DomainFormPanel.java
@@ -1,16 +1,25 @@
 package org.labkey.test.components.domain;
 
 import org.apache.commons.lang3.StringUtils;
+import org.json.JSONObject;
+import org.json.simple.JSONArray;
+import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
 import org.labkey.test.BootstrapLocators;
 import org.labkey.test.Locator;
 import org.labkey.test.WebDriverWrapper;
+import org.labkey.test.components.labkey.ui.core.Alert;
 import org.labkey.test.params.FieldDefinition;
 import org.labkey.test.selenium.WebElementWrapper;
+import org.openqa.selenium.Keys;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.ui.ExpectedConditions;
 
 import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -198,6 +207,15 @@ public class DomainFormPanel extends DomainPanel<DomainFormPanel.ElementCache, D
         return this;
     }
 
+    public File clickExportFields() throws Exception
+    {
+        getWrapper().scrollIntoView(elementCache().exportFieldsButton);
+        File[] exportFiles =  getWrapper().doAndWaitForDownload(()-> {
+            elementCache().exportFieldsButton.click();
+        }, 1);
+        return exportFiles[0];
+    }
+
     public DomainFormPanel setInferFieldFile(File file)
     {
         getWrapper().setFormElement(elementCache().fileUploadInput, file);
@@ -287,6 +305,9 @@ public class DomainFormPanel extends DomainPanel<DomainFormPanel.ElementCache, D
                 }, "New field didn't appear", 10000);
             }
         };
+
+        protected WebElement exportFieldsButton = Locator.tagWithClass("div", "domain-toolbar-export-btn")
+                .findWhenNeeded(this);
 
         protected void clearFieldCache()
         {
--- a/src/org/labkey/test/tests/DomainDesignerTest.java
+++ b/src/org/labkey/test/tests/DomainDesignerTest.java
@@ -1516,7 +1516,7 @@ public class DomainDesignerTest extends BaseWebDriverTest
 
         // create a list
         FieldDefinition.LookupInfo lookupInfo = new FieldDefinition.LookupInfo(getProjectName(), "lists", listName);
-        List<PropertyDescriptor> fields = importExportFields();
+        List<PropertyDescriptor> fields = importExportTestFields();
         TestDataGenerator dgen = new TestDataGenerator(lookupInfo)
                 .withColumns(fields);
         // collect the fields from the server response
@@ -1567,7 +1567,7 @@ public class DomainDesignerTest extends BaseWebDriverTest
         String sampleType = "importFieldsTestSampleType";
 
         FieldDefinition.LookupInfo lookupInfo = new FieldDefinition.LookupInfo(getProjectName(), "exp.materials", sampleType);
-        List<PropertyDescriptor> fields = importExportFields();
+        List<PropertyDescriptor> fields = importExportTestFields();
         fields.add(new FieldDefinition("name", FieldDefinition.ColumnType.String)); // need to add a 'name' field to a sampletype
         TestDataGenerator dgen = new TestDataGenerator(lookupInfo)
                 .withColumns(fields);
@@ -1621,7 +1621,40 @@ public class DomainDesignerTest extends BaseWebDriverTest
         assertThat(exportedFields.keySet(), is(roundTrippedFields.keySet()));
     }
 
-    private List<PropertyDescriptor> importExportFields()
+    @Test
+    public void testImportListFieldsToNonList() throws Exception
+    {
+        String listName = "listForFieldImportKeyTest";
+
+        // create a list
+        FieldDefinition.LookupInfo lookupInfo = new FieldDefinition.LookupInfo(getProjectName(), "lists", listName);
+        List<PropertyDescriptor> fields = importExportTestFields();
+        TestDataGenerator dgen = new TestDataGenerator(lookupInfo)
+                .withColumns(fields);
+        // collect the fields from the server response
+        List<PropertyDescriptor> createdFields = dgen.createList(createDefaultConnection(), "Key")
+                .getDomain().getFields();
+        DomainDesignerPage domainDesignerPage = DomainDesignerPage.beginAt(this, getProjectName(), "lists", listName);
+        DomainFormPanel domainFormPanel = domainDesignerPage.fieldsPanel();
+        File exportFile = domainFormPanel.clickExportFields();
+
+        String sampleTypeName = "sampleTypeWithNonKeyKeyField";
+        CreateSampleTypePage createSampleTypePage = CreateSampleTypePage.beginAt(this, getProjectName());
+        createSampleTypePage.setName(sampleTypeName)
+                .setDescription("created by exporting fields from a list with PK 'Key' field")
+                .getFieldsPanel()
+                .setInferFieldFile(exportFile);
+        createSampleTypePage.clickSave();
+
+        domainDesignerPage = DomainDesignerPage.beginAt(this, getProjectName(), "exp.materials", sampleTypeName);
+        File sampleTypeFields = domainDesignerPage.fieldsPanel().clickExportFields();
+        ArrayListMap<String, PropertyDescriptor> sampleFields = getFieldsFromExportFile(sampleTypeFields);
+        PropertyDescriptor keyField = sampleFields.get("Key");
+        assertFalse("Expect field import to disregard PK on fields imported to SampleType",
+                (Boolean)keyField.getAllProperties().get("isPrimaryKey"));
+    }
+
+    private List<PropertyDescriptor> importExportTestFields()
     {
         ConditionalFormatFilter gtFive = new ConditionalFormatFilter("5", Filter.Operator.GT);
         FieldDefinition.RangeValidator soonValidator = new FieldDefinition.RangeValidator("soonValidator", "between now and 2 days from now", "aaaa",
--- a/src/org/labkey/test/tests/DomainDesignerTest.java
+++ b/src/org/labkey/test/tests/DomainDesignerTest.java
@@ -1,6 +1,9 @@
 package org.labkey.test.tests;
 
 import org.hamcrest.CoreMatchers;
+import org.json.simple.JSONArray;
+import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.BeforeClass;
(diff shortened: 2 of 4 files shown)


src/org/labkey/test/components/domain/DomainFormPanel.java
```diff
@@ -288,6 +306,9 @@ public void click()
             }
         };
 
+        protected WebElement exportFieldsButton = Locator.tagWithClass("div", "domain-toolbar-export-btn")
+                .findWhenNeeded(this);
```
cnathe, line 310: The add field button finder uses `.findWhenNeeded(DomainFormPanel.this)` so that it has the given panel in scope when looking for the button. Is that needed here as well for the case when there are multiple domain form panels on the page?

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -1498,6 +1507,172 @@ public void addUpdateRemoveConditionalFormat() throws Exception
...
+                            .setValidators(List.of(soonValidator))));
+        dgen.createList(createDefaultConnection(), "Key");
+        DomainDesignerPage domainDesignerPage = DomainDesignerPage.beginAt(this, getProjectName(), "lists", listName);
+        DomainFormPanel domainFormPanel = domainDesignerPage.fieldsPanel();
+
+        File exportFile = domainFormPanel.clickExportFields();
+        List<PropertyDescriptor> createdFields = getFieldsFromExportFile(exportFile);
+
+        PropertyDescriptor stringField = createdFields.stream().filter(a-> a.getName().equals("textField"))
+                .findFirst().orElse(null);
```
cnathe: This filter to findFirst works, but we should expect a specific order to the fields that are exported and read in from the JSON file. So would it make sense to get the field at a specific index (zero here) since we are using an ArrayList in the getFieldsFromExportFile? This would also then provide a check that the exported fields are in the correct/expected ordering.

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -1498,6 +1507,172 @@ public void addUpdateRemoveConditionalFormat() throws Exception
...
+        assertNotNull("expected string field was not exported", stringField);
+        assertThat("description did not export", stringField.getDescription(), is("is texty"));
+        assertThat("label did not export", stringField.getLabel(), is("text field"));
```
cnathe: If you put the initial FieldDefinitions into variables, you could use them here instead of duplicating the text strings for the label/descriptions/etc.

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -1498,6 +1507,172 @@ public void addUpdateRemoveConditionalFormat() throws Exception
...
+                .findFirst().orElse(null);
+        assertNotNull("expected attachment field was not exported", attachmentField);
+        assertTrue(attachmentField.getHidden());
+
+        for (PropertyDescriptor intendedField : dgen.getColumns())
+        {
+            PropertyDescriptor exportedField = createdFields.stream().filter(a-> a.getName().equals(intendedField.getName()))
+                    .findFirst().orElse(null);
+            assertNotNull("expected field ["+intendedField.getName()+"] was not exported", exportedField);
+            verifyExpectedFieldProperties(intendedField, exportedField);
```
cnathe: I like this "iterate through the dgen.getColumns() and compare with the exportedField" approach. I'm curious why then, the above checks exist on a per field basis? From what I can see the verifyExpectedFieldProperties covers all of the things that those checks before the for loop are doing. Am I missing something?

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -1498,6 +1507,172 @@ public void addUpdateRemoveConditionalFormat() throws Exception
...
+                                .setURL("fake/list-def.url"),
+                        new FieldDefinition("attachmentField", FieldDefinition.ColumnType.Attachment)
+                                .setHidden(true),
+                        new FieldDefinition("decimalField", FieldDefinition.ColumnType.Decimal).
+                                setScale(1234).setFormat("0.####"),
+                        new FieldDefinition("intField", FieldDefinition.ColumnType.Integer)
+                                .setMvEnabled(true)
+                                .setConditionalFormats(List.of(new ConditionalFormat(List.of(gtFive), true, true, false))),
+                        new FieldDefinition("dateTimeField", FieldDefinition.ColumnType.DateAndTime)
+                                .setValidators(List.of(soonValidator))));
```
cnathe: Looks like this list of field definitions is a duplicate from the other test case. Could this be factored out to a function or variable to be reused in both places?

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -1498,6 +1507,172 @@ public void addUpdateRemoveConditionalFormat() throws Exception
...
+                        new FieldDefinition("attachmentField", FieldDefinition.ColumnType.Attachment)
+                            .setHidden(true),
+                        new FieldDefinition("decimalField", FieldDefinition.ColumnType.Decimal).
+                            setScale(1234).setFormat("0.####"),
+                        new FieldDefinition("intField", FieldDefinition.ColumnType.Integer)
+                            .setMvEnabled(true)
+                            .setConditionalFormats(List.of(new ConditionalFormat(List.of(eqFive), true, true, false))),
+                        new FieldDefinition("dateTimeField", FieldDefinition.ColumnType.DateAndTime)
+                            .setValidators(List.of(soonValidator))));
+        dgen.createList(createDefaultConnection(), "Key");
```
cnathe: The one "extra" thing for lists is that the exported field definitions JSON has an `isPrimaryKey` property set to true for the key field. Can we be sure to check that in this list definition export case?

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -1498,6 +1507,172 @@ public void addUpdateRemoveConditionalFormat() throws Exception
...
+        sampleTypeCreatePage.clickSave();
+
+        DomainDesignerPage sampleTypeDesignerPage = DomainDesignerPage.beginAt(this, getProjectName(), "exp.materials", roundTripSampleType);
+        File roundTripFile = sampleTypeDesignerPage.fieldsPanel().clickExportFields();
+        List<PropertyDescriptor> roundTrippedFields = getFieldsFromExportFile(roundTripFile);
+
+        for (PropertyDescriptor intendedField : dgen.getColumns())
+        {
+            String fieldName = intendedField.getName();
+            if (!fieldName.equals("name"))  // don't expect name col to export from sampletype, it is special
```
cnathe: The name field for sample type is special in that we add it automatically when the sample type is first saved. The export feature only exports whatever is "in view" for the domain form panel when that button is clicked. So in the sample type case, if you click export before initially saving the sample type, you probably would not get a "name" field, but if you save and then go back into the sample type and then export, you will.

cnathe (dismissed):

src/org/labkey/test/components/domain/DomainFormPanel.java
labkey-chrisj: The addFieldButton finder is enclosed in an enclosed type so it can override click(); the DomainFormPanel.this notation allows the enclosed class member to up-reference the containing class.  The exportFieldsButton doesn't need to do that, it's a direct member of the current DomainFormPanel instance.

src/org/labkey/test/tests/DomainDesignerTest.java
labkey-chrisj: I hadn't been enforcing order because field order here is initially established by the CreateDomainCommand, I don't know that the order isn't shuffled along the line while serializing.
I have added code to ensure that field order in the export file matches field order in the UI, also that it is maintained over a round-trip

src/org/labkey/test/tests/DomainDesignerTest.java
labkey-chrisj: I've taken that code out, it is redundant.  The validate function covers all of these points for all fields now

src/org/labkey/test/tests/DomainDesignerTest.java
labkey-chrisj: I wrote those first, then added the iterative/parameterized one after and forgot to remove them

src/org/labkey/test/tests/DomainDesignerTest.java
labkey-chrisj: Yes, but List.of(...) produces an immutable list, and I need to add a 'name' field for sampleTypes, so my refactor produces a mutable list

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -1498,6 +1507,172 @@ public void addUpdateRemoveConditionalFormat() throws Exception
...
+        JSONArray  jsonArray = (JSONArray) parser.parse(new FileReader(exportFile));
+
+        List<PropertyDescriptor> exportFields = new ArrayList<>();
+        for (int i=0; i < jsonArray.size(); i++)
+        {
+            exportFields.add(new PropertyDescriptor((JSONObject) jsonArray.get(i)));
+        }
+
+        return exportFields;
+    }
```
labkey-tchad: Have this return a `Map<String, PropertyDescriptor>`. The `fields.stream().filter(a-> a.getName().equals("intField"))...` pattern is a bit cumbersome and we have the field names readily available here.
```suggestion
    private Map<String, PropertyDescriptor> getFieldsFromExportFile(File exportFile) throws Exception
    {
        JSONParser parser = new JSONParser();
        JSONArray  jsonArray = (JSONArray) parser.parse(new FileReader(exportFile));

        Map<String, PropertyDescriptor> exportFields = new HashMap<>();
        for (int i=0; i < jsonArray.size(); i++)
        {
            PropertyDescriptor pd = new PropertyDescriptor((JSONObject) jsonArray.get(i));
            exportFields.put(pd.getName(), pd);
        }

        return exportFields;
    }
```

src/org/labkey/test/tests/DomainDesignerTest.java
labkey-chrisj: thanks, I'll add checks for this.
Unfortunately, that's not a field on the PropertyDescriptor class (tho really, it should be)

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -1498,6 +1507,172 @@ public void addUpdateRemoveConditionalFormat() throws Exception
...
+        assertThat(actualField.getMeasure(), is(intendedField.getMeasure()));
+        assertThat(actualField.getFormat(), is(intendedField.getFormat()));
+        for (ConditionalFormat intendedFormat : intendedField.getConditionalFormats())
+        {
+            ConditionalFormat actualFormat = intendedField.getConditionalFormats().stream()
+                    .filter(a-> a.toJSON().equals(intendedFormat.toJSON())).findFirst().orElse(null);
+            assertNotNull("conditional formats did not export with full fidelity", actualFormat);
+        }
+        // would like to do validators, but this part of the remoteAPI is incomplete; validators are settable
+        // on FieldDefinition, but not gettable on PropertyDescriptor
```
labkey-tchad, line 1717: There's no reason we couldn't add a constructor to `FieldDefinition` that takes a `JSONObject` similar to the one in `PropertyDescriptor`.

labkey-tchad (changes requested):

src/org/labkey/test/tests/DomainDesignerTest.java
labkey-chrisj: I've updated the test to get the fields from the domain-create response, instead of the ones I sent in.
This means I get back name and key fields in sampleType and list, respectively- and I validate export/import on all of them.

src/org/labkey/test/tests/DomainDesignerTest.java
labkey-chrisj: I want this, but am worried about feature-creep in the context of this story.  Perhaps in another story?

src/org/labkey/test/tests/DomainDesignerTest.java
labkey-tchad: That seems dangerous. If our API helper isn't setting some property correctly, it wouldn't be in the create response and we wouldn't end up validating it.

src/org/labkey/test/tests/DomainDesignerTest.java
labkey-chrisj: OK, I'll do this.
Unfortunately, this breaks my previous approach of ensuring that field ordering survives export/round-trip unless I use a map type that ensures key ordering

src/org/labkey/test/tests/DomainDesignerTest.java
labkey-chrisj: Good call.  I've factored them into a list and am using them to validate

src/org/labkey/test/tests/DomainDesignerTest.java
labkey-chrisj: OK, I'll also validate here that everything we send in comes back as well

labkey-chrisj: Doing so shows that the 'name' field isn't on the sampleType when it is created.
It's also not shown on the UI, not exported, cannot be imported, cannot be added via the UI... but the remoteAPI fails if no 'name' field is supplied when creating the sampleType
This has all kinds of code smells and I want this to be a bug, but something tells me it's by design

src/org/labkey/test/components/domain/DomainFormPanel.java
cnathe: Got it. Thanks for explaining.

src/org/labkey/test/tests/DomainDesignerTest.java
```diff
@@ -1498,6 +1509,196 @@ public void addUpdateRemoveConditionalFormat() throws Exception
...
+        String roundTripList = "roundTripListNameCreatedFromJsonFields";
+        EditListDefinitionPage listDefPage = _listHelper.beginCreateList(getProjectName(), roundTripList);
+        listDefPage.getFieldsPanel()
+                .setInferFieldFile(exportFile);
+        listDefPage.selectKeyField("Key");
+        listDefPage.clickSave();
+
+        domainDesignerPage = DomainDesignerPage.beginAt(this, getProjectName(), "lists", roundTripList);
+        assertThat("expect list created from export to have the same fields",
+                domainDesignerPage.fieldsPanel().fieldNames(), is(uiFields));
```
cnathe, line 1561: any reason to also check the field properties here? or do we think that the field names check is sufficient?

cnathe: oh nevermind, that is already being done in testFileImportFieldsToNewSampleType

cnathe (commented):
Test cases look good to me. I'll defer to Trey for final review/approval.

labkey-tchad (approved):
Good for now. Need a follow-on story to verify properties that aren't understood by `PropertyDescriptor`, other domain types (e.g. study datasets, assay batch/run/data, VarList), and column types (e.g. Lookup, File, Sample). We should also test some unusual/unsupported scenarios such as importing column lists with incompatible columns for a particular domain (different domains allow File, Attachment, or neither).

PR #528, opened 2020-11-17
new shared heatmap test component wrapper

labkey-chrisj:
#### Rationale
This change creates a shared component wrapper for the HeatMap component

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/744
https://github.com/LabKey/sampleManagement/pull/406

### DIFF
--- a/src/org/labkey/test/components/glassLibrary/grids/HeatMap.java
+++ b/src/org/labkey/test/components/glassLibrary/grids/HeatMap.java
@@ -18,6 +18,9 @@ import java.util.Optional;
 import static org.labkey.test.WebDriverWrapper.WAIT_FOR_JAVASCRIPT;
 import static org.labkey.test.util.TestLogger.log;
 
+/**
+ * Manipulates the HeatMap component from labkey-ui-components (components/src/internal/components/heatmap/HeatMap.tsx)
+ */
 public class HeatMap extends WebDriverComponent<HeatMap.ElementCache>
 {
     private final WebElement _el;
--- a/src/org/labkey/test/components/glassLibrary/grids/HeatMap.java
+++ b/src/org/labkey/test/components/glassLibrary/grids/HeatMap.java
@@ -1,0 +1,211 @@
+package org.labkey.test.components.glassLibrary.grids;
+
+import org.labkey.test.Locator;
+import org.labkey.test.WebDriverWrapper;
+import org.labkey.test.components.Component;
+import org.labkey.test.components.WebDriverComponent;
+import org.openqa.selenium.By;
+import org.openqa.selenium.StaleElementReferenceException;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.ui.ExpectedConditions;
+import org.openqa.selenium.support.ui.WebDriverWait;
+
+import java.util.List;
+import java.util.Optional;
+
+import static org.labkey.test.WebDriverWrapper.WAIT_FOR_JAVASCRIPT;
+import static org.labkey.test.util.TestLogger.log;
+
+public class HeatMap extends WebDriverComponent<HeatMap.ElementCache>
+{
+    private final WebElement _el;
+    private final WebDriver _driver;
+    private final Locator.XPathLocator _rowsLoc = Locator.tag("tbody").childTag("tr").withoutClass("grid-empty").withoutClass("grid-loading");
+    private final Locator _loadingRowLoc = Locator.css("tbody tr.grid-loading");
+    private final Locator _spinnerLoc = Locator.css("span i.fa-spinner");
+    private final Locator _emptyRowLoc = Locator.css("tbody tr.grid-empty");
+
+    protected HeatMap(WebElement element, WebDriver driver)
+    {
+        _el = element;
+        _driver = driver;
+    }
+
+    @Override
+    public WebElement getComponentElement()
+    {
+        return _el;
+    }
+
+    @Override
+    public WebDriver getDriver()
+    {
+        return _driver;
+    }
+
+    public String currentMonth()
+    {
+        return getColumnNames().get(12);
+    }
+
+    public String getColumnByIndex(int index)
+    {
+        return getColumnNames().get(index);
+    }
+
+    public List<String> getColumnNames()
+    {
+        List<WebElement> headerCells = elementCache().headerCells();
+        return getWrapper().getTexts(headerCells);
+    }
+
+    public Optional<WebElement> getOptionalRow(String linkText)
+    {
+        List<WebElement> rows = getRows();
+        if(rows.size() == 0)  // try again; we expect there to be rows if we're here
+        {
+            new WebDriverWait(getDriver(), 10).until(ExpectedConditions.numberOfElementsToBeMoreThan(By.cssSelector("tbody tr"), 0));
+            rows=getRows();
+        }
+        for (WebElement row : rows)
+        {
+            WebElement firstCell = Locator.tag("td").findElement(row);
+            if (Locator.linkWithText(linkText).existsIn(firstCell))
+            {
+                return Optional.of(row);
+            }
+        }
+        return Optional.empty();
+    }
+
+    public WebElement getEntityLink(String linkText)
+    {
+        WebElement row = getOptionalRow(linkText).get();
+        return Locator.linkWithText(linkText).waitForElement(row, WAIT_FOR_JAVASCRIPT);
+    }
+
+    public WebElement getSummaryLink(String rowLinkText)
+    {
+        WebElement row = getOptionalRow(rowLinkText).get();
+        return Locator.tagWithAttributeContaining("span", "title", "in the last 12 months")
+                .child("a").findElement(row);
+    }
+
+    public WebElement getCell(String rowLinkText, String headerText)
+    {
+        WebElement row = getOptionalRow(rowLinkText).get();
+        return Locator.css("td[headers=\"" + headerText + "\"]").findElement(row);
+    }
+
+    public String getCellColor(String rowLinkText, String headerText)
+    {
+        WebElement cell = getCell(rowLinkText, headerText);
+        WebElement backgroundCell = Locator.tag("div").findElement(cell);
+        return backgroundCell.getCssValue("background-color");
+    }
+
+    public String getToolTipTextForCell(String rowLinkText, String headerText)
+    {
+        WebElement cell = getCell(rowLinkText, headerText);
+        getWrapper().mouseOver(cell);
+        WebDriverWrapper.waitFor(()-> null != cell.getAttribute("aria-describedby"),
+                "MouseOver did not cause cell to show tooltip", 2000);
+        String toolTipId = cell.getAttribute("aria-describedby");
+        WebElement toolTip = Locator.id(toolTipId).findElement(getDriver());
+        return toolTip.getText();
+    }
+
+    public Boolean isLoaded()
+    {
+        return !_loadingRowLoc.existsIn(this) &&
+                !_spinnerLoc.existsIn(this) &&
+                Locator.tag("td").existsIn(this);
+    }
+
+    public Boolean hasData()
+    {
+        if (!isLoaded())
+            return false;
+        else
+        {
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/components/glassLibrary/grids/HeatMap.java
```diff
@@ -0,0 +1,211 @@
+package org.labkey.test.components.glassLibrary.grids;
```
labkey-tchad: This package should match the location of the product component in `labkey-ui-components`
```suggestion
package org.labkey.test.components.glassLibrary.heatmap;
```

src/org/labkey/test/components/glassLibrary/grids/HeatMap.java
```diff
@@ -0,0 +1,211 @@
...
+import org.openqa.selenium.support.ui.ExpectedConditions;
+import org.openqa.selenium.support.ui.WebDriverWait;
+
+import java.util.List;
+import java.util.Optional;
+
+import static org.labkey.test.WebDriverWrapper.WAIT_FOR_JAVASCRIPT;
+import static org.labkey.test.util.TestLogger.log;
+
+public class HeatMap extends WebDriverComponent<HeatMap.ElementCache>
```
labkey-tchad: Document the location of the wrapped product component
```suggestion

/**
 * Manipulates the HeatMap component from labkey-ui-components (components/src/internal/components/heatmap/HeatMap.tsx)
 */
public class HeatMap extends WebDriverComponent<HeatMap.ElementCache>
```

src/org/labkey/test/components/glassLibrary/grids/HeatMap.java
```diff
@@ -0,0 +1,211 @@
...
+    @Override
+    public WebDriver getDriver()
+    {
+        return _driver;
+    }
+
+    public String currentMonth()
+    {
+        return getColumnNames().get(12);
+    }
```
labkey-tchad: We shouldn't assume that a HeatMap has 12 months, or that all HeatMaps always have months on the X-axis, really.

labkey-tchad (changes requested):

src/org/labkey/test/components/glassLibrary/grids/HeatMap.java
labkey-chrisj: OK, I'll take this out, leave the` getColumnByIndex()` method there, and refactor usages of `currentMonth()` to use it instead

src/org/labkey/test/components/glassLibrary/grids/HeatMap.java
labkey-chrisj: done

src/org/labkey/test/components/glassLibrary/heatmap/HeatMap.java
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    public Optional<WebElement> getOptionalRow(String linkText)
+    {
+        List<WebElement> rows = getRows();
+        if(rows.size() == 0)  // try again; we expect there to be rows if we're here
+        {
+            new WebDriverWait(getDriver(), 10).until(ExpectedConditions.numberOfElementsToBeMoreThan(By.cssSelector("tbody tr"), 0));
+            rows=getRows();
+        }
```
labkey-tchad: Shouldn't `isLoaded` cover this?
This just adds a 10 second delay if there aren't any rows, which seems like a perfectly valid situation.

src/org/labkey/test/components/glassLibrary/heatmap/HeatMap.java
```diff
@@ -0,0 +1,221 @@
...
+        for (WebElement row : rows)
+        {
+            WebElement firstCell = Locator.tag("td").findElement(row);
+            if (Locator.linkWithText(linkText).existsIn(firstCell))
+            {
+                return Optional.of(row);
+            }
```
labkey-tchad: Squash this into one `Locator`.
```suggestion
            if (Locator.tag("td").append(Locator.linkWithText(linkText)).existsIn(row))
            {
                return Optional.of(row);
            }
```

src/org/labkey/test/components/glassLibrary/heatmap/HeatMap.java
```diff
@@ -0,0 +1,221 @@
...
+        }
+        return Optional.empty();
```
labkey-tchad: I'm not sure what making this `Optional` buys us. The main callers of this method just call `get` on it which will throw an empty `NoSuchElementException` if it isn't actually there.

src/org/labkey/test/components/glassLibrary/heatmap/HeatMap.java
```diff
@@ -0,0 +1,221 @@
...
+        return getColumnNames().get(index);
+    }
+
+    public List<String> getColumnNames()
+    {
+        List<WebElement> headerCells = elementCache().headerCells();
+        return getWrapper().getTexts(headerCells);
+    }
+
+    public List<String> getEntityNames()
```
labkey-tchad: We should keep this generic. HeatMaps aren't necessarily entity-related
```suggestion
    public List<String> getRowNames()
```

src/org/labkey/test/tests/component/HeatMapTest.java
```diff
@@ -0,0 +1,157 @@
...
+        HeatMap heatMap = testPage.getHeatMap();
+
+        assertTrue("Heatmap should be loaded and have data", heatMap.hasData());
+        for(String sampleTypeName : TEST_SAMPLE_TYPES.keySet())
+        {
+            String expectedSampleCount = Integer.toString(TEST_SAMPLE_TYPES.get(sampleTypeName));
+            assertThat("expect ["+expectedSampleCount+"] samples",
+                    heatMap.getSummaryLink(sampleTypeName).getText(), is(expectedSampleCount));
+        }
+        assertThat("", heatMap.getEntityNames(), hasItems(TEST_SAMPLE_TYPES.keySet().toArray(new String[0])));
```
labkey-tchad: Add non-blank message.
```suggestion
        assertThat("something", heatMap.getEntityNames(), hasItems(TEST_SAMPLE_TYPES.keySet().toArray(new String[0])));
```

labkey-tchad (approved):
Some more nits

src/org/labkey/test/components/glassLibrary/heatmap/HeatMap.java
labkey-chrisj: Nice catch. 
2 things about that; 

1. this wait-and-retry logic is from a time before we called waitForLoaded() in elementCache() so we probably don't need it any more,
2. If there aren't rows we should expect to see an alert saying there isn't data to show (so we probably can't get into that situation in the first place)

I'll pull the check-and-retry out for now, if it starts failing I will be surprised.

src/org/labkey/test/components/glassLibrary/heatmap/HeatMap.java
labkey-chrisj: There are 4 use cases in` SMAssayImportTest` that use it to verify that specific named assays are in the heatmap- the optional pattern is a holdover from using the `ResponsiveGrid` to find an` Optional<GridRow>` with that link in it- but yeah, we could do that check by confirming there's a row with that name in it, now that we expose` getRowNames()`

src/org/labkey/test/components/glassLibrary/heatmap/HeatMap.java
```diff
@@ -0,0 +1,224 @@
+package org.labkey.test.components.glassLibrary.heatmap;
+
+import org.labkey.test.Locator;
```
labkey-danield, line 3: Run SonarLint on this file. There are warning for unused imports, and other issues.

src/org/labkey/test/components/glassLibrary/heatmap/HeatMap.java
```diff
@@ -0,0 +1,224 @@
...
+     * Each row has an anchor tag in its first column; usually it's a link to a sampletype or an assay.
+     * We will refer to that link as the row's name
+     * @return
+     */
+    public List<String> getRowNames()
+    {
+        List<String> rowNames = new ArrayList<>();
+        for (WebElement row : getRows())
+        {
+            WebElement entityLink = Locator.xpath("//td/div/a").findElement(row);
```
labkey-danield: Curious why this locator is not a constant like the others.
    private final Locator _entityLink = Locator.xpath("//td/div/a");

src/org/labkey/test/components/glassLibrary/heatmap/HeatMap.java
labkey-danield: I've updated the SM test to not use the getOptionalRow. The only place this is now used is in this (HeatMap) class.

src/org/labkey/test/components/glassLibrary/heatmap/HeatMap.java
```diff
@@ -0,0 +1,224 @@
...
+        return Optional.empty();
+    }
+
+    public WebElement getEntityLink(String linkText)
+    {
+        WebElement row = getOptionalRow(linkText).get();
+        return Locator.linkWithText(linkText).waitForElement(row, WAIT_FOR_JAVASCRIPT);
+    }
+
+    public WebElement getSummaryLink(String rowLinkText)
```
labkey-danield: The parameter names should be more consistent. Here it is 'rowLinkText', for getOptionalRow the same value is called 'linkText' and in another method these values are returned as rowNames (getRowNames).

Since there is the function getRowNames I would go with rowNames for consistency.

src/org/labkey/test/components/glassLibrary/heatmap/HeatMap.java
```diff
@@ -0,0 +1,224 @@
...
+    {
+        WebElement row = getOptionalRow(rowLinkText).get();
+        return Locator.tagWithAttributeContaining("span", "title", " last 12 months")
+                .child("a").findElement(row);
```
labkey-danield: Alternatively couldn't you return just this:
`return Locator.tagWithAttributeContaining("a", "href", rowLinkText + "/runs").findElement(this);`

src/org/labkey/test/components/glassLibrary/heatmap/HeatMap.java
```diff
@@ -0,0 +1,224 @@
...
+    {
+        if (!isLoaded())
+            return false;
+        else
+        {
+            try {
+                List<WebElement> elements = _emptyRowLoc.findElements(getComponentElement());
+                return elements.size() == 0;
+            }
+            catch (StaleElementReferenceException e)
```
labkey-danield: Do you need to catch NoSuchElement in addition to StaleElement?

labkey-danield (changes requested):
I would ask that the parameter names be more consistent.

src/org/labkey/test/components/glassLibrary/heatmap/HeatMap.java
labkey-chrisj: the summary link doesn't always reflect runs.  It is the only link with a title attribute in the row element, however. Also, the a tags do not always have href values(these appear to be provisioned as part of app integration; they are absent in the core-components view which could be a thing to bring up about whether that's what we want there)

src/org/labkey/test/components/glassLibrary/heatmap/HeatMap.java
labkey-chrisj: It probably can't hurt

labkey-tchad: If either of those exceptions is thrown, they will be from the element returned by `getComponentElement`, not the empty row.

labkey-danield (approved):

src/org/labkey/test/components/glassLibrary/heatmap/HeatMap.java
labkey-chrisj: Yes, I think the try/catch here was vestigial from a time where these methods were handling a row element- so if we hit that exception we probably shouldn't swallow it

PR #532, opened 2020-11-21
Dismiss folder tree tooltip

labkey-tchad:
#### Rationale
Karl fixed some Ext4 tooltips. Now they're getting in the way occasionally.

#### Changes
* Move mouse after expanding folder tree to avoid activating tooltip
* Remove redundant method for expanding the folder tree


labkey-danield (approved):
Looks good to me. Now we just wait and see.

PR #534, opened 2020-11-23
FM Box Display Automation

labkey-danield:
#### Rationale
In the FreezerManager app there is a breadcrumb component that is used for navigation. It is a combination of links, texts and dropdown menus. It is possible that this control will be used outside of the Inventory module so it should go into the higher level testAutomation library.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/420

#### Changes
* Implement the FMBreadCrumbs component.

**Bread Crumb Component**
At it's most basic a bread crumb control could look like this:
image

This is just a `<ol>` element with several `<li><a>` elements.

In a more complex layout a breadcrumb can look like this:
image

This is still an `<ol>` with `<li>` elements, but now the `<li>` may contain a span (plain text) or a link. And some of the breadcrumbs will have a dropdown menu.

This test component will expose interaction with elements in the `<ol>` and give any the dropdown menu an element may have.


src/org/labkey/test/components/glassLibrary/navigation/FMBreadCrumbs.java
```diff
@@ -0,0 +1,210 @@
...
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * <p>
+ *     This is a navigational component for the breadcrumb.ts element <i>I think</i>.
+ * </p>
```
labkey-tchad: I think this is specific to Freezer Manager ([LocationBreadcrumb.tsx](https://github.com/LabKey/inventory/blob/develop/src/client/FreezerManager/pages/freezers/navigation/LocationBreadcrumb.tsx)). If so the test component should go in the inventory module with it.

src/org/labkey/test/components/glassLibrary/navigation/FMBreadCrumbs.java
```diff
@@ -0,0 +1,210 @@
...
+ */
+public class FMBreadCrumbs extends WebDriverComponent<FMBreadCrumbs.ElementCache>
+{
+    final WebElement _breadCrumbsElement;
+    final WebDriver _driver;
+
+    public FMBreadCrumbs(WebDriver driver)
+    {
+        this(Locators.component.findElement(driver), driver);
+    }
```
labkey-tchad: This sort of duplicates a globally scoped `FMBreadCrumbsFinder`. Consider removing this constructor in favor of using the Finder.

src/org/labkey/test/components/glassLibrary/navigation/FMBreadCrumbs.java
```diff
@@ -0,0 +1,210 @@
...
+     */
+    public List<String> getBreadCrumbs()
+    {
+        List<String> breadCrumbs = new ArrayList<>();
+        for(WebElement crumb : getBreadCrumbElements())
+        {
+            breadCrumbs.add(crumb.getText());
+        }
+
+        return breadCrumbs;
```
labkey-tchad: We have a helper for this sort of thing:
```suggestion
        return getWrapper().getTexts(getBreadCrumbElements());
```

src/org/labkey/test/components/glassLibrary/navigation/FMBreadCrumbs.java
```diff
@@ -0,0 +1,210 @@
...
+     */
+    public List<String> getBreadCrumbDropdownText(String breadCrumbText)
+    {
+        WebElement breadCrumb = elementCache().breadCrumb(breadCrumbText);
+        DropdownButtonGroup dropDownMenu = elementCache().dropDownMenu(breadCrumb);
+        if(!dropDownMenu.isExpanded())
+        {
+            dropDownMenu.expand();
+        }
+        return elementCache().dropDownMenu(breadCrumb).getMenuItemTexts();
```
labkey-tchad: It looks like `getMenuItemTexts` will open the menu if needed.
```suggestion
        return elementCache().dropDownMenu(breadCrumb).getMenuItemTexts();
```

labkey-tchad (commented):
I'd consider finding breadcrumb nodes by index instead of by text (or just alternative methods to do so). This would get real wonky if there are any duplicate values in the path.

src/org/labkey/test/components/glassLibrary/navigation/FMBreadCrumbs.java
labkey-danield: Moved.

src/org/labkey/test/components/glassLibrary/navigation/FMBreadCrumbs.java
labkey-danield: Changed.

src/org/labkey/test/components/glassLibrary/navigation/FMBreadCrumbs.java
labkey-danield: Removed redundant code.

src/org/labkey/test/components/glassLibrary/navigation/FMBreadCrumbs.java
labkey-danield: Yep. A copy and paste mistake.
Removed.

labkey-danield:
This PR is no longer valid. The code has been moved from TestAutomation to Inventory.

PR #536, opened 2020-11-25
Fb detail panel refactor

labkey-chrisj:
#### Rationale
This consolidates several redundant components 
Spec is [here](https://docs.google.com/document/d/18_52iMGac0xzTvhr9afFK7tOatUaGP-BEk0PhTLeRWs/edit)

#### Related Pull Requests
https://github.com/LabKey/sampleManagement/pull/413
https://github.com/LabKey/biologics/pull/747

#### Changes
mods to DetailTable, DetailTableEdit, 
new components DetailTableEdit and DetailContainer, 
delete old components DataPanel, EditingDataPanel, EntityEditFieldsPanel, DetailComponentTable
refactor usages in Biologics, SampleManager


src/org/labkey/test/components/glassLibrary/grids/DetailDataPanel.java
```diff
@@ -0,0 +1,122 @@
...
+        if (!isEditable())
+            throw new IllegalStateException("The current pane (with title ["+title+"]) is not editable");
+
+        elementCache().editButton().get().click();
+
+        return new DetailTableEdit.DetailTableEditFinder(getDriver())
+                .withTitle("Editing " + title).waitFor();
+    }
+
+    public DetailTable table()
```
labkey-danield: I believe our convention is that functions like this should be named `getTable`.

src/org/labkey/test/components/glassLibrary/grids/DetailDataPanel.java
```diff
@@ -0,0 +1,122 @@
...
+    @Override
+    protected ElementCache newElementCache()
+    {
+        return new ElementCache();
+    }
+
+    protected class ElementCache extends Component<?>.ElementCache
+    {
+        final WebElement heading = Locator.tagWithClass("div", "detail__edit--heading").findWhenNeeded(this);
+        public Optional<WebElement> editButton()
```
labkey-danield, line 81: It kind of looks like the main reason this is an Optional is to check for its existence in isEditable. You could achieve the same thing by making it a lazy element (findWhenNeeded) and calling isDisplayed which will return false if the element is not there.

src/org/labkey/test/components/glassLibrary/grids/DetailTable.java
```diff
@@ -120,12 +131,21 @@ static Locator fieldValue(String caption)
...
-            _locator= Locators.detailTable;
+            _locator= _baseLocator;
+        }
+
+        public DetailTableFinder withTitle(String title)
+        {
+            _locator = Locator.tagWithClass("div", "panel")
+                    .withChild(Locator.tagWithClass("div", "panel-heading").withText(title))
+                    .descendant(_baseLocator);
+            return this;
```
labkey-danield, line 150: Do you need the ("div", "panel") locator? Would this work?
`            _locator = Locator.tagWithClass("div", "panel-heading").withText(title)
                    .descendant(_baseLocator);
`
Just curious.

src/org/labkey/test/components/glassLibrary/grids/DetailTableEdit.java
```diff
@@ -12,15 +16,16 @@
...
 import java.util.List;
 
-public class DetailTableEdit extends WebDriverComponent
+public class DetailTableEdit extends WebDriverComponent<DetailTableEdit.ElementCache>
 {
-
-    private final WebElement _editPanel;
+    private final WebElement _formElement;
     private final WebDriver _driver;
+    private String _sourceTitle;
```
labkey-danield: This variable doesn't make sense to me. There is a setter for it:
`public DetailTableEdit setSourcePanelTitle(String sourcePanelTitle)`
The getter:`private String getSourceTitle()` uses the variable more as a flag (checks for null value) and if it is not null the getter ignores whatever value of the variable and uses the value from the panel header.
And since getter is private this a write-only variable, which is odd.

It feel likes you can get rid of this variable and change getSourceTitle to:
`    private String getSourceTitle()
    {
          return getTitle().replace("Editing ", "");
    }
`

src/org/labkey/test/components/glassLibrary/grids/DetailTableEdit.java
```diff
@@ -33,7 +38,20 @@ protected WebDriver getDriver()
     @Override
     public WebElement getComponentElement()
     {
-        return _editPanel;
+        return _formElement;
+    }
+
+    public DetailTableEdit setSourcePanelTitle(String sourcePanelTitle)
```
labkey-danield: This function can go away.

src/org/labkey/test/components/glassLibrary/grids/DetailTableEdit.java
```diff
@@ -121,6 +137,16 @@ public DetailTableEdit setTextField(String fieldCaption, String value)
         return this;
     }
 
+    public DetailTableEdit setInputByFieldName(String fieldName, String value)
```
labkey-danield, line 136: This and the setTextField function feel a little like duplicate functionality. I realize this uses the fieldName and setTextField uses the caption on the form. Maybe at some point in the future the duplication could be removed and a single method is present.

src/org/labkey/test/components/glassLibrary/grids/DetailTableEdit.java
```diff
@@ -252,24 +290,103 @@ public DetailTableEdit clearSelectValue(String fieldCaption)
         return this;
     }
 
-    protected static abstract class Locators
+    private String getSourceTitle()
```
labkey-danield, line 289: This can be reduced to just:
`return getTitle().replace("editing ", "");`

src/org/labkey/test/components/html/DetailContainer.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    static public class Locators
+    {
+        static public Locator.XPathLocator panelLoc = Locator.tagWithClass("div", "panel-default");
+        static public Locator.XPathLocator headerLoc = Locator.tagWithClass("div", "panel-heading");
+        static public Locator.XPathLocator bodyLoc = Locator.tagWithClass("div", "panel-body");
+        static public Locator panelWithTitle(String title)
+        {
+            return panelLoc.withChild(headerLoc.withText(title));
```
labkey-danield, line 60: Same question as before if you have a div.panel-heading (headerLoc), do you need the div.panel-default (panelLoc) to find it?

labkey-danield (changes requested):
You don't need the _sourceTitle variable in DetailTableEdit.

src/org/labkey/test/components/glassLibrary/grids/DetailTableEdit.java
labkey-chrisj: The intention for its use was to find the original table after finishing the edit.  but, if we can safely assume that removing the 'Editing ' prefix from the edit pane gets us the title of the origin detailTable, it's not necessary.
If the devs get creative or override that behavior and we need to fix it, I'm willing to expose a constructor that allows the finder to pass in the title to the source table, so it can be found when editing is done or cancelled.

src/org/labkey/test/components/glassLibrary/grids/DetailTableEdit.java
labkey-chrisj: There are some fields for which the text in the label or data-caption attribute just don't work; the special symbol for Extinction Coefficient seems to be a special weird thing that just isn't equal to what you put in code when you search for it in xpath
I honestly would have stuck with the one method, except that the one method just didn't work in some cases.  I would definitely like us to converge to a single implementation.  Until then, we can reference usages of `setInputbyFieldName` as likely spots for refactors in the product- some pretty obviously aren't using this component, but are instead hand-rolling its behavior

src/org/labkey/test/components/glassLibrary/grids/DetailDataPanel.java
labkey-chrisj: Really?  I thought that referencing it in order to call 'isDisplayed()' would trigger initialization... which (if it's not present) means having to handle the NoSuchElementException if its absence isn't a failure condition.
But yeah, having it be optional here is to support the case(s) where the user doesn't have edit permissions.

src/org/labkey/test/components/glassLibrary/grids/DetailDataPanel.java
labkey-chrisj: good call, refactoring

src/org/labkey/test/components/glassLibrary/grids/DetailTable.java
labkey-chrisj: I think it's best to have it, when finding the component by its titlebar. I don't think your suggestion will work- the table described by _baseLocator here does not descend from the panel-heading div, but it does from the panel.  FWIW, the panel-heading's followingSibling (div.panel-body) contains the table.  In any case, this locator finds the _baselocator-table descending from the top-level panel-default.

While there are containers out there that have the panel-body but no header, they all appear to have a containing panel-default.

src/org/labkey/test/components/glassLibrary/grids/DetailTableEdit.java
labkey-chrisj: okay

src/org/labkey/test/components/glassLibrary/grids/DetailTableEdit.java
labkey-chrisj: consider it done

src/org/labkey/test/components/html/DetailContainer.java
labkey-chrisj: I think yes.
having the panel-default be the componentElement here means the component can find the header and body elements within itself

Interestingly, I learned that the default locator for QueryGrid uses the panel-body element as its componentElement when I tried searching for one with that as the search scope (which failed, of course).  If the componentElement of the DetailContainer is the panel-default, that can be the searchScope for a responsiveGrid in it

src/org/labkey/test/components/glassLibrary/grids/DetailDataPanel.java
labkey-danield: Finding a lazy element (findWhenNeeded) handles the NoSuchElementException if you call isDisplayed, it will return false in that case.
Either way it is not a big deal, optional works just as well.

labkey-danield (approved):

PR #537, opened 2020-11-29
Remove non-existent actions from crawler exclusion list

labkey-adam:
#### Rationale
No need for crawler to exclude actions that don't exist

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1742


labkey-adam:
Probably easiest to review the two commits separately, since the first commit reordered the actions

labkey-adam:
The tool committed in the associated platform PR flagged all the remove actions as non-existent. Let me know if you think any of them needs to remain.

labkey-tchad (approved):
Thanks for the cleanup

PR #538, opened 2020-12-01
Scroll button on the grid into view.

labkey-danield:
#### Rationale
Sometimes the top of the grid will scroll under the page's header bar. Make sure a button on the top of the grid is clickable before trying to click it.

#### Related Pull Requests
* None

#### Changes
* Added ScrollIntoView call in the buttonClick method for the gridBar

### DIFF
--- a/src/org/labkey/test/components/glassLibrary/grids/GridBar.java
+++ b/src/org/labkey/test/components/glassLibrary/grids/GridBar.java
@@ -223,6 +223,7 @@ public class GridBar extends WebDriverComponent<GridBar.ElementCache>
         if(!button.findOptionalElement(this).isEmpty())
         {
             WebElement btn = button.waitForElement(this, 5_000);
+            getWrapper().scrollIntoView(btn);
             btn.click();
 
             if(doAction != null)


labkey-chrisj (approved):
good call here

labkey-tchad (commented):
You might be able to make a more general fix by making [`org.labkey.test.util.selenium.WebDriverUtils.ScrollUtil#scrollUnderFloatingHeader`](https://github.com/LabKey/testAutomation/blob/develop/src/org/labkey/test/util/selenium/WebDriverUtils.java#L41) know about app headers in addition to data region and regular labkey headers. `Locator.tag("div").withClasses("app-header-wrapper", "scrolled")` looks like it would work with Sample Manager. Not sure about FM or Biologics.

labkey-tchad:
Some extra context: We automatically attempt to scroll certain floating headers away when a click triggers a `ElementClickInterceptedException`. See `org.labkey.test.Locator#decorateWebElement` and `org.labkey.test.selenium.ReclickingWebElement#click`

labkey-danield:
I can see how using ScrollUtile would be more generic, but I think it is also something that should be applied to all of the clickable elements in the grid bar.

I've opened https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41963 and assigned it to test triage.

PR #549, opened 2020-12-07
Don't include modules that are not meant to be built in the teamCity distribution (once more)

labkey-susanh:
#### Rationale
(I merged the previous PR too soon, so I humbly request another review.)

Now that we correctly skip over building modules that are not meant to be built, we need to exclude them from the teamCity distribution as well.

#### Related Pull Requests
* https://github.com/LabKey/server/pull/18
* https://github.com/LabKey/gradlePlugin/pull/105
* #544

#### Changes
* Check for build viability before adding to distribution


labkey-danield (approved):
Third time is the charm?

PR #550, opened 2020-12-08
Don't set passwords for users

labkey-tchad:
#### Rationale
This test had been refactored to work against trial servers. This makes it work again. Trial servers don't have database authentication enabled, so we can't set user passwords there.

#### Related Pull Requests
* #546 

#### Changes
* Roll back most `DataReportsTest` changes from #525

### DIFF
--- a/src/org/labkey/test/tests/DataReportsTest.java
+++ b/src/org/labkey/test/tests/DataReportsTest.java
@@ -31,7 +31,6 @@ import org.labkey.test.WebTestHelper;
 import org.labkey.test.categories.DailyB;
 import org.labkey.test.categories.Reports;
 import org.labkey.test.components.html.BootstrapMenu;
-import org.labkey.test.util.APIUserHelper;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/DataReportsTest.java
```diff
@@ -547,8 +527,6 @@ public void testRReportShowSource()
 
         impersonateRole("Reader");
         clickFolder(getFolderName());
-
-        navigateToFolder(getProjectName(), getFolderName());
```
labkey-danield, line 551: I'm assuming this line and the goToProjectHome line below were removed because they are redundant. Is that correct?

labkey-danield (approved):

labkey-sweta (approved):

src/org/labkey/test/tests/DataReportsTest.java
labkey-tchad: Correct.

PR #557, opened 2020-12-16
Issue 39785: Don't return error page HTML from APIs that normally return JSON

labkey-jeckels:
#### Rationale
Server is changing to respond to 401, 404, and other error conditions with a more appropriate response content type

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1787

#### Changes
* Expect the default response type based on the request submitted
* Improve error messages when test fails

### DIFF
--- a/src/org/labkey/test/tests/ClientAPITest.java
+++ b/src/org/labkey/test/tests/ClientAPITest.java
@@ -28,6 +28,7 @@ import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.labkey.remoteapi.Command;
 import org.labkey.remoteapi.CommandException;
+import org.labkey.remoteapi.CommandResponse;
 import org.labkey.remoteapi.Connection;
 import org.labkey.remoteapi.PostCommand;
 import org.labkey.remoteapi.query.GetQueryDetailsCommand;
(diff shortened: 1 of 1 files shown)


labkey-nicka (approved):

src/org/labkey/test/tests/ClientAPITest.java
```diff
@@ -1448,47 +1452,61 @@ public void contentTypeResponseTest()
...
         Map<String, Object> expectedProps = new HashMap<>();
         expectedProps.put("success", false);
         expectedProps.put("exception", "No LabKey Server module registered to handle request for controller: bam");
 
-        runCommand(cn, command, "application/json", 404, expectedProps);
-        runCommand(cn, command, "text/xml", 404, expectedProps);
-        runCommand(cn, command, "text/html", 404, null);
+        runCommand(cn, command, "application/json", "text/html", 404, expectedProps);
+        runCommand(cn, command, "text/xml", "text/html", 404, expectedProps);
+        runCommand(cn, command, "text/html", "text/html", 404, null);
```
labkey-tchad, line 1463: This change seems to directly contradict the issue that it is testing for: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=30509

labkey-jeckels: Indeed. I talked with @labkey-nicka and we agree that the new behavior is more correct than the old. We were using the request content type as a proxy for what the response should look like, since we didn't have a way to pipe the actual response format in from the action into ExceptionUtil until my change in the platform PR here.

PR #558, opened 2020-12-16
sample type MV column test fixes

labkey-klum:
#### Rationale
Now that sample type MV columns are properly aliased in the virtual table we don't need to rely on the provisioned column names that will be case sensitive depending on the dialect.

If there are other places in the automation code that rely on the case sensitivity, please let me know and I'll get those fixed as well.

#### Changes
- Remove the conditional logic on the test cases for the indicator column name.


labkey-chrisj (approved):
So much cleaner

PR #562, opened 2020-12-16
Add test property to customize startup timeout

labkey-tchad:
#### Rationale
Servers sometimes take longer than a minute to startup. Particularly trial servers.

#### Changes
* Add test property: `webtest.server.startup.timeout`

### DIFF
--- a/src/org/labkey/test/TestProperties.java
+++ b/src/org/labkey/test/TestProperties.java
@@ -198,6 +198,24 @@ public abstract class TestProperties
         return "true".equals(System.getProperty("webtest.server.trial"));
     }
 
+    /**
+     * Parses system property 'webtest.server.startup.timeout' to determine maximum allowed server startup time.
+     * If property is not defined or is not an integer, it defaults to 60 seconds.
+     * @return Maximum number of seconds to wait for server startup
+     */
+    public static int getServerStartupTimeout()
+    {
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/TestProperties.java
```diff
@@ -198,6 +198,24 @@ public static boolean isTrialServer()
...
     }
 
+    /**
+     * Parses system property 'webtest.server.startup.timeout' to determine maximum allowed server startup time.
+     * If property is not defined or is not an integer, it defaults to 60 seconds.
+     * @return Maximum number of seconds to wait for server startup
+     */
+    public static int getServerStartupTimeout()
+    {
+        String property = System.getProperty("webtest.server.startup.timeout");
```
labkey-danield, line 208: Any value in doing this?
`System.getProperty("webtest.server.startup.timeout", 60)`

labkey-danield (approved):

src/org/labkey/test/TestProperties.java
labkey-tchad: `System.getProperty` only deals with Strings. We could do `System.getProperty("webtest.server.startup.timeout", "60")` but we'd still need to handle the `NumberFormatException` and use the default there as well.

PR #565, opened 2020-12-18
Add coverage for new SampleType export/import behavior

labkey-chrisj:
#### Rationale
This adds new test coverage for SampleType export and import, to provide coverage for derived and experiment run information

#### Related Pull Requests
N/A

#### Changes
adds test cases to SampleTypeFolderExportImportTest
adds functionality to SampleTypeAPIHelper
adds goToProjectFolder to LabkeySiteWrapper

### DIFF
--- a/src/org/labkey/test/util/exp/SampleTypeAPIHelper.java
+++ b/src/org/labkey/test/util/exp/SampleTypeAPIHelper.java
@@ -1,6 +1,5 @@
 package org.labkey.test.util.exp;
 
-import org.apache.commons.lang3.time.DateUtils;
 import org.junit.Assert;
 import org.labkey.remoteapi.CommandException;
 import org.labkey.remoteapi.Connection;
@@ -13,7 +12,6 @@ import org.labkey.test.params.experiment.SampleTypeDefinition;
 import org.labkey.test.util.TestDataGenerator;
 
 import java.util.Arrays;
-import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -26,20 +24,16 @@ public class SampleTypeAPIHelper
     public static final String SAMPLE_TYPE_DOMAIN_KIND = "SampleSet";
     public static final String SAMPLE_TYPE_DATA_REGION_NAME = "SampleSet";
     public static final String SAMPLE_TYPE_COLUMN_NAME = "Sample Set";
+    public static final String SAMPLE_NAME_EXPRESSION = "S-${now:date}-${dailySampleCount}";
 
     /**
      * Create a sample type in the specified container with the fields provided.
      *
      * @param containerPath Container in which to create the sample type.
-     * @param def domain properties for the new sample type.
+     * @param sampleTypeDefinition domain properties for the new sample type.
      * @return A TestDataGenerator for inserting rows into the created sample type.
      */
-    public static TestDataGenerator createEmptySampleType(String containerPath, SampleTypeDefinition def)
-    {
-        return createEmptySampleType(containerPath, def, null);
-    }
-
-    static public TestDataGenerator createEmptySampleType(String containerPath, SampleTypeDefinition sampleTypeDefinition, String dateFormatString)
+    static public TestDataGenerator createEmptySampleType(String containerPath, SampleTypeDefinition sampleTypeDefinition)
     {
         deleteDomain(new FieldDefinition.LookupInfo(containerPath, "samples", sampleTypeDefinition.getName()));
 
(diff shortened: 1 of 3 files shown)


src/org/labkey/test/util/exp/SampleTypeAPIHelper.java
```diff
@@ -33,15 +36,72 @@
      */
     public static TestDataGenerator createEmptySampleType(String containerPath, SampleTypeDefinition def)
     {
+        return createEmptySampleType(containerPath, def, null);
+    }
+
+    static public TestDataGenerator createEmptySampleType(String containerPath, SampleTypeDefinition sampleTypeDefinition, String dateFormatString)
```
labkey-danield: Should add a javadoc comment for this method that clarifies what the dateFormatString is used for.
Good choice on parameter name (sampleTypeDefinition) maybe should change the overloaded function to have same parameter name name for consistency and to help avoid confusion.

src/org/labkey/test/util/exp/SampleTypeAPIHelper.java
```diff
@@ -33,15 +36,72 @@
...
+    {
+        deleteDomain(new FieldDefinition.LookupInfo(containerPath, "samples", sampleTypeDefinition.getName()));
```
labkey-danield, line 38: Good addition.

src/org/labkey/test/util/exp/SampleTypeAPIHelper.java
```diff
@@ -33,15 +36,72 @@
...
         {
             throw new RuntimeException("Failed to create sample type.", e);
         }
 
+        if (null != dateFormatString)
+        {
+            dgen.addDataSupplier("DateColumn", () -> dgen.randomDateString(dateFormatString,
+                    DateUtils.addWeeks(new Date(), -39),
+                    new Date()));
+        }
```
labkey-danield: I'm unclear why this is part of createEmptySampleType. 
It isn't used in the creation of the sample type. It feels test specific and not something that is needed by all calls to createEmptySampleType.

src/org/labkey/test/util/exp/SampleTypeAPIHelper.java
```diff
@@ -33,15 +36,72 @@
...
+    {
+        return Arrays.asList(
+                new FieldDefinition("intColumn", FieldDefinition.ColumnType.Integer),
+                new FieldDefinition("decimalColumn", FieldDefinition.ColumnType.Decimal),
+                new FieldDefinition("stringColumn", FieldDefinition.ColumnType.String),
+                new FieldDefinition("sampleDate", FieldDefinition.ColumnType.DateAndTime),
+                new FieldDefinition("boolColumn", FieldDefinition.ColumnType.Boolean));
+    }
+
+    static public void deleteDomain(FieldDefinition.LookupInfo targetDomain)
```
labkey-danield: General lint comment: Reorder the modifiers to comply with the Java Language Specification. (public static void).

src/org/labkey/test/util/exp/SampleTypeAPIHelper.java
```diff
@@ -33,15 +36,72 @@
...
+    {
+        SampleTypeDefinition def = new SampleTypeDefinition(sampleTypeName);
+        def.setFields(fields);
+        if (null != parentAlias)
+            def.addParentAlias(parentAlias);
+        if (null != nameExpression)
+            def.setNameExpression(nameExpression);
+
+        return def;
+    }
```
labkey-danield: Why not add this as an overload to the constructor of SampleTypeDefinition?

labkey-danield (changes requested):

src/org/labkey/test/util/exp/SampleTypeAPIHelper.java
labkey-chrisj: It actually sets a data supplier on the TestDataGenerator it returns; as I think on it, the calling code could easily do that just as well.  If there was a way for TestDataGenerator to get the server dateFormatString independently it could do that by itself, but for now it will be something the test will have to do

src/org/labkey/test/util/exp/SampleTypeAPIHelper.java
labkey-chrisj: I'm eliminating the other overload, so dateFormatString isn't on this any more

src/org/labkey/test/util/exp/SampleTypeAPIHelper.java
labkey-chrisj: new SampleTypeDefinition(name).setFields(fields) is easy enough to call from test code.
Setting a parentAlias is simple enough for test code to handle

as a compromise I added a sample nameExpression to the helper class so test code won't have to be responsible for knowing what a valid name expression looks like

labkey-danield (approved):
Looks ok to me. I will leave the review of the test cases to the area owners.

PR #566, opened 2020-12-21
Fix test heap dump trigger

labkey-tchad:
#### Rationale
The heap dumping mechanism was created when our build was done in 'ant'. Gradle handles TeamCity properties differently, so '`enable.heap.dump`' was not getting passed through when running tests on TeamCity. Prepending "webtest." tells our Gradle plugin to pass the parameter through.

#### Changes
* Enable heap dumps with `webtest.enable.heap.dump`
* Add flag to prevent collection of more than one heap dump per test run

### DIFF
--- a/src/org/labkey/test/util/ArtifactCollector.java
+++ b/src/org/labkey/test/util/ArtifactCollector.java
@@ -165,7 +165,7 @@ public class ArtifactCollector
         if (_dumpedHeap || // Only one heap dump per suite (don't want to overload TeamCity)
             !isLocalServer() ||
             _test.isGuestModeTest() ||
-            isHeapDumpCollectionEnabled()
+            !isHeapDumpCollectionEnabled()
         )
         {
             return;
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/util/ArtifactCollector.java
```diff
@@ -161,26 +162,35 @@ public void dumpPageSnapshot(String fileSuffix, @Nullable String subdir, boolean
...
-        File destDir = ensureDumpDir();
         String dumpMsg = Locators.bodyPanel().childTag("div").findElement(_driver.getDriver()).getText();
-        String filename = dumpMsg.substring(dumpMsg.indexOf("HeapDump_"));
-        File heapDump = new File(TestFileUtils.getLabKeyRoot() + "/build/deploy", filename);
-        File destFile = new File(destDir, filename);
+        String filePrefix = "Heap dumped to ";
+        int prefixIndex = dumpMsg.indexOf(filePrefix);
+        if (prefixIndex < 0)
+        {
+            _test.checker().error("Unable to extract filename from page body. Update test or disable heap dump.\n" + dumpMsg);
```
labkey-danield: This failure message feels a little cryptic to me. I'm not sure I would know what to do if I saw this in a run failure. Specifically I am unclear on what or how I should update the test.

labkey-danield (approved):
Looks ok to me.

PR #567, opened 2020-12-22
Fb fm add sample to box automation

labkey-danield:
#### Rationale
Primarily had to update the EditableGrid to support FM AddSamples test. 

#### Related Pull Requests
* https://github.com/LabKey/inventory/pull/161

#### Changes
* Cleaned up some javadoc comments.
* Fixed check for read-only cells.
* Added EditableGrid.getCellValue to return the value of a cell.
* Added EditableGrid.getPickListForCell. Some cells have a pick-list this returns the displayed list.
* Clarified how the row number column is returned (the format is different in a location grid).
* Changed name of the function listOfnotPopulatedRows to listOfUnpopulatedRow in EditableGrid.
* Fixed a bug in assigning the metricUnit in SampleTypeDefinition.


src/org/labkey/test/components/glassLibrary/grids/EditableGrid.java
```diff
@@ -188,7 +203,7 @@ public int getRowCount()
         return getRows().size();
     }
 
-    public List<Integer> listOfnotPopulatedRows()
+    public List<Integer> listOfUnPopulatedRows()
```
labkey-chrisj: It may be preferable to call this getUnpopulatedRowIndices()

Also it will be helpful to document why this information is useful, and what exactly it means.  Perhaps the point to this code is to give callers an easy way to determine the index of the first empty row, in order to begin insert there?  If this is the case, we might be best-served to expose a getter for that index, and hide the internal workings of the component grid to prevent calling code from having to understand its inner workings

src/org/labkey/test/components/glassLibrary/grids/EditableGrid.java
```diff
@@ -297,20 +340,129 @@ public void setCellValue(int row, String columnName, Object value)
...
+    {
+        WebElement gridCell = getCell(row, columnName);
+        Actions builder = new Actions(getDriver());
+        builder.sendKeys(gridCell, Keys.ESCAPE).build().perform();
+
+        return this;
+    }
+
+    /**
+     * For the given row get the values displayed in the picklist for the given column.
```
labkey-chrisj: Are the pickList items the selectable options in a lookup list for the cell?  If so, it might be helpful to describe what a pickList is in those terms.

src/org/labkey/test/components/ui/EntityInsertPanel.java
```diff
@@ -82,7 +82,7 @@ public EntityInsertPanel addRecords(List<Map<String, Object>> records)
         getWrapper().setFormElement(elementCache().addRowsTxtBox, Integer.toString(records.size()));
         elementCache().addRowsButton.click();
 
-        List<Integer> rowIndices = elementCache().grid.listOfnotPopulatedRows();
+        List<Integer> rowIndices = elementCache().grid.listOfUnPopulatedRows();
 
         if(rowIndices.size() < records.size())
```
labkey-chrisj, line 87: TBH I think the grid should handle questions like this one (do I have enough open rows to accept this data?) by itself.  This is outside the scope of the current PR probably, but it occurred to me just now.
Also FWIW, the grid usually (but sometimes doesn't) adds rows automatically if the user pastes a set of data that fits within the columnSet, but exceeds the current rowCount.

labkey-chrisj (approved):

src/org/labkey/test/components/glassLibrary/grids/EditableGrid.java
labkey-danield: I've updated the function name to getEditableRowIndices, and the function below it to getReadonlyRowIndices and have added a javadoc above both.

I think that adding a getFirstEmptyRow would be useful.

The value in these functions, especially for Freezer Manager, is it returns a list of all the occupied spaces in a terminal location (readonly rows). This helps in validating that the grid reflects the same status as box or bag display.

src/org/labkey/test/components/ui/EntityInsertPanel.java
labkey-danield: That is a bit beyond the scope of this pr, this was only a function name change.

Freezer Manager definitely falls into the category of being a fixed size grid and not growing to match the pasted data. The grid represents a fixed size storage container.

src/org/labkey/test/components/glassLibrary/grids/EditableGrid.java
```diff
@@ -297,20 +340,129 @@ public void setCellValue(int row, String columnName, Object value)
...
+     * @return A reference to this EditableGrid.
+     */
+    public EditableGrid dismissPickList(int row, String columnName)
+    {
+        WebElement gridCell = getCell(row, columnName);
+        Actions builder = new Actions(getDriver());
+        builder.sendKeys(gridCell, Keys.ESCAPE).build().perform();
+
+        return this;
+    }
```
labkey-tchad: Is there a way to make this work without targeting a specific cell? Just to simplify it, since there can only be one pick-list open (I presume). Maybe something like:
```suggestion
    public EditableGrid dismissPickList()
    {
        Actions builder = new Actions(getDriver());
        builder.sendKeys(getComponentElement(), Keys.ESCAPE).build().perform();

        return this;
    }
```

labkey-tchad (approved):

src/org/labkey/test/components/glassLibrary/grids/EditableGrid.java
labkey-danield: I changed pick-list to dropdown. Wikipedia shows the two terms as being [interchangeable](https://en.wikipedia.org/wiki/Drop-down_list#:~:text=A%20drop%2Ddown%20list%20(abbreviated,it%20displays%20a%20single%20value.).

src/org/labkey/test/components/glassLibrary/grids/EditableGrid.java
labkey-danield: Changed.

PR #568, opened 2020-12-22
Fb responsive grid is loaded fix

labkey-danield:
#### Rationale
The ResponsiveGrid.isLoaded will sometimes fail with a NoSuchElement exception when it is trying to find the loadingGrid element.

#### Related Pull Requests
* None

#### Changes
* Change the isLoaded check to include getComponentElement().isDisplayed() && as the first check.


labkey-chrisj (approved):
Clever.

labkey-tchad (approved):

PR #575, opened 2021-01-01
Update tests for new Assay Type Select page

labkey-martyp:
#### Rationale
The related PRs created a new React Assay Type Select page.  This PR adds a new React test page and uses that when creating an assay.  Also updated remoteapi for uploading XAR files.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/1841
- https://github.com/LabKey/labkey-ui-components/pull/425

#### Changes
- Update ChooseAssayTypePage for manipulating this page
- Update AbstractAssayHelper.createAssayDesign to use new component page
- Update remotapi to updated apis


src/org/labkey/test/pages/ReactAssayTypeSelectPage.java
```diff
@@ -0,0 +1,86 @@
+package org.labkey.test.pages;
+
+import org.labkey.test.Locator;
+import org.labkey.test.components.Component;
+import org.labkey.test.components.WebDriverComponent;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+public class ReactAssayTypeSelectPage<EC extends ReactAssayTypeSelectPage.ElementCache> extends WebDriverComponent<EC>
```
labkey-chrisj: If this is a page, it should probably extend LabKeyPage- this has a name suggesting it's a page, but it extends WebDriverComponent, probably should call it AssayTypePicker, or something close to the component name

src/org/labkey/test/pages/ReactAssayTypeSelectPage.java
```diff
@@ -0,0 +1,86 @@
...
+    @Override
+    public WebDriver getDriver()
+    {
+        return driver;
+    }
+
+    public ReactAssayDesignerPage selectStandardAssay()
+    {
+        getWrapper().click(elementCache().stdAssayTab);
+        getWrapper().clickAndWait(elementCache().saveButton);
```
labkey-chrisj: Be cautious when using clickAndWait in react components, if this will be used in an app environment we can't always count on a page load event happening

src/org/labkey/test/pages/ReactAssayTypeSelectPage.java
```diff
@@ -0,0 +1,86 @@
...
+
+public class ReactAssayTypeSelectPage<EC extends ReactAssayTypeSelectPage.ElementCache> extends WebDriverComponent<EC>
+{
+    private final WebElement el;
+    private final WebDriver driver;
+
+    public ReactAssayTypeSelectPage(WebDriver driver)
+    {
+        this.driver = driver;
+        el = Locator.id("app").findElement(this.driver); // Full page component
```
labkey-chrisj: It's preferable to use the finder pattern in the component template- that gives calling code the option to use waitFor to synchronize until the component is present

labkey-chrisj (approved):

src/org/labkey/test/pages/ReactAssayTypeSelectPage.java
labkey-martyp: Ok converted this to a LabKeyPage and removed React from the name so it more closely matches the component

src/org/labkey/test/pages/ReactAssayTypeSelectPage.java
labkey-martyp: I added a TODO for now.  We can check this in the next story, integrating this component into Biologics

src/org/labkey/test/pages/ReactAssayTypeSelectPage.java
labkey-martyp: Updating to extend from LabKeyPage made this not necessary.  Removed

labkey-tchad (commented):
I think we already have a page for this (`ChooseAssayTypePage`).

labkey-tchad (changes requested):
See comment above. We should update the existing page instead of creating a new one.

labkey-martyp:
> I think we already have a page for this (`ChooseAssayTypePage`).

Thanks for the heads up.  I missed that this page already existed for some use cases.  I moved this functionality over to the ChooseAssayTypePage.  I deleted any functions that were not being used, instead of refactoring them.

labkey-tchad (approved):
Regarding the TODOs: The `clickAndWait`s here will almost certainly _not_ work in the Biologics app. You might want to break the methods up a bit more like they were in the original implementation (separate `selectAssayType` and `clickNext` methods) so that we can extend this class for Biologics and override as needed.
Also, does the button on this page actually say "Save"? If so, that seems a bit misleading since nothing is actually created when you select an assay type.

labkey-tchad (commented):
I just did a quick search of test code and I think `ContainerContextTest` needs to be updated. It will probably pass but only because it is looking for the wrong thing.

labkey-martyp:
It's not a save button, either Select X Assay or Import.  I updated the variable name.  Makes sense to split out the clickNext() function to be overridden.  I just added that to the TODO comment for the next story.  Without having this UI setup in Biologics I don't really know what will work for clickAndWait or what will need to be overridden.  But open to suggestions for this PR or the next story which I'm assuming @labkey-chrisj will be working on.

labkey-martyp:
> I just did a quick search of test code and I think `ContainerContextTest` needs to be updated. It will probably pass but only because it is looking for the wrong thing.

`AbstractAssayHelper.createAssayDesign()` used by `ContainerContextTest.testAssayList` has been updated in this PR.  Otherwise the behavior should be exactly the same as far as assay location.  Doesn't look like anything is actually selecting that in any tests.

labkey-tchad:
> > I just did a quick search of test code and I think `ContainerContextTest` needs to be updated. It will probably pass but only because it is looking for the wrong thing.
> 
> `AbstractAssayHelper.createAssayDesign()` used by `ContainerContextTest.testAssayList` has been updated in this PR. Otherwise the behavior should be exactly the same as far as assay location. Doesn't look like anything is actually selecting that in any tests.

That's true, but there's a bit where it [asserts that the Flow assay isn't available](https://github.com/LabKey/testAutomation/blob/develop/src/org/labkey/test/tests/ContainerContextTest.java#L459). The "`assertElementNotPresent`" will continue to pass, not because the assay isn't there but because the Locator is no longer valid.

PR #577, opened 2021-01-05
Item 8021: add configuration to build embedded distribution when property is present.

labkey-susanh:
#### Rationale
We want to be able to run tests on TeamCity using the embedded Tomcat instance, but don't always want to build the second distribution file, so we create a separate task that is configured to produce this distribution.  The output goes to a separate directory to avoid the problem of the deployDist task getting confused by the presence of more than one artifact with the same extension.

#### Related Pull Requests
* 

#### Changes
* Add embeddedDistribution task.


labkey-tchad (changes requested):
I think we can use just use the previous solution with a single distribution task. The embedded archive won't get created unless the flag is set so it won't interfere with our standard suites. I also updated the dependencies on TeamCity to grab `*-test.tar.gz` instead of `*.tar.gz`, so the embedded version won't get included even if we change things to not require that flag to build embedded distributions.

labkey-tchad (approved):

PR #579, opened 2021-01-06
Add test coverage for dataClass export/import

labkey-chrisj:
#### Rationale
This adds automated coverage for folder export and import/round-tripping of dataclasses, which will support LabBook's ability to archive referenced dataclasses.

#### Changes
New test for dataclass export/import
new API helper for dataclasses

### DIFF
--- a/src/org/labkey/test/tests/DataClassFolderExportImportTest.java
+++ b/src/org/labkey/test/tests/DataClassFolderExportImportTest.java
@@ -1,0 +1,351 @@
+package org.labkey.test.tests;
+
+import org.junit.BeforeClass;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.labkey.test.BaseWebDriverTest;
+import org.labkey.test.Locator;
+import org.labkey.test.TestFileUtils;
+import org.labkey.test.TestTimeoutException;
+import org.labkey.test.categories.DailyC;
+import org.labkey.test.components.CustomizeView;
+import org.labkey.test.params.FieldDefinition;
+import org.labkey.test.params.experiment.DataClassDefinition;
+import org.labkey.test.util.DataRegionTable;
+import org.labkey.test.util.LogMethod;
+import org.labkey.test.util.PortalHelper;
+import org.labkey.test.util.TestDataGenerator;
+import org.labkey.test.util.exp.DataClassAPIHelper;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.ui.ExpectedConditions;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.CoreMatchers.hasItems;
+import static org.junit.Assert.*;
+
+@Category({DailyC.class})
+public class DataClassFolderExportImportTest extends BaseWebDriverTest
+{
+    private static final String PROJECT_NAME = "DataClassExportFolderTest";
+    private static final String IMPORT_PROJECT_NAME = "DataClassImportFolderTest";
+
+    private final File DATAFILE_DIRECTORY = TestFileUtils.getSampleData("fileTypes");
+    private final File SAMPLE_CSV = new File(DATAFILE_DIRECTORY, "csv_sample.csv");
+    private final File SAMPLE_JPG = new File(DATAFILE_DIRECTORY, "jpg_sample.jpg");
+    private final File SAMPLE_PDF = new File(DATAFILE_DIRECTORY, "pdf_sample.pdf");
+    private final File SAMPLE_TIF = new File(DATAFILE_DIRECTORY, "tif_sample.tif");
+    private List<File> _attachments = List.of(SAMPLE_CSV, SAMPLE_JPG, SAMPLE_PDF, SAMPLE_TIF);
+
+    @Override
+    protected void doCleanup(boolean afterTest) throws TestTimeoutException
+    {
+        super.doCleanup(afterTest);
+    }
+
+    @BeforeClass
+    public static void setupProject()
+    {
+        DataClassFolderExportImportTest init = (DataClassFolderExportImportTest) getCurrentTest();
+
+        init.doSetup();
+    }
+
+    private void doSetup()
+    {
+        // Delete the import project if it exists.
+        _containerHelper.deleteProject(IMPORT_PROJECT_NAME, false);
+        // create the import project because we'll need it
+        _containerHelper.createProject(IMPORT_PROJECT_NAME);
+
+        PortalHelper portalHelper = new PortalHelper(this);
+        _containerHelper.createProject(PROJECT_NAME);
+
+        projectMenu().navigateToProject(PROJECT_NAME);
+        portalHelper.addWebPart("Data Classes");
+        portalHelper.addWebPart("Sample Types");
+        portalHelper.addWebPart("Experiment Runs");
+    }
+
+    @Before
+    public void preTest()
+    {
+        goToProjectHome();
+    }
+
+    @Test
+    public void testExportImportSimpleDataClass() throws Exception
+    {
+        String subfolder = "simpleDataClassExportFolder";
+        String subfolderPath = getProjectName() + "/" + subfolder;
+        String testDataClass = "testData?<>*/Class";    // having the dataClass with non-file-legal chars in it is intentional,
+                                                        // the import/export processes will write temp-files with names derived from
+                                                        // the dataClass name.  This ensures that the sanitized name on export can
+                                                        // be successfully matched up with the intended dataClass on import
+        String importFolder = "simpleDataClassImportFolder";
+
+        // create a subfolder in the import destination project
+        _containerHelper.createSubfolder(IMPORT_PROJECT_NAME, importFolder);
+        // create a subfolder in the project to hold types we'll export in this test method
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/util/exp/DataClassAPIHelper.java
```diff
@@ -0,0 +1,74 @@
...
+        TestDataGenerator dgen;
+        try
+        {
+            dgen = TestDataGenerator.createDomain(containerPath, dataClassDefinition);
+        }
+        catch (CommandException e)
+        {
+            throw new RuntimeException("Failed to create sample type.", e);
+        }
+        return dgen;
```
labkey-tchad: Why does this declare and return `dgen` outside of the try/catch?
```suggestion
        try
        {
            return TestDataGenerator.createDomain(containerPath, dataClassDefinition);
        }
        catch (CommandException e)
        {
            throw new RuntimeException("Failed to create sample type.", e);
        }
```

src/org/labkey/test/tests/DataClassFolderExportImportTest.java
```diff
@@ -0,0 +1,351 @@
...
+        // create the import project because we'll need it
+        _containerHelper.createProject(IMPORT_PROJECT_NAME);
+
+        PortalHelper portalHelper = new PortalHelper(this);
+        _containerHelper.createProject(PROJECT_NAME);
+
+        projectMenu().navigateToProject(PROJECT_NAME);
+        portalHelper.addWebPart("Data Classes");
+        portalHelper.addWebPart("Sample Types");
+        portalHelper.addWebPart("Experiment Runs");
```
labkey-tchad: Group multiple `addWebPart` calls to avoid entering/exiting page admin mode repeatedly.
```suggestion
        portalHelper.doInAdminMode(() -> {
            portalHelper.addWebPart("Data Classes");
            portalHelper.addWebPart("Sample Types");
            portalHelper.addWebPart("Experiment Runs");
        });
```

src/org/labkey/test/tests/DataClassFolderExportImportTest.java
```diff
@@ -0,0 +1,351 @@
...
+    {
+        DataClassFolderExportImportTest init = (DataClassFolderExportImportTest) getCurrentTest();
+
+        init.doSetup();
+    }
+
+    private void doSetup()
+    {
+        // Delete the import project if it exists.
+        _containerHelper.deleteProject(IMPORT_PROJECT_NAME, false);
```
labkey-tchad: Project deletion should happen in the `doCleanup` method.

src/org/labkey/test/tests/DataClassFolderExportImportTest.java
```diff
@@ -0,0 +1,351 @@
...
+
+        clickAndWait(Locator.linkWithText(testDataClass));
+        DataRegionTable sourceTable = DataRegionTable.DataRegion(getDriver()).withName("query").waitFor();
+        List<Map<String, String>> sourceRowData = new ArrayList<>();
+        for (int i=0; i<_attachments.size(); i++)
+        {
+            sourceTable.clickEditRow(i);
+            setFormElement(Locator.input("quf_attachmentColumn"), _attachments.get(i));
+            clickButton("Submit");
+        }
```
labkey-tchad: Does adding four attachments provide more interesting coverage that we would get from two or even just one? Seems like a bunch of unnecessary navigation.

src/org/labkey/test/tests/DataClassFolderExportImportTest.java
```diff
@@ -0,0 +1,351 @@
...
+        for (int i=0; i<expectedRecordCount; i++)
+        {
+            sourceRowData.add(sourceTable.getRowDataAsMap(i));
+        }
```
labkey-tchad: Consider adding a method to `DataRegionTable` to throw the whole thing into a `Map` instead of have tests gather it row-by-row. Something like:
```suggestion
        sourceRowData = sourceTable.getTableData();
```

src/org/labkey/test/tests/DataClassFolderExportImportTest.java
```diff
@@ -0,0 +1,351 @@
...
+            assertNotNull("expect all matching rows to come through", matchingMap);
+            assertThat("expect export and import values to be equivalent",
+                    exportedRow.get("intColumn"), equalTo(matchingMap.get("intColumn")));
+            assertThat("expect export and import values to be equivalent",
+                    exportedRow.get("stringColumn"), equalTo(matchingMap.get("stringColumn")));
+            assertThat("expect export and import values to be equivalent",
+                    exportedRow.get("decimalColumn"), equalTo(matchingMap.get("decimalColumn")));
+            assertThat("expect export and import values to be equivalent",
+                    exportedRow.get("attachmentColumn"), equalTo(matchingMap.get("attachmentColumn")));
+        }
```
labkey-tchad: I don't think we need to go through all this rigamarole to verify the data. `assertEquals` works pretty well on collections.
```suggestion
        // now ensure expected data in the sampleType made it to the destination folder
        assertEquals("Data class did not match after round tripping.", sourceRowData, destRowData);
```

src/org/labkey/test/tests/DataClassFolderExportImportTest.java
```diff
@@ -0,0 +1,351 @@
...
+            assertThat("expect export and import values to be equivalent",
+                    exportedRow.get("intColumn"), equalTo(matchingMap.get("intColumn")));
+
+            assertThat("expect required field values to be present",
+                    exportedRow.get("requiredField"), equalTo(matchingMap.get("requiredField")));
+            assertThat("expect missing value field values to be present",
+                    exportedRow.get("missingValueField"), equalTo(matchingMap.get("missingValueField")));
+            assertThat("expect missing value field values to be present",
+                    exportedRow.get("missingValueFieldMVIndicator"), equalTo(matchingMap.get("missingValueFieldMVIndicator")));
+        }
```
labkey-tchad: Use `assertEquals`
```suggestion
        assertEquals("Wrong data for data class after roundtripping." sourceRowData, destRowData);
```

labkey-tchad (dismissed):

src/org/labkey/test/tests/DataClassFolderExportImportTest.java
```diff
@@ -0,0 +1,351 @@
...
+                .getColumnDataAsText("Name");
+        for (String sourceRun : runNames)
+        {
+            assertThat("expect all runs to come through", importedRunNames, hasItems(sourceRun));
+        }
+
+        // we expect the sample types and experiment runs webparts to come across along with the folder import
+        clickAndWait(Locator.linkWithText(testDataClass));
+        DataRegionTable destSamplesTable = DataRegionTable.DataRegion(getDriver()).withName("query").waitFor();;
+
```
labkey-klum, line 147: seems fine but the webpart roundtripping wasn't specific to the data class export/import

labkey-klum (approved):
looks good from a coverage standpoint of test cases we wanted.

src/org/labkey/test/tests/DataClassFolderExportImportTest.java
```diff
@@ -0,0 +1,351 @@
...
+        final String REQUIRED_FIELD_NAME = "requiredField";
+        final String MISSING_FIELD_NAME = "missingField";
+        final String INDICATOR_FIELD_NAME = MISSING_FIELD_NAME + "MVIndicator";
+
+        String subfolder = "missingValueDataClassExportFolder";
+        String subfolderPath = getProjectName() + "/" + subfolder;
+        String testDataClass = "missing?Value*Data//Class";     // having the dataClass with non-file-legal chars in it is intentional,
+                                                                // the import/export processes will write temp-files with names derived from
+                                                                // the dataClass name.  This ensures that the sanitized name on export can
+                                                                // be successfully matched up with the intended dataClass on import
```
labkey-danield, line 186: This looks like a good thing to test for, would it make sense to put this in its own small test case/function? If the export fails then the test does not end up validating missing values (which the title says it will validate).

labkey-danield (approved):

src/org/labkey/test/tests/DataClassFolderExportImportTest.java
labkey-chrisj: I initially tried to add the files via remoteAPI/insertRows, but the RemoteAPI currently doesn't support doing that for Attachment type columns.  Here's an [issue ](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42191)to track that 
I settled on these 4 because I wanted to touch on the different way we present attachments in the grid column, but we could do 3 as easily.  (actually tho, this moves quickly enough that I wanted 4)

src/org/labkey/test/tests/DataClassFolderExportImportTest.java
labkey-chrisj: I'm not sure that it will.  I'm using non-file-char dataclass names in both scenarios to provide this coverage for dataclasses with different field types, including attachments (which I suspect are central to this scenario).

src/org/labkey/test/tests/DataClassFolderExportImportTest.java
labkey-chrisj: done

src/org/labkey/test/tests/DataClassFolderExportImportTest.java
labkey-chrisj: done

src/org/labkey/test/tests/DataClassFolderExportImportTest.java
labkey-chrisj: I like it, done

PR #581, opened 2021-01-11
Data region fixes to load Ext4 library now that it isn't always on the page

cnathe:
#### Rationale
Recent changes in platform removed the Ext4 library from automatically being loaded for every DataRegion. This means that this Ext4 usage from some other actions/buttons will need to make sure the Ext4 library is loaded lazily for this use case.

#### Previously Merged Pull Requests
* https://github.com/LabKey/platform/pull/1868
* https://github.com/LabKey/platform/pull/1874

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1875
* https://github.com/LabKey/dataintegration/pull/81
* https://github.com/LabKey/testAutomation/pull/581

#### Changes
* Update a couple of DataRegion test cases that now need to check for non-Ext4 alert messages

### DIFF
--- a/src/org/labkey/test/util/DataRegionTable.java
+++ b/src/org/labkey/test/util/DataRegionTable.java
@@ -31,6 +31,7 @@ import org.labkey.test.components.SummaryStatisticsDialog;
 import org.labkey.test.components.ext4.Window;
 import org.labkey.test.components.html.BootstrapMenu;
 import org.labkey.test.components.html.Checkbox;
+import org.labkey.test.components.labkey.LabKeyAlert;
 import org.labkey.test.components.study.DatasetFacetPanel;
 import org.labkey.test.components.study.ViewPreferencesPage;
 import org.labkey.test.pages.ImportDataPage;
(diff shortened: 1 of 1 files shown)


labkey-danield (approved):

src/org/labkey/test/util/DataRegionTable.java
```diff
@@ -997,9 +998,19 @@ public void removeColumn(String columnName, boolean errorExpected)
...
         {
-            Window removeError = new Window("Error", getDriver());
-            assertTrue(removeError.getBody().contains("You must select at least one field to display in the grid."));
-            removeError.clickButton("OK", true);
+            // If Ext4 is not on the page when Remove Column is clicked, this error will show as a bootstrap modal (i.e. LabKeyAlert)
+            LabKeyAlert alert = LabKeyAlert.getFinder(getDriver()).find();
+            if (alert != null)
+            {
+                assertTrue(alert.getText().contains("You must select at least one field to display in the grid."));
+                alert.accept();
```
labkey-tchad, line 1006: I think this should wait for a page load (assuming it behaves the same as the Ext4 dialog)
```suggestion
                doAndWaitForPageToLoad(() -> alert.accept());
```

cnathe: I don't believe the page load happens when this dialog closes for the "remove final column error case". It is just canceling the action from what I can tell in testing. 

The `true` param for the `removeError.clickButton()` Ext4 case is saying it should wait for the dialog to close (nothing to do with reload from what I can see. @labkey-tchad did you see some where in this code which looks to be checking for page load?

labkey-tchad: As you surmised, I was misinterpreting that `true` parameter. Thank you.

PR #590, opened 2021-01-15
Wait for elements to be present in EntityBulkUpdateDialog

labkey-chrisj:
#### Rationale
A recent [test failure](https://teamcity.labkey.org/viewLog.html?buildId=1227901&buildTypeId=LabKey_Trunk_Premium_ProductSuites_Biologics_BiologicsCPostgres&fromExperimentalUI=true#testNameId-795420342457657317) (a test failed because the calling code couldn't find a field-row on the EntityBulkUpdateDialog) suggests that earlier assumptions [that once the dialog has been found, it is completely rendered] were optimistic.  This adds a wait to any entry-point find methods, to give elements of the dialog to render before their absence is regarded as a failure.

#### Changes
This also takes some linting suggestions, e.g. converting instance-calls to waitFor() into static ones.


labkey-danield (approved):
This looks ok to me.

PR #591, opened 2021-01-16
Add redshift sample data with readme instructions for recreating/reloading cluster

labkey-jony:
This contains the SQL used to create the specific tables in Redshift along with the TSV files that contain the table data.

I've also added a readme file that goes over how to recreate the cluster and reload the data if the cluster ever needs to be recreated for any reason.


labkey-tchad:
Adding these files to the Redshift module's sampledata: https://github.com/LabKey/redshift/pull/13

PR #600, opened 2021-01-21
Fix RlabkeyTest

labkey-adam:
#### Rationale
`RlabkeyTest.testRlabkeyQueryApi()` started failing because of the new diImportHash column. The test creates an R report that queries list metadata, outputs specific columns, and checks expectations. A new column appearing meant we no longer met its expectations. Just need to bump the R script column indexes by one.


cnathe (approved):
Good catch!

PR #603, opened 2021-01-22
Create test pages for query actions

labkey-tchad:
#### Rationale
Redshift tests need to configure an external schema

#### Related Pull Requests
* 

#### Changes
* Create `InsertExternalSchemaPage` and update `ExternalSchemaTest` to use it


src/org/labkey/test/params/FieldDefinition.java
```diff
@@ -311,16 +311,22 @@ public String toString()
...
+        MultiLine("Multi-Line Text", "multiLine"),
         Integer("Integer", "int"),
+        LongInteger("Long Integer", "long"),
         String("Text", "string"),
         Subject("Subject/Participant", "string", "http://cpas.labkey.com/Study#ParticipantId", null),
         DateAndTime("Date Time", "dateTime"),
         Boolean("Boolean", "boolean"),
-        Double("Number (Double)", "float"),
+        Float("Float", "float"),
+        Double("Decimal (floating point)", "double"),
```
labkey-danield: It doesn't look like any of the changes in this file are needed for the other changes.
This file should be removed from this PR and put in a separate one, especially since these changes will break tests in other modules.

src/org/labkey/test/pages/query/SourceQueryPage.java
```diff
@@ -0,0 +1,95 @@
...
+    {
+        _ext4Helper.clickExt4Tab("Source");
+        return this;
+    }
+
+    public SourceQueryPage setSource(String sql)
+    {
+        viewSource();
+        setCodeEditorValue("queryText", "");
+        setCodeEditorValue("queryText", sql);
```
labkey-danield, line 44: I'm assuming that the double call here is to clear `queryText` and then set it to the value stored in `sql`. If that is not the case you may want to add a comment as to why the two calls.

src/org/labkey/test/pages/query/InsertExternalSchemaPage.java
```diff
@@ -0,0 +1,74 @@
...
+    {
+        Input userSchemaNameInput = Input.Input(Locator.name("userSchemaName"), getDriver()).findWhenNeeded();
+        //Input dataSourceCombo = Input.Input(Locator.name("dataSource"), getDriver()).findWhenNeeded(this);
+        Input sourceSchemaNameCombo = Input.Input(Locator.name("sourceSchemaName"), getDriver()).findWhenNeeded(this);
+        Input metaDataInput = Input.Input(Locator.name("metaData"), getDriver()).findWhenNeeded();
+
+        WebElement createButton = Locator.button("Create").findWhenNeeded(this);
+        WebElement cancelButton = Locator.button("Cancel").findWhenNeeded(this);
+    }
+}
```
labkey-danield, line 74: Do you need a waitForPage in any of these page objects?

labkey-danield (changes requested):

src/org/labkey/test/params/FieldDefinition.java
labkey-tchad: It didn't break anything in my experimentation but you're correct, it didn't end up being necessary for the related test.

src/org/labkey/test/pages/query/SourceQueryPage.java
labkey-tchad: Was trying to figure out a problem with detecting when the data grid would refresh (this was not the solution); I've removed the set to empty

src/org/labkey/test/pages/query/InsertExternalSchemaPage.java
labkey-tchad: None of the primary page content is dynamically loaded. The standard page load detection works for this page.

labkey-danield (approved):

PR #606, opened 2021-01-22
Fb biologics eval suite

labkey-chrisj:
#### Rationale
defines a list of test automation that can run against a remote trial/eval instance

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/773


labkey-tchad (changes requested):
No changes.

PR #609, opened 2021-01-26
Add logging and screen shots to help track down failure in ReportThumbnailTest.setThumbnailTypeToAuto.

labkey-danield:
#### Rationale
The ReportThumbnailTest.setThumbnailTypeToAuto intermittently fails on TeamCity (MSSQL/Windows runs). This looks like a product issue, the thumbnail does indeed not get set and the test is appropriately failing. This change adds extra logging and screenshot to help track the setting of the thumbnail and narrow down when it goes bad.

#### Related Pull Requests
* None.

#### Changes
* Added screen shots covering before and after setting the thumbnail in the report.
* Added some sanity checks to make sure the ui elements are set as expected.


labkey-tchad (approved):
It's possible that this is a server-side problem, we should get somebody familiar with the product side of thumbnail generation to add some debug logging.

PR #613, opened 2021-01-28
Time chart test fixes

labkey-adam:
#### Rationale
TimeChartVisitBasedTest and TimeChartAPITest started failing because I made an invalid `<specimen>` element in /data/study/study.xml in the related PR; it neglected to point to the specimens directory plus I failed to commit the corresponding specimen_settings.xml file. Instead of fixing this, though, I'm removing the reference altogether; this archive has never pointed to a specimen archive, so the tests aren't looking at specimen data and are almost certainly not checking for the "advanced" setting.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1933

#### Changes
* Remove invalid `<specimen>` element
* Remove unused study archive


labkey-danield (approved):
I'm fine with this change.

PR #614, opened 2021-01-29
Automation test for Chromatogram optimization library.

labkey-sweta:
#### Rationale
Automation test for Chromatogram optimization library.

#### Related Pull Requests
https://github.com/LabKey/targetedms/pull/325

#### Changes
build.gradle

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,8 +17,7 @@ project.configurations {
 
 project.dependencies {
 
-    implementation "org.xerial:sqlite-jdbc:3.7.2"
-    implementation("commons-dbcp:commons-dbcp:${commonsDbcpVersion}")
+    implementation "org.xerial:sqlite-jdbc:3.7.2" // declaring SQLite here to be used in targettedMS test
     api("junit:junit:${junitVersion}")
     api("org.seleniumhq.selenium:selenium-api:${seleniumVersion}")
     implementation("commons-io:commons-io:${commonsIoVersion}")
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -16,6 +16,9 @@ project.configurations {
 }
 
 project.dependencies {
+
+    implementation "org.xerial:sqlite-jdbc:3.7.2"
```
labkey-jeckels: Is there any way to declare this dependency in the TargetedMS module's test code, or does it have to be here in the central test harness? If it has to be here, please add a comment as to its use case.

build.gradle
```diff
@@ -16,6 +16,9 @@ project.configurations {
...
+    implementation("commons-dbcp:commons-dbcp:${commonsDbcpVersion}")
```
labkey-jeckels: See my comment in the other PR - I don't think this is necessary.

labkey-jeckels (changes requested):

build.gradle
labkey-sweta: It did not work with the dependency declaration mentioned in targettedMS module(since test is part of that module).
I am not aware of other way, let me know if there is another?

build.gradle
labkey-sweta: Removed.

labkey-jeckels:
@labkey-sweta Do you have unpushed changes? I'm only seeing the single commit to this branch.

labkey-jeckels (approved):

labkey-danield (approved):

PR #618, opened 2021-02-01
Fix Locators for R configuration selectors

labkey-tchad:
#### Rationale
Previous fix has the wrong locator for the pipeline and engine override selectors.

#### Related Pull Requests
* #604

#### Changes
* Update <select> locators to use 'name' instead of 'id'


labkey-danield (approved):
Sorry I didn't see this sooner.

PR #621, opened 2021-02-02
use case-insensitive text compare to await assay type button update

labkey-chrisj:
#### Rationale
A recent change to ChooseAssayPage caused it to wait for the select assay type button to contain the name of the assay before clicking it, but erroneously did the text comparison in a case-sensitive manner.  This caused multiple test failures. This change updates that behavior to disregard case, but still wait for the button's update.

### DIFF
--- a/src/org/labkey/test/pages/assay/ChooseAssayTypePage.java
+++ b/src/org/labkey/test/pages/assay/ChooseAssayTypePage.java
@@ -62,8 +62,9 @@ public class ChooseAssayTypePage extends LabKeyPage<ChooseAssayTypePage.ElementC
         waitForElementToBeVisible(elementCache().specialtySelectLocator);
         selectOptionByText(elementCache().specialtySelect, name);
 
-        waitFor(()-> elementCache().selectButton.getText().toLowerCase().contains(name.toLowerCase()),
-            "took too long for the select button text to contain the assay name", 2000);
+        waitFor(()->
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/pages/assay/ChooseAssayTypePage.java
```diff
@@ -62,8 +62,8 @@ public ReactAssayDesignerPage selectAssayType(String name)
         waitForElementToBeVisible(elementCache().specialtySelectLocator);
         selectOptionByText(elementCache().specialtySelect, name);
 
-        waitFor(()-> elementCache().selectButton.getText().toLowerCase().contains(name),
-            "took too long to update", 2000);
+        waitFor(()-> elementCache().selectButton.getText().toLowerCase().contains(name.toLowerCase()),
+            "took too long for the select button text to contain the assay name", 2000);
```
labkey-danield: Why not include the name in the error message?
```
waitFor(()-> elementCache().selectButton.getText().toLowerCase().contains(name.toLowerCase()),
            String.format("Took too long for the select button text to contain the assay name '%s'.", name), 2000);
```

labkey-danield (approved):

src/org/labkey/test/pages/assay/ChooseAssayTypePage.java
labkey-chrisj: That's a good idea, I'll do it

PR #622, opened 2021-02-02
Add case insensitive locator for attribute value.

labkey-danield:
#### Rationale
Added a slightly less strict locator that will check an attribute value but ignore the case. Sometimes this can happen when a field (column) has a display name that is a different case from the column name (Field_1 vs. field_1). 

#### Related Pull Requests
* None

#### Changes
* Added a locator that ignores the case of an attribute value.
* Used the locator in the DetailTableEdit.setBooleanField.
* Cleaned up some other code in setBooleanField.

### DIFF
--- a/src/org/labkey/test/components/glassLibrary/grids/DetailTableEdit.java
+++ b/src/org/labkey/test/components/glassLibrary/grids/DetailTableEdit.java
@@ -1,6 +1,6 @@
 package org.labkey.test.components.glassLibrary.grids;
 
-import org.eclipse.core.runtime.Assert;
+import org.junit.Assert;
 import org.labkey.test.BootstrapLocators;
 import org.labkey.test.Locator;
 import org.labkey.test.components.Component;
@@ -166,24 +166,17 @@ public class DetailTableEdit extends WebDriverComponent<DetailTableEdit.ElementC
     {
 
         WebElement fieldValueElement = elementCache().fieldValue(fieldCaption).findElement(getComponentElement());
-        Assert.isTrue(isEditableField(fieldValueElement), String.format("Field '%s' is not editable, and cannot be set.", fieldCaption));
+        Assert.assertTrue(String.format("Field '%s' is not editable and cannot be set.", fieldCaption), isEditableField(fieldValueElement));
 
         // The text used in the field caption and the value of the name attribute in the checkbox don't always have the same case.
         WebElement editableElement = fieldValueElement.findElement(Locator.tagWithAttributeIgnoreCase("input", "name", fieldCaption));
         String elementType = editableElement.getAttribute("type").toLowerCase().trim();
 
-        Assert.isTrue(elementType.equals("checkbox"), String.format("Field '%s' is not a checkbox, cannot be set to true/false.", fieldCaption));
+        Assert.assertEquals(String.format("Field '%s' is not a checkbox. Cannot be set to true/false.", fieldCaption), "checkbox", elementType);
 
         Checkbox checkbox = new Checkbox(editableElement);
 
-        if(!value)
-        {
-            checkbox.uncheck();
-        }
-        else
-        {
-            checkbox.check();
-        }
+        checkbox.set(value);
 
         return this;
     }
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/components/glassLibrary/grids/DetailTableEdit.java
```diff
@@ -164,38 +165,24 @@ public boolean getBooleanField(String fieldCaption)
...
-            }
+        if(!value)
+        {
+            checkbox.uncheck();
         }
         else
         {
-            throw new IllegalArgumentException("Field with caption '" + fieldCaption + "' is read-only. This field can not be set.");
+            checkbox.check();
         }
```
labkey-tchad: Checkbox has a `set` method:
```suggestion
        checkbox.set(value);
```

src/org/labkey/test/components/glassLibrary/grids/DetailTableEdit.java
```diff
@@ -164,38 +165,24 @@ public boolean getBooleanField(String fieldCaption)
...
+        Assert.isTrue(isEditableField(fieldValueElement), String.format("Field '%s' is not editable, and cannot be set.", fieldCaption));
 
-            WebElement fieldValueElement = elementCache().fieldValue(fieldCaption).findElement(getComponentElement());
+        // The text used in the field caption and the value of the name attribute in the checkbox don't always have the same case.
+        WebElement editableElement = fieldValueElement.findElement(Locator.tagWithAttributeIgnoreCase("input", "name", fieldCaption));
+        String elementType = editableElement.getAttribute("type").toLowerCase().trim();
 
-            WebElement editableElement = fieldValueElement.findElement(Locator.tagWithName("input", fieldCaption.toLowerCase()));
-            String elementType = editableElement.getAttribute("type").toLowerCase().trim();
+        Assert.isTrue(elementType.equals("checkbox"), String.format("Field '%s' is not a checkbox, cannot be set to true/false.", fieldCaption));
```
labkey-tchad: This is from non-standard assert library (`org.eclipse.core.runtime.Assert`)
```suggestion
        Assert.assertEquals(String.format("Field '%s' is not a checkbox, cannot be set to true/false.", fieldCaption), "checkbox", elementType);
```

src/org/labkey/test/components/glassLibrary/grids/DetailTableEdit.java
```diff
@@ -164,38 +165,24 @@ public boolean getBooleanField(String fieldCaption)
     public DetailTableEdit setBooleanField(String fieldCaption, boolean value)
     {
 
-        if(isFieldEditable(fieldCaption))
-        {
+        WebElement fieldValueElement = elementCache().fieldValue(fieldCaption).findElement(getComponentElement());
+        Assert.isTrue(isEditableField(fieldValueElement), String.format("Field '%s' is not editable, and cannot be set.", fieldCaption));
```
labkey-tchad: This is from non-standard assert library (`org.eclipse.core.runtime.Assert`)
```suggestion
        Assert.assertTrue(String.format("Field '%s' is not editable, and cannot be set.", fieldCaption), isEditableField(fieldValueElement));
```

labkey-tchad (approved):
A couple of fixes

src/org/labkey/test/components/glassLibrary/grids/DetailTableEdit.java
labkey-danield: Someday I will remember that.

src/org/labkey/test/components/glassLibrary/grids/DetailTableEdit.java
labkey-danield: D'oh! Fixed.

src/org/labkey/test/components/glassLibrary/grids/DetailTableEdit.java
labkey-danield: Fixed.

PR #623, opened 2021-02-02
Partial revert of RReportHelper timing change

labkey-tchad:
#### Rationale
Previous change was intended to fail fast if a `labkeyError` was found but it ended up causing some failures in `RLabKeyTest`

#### Related Pull Requests
* #539


labkey-tchad:
Verifying RLabKeyTest on TeamCity:
https://teamcity.labkey.org/project/LabKey_2011Release_Community_ModuleSuites_ApiSuites?branch=rHelperTiming&mode=builds

labkey-danield (approved):
It looks ok to me, but I'm sure the TeamCity runs will be a better judge if it is correct.

PR #629, opened 2021-02-04
Increase timeout for deleting project with many experiment runs

labkey-tchad:
#### Rationale
Workaround intermittent failure caused by Issue [42163: Deleting experiment properties is slow on SQL server](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42163)

#### Changes
* Increase timeout for `deleteContainer`

### DIFF
--- a/src/org/labkey/test/tests/elisa/ElisaMultiPlateAssayTest.java
+++ b/src/org/labkey/test/tests/elisa/ElisaMultiPlateAssayTest.java
@@ -13,6 +13,7 @@ import org.labkey.remoteapi.domain.Domain;
 import org.labkey.test.BaseWebDriverTest;
 import org.labkey.test.Locator;
 import org.labkey.test.TestFileUtils;
+import org.labkey.test.categories.Assays;
 import org.labkey.test.categories.DailyB;
 import org.labkey.test.pages.ReactAssayDesignerPage;
 import org.labkey.test.pages.assay.AssayDataPage;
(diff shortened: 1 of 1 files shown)


labkey-chrisj (approved):

src/org/labkey/test/tests/elisa/ElisaMultiPlateAssayTest.java
```diff
@@ -34,15 +33,26 @@
 import static org.hamcrest.CoreMatchers.hasItems;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.not;
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
```
labkey-danield, line 39: What's with all these added imports? Are they left over from some experimenting you did?

labkey-danield (approved):

src/org/labkey/test/tests/elisa/ElisaMultiPlateAssayTest.java
labkey-tchad: I have IntelliJ configured to "Optimize Imports" when I commit to get rid of unused imports. It also does other things, like, it expanded "`import static org.junit.Assert.*`" into the four methods that are actually used by that test.

PR #631, opened 2021-02-05
Make getBooleanField like the setter.

labkey-danield:
#### Rationale
Make getBooleanField like the setter. Remove the toLower call when getting the value of the name attribute.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/622 (already merged)

#### Changes
* Made the getter like the setter.
* Removed the call to lowercase the value of the name attribute.

### DIFF
--- a/src/org/labkey/test/components/glassLibrary/grids/DetailTableEdit.java
+++ b/src/org/labkey/test/components/glassLibrary/grids/DetailTableEdit.java
@@ -152,7 +152,13 @@ public class DetailTableEdit extends WebDriverComponent<DetailTableEdit.ElementC
      **/
     public boolean getBooleanField(String fieldCaption)
     {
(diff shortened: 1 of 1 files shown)


labkey-danield:
Running the Biologics, Inventory and SampleManager test suties.

labkey-tchad (approved):
`Checkbox` already has this functionality (`Checkbox.assertElementType`). However, it is currently disabled for fear of breaking tests that are using checkbox helpers for radio buttons. I think I did that with the intention of refactoring `WebDriverWrapper.setCheckbox` (which is widely misused for radio buttons) to use the `Checkbox` component. Since that never happened, I think that enabling `Checkbox.assertElementType` is **probably** safe but probably too risky to just throw in.

labkey-danield:
@labkey-tchad I agree that we should revisit the checkbox and it's assert check, and I also agree that it is a bit more than what I would like to do right now given the other code changes I am working on.
Doing an assert of the component element might be a general design approach to consider for other components as well.

PR #633, opened 2021-02-09
Add coverage for chart+assay run filters

labkey-chrisj:
#### Rationale
This adds coverage for chart/graph filters, to cover for issue 40384


src/org/labkey/test/tests/visualization/AssayRunFilterTest.java
```diff
@@ -0,0 +1,194 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+@Category({DailyC.class, Charting.class})
+public class AssayRunFilterTest extends BaseWebDriverTest
+{
+    @Override
+    protected void doCleanup(boolean afterTest) throws TestTimeoutException
```
labkey-nicka: This override does nothing and can be removed.

src/org/labkey/test/tests/visualization/AssayRunFilterTest.java
```diff
@@ -0,0 +1,194 @@
...
+import org.labkey.test.util.DataRegionTable;
+import org.labkey.test.util.PortalHelper;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+@Category({DailyC.class, Charting.class})
+public class AssayRunFilterTest extends BaseWebDriverTest
```
labkey-nicka: This test class seems really specific for testing this case. Did you consider adding this as another test case to an already existing Visualization test?

src/org/labkey/test/tests/visualization/AssayRunFilterTest.java
```diff
@@ -0,0 +1,194 @@
...
+    {
+        SaveProtocolCommand saveProtocolCommand = new SaveProtocolCommand(protocol);
+        ProtocolResponse savedProtocolResponse = saveProtocolCommand.execute(cn, getProjectName());
+        return savedProtocolResponse.getProtocol();
+    }
+
+    private AssayRunsPage goToAssayRunsPage(String assayName)
+    {
+        goToProjectHome();
+        clickAndWait(Locator.linkWithText("Assay List"));
```
labkey-nicka: Are there not utilities/components to do this navigation to an assay by name?

labkey-nicka (commented):
Thanks for adding this coverage. I have some feedback I'd like to discuss but overall it looks good.

src/org/labkey/test/tests/visualization/AssayRunFilterTest.java
labkey-chrisj: I did.  I also considered factoring the assay setup into a helper method instead of building it all up.
I can move all of this stuff into scatterplotTest and scrap this one

src/org/labkey/test/tests/visualization/AssayRunFilterTest.java
labkey-chrisj: If there are, I don't know about them.  We support this in biologics, but apparently not in LKS

src/org/labkey/test/tests/visualization/AssayRunFilterTest.java
labkey-chrisj: OK

src/org/labkey/test/tests/visualization/AssayRunFilterTest.java
labkey-nicka: I think making this a test case in the `ScatterPlotTest` is a good idea. If you think this will take more than an hour though I'd say just leave it as-is.

labkey-chrisj: It is, and I looked at tacking this scenario into scatterplottest- but it was more pain to try to add a test to than this was to do.  This runs fast and works.

src/org/labkey/test/tests/visualization/AssayRunFilterTest.java
labkey-chrisj: Not by name.  In LKS you navigate to the assay by knowing the protocol's Id.
There is an APIAssayHelper that promises to get a protocolId from the name, but we have the ID already in the test context.  I've added beginAts to the assay data and runs pages to make these clickAndWaits unnecessary

src/org/labkey/test/tests/visualization/AssayRunFilterTest.java
labkey-chrisj: OK, it's in there.
This eliminates an extra test class in that package, but it ties into a pattern I've been wanting to stamp out- specifically, the test where one method inherited via several base classes runs every test as a super-batch of aggregated tests that you can't run just one of.

labkey-nicka (approved):

src/org/labkey/test/tests/visualization/AssayRunFilterTest.java
labkey-tchad: > I've been wanting to stamp out- specifically, the test where one method inherited via several base classes runs every test as a super-batch of aggregated tests that you can't run just one of.

👍 👍 👍  Not suggesting any changes to this PR but wanted to give my hearty support for this sentiment

PR #634, opened 2021-02-09
Issue 42420: Support source query filter via ETL XML

labkey-jeckels:
#### Rationale
To make it easier to create ETLs that run against a remote server where the ETL developer may not have permission to create custom queries, it's more convenient to allow the ETL definition to add filters to the source table/query.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1979
* https://github.com/LabKey/dataintegration/pull/90

#### Changes
* Add filters to ETL XSD
* Parse and propagate them
* Apply them when fetching the source data


labkey-martyp (approved):
Looks good. @labkey-mohara here are a couple examples of this new feature

PR #636, opened 2021-02-15
Automation test for Add selection row

labkey-sweta:
#### Rationale
Automation test for bulk actions (delete and export)

#### Related Pull Requests
* 

#### Changes
src/org/labkey/test/tests/FieldEditorAddRowSelectionActionTest.java - New test added


src/org/labkey/test/components/domain/DomainFormPanel.java
```diff
@@ -207,6 +208,18 @@ public File clickExportFields() throws Exception
         return exportFiles[0];
     }
 
+    public DomainFormPanel clickDeleteFields()
+    {
+        getWrapper().scrollIntoView(elementCache().deleteFieldsButton);
+        elementCache().deleteFieldsButton.click();
```
labkey-danield, line 222: I believe that WebElement.click() automatically scrolls the element into view. The documentation for the method doesn't explicitly say, and comments on some of the boards indicate that it can vary from browser to browser. 

You shouldn't need to call getWrapper().scrollIntoView but neither does it do any real harm having it there.

src/org/labkey/test/components/domain/DomainFormPanel.java
```diff
@@ -207,6 +208,18 @@ public File clickExportFields() throws Exception
...
+
+        ModalDialog confirmDialog = new ModalDialog.ModalDialogFinder(getDriver())
+                .withTitle("Confirm Delete Selected Fields").timeout(1000).waitFor();
+        confirmDialog.dismiss("Yes, Delete Fields");
```
labkey-danield, line 226: Would there ever be a test case where you want to just return the dialog and not explicitly click "yes"?

src/org/labkey/test/tests/FieldEditorAddRowSelectionActionTest.java
```diff
@@ -0,0 +1,215 @@
...
+    @Override
+    public List<String> getAssociatedModules()
+    {
+        return Arrays.asList("study");
+    }
+
+    @Override
+    protected void doCleanup(boolean afterTest) throws TestTimeoutException
+    {
+        super.doCleanup(afterTest);
```
labkey-danield: If all you are doing in the doCleanup is calling super.doCleanup you shouldn't need to have this override.

labkey-danield (approved):
I think this looks OK, I'll let the devs comment on the test cases.

cnathe (commented):
The test cases look good to me. One very minor comment is that the test name including the word "Add" threw me off since there isn't any "adding of rows", I might suggest renaming to `FieldEditorRowSelectionActionTest`

RosalineP (approved):
Test cases also look good to me 👍

src/org/labkey/test/components/DomainDesignerPage.java
```diff
@@ -152,5 +158,8 @@ DomainFormPanel domainFormPanel(String title) // for situations with multiple do
             return new DomainFormPanel.DomainFormPanelFinder(getDriver()).withTitle(title)
                     .timeout(WAIT_FOR_JAVASCRIPT).findWhenNeeded(this);
         }
+
+        public final Checkbox selectAll = new Checkbox(Locator.tagWithAttributeContaining("input", "id", "domain-select-all-checkbox")
+                .findWhenNeeded(this));
```
labkey-tchad: Is this the correct scope? I think there could be multiple "select all" checkboxes on the page for multi-domain designers (such as assay result, run, and data domains).
I think the select all functionality belongs in `DomainFormPanel`

labkey-tchad (changes requested):

src/org/labkey/test/tests/FieldEditorAddRowSelectionActionTest.java
labkey-sweta: Removed

src/org/labkey/test/components/domain/DomainFormPanel.java
labkey-sweta: In future might be

src/org/labkey/test/components/DomainDesignerPage.java
labkey-sweta: Changed the location

labkey-tchad (approved):

PR #639, opened 2021-02-17
Add templates coverage in integration tests

labkey-chrisj:
#### Rationale
this branch updates the toggle locator in MultiMenu to be any tag type, to enable its use in Notebooks testing
If we decide we want to be able to set this locator to be more specific in its usages, cool

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/793

#### Changes
update multimenu toggle locator to be any tag type


labkey-danield (approved):
I'm not really a fan of wild cards in locators, but neither can I say this is a bad change. If this helps use this component in the Notebooks test then I'm ok with it, but we should keep an eye out for any odd failures that might be caused by it.

PR #640, opened 2021-02-17
Add Option Buttons for Derived or Pooled Sample Creation

labkey-danield:
#### Rationale
The bulk insert dialog has changed to have options when creating samples that have parents. The created samples can be derived (one sample per parent) or pooled (one sample multiple parents). These options are available to everyone but may not be turned on for everyone. Sample Manager will have them, Biologics does not have them yet.
Screen Shot 2021-02-17 at 10 01 58 AM

#### Related Pull Requests
* None at this time.

#### Changes
* Added creation type option buttons and associated methods to the EntityBulkInsertDialog.
* Replaced parameter in EntityBulkInsertDialog constructor. Previous parameter was not used.
* Added a selectAll method to the editable grid. Used in SampleManager test for this feature.

### DIFF
--- a/src/org/labkey/test/components/glassLibrary/grids/EditableGrid.java
+++ b/src/org/labkey/test/components/glassLibrary/grids/EditableGrid.java
@@ -119,6 +119,19 @@ public class EditableGrid extends WebDriverComponent<EditableGrid.ElementCache>
         return this;
     }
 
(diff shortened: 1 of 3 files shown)


labkey-danield:
I have run the Biologics test on TeamCity to make sure this change does not break existing tests.

labkey-tchad (approved):

PR #642, opened 2021-02-17
Conditionalize specimen verification in tests

labkey-adam:
#### Rationale
We want study tests to run both with and without the specimen module present. This PR conditionalizes specimen setup and verification steps based on whether the module is currently deployed.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2005

### DIFF
--- a/src/org/labkey/test/tests/StudyBaseTest.java
+++ b/src/org/labkey/test/tests/StudyBaseTest.java
@@ -104,7 +104,8 @@ public abstract class StudyBaseTest extends BaseWebDriverTest
         }
 
         _containerHelper.createSubfolder(getProjectName(), getFolderName(), "Study");
-        _containerHelper.enableModule("Specimen");
+        if (_studyHelper.isSpecimenModulePresent())
+            _containerHelper.enableModule("Specimen");
         new ApiPermissionsHelper(this).checkInheritedPermissions();
     }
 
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/util/StudyHelper.java
```diff
@@ -48,6 +49,7 @@
     public static final File SPECIMEN_ARCHIVE_B = TestFileUtils.getSampleData("study/specimens/sample_b.specimens");
 
     protected BaseWebDriverTest _test;
+    private Boolean _specimenModulePresent = null;
```
labkey-tchad: Make this static?

labkey-tchad (approved):

src/org/labkey/test/tests/StudyBaseTest.java
```diff
@@ -212,6 +214,11 @@ protected void exportStudy(boolean zipFile, boolean exportPhi, PhiSelectType exp
         waitForText("Visit Map", "Cohort Settings", "QC State Settings", "CRF Datasets", "Assay Datasets",
                 "Dataset Data", "Specimens", "Specimen Settings", "Participant Comment Settings");
```
labkey-tchad: Remove "Specimens" and "Specimen Settings" from the list of text to wait for.

labkey-tchad (changes requested):

src/org/labkey/test/util/StudyHelper.java
labkey-adam: Sure

src/org/labkey/test/tests/StudyBaseTest.java
labkey-adam: Oops

labkey-tchad (approved):

PR #644, opened 2021-02-20
Specimen test fixes

labkey-adam:
#### Rationale
Update some tests to accommodate recent specimen behavior changes

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2014

#### Changes
* AdvancedImportOptionsTest: Update archive so it enabled specimen module on import
* Fix several failing publish tests - specimen presence in publish wizard and export needs to be conditional on specimen module being active in current folder (not just present in the deployment)


src/org/labkey/test/util/StudyHelper.java
```diff
@@ -523,6 +523,15 @@ public boolean isSpecimenModulePresent()
         return _specimenModulePresent;
     }
 
+    public boolean isSpecimenModuleActive()
+    {
+        if (isSpecimenModulePresent())
```
labkey-tchad: Is this inverted?
```suggestion
        if (!isSpecimenModulePresent())
```

labkey-adam: Oops... thanks!

labkey-tchad (approved):

PR #650, opened 2021-02-27
Test fix for app NavBar menu-section-header locator

cnathe:
#### Rationale
Several SM tests are failing with the following error: `Element <span class="menu-section-header"> could not be scrolled into view`
See related changes in https://github.com/LabKey/labkey-ui-components/pull/453

#### Changes
* Update locator for ProductMenu.menuSection


Sigmonia (approved):

labkey-nicka (approved):
Thanks Cory.

PR #655, opened 2021-03-03
Automation for LabKey R&D UX - Field Editor Updates 2020, Summary View

labkey-sweta:
#### Rationale
1. Adding test coverage for toggle between Summary and detail mode
2. Links in the detail mode

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/503
* https://github.com/LabKey/labkey-ui-components/pull/464
* https://github.com/LabKey/platform/pull/2066

#### Changes

1. src/org/labkey/test/components/domain/DomainFormPanel.java
2. src/org/labkey/test/tests/FieldEditorRowSelectionActionTest.java

### DIFF
--- a/src/org/labkey/test/tests/FieldEditorRowSelectionActionTest.java
+++ b/src/org/labkey/test/tests/FieldEditorRowSelectionActionTest.java
@@ -12,7 +12,6 @@ import org.labkey.serverapi.reader.Readers;
 import org.labkey.test.BaseWebDriverTest;
 import org.labkey.test.Locator;
 import org.labkey.test.TestFileUtils;
-import org.labkey.test.TestTimeoutException;
 import org.labkey.test.categories.DailyA;
 import org.labkey.test.components.DomainDesignerPage;
 import org.labkey.test.components.domain.DomainFormPanel;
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/tests/FieldEditorRowSelectionActionTest.java
```diff
@@ -27,10 +26,14 @@
 import java.util.List;
 
 @Category(DailyA.class)
-@BaseWebDriverTest.ClassTimeout(minutes = 2)
+@BaseWebDriverTest.ClassTimeout(minutes = 3)
 public class FieldEditorRowSelectionActionTest extends BaseWebDriverTest
 {
     private final static String PROJECT_NAME = "Field Editor Row Selection Action Test";
+    private final static List<String> expectedHeaders = Arrays.asList("Name", "Range URI", "Required", "Lock Type", "Lookup Container", "Lookup Schema", "Lookup Query",
```
RosalineP: Should we include a test in a folder in which the `Ontology` module is not enabled, in order to ensure `"Source Ontology", "Concept Import Column", "Concept Label Column", "Principal Concept Code"` columns accurately do not get rendered, or would that be an overzealous test?

src/org/labkey/test/tests/FieldEditorRowSelectionActionTest.java
```diff
@@ -83,6 +88,19 @@ public void testListActions() throws Exception
 
         checker().verifyTrue("Exported Fields are not same as UI", uiFields.equals(exportedFields));
 
+        log("Verifying the toggle between summary view and detail view list domain");
```
RosalineP, line 90: The section here, and the block at line 127, and at 200, have a lot of similarities and the differences they do have look like they could be adjusted via function arguments, if they were modularized into a function `testSummaryViewActions` or something.

Would it be good style to group these three chunks into one function, or in tests is it preferable to have the code be more explicit like this?

RosalineP (commented):
Just a couple comments! Thanks for doing the tests on this 👍

Please leave merge task to me, as I will be going through the ui-components publishing process on these branches due to Product drift necessitating a Product update between feature merge and Automation Test merge.

src/org/labkey/test/tests/FieldEditorRowSelectionActionTest.java
labkey-sweta: Done.

src/org/labkey/test/tests/FieldEditorRowSelectionActionTest.java
labkey-sweta: I thought about it, it was mostly assert statements and not a lot of action having. Decided to keep them in the test.

Let me know if you feel strongly to move.

RosalineP (approved):
Looks good!

PR #656, opened 2021-03-05
Update bulk insert dialog to allow titles not specific to Samples

labkey-chrisj:
#### Rationale
To support entity bulk insert of assay run data, the EntityBulkInsertDialog can't have a hard-coded title specific to SampleTypes.  This parameterizes the title of this shared component wrapper, and calls for updating calling page classes.

#### Related Pull Requests
Biologics https://github.com/LabKey/biologics/pull/816
SampleManager https://github.com/LabKey/sampleManagement/pull/508

#### Changes
Allow the caller to specify the title of the EntityBulkInsertDialog

### DIFF
--- a/src/org/labkey/test/components/ui/entities/EntityBulkInsertDialog.java
+++ b/src/org/labkey/test/components/ui/entities/EntityBulkInsertDialog.java
@@ -20,12 +20,7 @@ public class EntityBulkInsertDialog extends ModalDialog
 
     public EntityBulkInsertDialog(WebDriver driver)
     {
-        this(new ModalDialogFinder(driver).withTitle("Bulk Creation of"));
-    }
-
-    public EntityBulkInsertDialog(WebDriver driver, String dialogTitle)
-    {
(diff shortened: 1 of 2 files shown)


labkey-tchad (changes requested):
Tests shouldn't have to know the title of the dialog, that's the whole point of having components. I would say just find a the modal dialog regardless of title or find a partial title value that works more broadly (Maybe just `"Bulk"` would work)

labkey-chrisj:
What I've done requires the page to know the title, which I figured would be acceptable.
The original state here was hard-coded to only be useful for bulk uploads of SampleType, which also kind of defeats the point of having a shared component able to handle cases that aren't the one it's hard-coded for.

Okay, I'll go back to hard-coding something else that will hopefully work for both known cases, right up until the point where we want to use this thing for anything else

labkey-chrisj:
Without this change, EntityBulkInsertDialog cannot be used for assay data in Biologics and should perhaps be called SampleTypeBulkInsertDialog

labkey-tchad (approved):

PR #659, opened 2021-03-10
Issue 42386: Don't show reserved fields as inferred fields

labkey-susanh:
#### Rationale
When inferring fields from an uploaded file, the file may reasonably contain some of the fields that are reserved and will already be created for a particular domain.  Currently, the user has to remove these fields manually after the server infers the full set of fields.  We change this so we will display a message indicating that some fields in the file were recognized as reserved and thus not included in the field listing.  Thus a user has a better chance of creating (or updating) a domain successfully the first time through.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2064
* https://github.com/LabKey/labkey-ui-components/pull/463
* https://github.com/LabKey/sampleManagement/pull/506

#### Changes
* Add test for list, sample type, and data class that assure we are ignoring reserved fields when inferring from a file for each of these domains.  
* Assay result domain is covered by a test in Sample Manager.  See the related PR.

### DIFF
--- a/src/org/labkey/test/components/domain/DomainFormPanel.java
+++ b/src/org/labkey/test/components/domain/DomainFormPanel.java
@@ -317,6 +317,11 @@ public class DomainFormPanel extends DomainPanel<DomainFormPanel.ElementCache, D
         return getPanelAlertWebElement().getText();
     }
 
+    public String getPanelAlertText(int index)
+    {
+        return getPanelAlertWebElement(index).getText();
+    }
+
     /**
      * There may be an element in the alert that a test will need to interact with so return the alert element and let
      * the test find the control it needs.
@@ -324,12 +329,19 @@ public class DomainFormPanel extends DomainPanel<DomainFormPanel.ElementCache, D
      */
     public WebElement getPanelAlertWebElement()
     {
+        return getPanelAlertWebElement(0);
+    }
+
+    public WebElement getPanelAlertWebElement(int index)
+    {
+
         getWrapper().waitFor(()-> BootstrapLocators.infoBanner.existsIn(getDriver()),
                 "the info alert did not appear as expected", 1000);
 
         // It would be better to not return a raw WebElement but who knows what the future holds, different alerts
         // may show different controls.
-        return BootstrapLocators.infoBanner.existsIn(getDriver()) ? BootstrapLocators.infoBanner.findElement(getDriver()) : null;
+        return BootstrapLocators.infoBanner.existsIn(getDriver()) ?
+                BootstrapLocators.infoBanner.findElements(getDriver()).get(index) : null;
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


src/org/labkey/test/components/domain/DomainFormPanel.java
```diff
@@ -317,19 +317,31 @@ public String getPanelAlertText()
...
+    {
+
         getWrapper().waitFor(()-> BootstrapLocators.infoBanner.existsIn(getDriver()),
                 "the info alert did not appear as expected", 1000);
 
         // It would be better to not return a raw WebElement but who knows what the future holds, different alerts
         // may show different controls.
-        return BootstrapLocators.infoBanner.existsIn(getDriver()) ? BootstrapLocators.infoBanner.findElement(getDriver()) : null;
+        return BootstrapLocators.infoBanner.existsIn(getDriver()) ?
+                BootstrapLocators.infoBanner.findElements(getDriver()).get(index) : null;
```
labkey-danield: This works and is fine as is. The findElements method will return an empty list if there are no elements so the conditional check could be:
```
 return !BootstrapLocators.infoBanner.findElements(getDriver()).isEmpty() ?
                BootstrapLocators.infoBanner.findElements(getDriver()).get(index) : null;
```

src/org/labkey/test/tests/DataClassTest.java
```diff
@@ -128,6 +131,38 @@ public void testReservedFieldNames()
...
+
+        log("Infer fields from a file that contains some reserved fields.");
+        DomainFormPanel domainForm = createPage
+                .getDomainEditor()
+                .setInferFieldFile(inferenceFile);
+        checker().verifyEquals("Reserved field warning not as expected",  expectedInfoMsg, domainForm.getPanelAlertText());
+        createPage.clickSave();
+        DataRegionTable drt = DataRegion(getDriver()).find();
+        checker().verifyTrue("Data class not found in list of data classes", drt.getColumnDataAsText("Name").contains(name));
+        log("End of test.");
```
labkey-danield: I get dinged in my PR feedback when I add this. :)

labkey-danield (approved):

src/org/labkey/test/components/domain/DomainFormPanel.java
labkey-tchad: Should this not be scoped to the panel? Also, you can use `findOptionalElement` to avoid finding the banner twice.
```suggestion
        return BootstrapLocators.infoBanner.index(index).findOptionalElement(this).orElse(null);
```

labkey-susanh: Thanks.  Also, is it preferable to get an index out of bounds error if there are not as many panels as expected or a potential null pointer exception from the caller of this?  Or should this simply fail with an assertion error if there are fewer than `index` elements?

src/org/labkey/test/tests/DataClassTest.java
labkey-susanh: Just following examples from SM :).

src/org/labkey/test/components/domain/DomainFormPanel.java
labkey-susanh: Ah, that's nice.  I'll do that.

PR #661, opened 2021-03-11
Improve timing when interacting with edit-in-place inputs

labkey-tchad:
#### Rationale
The `WorkbookTest` fails occasionally when interacting with 'edit-in-place' elements. Trying some tweaks to the test interaction.

#### Changes
* Tweak interaction with edit-in-place inputs for workbook name and description

### DIFF
--- a/src/org/labkey/test/tests/WorkbookTest.java
+++ b/src/org/labkey/test/tests/WorkbookTest.java
@@ -31,6 +31,7 @@ import org.labkey.test.util.WorkbookHelper;
 import org.labkey.test.util.WorkbookHelper.WorkbookFolderType;
 import org.openqa.selenium.Keys;
 import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.ui.ExpectedConditions;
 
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/WorkbookTest.java
```diff
@@ -89,19 +90,21 @@ public void testSteps()
         assertEquals("id's generated when workbooks are created should be sequential", Arrays.asList(1, 2, 3), ids);
 
         // Edit Workbook Name
-        waitAndClick(Locator.css(".wb-name + .labkey-edit-in-place"));
-        WebElement nameInput = waitForElement(Locator.css(".wb-name + .labkey-edit-in-place + input"));
+        WebElement wbNameEl = Locator.css(".wb-name + .labkey-edit-in-place").waitForElement(shortWait());
```
labkey-chrisj, line 93: this seems like the sort of thing that would go nicely into a component, if we foresee needing to do that again

labkey-chrisj (approved):

PR #664, opened 2021-03-11
Fix SpecimenTest failure

labkey-tchad:
#### Rationale
`SpecimenTest` reconfigures notification settings a couple of times but sometimes it doesn't stick. This is due to the way the form on this page works. There is a fancy input on the page that takes a moment to render. If the test clicks save before the input is there, it submits an empty value, which produces the error below. Waiting for the input to render before clicking save should remedy this.

#### Changes
* Wait for form to render before clicking "Save"

image


labkey-chrisj (approved):
this makes sense

PR #669, opened 2021-03-12
Update entity insert components

labkey-chrisj:
#### Rationale
This updates `EntityInsertPanel `to wait to be loaded while instantiating, and hopefully surfaces the fileUpload panel for that mode of its function.  It also adds a finder/wait to the BulkInsertDialog to support finding error/warning messages

#### Related Pull Requests
https://github.com/LabKey/biologics/tree/fb_bio_media_sample_crud
This is a product feature branch that migrates Biologics to use the FileUpload components for rawMaterials (like it does Samples) and retire its older sampleTypeUploadPage.
image

#### Changes
Promote Grid getter to be a class function vs. being a member of the elementCache- having it outside of the cache allows us to use cache initialization as the gate to make the component waitforLoaded

Some past iteration of the component appears to have dropped support for the FileUpload mode of its operation, this will hopefully restore that.  This introduces the notion of selecting the input operational mode (grid, or file) and surfacing the fileUploadPanel as a subcomponent 
image

Adds the ability for Bulk Insert dialog to report when it has an input warning: 
image

### DIFF
--- a/src/org/labkey/test/components/ui/entities/EntityBulkInsertDialog.java
+++ b/src/org/labkey/test/components/ui/entities/EntityBulkInsertDialog.java
@@ -14,6 +14,7 @@ import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 
 import java.util.List;
+import java.util.Optional;
 
 public class EntityBulkInsertDialog extends ModalDialog
 {
@@ -192,6 +193,18 @@ public class EntityBulkInsertDialog extends ModalDialog
         return elementCache().checkBox(fieldKey).get();
     }
 
+    public Optional<WebElement> validationMessage()
+    {
+        return elementCache().validationMessage.findOptionalElement(this);
+    }
+
+    public String waitForValidationError()
+    {
+        WebDriverWrapper.waitFor(()-> validationMessage().isPresent(),
+                "Field validation error did not appear", 2000);
+        return validationMessage().get().getText();
+    }
+
     public void clickAddRows()
     {
         elementCache().addRowsButton.click();
@@ -253,6 +266,8 @@ public class EntityBulkInsertDialog extends ModalDialog
 
     protected class ElementCache extends ModalDialog.ElementCache
     {
+        public Locator validationMessage = Locator.tagWithClass("span", "validation-message");
+
         public WebElement formRow(String fieldKey)
         {
             return Locator.tagWithClass("div", "row")
--- a/src/org/labkey/test/components/ui/entities/EntityBulkInsertDialog.java
+++ b/src/org/labkey/test/components/ui/entities/EntityBulkInsertDialog.java
@@ -193,11 +193,20 @@ public class EntityBulkInsertDialog extends ModalDialog
         return elementCache().checkBox(fieldKey).get();
     }
 
+    /**
+     * Finds a validation/error message in the dialog, if one exists.
+     * @return
+     */
     public Optional<WebElement> validationMessage()
     {
         return elementCache().validationMessage.findOptionalElement(this);
     }
 
+    /**
+     * Gets the text value of an error/warning message in the dialog, if it exists within 2 seconds.
+     * Treats non-existence of the error message as a failure
+     * @return
+     */
     public String waitForValidationError()
     {
         WebDriverWrapper.waitFor(()-> validationMessage().isPresent(),
--- a/src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
+++ b/src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
@@ -1,9 +1,11 @@
 package org.labkey.test.components.ui.entities;
 
 import org.labkey.test.Locator;
+import org.labkey.test.WebDriverWrapper;
 import org.labkey.test.components.Component;
 import org.labkey.test.components.WebDriverComponent;
 import org.labkey.test.components.react.ReactSelect;
+import org.labkey.test.components.ui.files.FileUploadPanel;
 import org.labkey.test.components.ui.grids.EditableGrid;
 import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.WebDriver;
@@ -86,7 +88,7 @@ public class EntityInsertPanel extends WebDriverComponent<EntityInsertPanel.Elem
         getWrapper().setFormElement(elementCache().addRowsTxtBox, Integer.toString(records.size()));
         elementCache().addRowsButton.click();
 
-        List<Integer> rowIndices = elementCache().grid.getEditableRowIndices();
+        List<Integer> rowIndices = grid().getEditableRowIndices();
 
         if(rowIndices.size() < records.size())
         {
@@ -106,7 +108,7 @@ public class EntityInsertPanel extends WebDriverComponent<EntityInsertPanel.Elem
 
     public EntityInsertPanel setRecordValues(Map<String, Object> columnValues)
     {
-        int insertRowIndex = elementCache().grid.getEditableRowIndices().get(0);
+        int insertRowIndex = grid().getEditableRowIndices().get(0);
         return setRecordValues(columnValues, insertRowIndex);
     }
 
@@ -114,7 +116,7 @@ public class EntityInsertPanel extends WebDriverComponent<EntityInsertPanel.Elem
     {
         for(String columnName : columnValues.keySet())
         {
-            elementCache().grid.setCellValue(row, columnName, columnValues.get(columnName));
+            grid().setCellValue(row, columnName, columnValues.get(columnName));
         }
 
         return this;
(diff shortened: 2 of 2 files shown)


labkey-chrisj:
https://teamcity.labkey.org/project.html?projectId=LabKey_Trunk_Premium_SampleManager&tab=projectOverview&branch_LabKey_Trunk_Premium_SampleManager=bio_media_sample_crud  Should have test results from SM with this iteration of these components

src/org/labkey/test/components/ui/entities/EntityBulkInsertDialog.java
```diff
@@ -192,6 +193,18 @@ public boolean getBooleanField(String fieldKey)
         return elementCache().checkBox(fieldKey).get();
     }
 
+    public Optional<WebElement> validationMessage()
+    {
+        return elementCache().validationMessage.findOptionalElement(this);
+    }
+
+    public String waitForValidationError()
```
labkey-danield, line 210: Add a javadoc comment for this. Is the idea that you would call one of the 'set' functions with an invalid value and then call this?

src/org/labkey/test/components/ui/entities/EntityBulkInsertDialog.java
```diff
@@ -192,6 +193,18 @@ public boolean getBooleanField(String fieldKey)
...
+    {
+        WebDriverWrapper.waitFor(()-> validationMessage().isPresent(),
+                "Field validation error did not appear", 2000);
+        return validationMessage().get().getText();
```
labkey-danield, line 214: I think one of the things Trey comments on frequently is not to find the element twice if you don't have to. So to that end would this work:

```
Optional<WebElement> element = validationMessage();
WebDriverWrapper.waitFor(()-> element.isPresent(),
                "Field validation error did not appear", 2000);
        return element.get().getText();
```

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
```diff
@@ -204,14 +280,33 @@ protected ElementCache newElementCache()
 
     protected class ElementCache extends Component.ElementCache
     {
+        public ElementCache()
+        {
+            waitForLoaded();
+        }
```
labkey-danield, line 286: I read the general comments for this PR, but I don't think this is correct. The waitForLoaded in the ElementCache waits for elements that are not in the elementCache, because of that the waitForLoad should be at the panel.

Why not make waitForLoaded public and then have consumers wait until it is loaded? Ideally that should be the page(s) that expose this panel. Or call the waitForLoaded in the EntityInsertPanel constructor.

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
```diff
@@ -102,73 +108,105 @@ public EntityInsertPanel addRecords(List<Map<String, Object>> records)
...
-        return elementCache().grid;
+        showGrid();
+        return grid();
+    }
+
+    public FileUploadPanel getFileUploadPanel()
+    {
+        showFileUpload();
+        return fileUploadPanel();
+    }
```
labkey-danield, line 135: If I understand what component this is address, this is incorrect. The upload panel is not a sub panel in the insert panel, it is a separate panel on the page. This should not be nested here. It would be helpful if you provided a screenshot, or some other reference, in your PR description to clarify.

SampleManager has these as two different pages (similar to how other automation in LabKey is done). SampleImportPage covers this functionality.

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
```diff
@@ -102,73 +108,105 @@ public EntityInsertPanel addRecords(List<Map<String, Object>> records)
...
+
+    private EditableGrid grid()
+    {
+        return new EditableGrid.EditableGridFinder(_driver).timeout(WAIT_FOR_JAVASCRIPT).waitFor(this);
+    }
```
labkey-danield, line 140: I'm still not clear how this is more advantageous than putting this in the elementCache. It would be the same as the other functions in that class that return an element.

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
```diff
@@ -102,73 +108,105 @@ public EntityInsertPanel addRecords(List<Map<String, Object>> records)
...
+
+    private FileUploadPanel fileUploadPanel()
+    {
+        return new FileUploadPanel.FileUploadPanelFinder(_driver).timeout(WAIT_FOR_JAVASCRIPT).waitFor(this);
     }
```
labkey-danield, line 145: See my comment above.

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
```diff
@@ -196,6 +234,44 @@ protected boolean isVisible(Locator locator)
...
+        WebDriverWrapper.waitFor(()-> isGridVisible(),
+                "the grid did bot become visible", 2000);
+        return this;
+    }
+
+    private EntityInsertPanel showFileUpload()
+    {
+        if (!isFileUploadVisible())
+            elementCache().modeSelectListItem("from File")
+                    .waitForElement(this, 2000).click();
```
labkey-danield, line 252: Again if I understand the component (please add a screen shot to clarify), this is dangerous because this is clicking a tab to switch panels, and this is being done in a panel. This is a function more appropriate at a page.

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
```diff
@@ -196,6 +234,44 @@ protected boolean isVisible(Locator locator)
...
+
+        WebDriverWrapper.waitFor(()-> isFileUploadVisible(),
+                "the file upload panel did bot become visible", 2000);
+        return this;
+    }
+
+    private void waitForLoaded()
+    {
+        WebDriverWrapper.waitFor(()-> isGridVisible() || isFileUploadVisible(),
+                "The insert panel did not become loaded", WAIT_FOR_JAVASCRIPT);
```
labkey-danield, line 262: This is part of the danger of treating this as a nested panel. It assumes that this starts on the grid and then when calling isFileUploadVisible the view will change by clicking a tab. Won't this by default then leave you on the file upload panel?

labkey-danield (changes requested):
Can you please add a screen shot to the PR description to provide context. If I understand the component correctly I don't think this is the correct design. You. are nesting panels that are not nested, they should be two separate panels on a page.

src/org/labkey/test/components/ui/entities/EntityBulkInsertDialog.java
labkey-chrisj: Anything you do that will cause the dialog to show a warning/input error should show this.  For example, if you set quantity to anything >1000, you will see one of these.  Or, we could just not put this on the component and require tests to do it themselves

src/org/labkey/test/components/ui/entities/EntityBulkInsertDialog.java
labkey-chrisj: Will that actually work?  If you assign the value of an element before its target exists in the DOM will the element magically update itself it it appears in the DOM later? I would expect that to just be false unless calling `.isPresent()` causes webdriver to re-evaluate the DOM for it again.

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
labkey-chrisj: It happens that if the constructor-time call to waitForLoaded() references members of the elementCache (which aren't there yet), we end up with infinite recursion/stackOverflow.  That's why I moved the grid and fileupload panels out of the elementCache, to use finder/wait methods.

I've had (and lost) this argument about having waitForLoaded be public, with Trey.  I did it this way because it's how ResponsiveGrid, ReactSelect do it- as an internal block the component does.
Alternatively, we could call waitForLoaded() at the entry to every public method that might require it, or we could add defensive waits for every element of the elementCache

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
labkey-chrisj: I've included a screen-grab of the file-upload panel part of the EntityUploadPanel from a non-biologics/non-SampleManager view, to illustrate that file upload is a selectable mode of the component, the panel is there in it

Or, we could just not do this and only support grid functions in this component, it's not a hill I need to die on here
I only added it here because I wanted a functional waitForLoaded() function that wouldn't be broken if the page for some reason opened the component in file upload mode

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
labkey-chrisj: It beats infinite recursion.
Or, we could write a waitForLoaded method that doesn't use the grid or the fileUploadPanel and still call it from the elementCache constructor, or we could put the waitForLoaded in the component constructor but that puts the wait up front instead of deferring it until you try to do something with the instance.
Also, first-order getters like these (which don't cache) also don't risk staleness in a component if a test toggles to the other mode the component supports

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
labkey-chrisj: the file upload container component is a child of the file upload pane, which is a child of the EntityInsertPanel component
image

labkey-chrisj: My thought is that our component wrapper classes shouldn't require a page to do things that the component does.

labkey-chrisj:
If this is going to turn into a drawn-out thing, I can just toss this work and write page classes for Biologics that will make sure they don't do things that will case the component to fall over.  I only did this in the component class because we've agreed as a team that components should handle their own readiness internally

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
labkey-chrisj: A quick look through EntityInsertPanel.tsx   [labkey-ui-components/packages/components/src/internal/components/entities/EntityInsertPanel.tsx]
tells me that file upload/add/remove/submit are parts of what it does.

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
labkey-chrisj: The tab-switch is, as I understand, part of the component.  (This is based on my conversations with Cory about testing these in the core-components.view of every labkey folder)  Here it is, in the core-components view: 
image

If this is a hill you're willing to die on, fine. we can scrap this branch and I'll write pages in Biologics that don't expect the component to do this kind of thing

labkey-chrisj:
This is closed to prevent dependency on this branch holding up product feature work

src/org/labkey/test/components/ui/entities/EntityBulkInsertDialog.java
labkey-tchad: Chris is correct. `validationMessage` will find the message or not and `element.isPresent` would never change. If `validationMessage` returned a `LazyWebElement`, you could find it once and wait for `element.isDisplayed`.

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
labkey-tchad: I like the pattern of having the element cache constructor provide `waitForLoaded` functionality because it is more compatible with lazy initialization. It just requires you to move all of the things you call during the wait (`isGridVisible` and `isFileUploadVisible`) into the element cache to avoid the stack overflow you mentioned. The the constructor would be like:
```
public ElementCache()
{
    WebDriverWrapper.waitFor(()-> isGridVisible() || isFileUploadVisible(),
                "The insert panel did not become loaded", WAIT_FOR_JAVASCRIPT);
}
```

PR #672, opened 2021-03-18
Test Automation for Issue 41202: Combination of Submitter and Reader …

labkey-danield:
Migrated the existing StudySecurityTest from BaseStudyTest to BaseWebDriverTest.
Refactored tests to work with new base class.
Filled out StudySecurityPage
Refactored tests to use StudySecurityPage.

#### Rationale
This is being done as part of the automation for Issue 41202: Combination of Submitter and Reader Role does not see the "Insert Rows" icon
Only the very top part of the Study Security form was complete, the test need the other part of the for as well. The form is composed of panels but this does not break the panels out into component parts (primarily because I do not believe they are used anywhere else).
The import/export part of the form has not been completed.
Screen Shot 2021-03-17 at 8 32 46 PM

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2106

#### Changes
Implemented the Study Security for groups panel:
Screen Shot 2021-03-17 at 8 33 37 PM

Implemented Per Dataset Permissions panel
Screen Shot 2021-03-17 at 8 35 05 PM

### DIFF
--- a/src/org/labkey/test/pages/study/StudySecurityPage.java
+++ b/src/org/labkey/test/pages/study/StudySecurityPage.java
@@ -15,21 +15,30 @@
  */
 package org.labkey.test.pages.study;
 
+import org.junit.Assert;
 import org.labkey.test.Locator;
 import org.labkey.test.WebDriverWrapper;
 import org.labkey.test.WebTestHelper;
 import org.labkey.test.components.html.SelectWrapper;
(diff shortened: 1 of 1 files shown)


labkey-danield:
I did it again! I did a pr into develop and not 21.3-SNAPSHOT.

PR #678, opened 2021-03-23
EntityInsertPanel updates

labkey-chrisj:
#### Rationale
Currently the `EntityInsertPanel `test wrapper has no concept of waiting for itself to be ready, which does not conform to our standards.  Several current biologics test failures are likely to be resolved by making the component handle its own loadedness requirements.
It also doesn't have a concept of having both grid-insert and file-upload modes (or for that matter, copy/paste, which is shown when inserting assay run data), despite these being behaviors implemented by the component ([here](https://github.com/LabKey/labkey-ui-components/blob/master/packages/components/src/internal/components/entities/EntityInsertPanel.tsx))

It may be that we could stand to add some clarity around what exactly is (and what isn't) on the component, when considering which behaviors we ought to implement on component wrappers like this one.  For example, the targetEntitySelect and parentSelection button/selects are optional, but we wrap them in the test wrapper.  The submit and cancel buttons are always there, (and per Cory, part of the component) but we regard them as part of the page, never of the component.  Likewise, the copy/paste (text-paste) panel of it is there on assay upload but is not handled by the component.  Perhaps our design discussions will benefit from dev input on what is/what isn't the component.

The file-upload side is already tested in SM as a part of the pages, but because this component has multiple possible modes and state combinations of them, it seemed prudent to have `waitForLoaded()` acknowledge these possible states and handle them accordingly.  

Here's what `EntityInsertPanel `looks like in grid-mode, before the target type is selected: 
image

Here's what we get when the target type is selected:
image

Here's what we get when the file upload tab is selected
image

Here's what we see when a file has been attached to the form (exactly what we see in DomainDesigner when inferring fields from file and uploading file data there)
image

Here's what we see when uploading assay data and there is no target-entitySelect (but there is a copy-and-paste upload mode)
image
This view brings to light that the component test wrapper also doesn't have a concept of the component's copy-paste functionality (which I'd very much like to add)

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/842

#### Changes

- new method, waitForLoaded (called from the ElementCache constructor)
- new getters for `targetTypeSelect`, `editableGrid`, `fileUploadPanel`
- defensively invalidate elementCache when switching between grid and file upload modes
- This branch fixes a bug in TestDataGenerator, which was causing data written in writeTsvData to be null-discovered while writing a test case for file upload
- New test cases (in Biologics, and in EntityInsertPanelTest) to exercise file upload behavior


labkey-chrisj:
SM tests with these changes https://teamcity.labkey.org/project.html?projectId=LabKey_Trunk_Premium_SampleManager&tab=projectOverview&branch_LabKey_Trunk_Premium_SampleManager=entity_insert_pane_updates
Biologics tests with them https://teamcity.labkey.org/project.html?projectId=LabKey_Trunk_Premium_ProductSuites_Biologics&branch_LabKey_Trunk_Premium_ProductSuites_Biologics=entity_insert_pane_updates

src/org/labkey/test/components/ui/entities/EntityBulkInsertDialog.java
```diff
@@ -192,6 +193,18 @@ public boolean getBooleanField(String fieldKey)
         return elementCache().checkBox(fieldKey).get();
     }
 
+    public Optional<WebElement> validationMessage()
```
labkey-chrisj, line 200: This warning is shown if a test or user attempts to specify more than 1000 things to insert.
`EntityBulkInsertDialog `is the dialog that is shown when the 'bulk insert' button on `EntityInsertPanel `is clicked.

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
```diff
@@ -45,6 +48,11 @@ public WebDriver getDriver()
         return _driver;
     }
 
+    public ReactSelect entityTypeSelect()
+    {
+        return ReactSelect.finder(getDriver()).withIdStartingWith("targetEntityType").waitFor();
+    }
+
```
labkey-chrisj, line 56: Our existing test code relies on the test (or page) knowing the label text of the targetEntityType select's label.
I've chosen to add this method instead of replacing the one below (`geEntityTypeSelect(String label)`) in order to give us options about whether or not to refactor test code using it.  If it's preferable to replace older usages with this one, the refactor will be pretty easy and pleasant to do.

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
```diff
@@ -139,6 +147,13 @@ private EditableGrid grid()
         return new EditableGrid.EditableGridFinder(_driver).timeout(WAIT_FOR_JAVASCRIPT).waitFor(this);
     }
 
+    public EntityInsertPanel setUpdateDataForFileUpload(boolean checked)
```
labkey-chrisj, line 160: This checkbox appears on the file upload pane, and configures whether or not the file import process should overwrite or update existing records

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
```diff
@@ -234,36 +249,56 @@ protected boolean isVisible(Locator locator)
         }
     }
```
labkey-chrisj, line 267: My investigation of switching panes between grid/file upload shows that the switch doesn't show/hide DOM elements, it  destroys them but the DOM flutter occurs within the ComponentElement so it doesn't cause the componentElement to go stale.  Interestingly, state is well-preserved between panes (if you've attached a file to the fileUpload panel and switch to grid and then come back, the file remains attached and the preview is still there).  Invalidating the cache here seems to be a best defensive practice

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
```diff
@@ -287,8 +287,8 @@ private void waitForLoaded()
         WebDriverWrapper.waitFor(()-> {
```
labkey-chrisj, line 306: This change acknowledges the use case for assay run updates;  

- when uploading assay run data, no entityTypeSelect is present so there needs to be a condition here that doesn't require an entityTypeSelect to be shown.
- Likewise, the grid isn't shown unless the entityTypeSelect has a value/the component has a target to get the target table schema from- so isGridVisible() will always mean the entityTypeSelect has a value.
- If the user arrives at a page without the URL param to populate entityTypeSelect, it will be empty (this case happens in the core-components view, and can happen in SM and Biologics.

src/org/labkey/test/tests/component/EntityInsertPanelTest.java
```diff
@@ -109,6 +131,18 @@ private boolean isSubmitEnabled()
         return null == submitButton().getAttribute("disabled");
     }
 
+    public void clickFileImport()
```
labkey-chrisj, line 133: It turns out that the buttons for file import and cancel are different buttons than the ones for grid import and cancel.  They follow very similar logic governing whether they are enabled or not based on the state of the component; I believe there's a strong argument in favor of having the test component wrapper implement them but it's not a hill I'm willing to die on

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -304,7 +304,8 @@ public String writeTsvContents()
         {
```
labkey-chrisj, line 310: The problem here was that the key on the row is somehow cased differently than its name in `_indices`.  Thus, calling `row.get(_indices.get(index).getName())` returned null because of case mismatching.

One potentially problematic issue with this solution is that it toLowerCases `keyAtIndex` and that might not be right(but it was the only way I could make this work).  
The key of the row is the column name, in this case if I passed the name pulled out of the PropertyDescriptor and it missed because the values in `row `were all lower case, the names were camel-cased.  I've had success in the past resolving case-insensitive key mismatches by streaming over a set to find the one that matches in the same case and then passing that as the key, but I suspect there's a better long-term fix to this than that.

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
labkey-tchad: I'd say [deprecate](https://docs.oracle.com/javase/tutorial/java/annotations/predefined.html) the old one and we can refactor usages at our leisure while redirecting folks to use the correct method going forward.

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
```diff
@@ -102,73 +117,126 @@ public EntityInsertPanel addRecords(List<Map<String, Object>> records)
...
         }
 
         return this;
     }
 
     public EditableGrid getEditableGrid()
+    {
+        showGrid();
+        return grid();
+    }
```
labkey-tchad, line 147: Is the extra `grid()` method really necessary? Baking `showGrid` into the grid getter is a good idea but I don't think we need a separate method without it.

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
```diff
@@ -196,6 +264,64 @@ protected boolean isVisible(Locator locator)
...
     }
 
+    public EntityInsertPanel showGrid()
+    {
+        if (!isGridVisible())
+            elementCache().modeSelectListItem("from Grid")
+                    .waitForElement(this, 2000).click();
+        clearElementCache();
+        WebDriverWrapper.waitFor(()-> isGridVisible(),
+                "the grid did bot become visible", 2000);
```
labkey-tchad: Hide this all under the `if` so that we don't clear the cache unnecessarily.
```suggestion
        if (!isGridVisible())
        {
            elementCache().modeSelectListItem("from Grid")
                    .waitForElement(this, 2000).click();
            clearElementCache();
            WebDriverWrapper.waitFor(()-> isGridVisible(),
                "the grid did bot become visible", 2000);
        }
```

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
```diff
@@ -196,6 +264,64 @@ protected boolean isVisible(Locator locator)
...
+
+    /**
+     * finds the mode select tabs, to switch between grid input and file upload
+     * @param containsText
+     * @return
+     */
+    private Locator.XPathLocator modeSelectListItem(String containsText)
+    {
+        return Locator.tagWithClass("li", "list-group-item").containing(containsText);
+    }
```
labkey-tchad, line 327: Why isn't this in the element cache?

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
```diff
@@ -196,6 +264,64 @@ protected boolean isVisible(Locator locator)
...
+        if (!isFileUploadVisible())
+        {
+            var toggle = elementCache().modeSelectListItem("from File")
+                    .waitForElement(this, 2000);
+            getWrapper().shortWait().until(ExpectedConditions.elementToBeClickable(toggle));
+            toggle.click();
+        }
+        clearElementCache();
+        WebDriverWrapper.waitFor(()-> isFileUploadVisible(),
+                "the file upload panel did bot become visible", 2000);
```
labkey-tchad: No need to clear the cache and wait for the panel if the panel is already in file upload mode. Put all of this under the `if`:
```suggestion
        if (!isFileUploadVisible())
        {
            var toggle = elementCache().modeSelectListItem("from File")
                    .waitForElement(this, 2000);
            getWrapper().shortWait().until(ExpectedConditions.elementToBeClickable(toggle));
            toggle.click();
            clearElementCache();
            WebDriverWrapper.waitFor(()-> isFileUploadVisible(),
                    "the file upload panel did bot become visible", 2000);
        }
```

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
```diff
@@ -204,19 +330,45 @@ protected ElementCache newElementCache()
...
+            return bulkInsertBtnLoc.findElement(this);
+        }
+        WebElement bulkUpdateBtn()
+        {
+            return Locator.button("Bulk update").findElement(this);
+        }
+        WebElement deleteRowsBtn()
+        {
+            return deleteRowsBtnLoc.findElement(this);
+        }
```
labkey-tchad: Is switching these buttons to getters necessary? The element cache gets cleared when switching modes.

src/org/labkey/test/util/TestDataGenerator.java
labkey-tchad: You could use `CaseInsensitiveHashMap` (from our remoteapi) for the rows. It exists for precisely this sort of thing.

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
labkey-chrisj: done

src/org/labkey/test/util/TestDataGenerator.java
labkey-chrisj: If I make rows into a List of CaseInsensitiveHashMap, I run into problems trying to pass rows into an insertRowsCommand- it expects a List<Map<String, Object>> 
...unless there's a ready conversion maybe?

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
labkey-chrisj: It shouldn't be necessary, now that waitForLoaded doesn't rely on it being present.  I've realigned getEditableGrid() to return the cached grid, and have replaced calls to grid() to call getEditableGrid() or else use the elementCache

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
labkey-chrisj: good catch, thanks

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
labkey-chrisj: Probably not.  I'll switch them back to cached lazyWebElements, with short searchTimeouts

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
labkey-chrisj: It's referenced in methods that are called by waitForReady.  `isDeleteRowsVisible` and `isFileUploadVisible `both use it- because waitForReady is called from the elementCache constructor, it's not available yet.

src/org/labkey/test/util/TestDataGenerator.java
labkey-chrisj: I figured out why the case was mismatched- when generating rows with random data, I was creating them with column name and type toLowerCase(), which caused the mismatch here.  By no longer doing that (and by instead switching over columnType.toLowerCase when setting dataSuppliers for columns to ensure a match) I think I've got a fix.

I'd like to discuss using a CaseInsensitiveHashMap here, but I worry that it will require changing a lot of test code (and our remoteAPI wrappers, perhaps)

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
labkey-chrisj: it is referenced by methods called from waitForReady.  Because that's called from the elementCache constructor, it doesn't work for it to be in the constructor

src/org/labkey/test/util/TestDataGenerator.java
labkey-tchad: I don't think `columnType.toLowerCase` will make any difference here. The mismatch is with the name, not the type, right?
As for using CIHashMap; it wouldn't require much refactor. I think the only change needed would be to update `addCustomRow` and `generateRows` to use it. (See #686)

labkey-tchad (approved):
Consider #686

PR #684, opened 2021-03-30
Test fix

labkey-sweta:
#### Rationale
Changing the column name from CreatedBy to createdBy.

#### Related Pull Requests
* 

#### Changes
src/org/labkey/test/tests/CustomizeViewTest.java


labkey-tchad (approved):

labkey-matthewb (commented):
I think this is a server bug with List column names.

labkey-matthewb:
> I think this is a server bug with List column names.

fix on its way

labkey-sweta:
Fixed as part of product bug fix.

PR #686, opened 2021-03-31
Use case insensitive map for test data

labkey-tchad:


labkey-chrisj (approved):
I like this, but when I tried doing it I made _rows into a list of CaseInsensitiveHashMap and that became a problem when passing them into insertRows().
If this will work, fantastic!

PR #687, opened 2021-04-01
Fb update test study archive

labkey-danield:
#### Rationale
Support for importing (and converting) the old chart formats has been removed. Any test cases that used these old archives needed to be updated, to ignore the charts that are no longer being imported, or the archive needs to be recreated with the new chart format.

#### Related Pull Requests
* None

#### Changes
* Added a Read_Me.txt file that list the study archives in the testAutomation/data/studies directory and the test classes that may use them.
* Removed references to charts that are no longer being imported from the DataViewsTests and ReportAndDatasetNotificationTest. These tests where not interested in the content of the charts only in the items imported, so the reference to the chart can safely be removed.
* Updated the LabKeyDemoStudyWithCharts.folder.zip file. The ReportThumbnailTest uses the chart that was no longer be imported so that needed to be recreated and the study exported (archived) again.


data/studies/Read_Me.txt
```diff
@@ -0,0 +1,151 @@
...
+		StudyScheduleTest.java (DailyC)
+		StudySimpleExportTest.java (DailyC)
+		StudyTest.java (DailyC)
+		StudyVisitTagTest.java (DailyC)
+		TimeChartImportTest.java (DailyC)
+	StudyDatasetsTest.java (DailyA)
+	TruncationTest.java (DailyA)
+
+CAVDTestStudy.folder.zip
+	Not used?
```
labkey-tchad: This is used by the `CAVDStudyTest` in [customModules](https://github.com/LabKey/customModules/blob/develop/viscstudies/test/src/org/labkey/test/tests/viscstudies/CAVDStudyTest.java). We should probably move this file (not sure why it ended up here after the big sampledata migration.

labkey-tchad (approved):
Consider checking in the unzipped version of `LabKeyDemoStudyWithCharts.folder.zip` and using `org.labkey.test.util.StudyImporter#zipAndImportStudy` to import it.

data/studies/Read_Me.txt
labkey-danield: Updated.

labkey-adam:
You might consider documenting usages of the exploded archives as well: /data/study, /data/studies/ExtraKeyStudy, and various ones under data/FolderExport

labkey-adam (approved):

PR #688, opened 2021-04-01
Create temp dir

labkey-tchad:
#### Rationale
`TestFileUtils.saveFile` assumes that the parent directory already exists.

#### Related Pull Requests
* #678


src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -322,10 +322,12 @@ public String writeTsvContents()
      * Creates a file containing the contents of the current rows, formatted in TSV.
      * The file is written to the test temp dir
      * @param fileName  the name of the file, e.g. 'testDataFileForMyTest.tsv'
-     * @return
+     * @return File object pointing at created TSV
      */
     public File writeData(String fileName)
     {
```
labkey-chrisj, line 328: I like this.
Would it make sense to push this down into testFileUtils, maybe do something similar with the projectFilesRoot location?

labkey-chrisj (approved):

PR #690, opened 2021-04-06
Use 'startsWith' to find responsiveGrid header cell

labkey-chrisj:
#### Rationale
Now that the grid supports PHI/redacting fields, it's time for the wrapper to support that.
image
(the far-right cell pictured here, 'Sequence', has a class decoration on the cell to signal that the contents of this column are redacted for PHI reasons.  This change supports surfacing this data, along with the title attribute (which is used to message why).

While testing, I found a problem in the way we find header cells; we've been using 'contains' to find the first matching cell with that text in it, but in this case it returns the 'Prot Sequences' header cell, because it has 'Sequence' in it.  Switching to use 'startsWith' un-blocks that.  I've created this PR to test just that change so we can find the test (if it exists) that lives or dies on a 'contains' matching.

#### Related Pull Requests
(soon.  tests for neurimmune/biologics story, to support hiding sequence data from non-PHI users.   It uses the new methods on ResponsiveGrid in this PR)

#### Changes
Use 'startsWith' for text matching when finding grid header cell
surface ability to check for whether or not the grid is redacting content for PHI reasons
provide means to check for the title attribute (where an explanation for PHI transform will likely be)


labkey-chrisj:
ugh, all these merge branches coming along with branch creation.  If anyone can suggest a way to limit that sort of thing, I'd love to hear it

labkey-chrisj:
biologics tests with this change: https://teamcity.labkey.org/project/LabKey_Trunk_Premium_ProductSuites_Biologics?branch=grid_header_cell_help&buildTypeTab=overview&mode=builds

sampleManager tests with this change: https://teamcity.labkey.org/project/LabKey_Trunk_Premium_SampleManager?branch=grid_header_cell_help&buildTypeTab=overview&mode=builds#all-projects

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
```diff
@@ -522,6 +522,32 @@ public boolean gridMessagePresent()
...
+     * @return  the contents of the 'title' attribute of the cell, or null if the attribute is
+     * not present.
+     */
+    public String getColumnTitleAttribute(String columnText)
+    {
+        var cell = elementCache().getColumnHeaderCell(columnText);
+        if (null == cell.getAttribute("title"))
+            return null;
+        else
+            return cell.getAttribute("title");
```
labkey-danield: Why not just do:
`return elementCache().getColumnHeaderCell(columnText).getAttribute("title");`

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
```diff
@@ -572,7 +598,8 @@ protected final WebElement getColumnHeaderCell(String headerText)
         {
             if (!headerCells.containsKey(headerText))
             {
-                WebElement headerCell = Locator.xpath("//th[./span[contains(text(), '" + headerText + "')]]").findElement(this);
+                WebElement headerCell = Locator.tagWithClass("th", "grid-header-cell")
+                        .withChild(Locator.tag("span").startsWith(headerText)).findElement(this);
```
labkey-danield, line 598: Wouldn't `Locator.tag("span").withText(headerText))` be a better option? This would remove any ambiguity as to which column to get.

labkey-danield (approved):

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
labkey-chrisj: I think I was just worried about the attribute being null and trying to be clever
your suggestion is better, actually

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
labkey-chrisj: I tried finding it with `withText ` initially, but for some reason it didn't work.  I resorted to using `startsWith` when it worked for every use case I could contrive, but I didn't thoroughly investigate why 'withText' didn't work even when calling 'getText' on the found element returned the same text.  I suspect that when searching, text within contained elements might matter (most of these header cells have menus in a contained element in them)?

labkey-adam:
@labkey-chrisj Not sure exactly how and when you created this branch, but this often happens when you commit a change before pulling the latest from changes remote. So, always pull before commit/push.

labkey-tchad:
> ugh, all these merge branches coming along with branch creation. If anyone can suggest a way to limit that sort of thing, I'd love to hear it

You need to delete and recreate you local 'develop' branch. It looks like your local develop has commits that came from 'fb_entity_insert_panel_updates' and 'fb_bio_media_sample_crud'. Not quite sure how they would have gotten there.

PR #692, opened 2021-04-07
LabKey UX - Query metadata editor Automation test

labkey-sweta:
#### Rationale
Automation test coverage for migrated query metadata editor.

#### Related Pull Requests
* 

#### Changes

1. Edit metadata page : src/org/labkey/test/pages/query/EditMetadataPage.java
2. Test : src/org/labkey/test/tests/EditMetadataDomainTest.java

### DIFF
--- a/src/org/labkey/test/pages/query/EditMetadataPage.java
+++ b/src/org/labkey/test/pages/query/EditMetadataPage.java
@@ -1,0 +1,88 @@
+package org.labkey.test.pages.query;
+
+import org.labkey.test.BaseWebDriverTest;
+import org.labkey.test.Locator;
+import org.labkey.test.WebDriverWrapper;
+import org.labkey.test.components.bootstrap.ModalDialog;
+import org.labkey.test.components.domain.BaseDomainDesigner;
+import org.labkey.test.components.domain.DomainFormPanel;
+import org.labkey.test.components.react.ReactSelect;
+import org.labkey.test.util.DataRegionTable;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+import static org.labkey.test.WebDriverWrapper.WAIT_FOR_JAVASCRIPT;
+
+public class EditMetadataPage extends BaseDomainDesigner<EditMetadataPage.ElementCache>
+{
+    public EditMetadataPage(WebDriver driver)
+    {
+        super(driver);
+    }
+
+
+    public DataRegionTable clickViewData(BaseWebDriverTest test, String schemaName, String queryName)
+    {
+        return test.viewQueryData(schemaName, queryName);
+    }
+
+    public void clickEditSourcePage()
+    {
+        elementCache().editSourceBtn.click();
+    }
+
+    public EditMetadataPage clickSave()
+    {
+        elementCache().saveBtn.click();
+        return this;
+    }
+
+    public void clickResetToDefault()
+    {
+        WebDriverWrapper.waitFor(elementCache().resetToDefaultBtn::isEnabled,
+                "The alias button did not become enabled as expected", 2000);
+        elementCache().resetToDefaultBtn.click();
+        new ModalDialog(new ModalDialog.ModalDialogFinder(getDriver()).withTitle("Confirm Reset")).dismiss("Reset");
+    }
+
+    public EditMetadataPage clickAliasField(String fieldName)
+    {
+        WebDriverWrapper.waitFor(elementCache().aliasFieldBtn::isEnabled,
+                "The alias button did not become enabled as expected", 2000);
+        elementCache().aliasFieldBtn.click();
+        ModalDialog dialog = new ModalDialog(new ModalDialog.ModalDialogFinder(getDriver()).withTitle("Choose a field to wrap"));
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/pages/query/EditMetadataPage.java
```diff
@@ -0,0 +1,88 @@
...
+{
+    public EditMetadataPage(WebDriver driver)
+    {
+        super(driver);
+    }
+
+
+    public DataRegionTable clickViewData(BaseWebDriverTest test, String schemaName, String queryName)
+    {
+        return test.viewQueryData(schemaName, queryName);
```
labkey-danield, line 26: This threw me for a bit at first, given the function name I was expecting some kind of click being done on an element in the elementCache.

Given that this is calling a method off of the BaseWebDriverTest do you really need this function in the page? Wouldn't it make more sense to just call `viewQueryData(schemaName, queryName)` in the test?

Actually it looks like this method isn't called at all by the test, so that may be another argument for not needing it.

src/org/labkey/test/pages/query/EditMetadataPage.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    public void clickEditSourcePage()
+    {
+        elementCache().editSourceBtn.click();
+    }
```
labkey-danield, line 32: Should this return a reference to some page? Also looks like this isn't called from the test as well. Maybe this isn't needed?

src/org/labkey/test/pages/query/EditMetadataPage.java
```diff
@@ -0,0 +1,88 @@
...
+        return this;
+    }
+
+    public void clickResetToDefault()
+    {
+        WebDriverWrapper.waitFor(elementCache().resetToDefaultBtn::isEnabled,
+                "The alias button did not become enabled as expected", 2000);
+        elementCache().resetToDefaultBtn.click();
+        new ModalDialog(new ModalDialog.ModalDialogFinder(getDriver()).withTitle("Confirm Reset")).dismiss("Reset");
+    }
```
labkey-danield, line 46: In this case maybe the function should just be called `resetToDefault`, because it does more than just click the button it also responds to the dialog. 

Alternatively you could leave it named as is and return the dialog and let the test deal with it.

Generally when I see a "click" method I'm expecting some singular action that results in something like a page navigation, form submit or returning some prompt I need to interact with. This might be something to discuss during our next coding standards discussion.

src/org/labkey/test/pages/query/EditMetadataPage.java
```diff
@@ -0,0 +1,88 @@
...
+    {
+        WebDriverWrapper.waitFor(elementCache().aliasFieldBtn::isEnabled,
+                "The alias button did not become enabled as expected", 2000);
+        elementCache().aliasFieldBtn.click();
+        ModalDialog dialog = new ModalDialog(new ModalDialog.ModalDialogFinder(getDriver()).withTitle("Choose a field to wrap"));
+        ReactSelect.finder(getDriver()).findWhenNeeded(dialog).select(fieldName);
+        dialog.dismiss("OK");
+
+        return this;
+    }
```
labkey-danield, line 58: Same kind of comment as above maybe this would be better named `setAliasField`.

src/org/labkey/test/pages/query/EditMetadataPage.java
```diff
@@ -0,0 +1,88 @@
...
+                .timeout(WAIT_FOR_JAVASCRIPT)
+                .findWhenNeeded(this);
+
+        protected Locator.XPathLocator queryMetadataButtonPanelLocator()
+        {
+            return Locator.byClass("query-metadata-editor-buttons");
+        }
+
+    }
+}
```
labkey-danield, line 88: There is nothing wrong with having the macro like functions you have created for this page (ResetToDefault, AliasField) that choose the default action for any confirmation dialogs that appear. However keep in mind it does limit the kind of tests that can use the form. For example if there is a test that wants to validate that canceling out of a dialog works as expected, or one that expects an error message when updating, they will not be able to use the methods already defined.

This approach may be perfectly ok for this feature/test but it is something to keep in mind for the future.

labkey-danield (approved):
I have some comments on function names, but otherwise I think it is ok.

It would be helpful if next time you could include a screenshot of the page you are interacting with.

src/org/labkey/test/pages/query/EditMetadataPage.java
labkey-sweta: I added them for future reference. We can update the return type as per the requirement.

src/org/labkey/test/pages/query/EditMetadataPage.java
labkey-sweta: I added the clickViewData() method for consistency.

src/org/labkey/test/pages/query/EditMetadataPage.java
labkey-sweta: Agree

src/org/labkey/test/pages/query/EditMetadataPage.java
labkey-sweta: ok

PR #696, opened 2021-04-13
AppsMenu test components

labkey-chrisj:
#### Rationale
Adds component wrappers for the AppsMenu, which provides navigation between LKB/LKSM/LKS pages

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/857  <- contains the tests that will use the components defined here

#### Changes
Because the menu-expand piece of this is different in LKSM/LKB than it is in LKS, there are two classes to wrap the base menu part- AppsMenu and LKAppsMenu.

Expanding the base menu shows the ProductsNavContainer, which looks like this:
image

Clicking the "LabKey Server" link shows the LKSNavContainer, which looks like this: 
image

Clicking either SM or Biologics from the ProductsNavContainer shows the ProjectsNavContainer, which contains a list of projects of that kind and looks like this: 
image

Clicking a project link in the ProjectsNavContainer shows the LeafNavContainer, which contains a list of product sections (shown here: a Biologics project) 
image

All of the NavContainer-named classes derive from BaseNavContainer.
Those with back-caret titles support navigating back to the previous context


src/org/labkey/test/components/ui/navigation/apps/LKAppsMenu.java
```diff
@@ -0,0 +1,69 @@
...
+import org.labkey.test.Locator;
+import org.labkey.test.components.html.BaseBootstrapMenu;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+/**
+ * Wraps the expand/collapse toggle to show/hide the Apps-navigation menu in LabKey pages
+ * In SM or Biologics views, AppsMenu is the analog of this component
+ */
+public class LKAppsMenu extends BaseBootstrapMenu
```
labkey-danield: It kind of feels like this could be merged with AppsMenu. If I understand correctly this is the "button" on the title bar, maybe if the locators in the getToggleLocator and in the finder used an 'or' condition you could combine these two files (`Locator.xpath("(//li[@id='headerProductDropdown']/a) | (//button[@id='product-navigation-button'])")`).

Maybe I'm missing something.

labkey-danield (approved):

src/org/labkey/test/components/ui/navigation/apps/LKAppsMenu.java
labkey-chrisj: This is a good suggestion, I just worried that if I tried to get too fancy with those two parts you guys would ask me to separate them.
I'm willing to try to unify these two into one that can straddle both worlds, tho

PR #699, opened 2021-04-15
Issue41828: External Specimen Importer Interstitial Page Text

RosalineP:
#### Rationale
For the situation in which the `professional` module was present but not enabled, we previously displayed a message which could lead to confusion about whether specimen import is possible.  

Previous:
Screen Shot 2021-04-14 at 1 24 13 PM

Current:
For when `professional` is inactive but present, we display:
Professional off

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2184

#### Changes
* Update 'upsell' element to 'enable module' element

### DIFF
--- a/src/org/labkey/test/pages/study/ConfigureImporterPage.java
+++ b/src/org/labkey/test/pages/study/ConfigureImporterPage.java
@@ -35,7 +35,7 @@ public class ConfigureImporterPage extends LabKeyPage<ConfigureImporterPage.Elem
     @Override
     protected void waitForPage()
     {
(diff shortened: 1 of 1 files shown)


labkey-chrisj (approved):
Looks good, except for the [build break](https://teamcity.labkey.org/viewLog.html?buildId=1382715&buildTypeId=LabKey_Trunk_Premium_VerifyTestBuild).

55cdd70f998674a/server/modules/professional/test/src/org/labkey/test/tests/professional/ProfessionalImporterProviderUpsellTest.java:56: error: cannot find symbol
                configPage.isUpsellBannerShown());

PR #703, opened 2021-04-21
make EntityBulkInsertDialog wait to be ready

labkey-chrisj:
#### Rationale
Recent test failures suggest that changes to this component are causing us to try to interact with fields in this dialog before the dialog is ready for it.  The addition of options for derivatives/pooled samples may have slowed rendering down enough that the tests are getting to the dialog while it is still being drawn/animated in

#### Related Pull Requests
n/a

#### Changes
bake a call to super.waitForReady into newElementCache
add separate wait for cancel/submit buttons to be present to the elementCache c'tor
move getters for select fields into the elementCache

note that the call to `super.waitForReady` feels ever so slightly hacky (if we had the right type-parameterization on `EntityBulkInsertDialog `and its finder, `super.newElementCache` would do that for free!  But, my attempts to fix that ran into troubles


labkey-chrisj:
Wow, that was super-weird. Had to delete my local branch/pull fresh and re-push the same changes for my local changes to actually be pushed

labkey-tchad (approved):

labkey-danield (approved):

PR #704, opened 2021-04-21
Java API test Commands

labkey-tchad:
#### Rationale
Interacting with `core-saveModuleProperties` and `query-insertExternalSchema` through the UI is time consuming and often unnecessary. Setting up folders in the `UserReg-WS` module requires hitting both of these actions so we can save some time for those tests by using the API.

#### Related Pull Requests
* FDA-MyStudies/UserReg-WS#8

#### Changes
* Add `SaveModulePropertiesCommand` and minor refactors to UI module property helpers
* Add `InsertExternalSchemaCommand`

### DIFF
--- a/src/org/labkey/remoteapi/core/SaveModulePropertiesCommand.java
+++ b/src/org/labkey/remoteapi/core/SaveModulePropertiesCommand.java
@@ -55,6 +55,7 @@ public class SaveModulePropertiesCommand extends PostCommand<CommandResponse>
             propJson.put("moduleName", prop.getModuleName());
             propJson.put("propName", prop.getPropertyName());
             propJson.put("value", prop.getValue());
+            properties.add(propJson);
         }
         json.put("properties", properties);
         return json;
--- a/src/org/labkey/remoteapi/core/SaveModulePropertiesCommand.java
+++ b/src/org/labkey/remoteapi/core/SaveModulePropertiesCommand.java
@@ -1,0 +1,62 @@
+package org.labkey.remoteapi.core;
+
+import org.json.simple.JSONArray;
+import org.json.simple.JSONObject;
+import org.labkey.remoteapi.CommandException;
+import org.labkey.remoteapi.CommandResponse;
+import org.labkey.remoteapi.Connection;
+import org.labkey.remoteapi.PostCommand;
+import org.labkey.remoteapi.security.GetContainersCommand;
+import org.labkey.remoteapi.security.GetContainersResponse;
+import org.labkey.test.params.ModuleProperty;
+
+import java.io.IOException;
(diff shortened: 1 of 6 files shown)


src/org/labkey/remoteapi/core/SaveModulePropertiesCommand.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public JSONObject getJsonObject()
+    {
+        JSONObject json = new JSONObject();
+        JSONArray properties = new JSONArray();
+        for (ModuleProperty prop : _moduleProperties)
+        {
+            JSONObject propJson = new JSONObject();
+            propJson.put("userId", 0);
```
labkey-adam, line 53: Do you understand the purpose of this "userId"... the action blows up if it's missing or invalid (not a real user), but then it otherwise ignores this value.

labkey-adam (approved):

src/org/labkey/remoteapi/core/SaveModulePropertiesCommand.java
labkey-tchad: I just saw it in the POST through the UI and matched that.
Looks like it's a vestigial remnant that snuck by a couple of old refactors.

labkey-tchad: This comment implies that there was once a intention to allow user-specific module properties: https://github.com/LabKey/platform/blob/develop/core/webapp/ModulePropertiesAdminPanel.js#L237

labkey-adam: Cool, thanks. This doesn't hold up your merge, of course. I'll check in with Ben & Josh to verify we should rip out the now pointless server-side check.

src/org/labkey/remoteapi/query/InsertExternalSchemaCommand.java
```diff
@@ -0,0 +1,166 @@
...
+        GetQueriesResponse response = command.execute(connection, folderPath);
+        try
+        {
+            Assert.assertEquals(String.format("Wrong tables published to %s schema '%s'.", _params.schemaType(), _params.userSchemaName),
+                    new HashSet<>(_params.tables), new HashSet<>(response.getQueryNames()));
+        }
+        catch (AssertionError e)
+        {
+            throw new CommandException(e.getMessage());
+        }
```
labkey-danield, line 69: Just curious...
I'm assuming that you are catching this and throwing it as a CommandException for consistency (everything throws CommandException), is that why?

labkey-danield (approved):

PR #712, opened 2021-05-04
Add test helper method for conversion error messages. Use it.

labkey-adam:
#### Rationale
Conversion error message verification is scattered about the tests and the product wording is changing. Add a helper method that generates the expected message.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2232

### DIFF
--- a/src/org/labkey/test/BaseWebDriverTest.java
+++ b/src/org/labkey/test/BaseWebDriverTest.java
@@ -2438,6 +2438,13 @@ public abstract class BaseWebDriverTest extends LabKeySiteWrapper implements Cle
         return exportIconLoc.findElements(getDriver()).size();
     }
 
+    // Note: Keep in sync with OntologyManager.getStandardConversionErrorMessage()
+    public String getConversionErrorMessage(Object value, String fieldName, Class<?> targetClass)
+    {
+        String fromType = (value instanceof String) ? "" : "(" + (value.getClass().getSimpleName() + ") ");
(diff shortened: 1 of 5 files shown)


labkey-danield (approved):

src/org/labkey/test/BaseWebDriverTest.java
```diff
@@ -2438,6 +2438,13 @@ public int getExportScriptIconCount(String chartParentCls)
         return exportIconLoc.findElements(getDriver()).size();
     }
 
+    // Note: Keep in sync with OntologyManager.getStandardConversionErrorMessage()
+    public String getConversionErrorMessage(Object value, String fieldName, Class<?> targetClass)
+    {
+        String fromType = (value instanceof String) ? "" : "(" + (value.getClass().getSimpleName() + ") ");
+        return "Could not convert value " + fromType + "'" + value + "' for field '" + fieldName + "'; expected type " + targetClass.getSimpleName();
+    }
+
```
labkey-tchad: Please move this to `LabKeySiteWrapper`. BWDT should just be methods for test setup, teardown, failure handling, and the like (yes, there is much cleanup to do in this regard).
I would suggest making it static and putting it in some utility class but I don't see any great candidates for where to move it.

labkey-tchad (approved):

src/org/labkey/test/BaseWebDriverTest.java
labkey-adam: Moved

PR #713, opened 2021-05-04
Add filewatcher-test-custom-fields.pipeline.xml

labkey-alan:
#### Rationale
This new file watcher definition serves as our only example of using custom fields in script based pipelines, and should be used in a future test to test the rendering of custom fields.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2234
* https://github.com/LabKey/premium/pull/245/files

#### Changes
* Add filewatcher-test-custom-fields.pipeline.xml
* Update PipelineTriggerWizard.java to be compatible with new UI

### DIFF
--- a/src/org/labkey/test/components/pipeline/PipelineTriggerWizard.java
+++ b/src/org/labkey/test/components/pipeline/PipelineTriggerWizard.java
@@ -17,20 +17,18 @@ package org.labkey.test.components.pipeline;
 
 import org.jetbrains.annotations.NotNull;
 import org.labkey.test.Locator;
-import org.labkey.test.Locators;
 import org.labkey.test.WebDriverWrapper;
 import org.labkey.test.WebTestHelper;
 import org.labkey.test.components.Component;
 import org.labkey.test.components.WebDriverComponent;
-import org.labkey.test.components.bootstrap.ModalDialog;
 import org.labkey.test.components.html.Checkbox;
 import org.labkey.test.components.html.Input;
 import org.labkey.test.components.html.OptionSelect;
-import org.labkey.test.components.labkey.LabKeyAlert;
+import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 
-import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/components/pipeline/PipelineTriggerWizard.java
```diff
@@ -183,11 +189,17 @@ public PipelineTriggerWizard setParameterFunction(String function)
         return this;
     }
 
-    public PipelineTriggerWizard addCustomParamter(String key, String value, @NotNull Integer index)
+    public PipelineTriggerWizard showAdvanced()
+    {
+        elementCache().showAdvanced.click();
+        return this;
```
labkey-chrisj, line 195: This sort of method usually does better if there's a related check to see if the desired state is already present, like 'isAdvancedShown()'
so, if (!isAdvancedShown())
     elementCache.showAdvanced.click()
     waitFor(()-> isAdvancedShown, "didn't happen in time", 2000)
might be how showAdvanced() might work

src/org/labkey/test/components/pipeline/PipelineTriggerWizard.java
```diff
@@ -233,12 +245,12 @@ protected ElementCache newElementCache()
 
     protected class ElementCache extends Component.ElementCache
     {
-        public ElementCache()
-        {
-            if (getWrapper().getUrlParameters().containsKey("rowId"))
-                // Wait for saved config to populate form
-                Locators.pageSignal("triggerConfigLoaded").waitForElement(getDriver(), 10000);
-        }
+//        public ElementCache()
```
labkey-chrisj, line 248: probably best to remove commented code, but it's worth considering whether or not removing this synchronizing code could cause downstream race conditions

labkey-chrisj (approved):

src/org/labkey/test/components/pipeline/PipelineTriggerWizard.java
labkey-alan: We no longer asynchronously load the saved config, it's injected by the JSP now, so there shouldn't be any issue.

PR #714, opened 2021-05-05
Ontology annotation test components and wrappers

labkey-chrisj:
#### Rationale
Contains new component wrappers for ontology-specific UI, also updates to DomainDesigner components (and FieldDefinition) to support ontology 

#### Related Pull Requests
https://github.com/LabKey/ontology/pull/39 <-- contains the test class, + page for concept browser

#### Changes
modal dialog for ontologyPicker (which currently assumes there's only one ontology, doesn't pick the intended one yet)
wrappers for subcomponents on the browse-concept page and the ontology picker dialog:
**OntologyTreeSearch** - wraps the search bar in the picker dialog, uses **TreeSearchResult** to wrap search results
image
- notable support: find item by label, or by code

**OntologyTreePanel** - wraps the treeview, uses **TreeNode** to wrap nodes in it
image
-currently supports: can select a node given a List<String> path to it
-TreeNode can find its own children, or pick one with a given title

**ConceptInfoTabs** - wraps the tabs on the right in the picker dialog (or on the browse concept page)
image
Can obtain selected concept title, selected concept code, synonyms, and path
Does not (yet) navigate to alternate path

Added fields to FieldDefinition.java, to support ontology concepts
Added support in DomainFieldRow, to interact with ontology field properties UI when defining a domain

### DIFF
--- a/src/org/labkey/test/util/search/SearchAdminAPIHelper.java
+++ b/src/org/labkey/test/util/search/SearchAdminAPIHelper.java
@@ -62,7 +62,7 @@ public abstract class SearchAdminAPIHelper
         // Invoke a special server action that waits until all previous indexer tasks are complete, even wait for background indexing tasks to complete (e.g. deleteContainer)
         var cmd = new PostCommand("search", "waitForIndexer");
         cmd.setTimeout(timeout);
-        cmd.setParameters(Map.of("priority","background"));
+        cmd.setParameters(new HashMap<>(Map.of("priority", "background")));
 
         executeWaitForIndexer(cmd);
     }
--- a/src/org/labkey/test/util/search/SearchAdminAPIHelper.java
+++ b/src/org/labkey/test/util/search/SearchAdminAPIHelper.java
@@ -16,6 +16,7 @@
 package org.labkey.test.util.search;
 
 import org.apache.http.HttpStatus;
+import org.labkey.remoteapi.PostCommand;
 import org.labkey.test.WebTestHelper;
 import org.labkey.test.util.LogMethod;
 import org.labkey.test.util.LoggedParam;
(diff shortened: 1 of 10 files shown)


src/org/labkey/test/util/search/SearchAdminAPIHelper.java
```diff
@@ -28,26 +29,54 @@
...
+
+    @LogMethod(quiet = true)
+    public static void waitForIndexerBackground(int timeout)
     {
         // Invoke a special server action that waits until all previous indexer tasks are complete, even wait for background indexing tasks to complete (e.g. deleteContainer)
-        int response = WebTestHelper.getHttpResponse(WebTestHelper.buildURL("search", "waitForIndexer", Map.of("priority","background"))).getResponseCode();
-        assertEquals("WaitForIndexer action timed out", HttpStatus.SC_OK, response);
+        var cmd = new PostCommand("search", "waitForIndexer");
+        cmd.setTimeout(timeout);
+        cmd.setParameters(Map.of("priority","background"));
```
labkey-tchad: Commands get crabby sometimes if you pass an immutable Map to `setParameters`. You need to pass in a mutable collection.
```suggestion
        cmd.setParameters(new HashMap<>(Map.of("priority", "background")));
```

src/org/labkey/test/util/search/SearchAdminAPIHelper.java
```diff
@@ -28,26 +29,54 @@
...
+    private static void executeWaitForIndexer(PostCommand cmd)
+    {
+        try
+        {
+            var response = cmd.execute(WebTestHelper.getRemoteApiConnection(), null);
+            assertEquals("WaitForIndexer action timed out", HttpStatus.SC_OK, response.getStatusCode());
+        } catch (Exception cmdException)
+        {
+            throw new RuntimeException("an error occurred while waiting for search indexing to finish", cmdException);
+        }
```
labkey-tchad, line 81: I wouldn't bother breaking `executeWaitForIndexer` out to a separate method. It's short and only used in one place.

labkey-tchad (changes requested):

src/org/labkey/test/util/search/SearchAdminAPIHelper.java
labkey-tchad: Sorry. You had it right the first time. I misread the code. Didn't realize that method was getting called twice.

labkey-chrisj: hahahahahahaa ok I'll put it back

labkey-tchad (approved):

PR #715, opened 2021-05-06
Automation test for Sample type support for Link to Study.

labkey-sweta:
#### Rationale
Automation test for Sample type support for Link to Study.

#### Related Pull Requests
* 

#### Changes
New test class src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java

### DIFF
--- a/src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
+++ b/src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
@@ -93,7 +93,6 @@ public class SampleTypeLinkToStudyTest extends BaseWebDriverTest
                 .setFields(List.of(
                         new FieldDefinition("VisitDate", FieldDefinition.ColumnType.VisitDate),
                         new FieldDefinition("ParticipantId", FieldDefinition.ColumnType.Subject))), data2);
-
     }
 
     @Test
@@ -172,7 +171,6 @@ public class SampleTypeLinkToStudyTest extends BaseWebDriverTest
         for (int i = 0; i < numOfRowsToBeLinked; i++)
             samplesTable.checkCheckbox(i);
 
-
         samplesTable.clickHeaderButtonAndWait("Link to Study");
 
         log("Link to study: Choose target");
@@ -208,7 +206,6 @@ public class SampleTypeLinkToStudyTest extends BaseWebDriverTest
             table.checkCheckbox(i);
         table.clickHeaderButton("Recall");
         acceptAlert();
-
     }
 
     @Override
--- a/src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
+++ b/src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
@@ -10,26 +10,28 @@ import org.labkey.test.Locator;
 import org.labkey.test.TestTimeoutException;
 import org.labkey.test.categories.DailyC;
 import org.labkey.test.params.FieldDefinition;
+import org.labkey.test.params.experiment.SampleTypeDefinition;
 import org.labkey.test.util.DataRegionTable;
 import org.labkey.test.util.PortalHelper;
-import org.labkey.test.util.TestDataGenerator;
+import org.labkey.test.util.SampleTypeHelper;
 
 import java.io.IOException;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
 import java.util.Arrays;
 import java.util.List;
-import java.util.Map;
-
-import static org.labkey.test.util.exp.SampleTypeAPIHelper.SAMPLE_TYPE_DOMAIN_KIND;
 
 @Category({DailyC.class})
 @BaseWebDriverTest.ClassTimeout(minutes = 2)
 public class SampleTypeLinkToStudyTest extends BaseWebDriverTest
 {
     final static String SAMPLE_TYPE_PROJECT = "Sample Type Test Project";
-    final static String LINKED_STUDY_PROJECT = "Linked Study Test Project";
-
+    final static String VISIT_BASED_STUDY = "Visit Based Study Test Project";
+    final static String DATE_BASED_STUDY = "Date Based Study Test Project";
     final static String SAMPLE_TYPE1 = "Sample type 1";
     final static String SAMPLE_TYPE2 = "Sample type 2";
+
+    protected DateTimeFormatter _dateTimeFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
 
     @BeforeClass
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
```diff
@@ -0,0 +1,194 @@
...
+        log("Linking sample types one row to study");
+        linkToStudy(SAMPLE_TYPE1, 2, Arrays.asList("P1", "P2"), Arrays.asList("1", "2"));
+
+        log("Verifying the linked sample type in study");
+        goToProjectHome(LINKED_STUDY_PROJECT);
+        clickAndWait(Locator.linkWithText(SAMPLE_TYPE1));
+        DataRegionTable table = new DataRegionTable("Dataset", getDriver());
+
+        checker().verifyEquals("Incorrect number of rows linked", 2, table.getDataRowCount());
+        checker().verifyEquals("Incorrect Participant ID's", Arrays.asList("P1", "P2"), table.getColumnDataAsText("ParticipantId"));
```
labkey-danield: Does order matter here? I'm not sure if checker().verifyEquals will pass if the order is different. if not you may want to sort the list () before passing in.

```
List<String> actualIds = table.getColumnDataAsText("ParticipantId"));
List<String> expectedIds = Arrays.asList("P1", "P2");
Collections.sort(actualIds);
Collections.sort(expectedIds);
checker().verifyEquals("Incorrect Participant ID's", expectedIds, actualIds);
```

Also I'm not sure you need both checks, number of row and content of column. If the number of rows is different than expected, then the content of the column will not be as expected either.

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
```diff
@@ -0,0 +1,194 @@
...
+            samplesTable.checkCheckbox(i);
+        }
+
+        samplesTable.clickHeaderButtonAndWait("Link to Study");
+
+        log("Link to study: Choose target");
+        selectOptionByText(Locator.id("targetStudy"), "/" + LINKED_STUDY_PROJECT + " (" + LINKED_STUDY_PROJECT + " Study)");
+        clickButton("Next");
+
+        log("Lick to Study: Verify Results");
```
labkey-danield: I think you want to "Click to study". :)

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
```diff
@@ -0,0 +1,194 @@
...
+
+    }
+
+    @Test
+    public void testLinkToStudy()
+    {
+        goToProjectHome(SAMPLE_TYPE_PROJECT);
+
+        log("Linking sample types one row to study");
+        linkToStudy(SAMPLE_TYPE1, 2, Arrays.asList("P1", "P2"), Arrays.asList("1", "2"));
```
labkey-klum: We should also test setting up the source samples with fields of types : subject & visit (ID or date) as well. I think it's fine to manually specify the subject/timepoint info in the confirm view, but we also want to verify that we can use the new datatypes to identify this information.

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
```diff
@@ -0,0 +1,194 @@
...
+
+        log("Verifying the linked sample type in study");
+        goToProjectHome(LINKED_STUDY_PROJECT);
+        clickAndWait(Locator.linkWithText(SAMPLE_TYPE1));
+        DataRegionTable table = new DataRegionTable("Dataset", getDriver());
+
+        checker().verifyEquals("Incorrect number of rows linked", 2, table.getDataRowCount());
+        checker().verifyEquals("Incorrect Participant ID's", Arrays.asList("P1", "P2"), table.getColumnDataAsText("ParticipantId"));
+
+        log("Verifying log entries");
```
labkey-klum, line 114: It might also be beneficial to ensure that we can link the same sample to multiple studies and verify that the links are rendered correctly.

labkey-klum (changes requested):
looks mostly good, with the exception of a few scenarios we can improve.

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
```diff
@@ -0,0 +1,194 @@
...
+        goToProjectHome(SAMPLE_TYPE_PROJECT);
+        clickAndWait(Locator.linkWithText(SAMPLE_TYPE2));
+        String expectedComment = "1 row(s) were recalled from a study to the sample type: " + SAMPLE_TYPE2;
+        verifyLinkToHistory(expectedComment);
+        verifyAuditLogEvents(expectedComment);
+    }
+
+    private void linkToStudy(String sampleName, int numOfRowsToBeLinked, List<String> participantID, List<String> visitId)
+    {
```
labkey-danield, line 168: Not a big deal but it looks like numOfRowsToBeLinked is equal to participantID.size() or VisitId.size(), so you don't technically need that parameter.

You could also (if you wanted) merge participantID and VisitId into a Map<String, String> where the participantID is the key. It doesn't do a lot, just an alternative way to implement the function.

I would suggest more consistent variable naming: participant**ID** vs visit**Id**.

labkey-danield (approved):
Just some superficial comments. I will let Rosaline comment on the scope of the test cases.

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
labkey-sweta: Yes offcourse

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
labkey-sweta: Done

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
labkey-sweta: Order is always the same. Thank you

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
labkey-sweta: Changed the method signature itself. Thank you

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
labkey-sweta: Done

labkey-klum (approved):

RosalineP (approved):

PR #716, opened 2021-05-06
Update SiteWrapper to be aware of LKSM App page.

labkey-danield:
#### Rationale
In order to run the LKSM automation against a trial instance the siteWrapper needs to be able to identify when a login has succeeded. The LKSM App page does not have the traditional LKS elements so added a check to for the user menu that is specific to the App UI. Hopefully it is general enough that if other apps are created that share the same header menu they will just work when running against a trial instance.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/564

#### Changes
* Update LabKeySiteWrapper.simpleSignIn to look for user menu in LKSM App.

### DIFF
--- a/src/org/labkey/test/LabKeySiteWrapper.java
+++ b/src/org/labkey/test/LabKeySiteWrapper.java
@@ -48,6 +48,7 @@ import org.labkey.remoteapi.query.SelectRowsResponse;
 import org.labkey.test.components.core.ProjectMenu;
 import org.labkey.test.components.dumbster.EmailRecordTable;
 import org.labkey.test.components.html.SiteNavBar;
+import org.labkey.test.components.ui.navigation.UserMenu;
 import org.labkey.test.pages.core.admin.CustomizeSitePage;
 import org.labkey.test.pages.core.admin.ShowAdminPage;
 import org.labkey.test.pages.user.UserDetailsPage;
@@ -133,7 +134,7 @@ public abstract class LabKeySiteWrapper extends WebDriverWrapper
             executeScript("window.onbeforeunload = null;"); // Just get logged in, ignore 'unload' alerts
             beginAt(WebTestHelper.buildURL("login", "login"));
             waitForAnyElement("Should be on login or Home portal", Locator.id("email"), SiteNavBar.Locators.userMenu,
-                    Locator.tagWithId("a", "user-menu-dropdown"));
+                    UserMenu.AppUserMenu());
         }
 
         if (PasswordUtil.getUsername().equals(getCurrentUser()))
@@ -156,7 +157,7 @@ public abstract class LabKeySiteWrapper extends WebDriverWrapper
                 if (isElementPresent(SiteNavBar.Locators.userMenu))
                     return true;
 
-                if(isElementPresent(Locator.tagWithId("a", "user-menu-dropdown")))
+                if(isElementPresent(UserMenu.AppUserMenu()))
                 {
                     goToHome();
                     return true;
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/LabKeySiteWrapper.java
```diff
@@ -132,7 +132,8 @@ public void simpleSignIn()
         {
             executeScript("window.onbeforeunload = null;"); // Just get logged in, ignore 'unload' alerts
             beginAt(WebTestHelper.buildURL("login", "login"));
-            waitForAnyElement("Should be on login or Home portal", Locator.id("email"), SiteNavBar.Locators.userMenu);
+            waitForAnyElement("Should be on login or Home portal", Locator.id("email"), SiteNavBar.Locators.userMenu,
+                    Locator.tagWithId("a", "user-menu-dropdown"));
```
labkey-tchad: Add a static `getLocator` method to `org.labkey.test.components.ui.navigation.UserMenu` rather than having this "user-menu-dropdown" id defined in several places.

src/org/labkey/test/LabKeySiteWrapper.java
```diff
@@ -154,6 +155,13 @@ public void simpleSignIn()
             {
                 if (isElementPresent(SiteNavBar.Locators.userMenu))
                     return true;
+
+                if(isElementPresent(Locator.tagWithId("a", "user-menu-dropdown")))
+                {
+                    goToHome();
+                    return true;
+                }
+
```
labkey-tchad, line 161: This should make sure the standard userMenu is present after navigating. Either remove the return line and move this `if` to the start of the `waitFor`; or have it assert that the standard userMenu is present after navigating home.

labkey-danield: Fixed.

src/org/labkey/test/LabKeySiteWrapper.java
labkey-danield: Fixed, but not sure I fixed it the way you suggested. I would like to review this with you @labkey-tchad to make sure the fix is correct.

src/org/labkey/test/components/ui/navigation/UserMenu.java
```diff
@@ -45,4 +45,14 @@ protected Locator getToggleLocator()
     {
         return Locator.id(MENU_ID);
     }
+
+    /**
+     * The button for the drop down user menu. The user menu on the SampleManager app is an example of this button.
+     *
+     * @return A locator to the drop down user menu button.
+     */
+    public static Locator AppUserMenu()
```
labkey-tchad: Java method names shouldn't be capitalized.

src/org/labkey/test/components/ui/navigation/UserMenu.java
```diff
@@ -45,4 +45,14 @@ protected Locator getToggleLocator()
...
+    {
+        return Locator.tagWithId("a", "user-menu-dropdown");
```
labkey-tchad: Use the `MENU_ID` defined in this class.
```suggestion
        return Locator.tagWithId("a", MENU_ID);
```

Also, have `getToggleLocator` call and return this method.

labkey-tchad (approved):
A couple of minor changes.

src/org/labkey/test/components/ui/navigation/UserMenu.java
labkey-danield: Yep, linter said the same thing. Fixed.

src/org/labkey/test/components/ui/navigation/UserMenu.java
labkey-danield: Updated.

PR #726, opened 2021-05-18
TabbedGridPanel

labkey-nicka:
#### Rationale
Adds a component wrapper for `TabbedGridPanel` component that is a part of our `@labkey/components` package. Done in collaboration with @labkey-chrisj.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/528
* https://github.com/LabKey/biologics/pull/885
* https://github.com/LabKey/sampleManagement/pull/572
* https://github.com/LabKey/testAutomation/pull/726

#### Changes
* Adds a `TabbedGridPanel` component wrapper for interacting with the related `@labkey/component` React component on app pages.

### DIFF
--- a/src/org/labkey/test/components/ui/grids/TabbedGridPanel.java
+++ b/src/org/labkey/test/components/ui/grids/TabbedGridPanel.java
@@ -1,0 +1,131 @@
+package org.labkey.test.components.ui.grids;
+
+import org.labkey.test.Locator;
+import org.labkey.test.components.Component;
+import org.labkey.test.components.WebDriverComponent;
+import org.labkey.test.components.html.Input;
+import org.labkey.test.pages.LabKeyPage;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.ui.ExpectedConditions;
+
+import java.util.List;
+
+import static org.labkey.test.WebDriverWrapper.WAIT_FOR_JAVASCRIPT;
+import static org.labkey.test.components.html.Input.Input;
+
+/**
+ * TabbedGridPanel wraps components/src/public/QueryModel/TabbedGridPanel.tsx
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/components/ui/grids/TabbedGridPanel.java
```diff
@@ -0,0 +1,126 @@
...
+            getWrapper().waitFor(()-> isSelected(tabText), "tab did not become selected in time", 2000);
+        }
+
+        return getSelectedGrid();
+    }
+
+    public QueryGrid getSelectedGrid()
+    {
+        return new QueryGrid.QueryGridFinder(getDriver()).waitFor(this);
+    }
```
labkey-tchad, line 70: If the class documentation is accurate and the grids are only shown and hidden when switching tabs, I would expect this to always return the first grid regardless of which tab is selected. I don't see anything in `QueryGrid.QueryGridFinder` that would restrict this to finding only the visible grid.

labkey-tchad (changes requested):
Can we add one of these to `LabKeyUIComponentsPage.tsx` with a corresponding component test and update to `CoreComponentsTestPage`?

labkey-nicka:
> Can we add one of these to `LabKeyUIComponentsPage.tsx` with a corresponding component test and update to `CoreComponentsTestPage`?

We can, however, I'd prefer to do that with a future PR. This panel is now in use in the related PRs and I can get something added to the backlog to support putting this into the components page.

src/org/labkey/test/components/ui/grids/TabbedGridPanel.java
labkey-chrisj: It turns out that the comment isn't accurate, the grid component switching appears to be handled as virtual-dom actions- meaning there's only one grid at a time

labkey-tchad (approved):

PR #725, opened 2021-05-18
DomainDesignerTest - fix for FieldValidator.toJSONObject() to include "expression"

cnathe:
#### Rationale
A few DomainDesignerTest cases started failing after this PR was merged: https://github.com/LabKey/testAutomation/pull/714.
There was a necessary change in that PR so that the FieldDefinition.java test class would properly call the toJSONObject method. This PR fixes up the FieldValidator.toJSONObject part of that class to include the required "expression" property.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/714

#### Changes
* Update for FieldValidator.toJSONObject() to include "expression"

PR #727, opened 2021-05-18
Fix errors when defining domains with lookups over API

labkey-tchad:
#### Rationale
Until recently, `FieldDefinition.toJSONObject()` was getting inadvertently bypassed. Now that it is getting invoked as designed, some tests are revealing that the current implementation has some issues. Overriding `getAllProperties` interferes with the functionality of PropertyDescriptor less than overriding `toJSONObject`.

#### Related Pull Requests
* #714 

#### Changes
* Refactor `FieldDefinition` to override `getAllProperties` instead of `toJSONObject`

### DIFF
--- a/src/org/labkey/test/params/FieldDefinition.java
+++ b/src/org/labkey/test/params/FieldDefinition.java
@@ -22,33 +22,26 @@ import org.labkey.remoteapi.domain.PropertyDescriptor;
 import org.labkey.remoteapi.query.Filter;
 import org.labkey.test.components.html.OptionSelect;
 
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 public class FieldDefinition extends PropertyDescriptor
 {
     // for UI helpers
     private ColumnType _type;
     private LookupInfo _lookup;
-
-    // Field properties not supported by PropertyDescriptor
-    private Integer _scale;
-    private Boolean _shownInDetailsView;
-    private Boolean _shownInInsertView;
-    private Boolean _shownInUpdateView;
-    private Boolean _isPrimaryKey;
-    private Boolean _lookupValidatorEnabled;
-    private String _url;
-    private List<FieldValidator<?>> _validators;
-    private String _importAliases;
-    private String _sourceOntology;
-    private String _conceptLabelColumn;
-    private String _conceptImportColumn;
-    private String _principalConceptCode;
     private String _principalConceptSearchExpression;
+
+    // Stash validator collection to avoid having to convert back from JSON Maps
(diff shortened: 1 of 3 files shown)


labkey-chrisj:
If this hasn't been mentioned for you yet, the BVT failures in DomainDesignerTest were addressed in this PR: https://github.com/LabKey/testAutomation/pull/725

It turns out that the expression part of a range validator is required, but we weren't writing that into JSON

src/org/labkey/test/params/FieldDefinition.java
```diff
@@ -195,147 +170,152 @@ public PropertyDescriptor setLookup(String schema, String query, String containe
...
     {
-        super.setHidden(hidden);
+        setFieldProperty("scale", scale);
         return this;
     }
 
     public Boolean isPrimaryKey()
     {
-        return  _isPrimaryKey;
+        return (Boolean) getFieldProperty("isPrimaryKey");
```
labkey-danield, line 219: I'm sure this works, but just curious if it would be better/safer to do `Boolean.valueOf(getFieldProperty("isPrimaryKey"))`?

src/org/labkey/test/util/ListHelper.java
```diff
@@ -507,14 +507,20 @@ public RangeValidator(String name, String description, String message, RangeType
     @Deprecated
     public static class ListColumn extends FieldDefinition
     {
-        public ListColumn(String name, String label, ListColumnType type, String description, String format, LookupInfo lookup, FieldValidator validator, String url, Integer scale)
+        public ListColumn(String name, String label, ListColumnType type, String description, String format, LookupInfo lookup, FieldValidator<?> validator, String url, Integer scale)
```
labkey-danield, line 510: I realize we may not have the time right now, but since this is deprecated, shouldn't we stop fixing it and move existing code to it's replacement?

labkey-danield (approved):
This looks ok to me, but I noticed there are two BVT failures. Are they related to this change?

src/org/labkey/test/util/ListHelper.java
labkey-tchad: We definitely should. I started pulling that thread but quickly realized it would take over the PR.

labkey-chrisj (approved):
Thanks for doing this; these changes address questions I've had for a while but have been reticent to touch.

src/org/labkey/test/params/FieldDefinition.java
labkey-tchad: We have full control of what goes into the property Map. Somebody would have to go out of their way to make this not work.

PR #732, opened 2021-05-21
ontology picker test components update

labkey-chrisj:
#### Rationale
Updates to components and page classes to support more granular search testing and selection when multiple ontologies are present

#### Related Pull Requests
https://github.com/LabKey/ontology/pull/42

#### Changes

- Add ontology select to Concept Picker dialog to enable testing to cover scenarios in which there are multiple ontologies to choose from
- surface search component to allow test to do more granular search behavior, add better synchronization to search behavior
- enable test code to get available options from domainFieldRow's ontology select

### DIFF
--- a/src/org/labkey/test/components/domain/DomainFieldRow.java
+++ b/src/org/labkey/test/components/domain/DomainFieldRow.java
@@ -506,6 +506,10 @@ public class DomainFieldRow extends WebDriverComponent<DomainFieldRow.ElementCac
         return this;
     }
 
+    /**
+     * Allows test code to get which ontologies are available for selection
+     * @return a list of the full text shown in the options
+     */
     public List<String> getConceptImportSelectOptions()
     {
         expand();
--- a/src/org/labkey/test/components/domain/DomainFieldRow.java
+++ b/src/org/labkey/test/components/domain/DomainFieldRow.java
@@ -9,6 +9,7 @@ import org.labkey.test.components.html.Checkbox;
 import org.labkey.test.components.html.Input;
 import org.labkey.test.components.html.RadioButton;
 import org.labkey.test.components.html.SelectWrapper;
+import org.labkey.test.components.react.ReactSelect;
 import org.labkey.test.components.ui.ontology.ConceptPickerDialog;
 import org.labkey.test.params.FieldDefinition;
 import org.labkey.test.util.LabKeyExpectedConditions;
@@ -503,6 +504,12 @@ public class DomainFieldRow extends WebDriverComponent<DomainFieldRow.ElementCac
         expand();
         elementCache().getConceptImportFieldSelect().selectByVisibleText(importField);
         return this;
+    }
+
+    public List<String> getConceptImportSelectOptions()
+    {
+        expand();
+        return getWrapper().getTexts(elementCache().getOntologySelect().getOptions());
     }
 
     public DomainFieldRow setConceptLabelField(String labelField)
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/components/domain/DomainFieldRow.java
```diff
@@ -505,6 +506,12 @@ public DomainFieldRow setConceptImportField(String importField)
         return this;
     }
 
+    public List<String> getConceptImportSelectOptions()
```
labkey-chrisj, line 513: This change enables the test to get which ontologies are available/selectable in the field row

src/org/labkey/test/components/ui/ontology/ConceptInfoTabs.java
```diff
@@ -71,6 +72,8 @@ public ConceptInfoTabs showPathInformation()
     public String getTitle()
```
labkey-chrisj, line 77: If no concept is selected, the 'none selected' element will be present.  surfacing its text in `getTitle()` in this case makes sense, I think (especially because it handles the case where the title element isn't present)

src/org/labkey/test/components/ui/ontology/ConceptInfoTabs.java
```diff
@@ -123,6 +126,10 @@ protected ElementCache newElementCache()
             return Locator.tagWithClass("ul", "synonyms-text").child(Locator.tag("li"))
                     .findElements(overviewPane);
         }
+        Optional<WebElement> noneSelectedElement()
```
labkey-chrisj, line 136: This element will be present when no concept is selected

src/org/labkey/test/components/ui/ontology/OntologyTreeSearch.java
```diff
@@ -55,12 +58,76 @@ public OntologyTreeSearch selectItemWithLabel(String searchExpression, String la
...
+        var input = elementCache().searchInput;
+        String placeholder = input.getComponentElement().getAttribute("placeholder");
+        input.set("");
+        // wait for the results container to collapse, and for the placeholder for the input to be shown
+        getWrapper().waitFor(()-> !isResultContainerExpanded() &&
+                Locator.css("input:placeholder-shown").withAttribute("placeholder", placeholder)
+                        .existsIn(this),  2000);
+        return this;
+    }
+
```
labkey-chrisj, line 103: Because I wanted to write tests that would search on a repeated basis, I needed a way to wait for search results to appear before returning them.  This also called for a way to clear those results, so I could begin from a state in which there were no searchResults to find.

src/org/labkey/test/components/ui/ontology/OntologyTreeSearch.java
```diff
@@ -55,12 +58,76 @@ public OntologyTreeSearch selectItemWithLabel(String searchExpression, String la
...
+    /**
+     * when more search hits exist than are shown (max of 20 can be shown), an added search result footer element
+     * will appear with text advising the user to refine their search
+     * @return
+     */
+    Optional<WebElement> searchResultFooterElement()
+    {
+        return Locator.tagWithClass("div", "result-footer").findOptionalElement(elementCache().resultContainer());
+    }
+
```
labkey-chrisj, line 125: The search result footer appears only when there are more search 'hits' than can be shown.  (the control will show a maximum of 20, and then it will show this footer element to advise the user to refine their search).

labkey-tchad (approved):
Looks good but most of the GitHub comments you added should be added to JavaDoc in the code itself.

labkey-chrisj:
Thanks!  I'll go through these and add javadocs accordingly.

PR #740, opened 2021-05-27
43216: Display loading errors for faceted filter pane

labkey-nicka:
#### Rationale
Streamlines `DataRegionTable` handling due to no longer having a need for column filtering to know if the column is PHI or not. Last usages of some deprecated filter methods were removed in the compliance module resulting in a cleanup that could be done on aisle `setFilter()`.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2286
* https://github.com/LabKey/testAutomation/pull/740
* https://github.com/LabKey/compliance/pull/115
* https://github.com/LabKey/tnprc_ehr/pull/538

#### Changes
* Remove `isPHILoggingColumn` bit from `DataRegionTable.setFilters()` methods. Remove `@deprecated` variants that are no longer in use.


labkey-chrisj (approved):
this looks good

PR #745, opened 2021-06-01
Exclude temporary "specimen2" actions from crawler

labkey-adam:
#### Rationale
We're in the process of moving actions from the "specimen" controller in the study module to the "specimen2" controller in the specimen module. For now, exclude both sets.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2305

### DIFF
--- a/src/org/labkey/test/util/Crawler.java
+++ b/src/org/labkey/test/util/Crawler.java
@@ -226,22 +226,31 @@ public class Crawler
             new ControllerActionId("specimen", "downloadSpecimenList"),
             new ControllerActionId("specimen", "emailLabSpecimenLists"),
             new ControllerActionId("specimen", "getSpecimenExcel"),
+
+            // We're in the process of moving actions from the "specimen" controller in the study module to the
+            // "specimen2" controller in the specimen module. For now, exclude both sets. After migration is complete,
+            // eliminate the "specimen2" variants.
+            new ControllerActionId("specimen2", "download"),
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/util/Crawler.java
```diff
@@ -226,22 +226,31 @@ public Crawler(BaseWebDriverTest test, Collection<String> projects, Duration cra
...
             new ControllerActionId("specimen", "emailLabSpecimenLists"),
             new ControllerActionId("specimen", "getSpecimenExcel"),
+
+            // We're in the process of moving actions from the "specimen" controller in the study module to the
+            // "specimen2" controller in the specimen module. For now, exclude both sets. After migration is complete,
+            // eliminate the "specimen2" variants.
+            new ControllerActionId("specimen2", "download"),
+            new ControllerActionId("specimen2", "downloadSpecimenList"),
+            new ControllerActionId("specimen2", "emailLabSpecimenLists"),
+            new ControllerActionId("specimen2", "getSpecimenExcel"),
```
labkey-tchad, line 236: If you want to skip the "specimen2" actions entirely, you could add it to `getExcludedControllers`.

labkey-tchad (approved):

src/org/labkey/test/util/Crawler.java
labkey-adam: We want most of them to be tested; just excluding the same problematic ones as they shift to "specimen2" and then back to "specimen"

PR #749, opened 2021-06-02
Issue 42661: Make crawler handle file downloads

labkey-tchad:
#### Rationale
[Issue 42661: Make crawler handle file downloads, remove export actions from exclude list](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42661)

#### Changes
* Add `beginAt` that conditionally allows downloads
* Add `doAndMaybeWaitForPageToLoad` to conditionally wait for page load
* Stop excluding download actions from crawler

### DIFF
--- a/src/org/labkey/test/util/Crawler.java
+++ b/src/org/labkey/test/util/Crawler.java
@@ -156,8 +156,6 @@ public class Crawler
             new ControllerActionId("admin", "deleteFolder"),
             new ControllerActionId("admin", "doCheck"),
             new ControllerActionId("admin", "dumpHeap"),
-            new ControllerActionId("admin", "exportQueries"), // download action
-            new ControllerActionId("admin", "getSchemaXmlDoc"), // download action
             new ControllerActionId("admin", "mapNetworkDrive"), // 404 on non-Windows
             new ControllerActionId("admin", "memTracker"),
             new ControllerActionId("admin", "queryStackTraces"),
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/util/Crawler.java
```diff
@@ -168,99 +166,32 @@ public Crawler(BaseWebDriverTest test, Collection<String> projects, Duration cra
...
-            new ControllerActionId("targetedms", "showRetentionTimesChart"),
-
 
             // Disable crawler for single-page apps until we make `beginAt` work with them
             new ControllerActionId("biologics", "app"),
             new ControllerActionId("cds", "app"),
             new ControllerActionId("samplemanager", "app"),
 
-            // Actions that error with no parameters. Generally linked from admin-spider.view
-            new ControllerActionId("user", "changeEmail"), // NotFoundException from changeEmail.jsp
```
labkey-jeckels, line 262: Should this be removed? Sounds like it's trying to avoid a 404 error

labkey-jeckels (approved):

labkey-jeckels:
I'm not all that savvy about the best ways to wait for elements, so if you know more @labkey-sweta that would be great to review. But overall this is great!

@labkey-tchad do we need to worry about cleaning up the downloaded files? Maybe I missed code that deletes them

src/org/labkey/test/util/Crawler.java
labkey-tchad: The crawler has been improved since that exclusion was added. It now knows to ignore 404s from links found on 'admin-spider' (#70).

labkey-sweta (approved):

PR #750, opened 2021-06-03
Update DetailDataPanel and DetailTable to work with aliquots.

labkey-danield:
#### Rationale
The DetailDataPanel and DetailTable are used in the overview page of samples. If the sample is an aliquot the DetailDataPanel will have more than one DetailTable on it.

#### Related Pull Requests
* https://github.com/LabKey/biologics/pull/901
* https://github.com/LabKey/sampleManagement/pull/586

#### Changes
* Fixed a bug in the DetailTable where getting a field value was not scoped to look only at the table (it was looking at the larger container).
* Changed DetailDataPanel to look for multiple panels.
* Added methods to get the DetailTables that show up for aliquots.

### Details
If a sample is not an aliquot, which will be most samples, the **DetailDataPanel** looks like this:
Screen Shot 2021-06-02 at 6 19 39 PM

The method `DetailDataPanel.getTable` returns the DetailTable that is shown.

If the sample is an aliquot the DetailDataPanel will look like this:
Screen Shot 2021-06-02 at 6 19 14 PM

There are three tables shown. The top panel is aliquot data. The middle table, which is two lines, is meta data for the aliquot. The bottom table is sample data which is as before.

This update adds methods to get the tables that are shown for aliquots.

The **DetailTable** also had to be updated to account for the scenario where a field is a very simple `td` and does not have attributes to identify it.

The aliquot tables basically have this html:
`<table class="table table-responsive table-condensed detail-component--table__fixed sample-aliquots-details-meta-table"><body><tr><td>Sample description</td><td><span class="detail-display">Description for sample STA01_1</span></td></tr></tbody></table>`

This change has been run against LKB, ELN, LKB-Recipe, LKSM & LKFM.

This differs from what was previously expected for a detail which assumes the td containing the value would have attributes `data-caption` or `data-fieldkey` like this:
`<td data-caption="Str" data-fieldkey="str"><span class="detail-display">This is sample 1.</span></td>`

The update to the DetailTable makes the getValue method more flexible so it can get the value for a field that does not have these attributes.

### DIFF
--- a/src/org/labkey/test/components/ui/grids/DetailDataPanel.java
+++ b/src/org/labkey/test/components/ui/grids/DetailDataPanel.java
@@ -85,7 +85,8 @@ public class DetailDataPanel extends WebDriverComponent<DetailDataPanel.ElementC
     }
 
     /**
-     * If this sample is an aliquot there will be two tables in the panel, the first table will contain the aliquot information.
+     * If this sample is an aliquot there will be multiple tables in the panel, the first table will contain the aliquot
+     * information (under the 'Aliquot Data' header). This will return that table.
      *
      * @return A {@link DetailTable}.
      */
@@ -98,6 +99,24 @@ public class DetailDataPanel extends WebDriverComponent<DetailDataPanel.ElementC
 
         // As of release 21.05 if this sample is an aliquot, then the table with the aliquot information is the first table in the panel.
         return tables.get(0);
+    }
+
+    /**
+     * If this sample is an aliquot there will be multiple tables in the panel, the second table is under the 'Original
+     * Sample Data' header, but is a separate table from the other original data displayed and may only be one or two
+     * lines. The fields 'Original sample' and 'Sample description' will show up here.
+     *
+     * @return A {@link DetailTable}.
+     */
+    public DetailTable getAliquotDetailMetaTable()
+    {
+        List<DetailTable> tables = elementCache().detailTables();
+
+        if(tables.size() == 1)
+            throw new RuntimeException("There does not appear to be an 'Aliquot (details meta)' table in this panel.");
+
+        // As of release 21.05 if this sample is an aliquot, then the table with the aliquot information is the first table in the panel.
+        return tables.get(1);
     }
 
     @Override
--- a/src/org/labkey/test/components/ui/grids/DetailDataPanel.java
+++ b/src/org/labkey/test/components/ui/grids/DetailDataPanel.java
@@ -6,6 +6,7 @@ import org.labkey.test.components.WebDriverComponent;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 
+import java.util.List;
 import java.util.Optional;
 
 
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/components/ui/grids/DetailDataPanel.java
```diff
@@ -64,9 +71,53 @@ public DetailTableEdit clickEdit()
...
+     * information (under the 'Aliquot Data' header). This will return that table.
+     *
+     * @return A {@link DetailTable}.
+     */
+    public DetailTable getAliquotTable()
+    {
+        List<DetailTable> tables = elementCache().detailTables();
+
+        if(tables.size() == 1)
+            throw new RuntimeException("There does not appear to be an 'Aliquot' table in this panel.");
```
labkey-tchad: Consider throwing `NoSuchElementException` like you did with `DetailTable.getField`
```suggestion
            throw new NoSuchElementException("There does not appear to be an 'Aliquot' table in this panel.");
```

src/org/labkey/test/components/ui/grids/DetailDataPanel.java
```diff
@@ -64,9 +71,53 @@ public DetailTableEdit clickEdit()
...
+     * lines. The fields 'Original sample' and 'Sample description' will show up here.
+     *
+     * @return A {@link DetailTable}.
+     */
+    public DetailTable getAliquotDetailMetaTable()
+    {
+        List<DetailTable> tables = elementCache().detailTables();
+
+        if(tables.size() == 1)
+            throw new RuntimeException("There does not appear to be an 'Aliquot (details meta)' table in this panel.");
```
labkey-tchad: Consider
```suggestion
            throw new NoSuchElementException("There does not appear to be an 'Aliquot (details meta)' table in this panel.");
```

src/org/labkey/test/components/ui/grids/DetailTable.java
```diff
@@ -120,17 +147,33 @@ public void clickField(String fieldCaption)
...
+        public final WebElement dataCaptionField(String caption)
+        {
+            return Locator.tagWithAttribute("td", "data-caption", caption).findWhenNeeded(this);
+        }
+
+        // Some tables will show a value in a td with no attributes, use the td that has the text (label) to find the value.
+        public final WebElement siblingField(String caption)
+        {
+            return Locator.tagContainingText("td", caption).followingSibling("td").findWhenNeeded(this);
+        }
```
labkey-tchad, line 173: Consider unioning these Locators if there's no need to distinguish between the two.
```suggestion
        public final WebElement dataCaptionField(String caption)
        {
            return Locator.XPathLocator.union(
                Locator.tagWithAttribute("td", "data-caption", caption),
                Locator.tagContainingText("td", caption).followingSibling("td")) // Some tables will show a value in a td with no attributes, use the td that has the text (label) to find the value.
            .findElement(this);
        }
```
Also, is there some distinguishing CSS class or attribute that we could add to the sibling Locator? (e.g. `Locator.tagContainingText("td", caption).followingSibling("td").withAttribute("data-caption")`)

labkey-tchad (approved):
Just some minor suggestions

src/org/labkey/test/components/ui/grids/DetailDataPanel.java
labkey-danield: Changed.

src/org/labkey/test/components/ui/grids/DetailDataPanel.java
labkey-danield: Changed.

src/org/labkey/test/components/ui/grids/DetailTable.java
labkey-danield: Unfortunately there is not a consistent attribute in the sibling locator. It can range form what you have mentioned (data-caption) to "detail-display" to a td with no attributes.

PR #754, opened 2021-06-03
Item 8855: Use user authentication attributes as RStudio pro use name - automated tests

XingY:
#### Rationale
In order for LDAP test utils to be accessible to RStudio test suite, we will move those util from ldap module to testAutomation

#### Related Pull Requests
* https://github.com/LabKey/rstudio/pull/88
* https://github.com/LabKey/ldap/pull/50
* https://github.com/LabKey/testAutomation/pull/754
* 
#### Changes
*

### DIFF
--- a/src/org/labkey/test/components/ui/OmniBox.java
+++ b/src/org/labkey/test/components/ui/OmniBox.java
@@ -222,13 +222,9 @@ public class OmniBox extends WebDriverComponent<OmniBox.ElementCache>
 
     private OmniBox setText(String inputValue)
     {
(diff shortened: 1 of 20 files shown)


labkey-tchad (approved):

labkey-tchad (changes requested):
Looks like the package paths got messed up in the move.
The package structure should be the same as it was in the ldap module. There is an extra `pages/ldap/` in the new classes.

labkey-tchad (approved):

PR #756, opened 2021-06-05
Item #8853: Adjust Concept dialog utilities with new feature work

Sigmonia:
#### Rationale
Concept dialog test utilities updates to support Item #8853: Ontology, Form insert and updates

#### Related Pull Requests
* https://github.com/LabKey/ontology/pull/43

#### Changes
* Look for Ontology selector for concept code columns
* adjusted waiting and dialogs ready checks

### DIFF
--- a/src/org/labkey/test/components/ui/ontology/ConceptPickerDialog.java
+++ b/src/org/labkey/test/components/ui/ontology/ConceptPickerDialog.java
@@ -1,5 +1,6 @@
 package org.labkey.test.components.ui.ontology;
 
+import org.labkey.test.Locator;
 import org.labkey.test.components.bootstrap.ModalDialog;
 import org.labkey.test.components.react.ReactSelect;
 import org.openqa.selenium.WebDriver;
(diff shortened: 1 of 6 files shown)


src/org/labkey/test/components/ui/ontology/ConceptPickerDialog.java
```diff
@@ -116,6 +115,12 @@ public ConceptPickerDialog selectOntology(String ontology)
         return select.getOptions();
     }
 
+    public ConceptPickerDialog waitForActiveTreeNode()
```
labkey-chrisj, line 119: getVisibleActiveNode() should already wait for the node, if all you need is the wait/sync

src/org/labkey/test/components/ui/ontology/OntologyTreePanel.java
```diff
@@ -71,6 +79,8 @@ protected ElementCache newElementCache()
     protected class ElementCache extends Component<?>.ElementCache
     {
         WebElement nodeContainer = Locator.tag("ul").parent().findWhenNeeded(this);
+        Locator activeNode = Locator.tagWithClass("div", "filetree-node-active");
```
labkey-chrisj, line 82: does this mean the finder for TreeNodeFinder.activeOnly() is looking for a different locator now?
To this point, I've been using TreeNodeFinder to build node locators

src/org/labkey/test/params/FieldDefinition.java
```diff
@@ -313,13 +314,19 @@ public FieldDefinition setPrincipalConceptCode(String principalConceptCode)
...
 
+    public String getPrincipalConceptSearchSourceOntology()
+    {
+        return _principalConceptSearchSourceOntology;
+    }
+
     public String getPrincipalConceptSearchExpression()
     {
         return _principalConceptSearchExpression;
     }
```
labkey-chrisj, line 326: I think it will be helpful to note that the searchExpression is to support the UI case/not be passed in as json to the remoteAPI.  We only use the searchExpression to give the UI the ability to obtain the conceptCode, when we're using the API we give it the conceptCode directly

labkey-chrisj (approved):

PR #761, opened 2021-06-10
Fb session logout blur test

labkey-chrisj:
#### Rationale
This adds a modal dialog test wrapper for the modal dialog shown when your session has been closed (whether by API logout, or having signed out in a different browser/tab).

#### Related Pull Requests
https://github.com/LabKey/inventory/pull/255 <- tests for FM compliance
https://github.com/LabKey/compliance/pull/117 <- tests for LKS compliance
https://github.com/LabKey/sampleManagement/pull/595 <- tests for SM compliance

#### Changes
The modal dialog wrapper has a check for the body/header/footer elements we expect to have a blurred style (or not) depending on how the session settings are set on the complianceSettings page

[spec here](https://docs.google.com/document/d/1BXAifNJdfBhsP6zg2Xv5pbnUXi3ptiv1rX1o6A5e4HY/edit#)


src/org/labkey/test/components/bootstrap/LoggedOutDialog.java
```diff
@@ -0,0 +1,98 @@
...
+
+        // if any of these bodyParts are present and don't have the lk-content-blur style applied, return false
+        // otherwise, true
+        for(Boolean fact : factors)
+        {
+            if(!fact)
+                return false;
+        }
+        return true;
+    }
```
labkey-danield, line 85: Interesting approach to put this check with dialog, it does make it easy to find.

Is there a way to check that this is the top level dialog/component? The scenario: a page has a dialog open then gets the timeout / logged out prompt. This blurred should be rendered on top of that dialog.

labkey-chrisj: I haven't come up with a universal way of identifying whatever the top-level component is.
The top of the DOM hierarchy (the 'body' element under /html) will typically get a style to note that it is under a modal dialog when it is, but I haven't seen this happen when the top element is something like a browser alert or a non-bootstrap dialog.

I've focused on the lk-header-ct/lk-body-ct/footer elements because they are the ones we style behind the dialog, not as a guarantee future labkey won't surface stuff that isn't contained in them

src/org/labkey/test/components/bootstrap/LoggedOutDialog.java
```diff
@@ -0,0 +1,126 @@
...
+    public WebElement getComponentElement()
+    {
+        return _el;
+    }
+
+    @Override
+    public WebDriver getDriver()
+    {
+        return _driver;
+    }
```
labkey-tchad: I think these `getComponentElement` and `getDriver` overrides are unnecessary. They just do the same thing as the methods in `ModalDialog`.

src/org/labkey/test/components/bootstrap/LoggedOutDialog.java
```diff
@@ -0,0 +1,126 @@
...
+        }
+        return true;
+    }
+
+
+    protected ElementCache elementCache()
+    {
+        return new ElementCache();
+    }
+
```
labkey-tchad: Overriding `elementCache` like this clobbers any actual caching we might do. The correct pattern for when there isn't a type parameter for the element cache would be:
```suggestion
    protected ElementCache elementCache()
    {
        return (ElementCache) super.elementCache();
    }

    protected ElementCache newElementCache()
    {
        return new ElementCache();
    }
```

src/org/labkey/test/components/bootstrap/LoggedOutDialog.java
```diff
@@ -0,0 +1,126 @@
...
+            factors.add(lkHeader.get().getAttribute("class").contains("lk-content-blur"));
+
+        var lkBody = elementCache().lkBody.findOptionalElement(getDriver());
+        if (lkBody.isPresent())
+            factors.add(lkBody.get().getAttribute("class").contains("lk-content-blur"));
+
+        var footerBlock = elementCache().footerBlock.findOptionalElement(getDriver());
+        if (footerBlock.isPresent())
+            factors.add(footerBlock.get().getAttribute("class").contains("lk-content-blur"));
+
```
labkey-tchad: These checks are not reliable; if the header/body/footer Locators become invalid we would never know whether they were being properly blurred. I'd say to have this take a list of Locators for things that should be blurred and have this method be strict about the elements being present. Also, why not return immediately as soon as we find a non-blurred element?
```suggestion
    public boolean isContentBlurred(Locator... shouldBeBlurred)
    {
        for (Locator loc : shouldBeBlurred)
        {
            if (!loc.findElement(getDriver()).getAttribute("class").contains("lk-content-blur"))
                return false;
        }
        return true;
```

labkey-tchad (changes requested):

src/org/labkey/test/components/bootstrap/LoggedOutDialog.java
labkey-chrisj: Header and footer aren't present in react Apps (like SM, FM) so this logic wouldn't work for this feature in them.
I could write separate methods for checking in-app vs. in LKS that could be strict in this way
Or I could write a method that requires them if we're in an app
...I consolidated the code to work in these cases after my first approach (verify different situations separately) was rejected in favor of not duplicating code, so at this point you people are hard to please hahaha

src/org/labkey/test/components/bootstrap/LoggedOutDialog.java
labkey-chrisj: gone

src/org/labkey/test/components/bootstrap/LoggedOutDialog.java
labkey-chrisj: done

src/org/labkey/test/components/bootstrap/LoggedOutDialog.java
labkey-chrisj: I've implemented the changes we discussed offline; now the helper takes a list of Locator of elements to check for styling and will fail if they aren't present.  I've updated the tests to specify the top-level elements expected to be present (for the app, or for LKS) and added an additional check to make sure we know it if these top-level element classes are changed out from under us

labkey-tchad (approved):

src/org/labkey/test/components/bootstrap/LoggedOutDialog.java
```diff
@@ -0,0 +1,118 @@
...
+    {
+        // if any of these bodyParts are present and don't have the lk-content-blur style applied, return false
+        // otherwise, true
+        for(Locator loc : elementsToBeBlurred)
+        {
+            if(!loc.findElement(getDriver()).getAttribute("class").contains("lk-content-blur"))
+                return false;
+        }
+        return true;
+    }
```
labkey-tchad, line 85: Since we check the blurred an unblurred conditions, I suggest we check all of the elements and have it throw an error if there is a mix of blurriness.

PR #763, opened 2021-06-14
Item 9012: Update application navigation menu behavior

labkey-susanh:
#### Rationale
We are simplifying the product/application navigation menu to better suit the majority use case for this menu.  In particular, we are removing the option to navigate to products in a different container from this menu, with the only exception being navigating to the Home folder.  

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2343
* https://github.com/LabKey/biologics/pull/909
* https://github.com/LabKey/sampleManagement/pull/594
* https://github.com/LabKey/labkey-ui-components/pull/555

#### Changes
* Remove project navigation component
* Update locators
* Add locator for new link in application menu footer
* Update LookAndFeelSettings page for new setting related to application menu.


src/org/labkey/test/components/ui/navigation/apps/LeafNavContainer.java
```diff
@@ -33,14 +34,30 @@ public void clickItem(String itemText)
...
+            while(currentUrl.equals(getDriver().getCurrentUrl()) && (tries < 10))
+            {
+                wait(wait/10);
+                tries++;
+            }
+        }
+        catch (InterruptedException e)
+        {
+            throw new NotFoundException("Navigation to " + itemText + " was interrupted.");
+        }
```
labkey-danield: Would this be cleaner?
```
        String currentUrl = getDriver().getCurrentUrl();
        WebDriverWrapper.waitFor(()->!currentUrl.equals(getDriver().getCurrentUrl()), "Did not navigate, or some such thing.", 1_000);

```

src/org/labkey/test/components/ui/navigation/apps/ProductListItem.java
```diff
@@ -44,6 +44,11 @@ public void clickNavIcon()
         elementCache().navIcon.click();
     }
 
+    public boolean isEnabled()
+    {
+        return !"labkey-page-nav-disabled".equals(_el.getAttribute("class"));
```
labkey-danield: Do we have to worry about the class attribute having anything more than "labkey-page-nav-disabled"? If so might change this to:
`return !_el.getAttribute("class").contains("labkey-page-nav-disabled");`

labkey-danield (approved):
I only had a couple of comments, @labkey-chrisj might have more input since he is the developer of the nav library code.

src/org/labkey/test/components/ui/navigation/apps/LeafNavContainer.java
labkey-susanh: Yes, much better.  Thanks.

src/org/labkey/test/components/ui/navigation/apps/ProductListItem.java
labkey-susanh: Currently, equals works, but checking contains is probably better for future-proofing.

PR #768, opened 2021-06-18
Test coverage for issue42857

labkey-chrisj:
#### Rationale
One of our clients wants to bulk-import samples using a nameExpression that can resolve existing values via a lookup to another domain.  This test attempts to do that, and in the attempt appears to have found an [adjacent issue ](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43394)

The test creates an integer-keyed list called Colors, with string columns ColorName and Code
values for ColorName and Code look like 'red' and 'rd', respectively.

It then creates a sampleType with a column named 'Color' that looks up the 'Colors' list (in the same project), and gives the sampleType a nameExpression with value 'TEST-${Color/Code}.
Via the bulk insert UI, it attempts to paste in a new Sample with value in the 'Color' column matching a value in the list's 'Code' column.

#### Related Pull Requests
n/a

#### Changes
just the new test case

### DIFF
--- a/src/org/labkey/test/params/FieldDefinition.java
+++ b/src/org/labkey/test/params/FieldDefinition.java
@@ -112,13 +112,6 @@ public class FieldDefinition extends PropertyDescriptor
     }
 
     @Override
-    public FieldDefinition setRangeURI(String rangeURI)
-    {
-        super.setRangeURI(rangeURI);
-        return this;
-    }
-
-    @Override
     public FieldDefinition setMvEnabled(Boolean mvEnabled)
     {
         super.setMvEnabled(mvEnabled);
--- a/src/org/labkey/test/params/FieldDefinition.java
+++ b/src/org/labkey/test/params/FieldDefinition.java
@@ -112,6 +112,13 @@ public class FieldDefinition extends PropertyDescriptor
     }
 
     @Override
+    public FieldDefinition setRangeURI(String rangeURI)
+    {
+        super.setRangeURI(rangeURI);
+        return this;
+    }
+
+    @Override
     public FieldDefinition setMvEnabled(Boolean mvEnabled)
     {
         super.setMvEnabled(mvEnabled);
--- a/src/org/labkey/test/tests/SampleTypeNameExpressionTest.java
+++ b/src/org/labkey/test/tests/SampleTypeNameExpressionTest.java
@@ -21,12 +21,16 @@ import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.labkey.test.BaseWebDriverTest;
+import org.labkey.test.Locator;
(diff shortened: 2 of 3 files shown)


src/org/labkey/test/tests/SampleTypeNameExpressionTest.java
```diff
@@ -130,6 +135,54 @@ public void testParentAliasExpression()
...
+        String pasteData = "ColorLookup\tNoun\n" +
+                "red\tryder\n" +
+                "green\tgiant\n" +
+                "blue\tangel\n" +
+                "yellow\tjersey";
+
+        // now create a sampleType with a Color column that looks up to Colors
+        var sampleTypeDef = new SampleTypeDefinition(nameExpSamples)
+                .setFields(List.of(new FieldDefinition("ColorLookup", colorsLookup).setRangeURI(FieldDefinition.ColumnType.Integer.getRangeURI()),
+                        new FieldDefinition("Noun", FieldDefinition.ColumnType.String)))
```
labkey-tchad: The correct pattern to set the type of a lookup is to call `setTableType`. This `setRangeURI` isn't necessary.

src/org/labkey/test/tests/SampleTypeNameExpressionTest.java
```diff
@@ -130,6 +135,54 @@ public void testParentAliasExpression()
                 "Parent", "Jessi");
     }
 
+    // Issue 42857: samples: bulk import with name expression containing lookup fails to convert lookup string value
+    @Test
+    public void testLookupNameExpression() throws Exception
+    {
+        String lookupList = "Colors";
+        FieldDefinition.LookupInfo colorsLookup = new FieldDefinition.LookupInfo(getProjectName(), "lists", lookupList);
```
labkey-tchad:
```suggestion
        FieldDefinition.LookupInfo colorsLookup = new FieldDefinition.LookupInfo(getProjectName(), "lists", lookupList).setTableType(FieldDefinition.ColumnType.Integer);
```

src/org/labkey/test/params/FieldDefinition.java
```diff
@@ -111,6 +111,13 @@ public FieldDefinition setFormat(String format)
         return this;
     }
 
+    @Override
+    public FieldDefinition setRangeURI(String rangeURI)
+    {
+        super.setRangeURI(rangeURI);
+        return this;
+    }
+
```
labkey-tchad: Don't need this now, right?
```suggestion
```

labkey-tchad (changes requested):

src/org/labkey/test/params/FieldDefinition.java
labkey-chrisj: Let's hope you're right

labkey-tchad (approved):

PR #773, opened 2021-06-21
Reset header and footer after tests

labkey-tchad:
#### Rationale
Some tests expect default header and footer settings.

#### Related Pull Requests
* https://github.com/LabKey/compliance/pull/117

#### Changes
* Call `premium-adminConsoleConfigurePageElements` from test scrubber


labkey-chrisj (approved):
Thanks for doing this!

PR #779, opened 2021-06-23
Automation test

labkey-sweta:
#### Rationale
1. Final Polish for Study - QC state and Folder import/export.
2. Publish History support for Sample Types - audit log verification

#### Related Pull Requests
* 

#### Changes
1. src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java - New test cases
2. src/org/labkey/test/pages/admin/ExportFolderPage.java - setting sample dataset data and definition.

### DIFF
--- a/src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
+++ b/src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
@@ -1,6 +1,7 @@
 package org.labkey.test.tests;
 
 import org.jetbrains.annotations.Nullable;
+import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
```diff
@@ -357,6 +356,106 @@ public void testAssaySupport()
         new DataRegionTable("Data", getDriver()).clickHeaderButtonAndWait("Link to Study");
     }
 
+    @Test
+    public void testQCState()
```
labkey-klum, line 360: I'm curious why this test is here? It doesn't seem to have any relationship to sample types or link to study (other than it's a sample type dataset). There are other QC state related tests in `StudyTest.class`. Also, is there not an existing test that accomplishes the same scenario?

labkey-klum (approved):

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
labkey-sweta: This was the part i tested as Final Polish for Study.

RosalineP: Perhaps the test could be pared down to retain the part where it checks the `Data linked to this study` option in Manage Dataset QC States is in fact set to the state indicated in the dropdown. The other part, changing the QC state of one row, could probably be omitted.

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
```diff
@@ -357,6 +356,106 @@ public void testAssaySupport()
...
+        selectOptionByText(Locator.name("newState"), "Approved");
+        setFormElement(Locator.name("comments"), "Approved");
+        clickButton("Update Status");
+
+        new BootstrapMenu.BootstrapMenuFinder(getDriver()).withButtonTextContaining("QC State").find().clickSubMenu(true, "All data");
+        checker().verifyEquals("Incorrect value for QC state after update", Arrays.asList("Approved?", "Pending Review?"), table.getColumnDataAsText("QCState"));
+    }
+
+    @Test
+    public void testFolderImportExport()
```
RosalineP, line 397: Looks good 👍

RosalineP (approved):

PR #782, opened 2021-06-30
Automation test coverage for issue 42937 

labkey-sweta:
#### Rationale
Automation test coverage for issue 42937 

#### Related Pull Requests
https://github.com/LabKey/platform/pull/2412

#### Changes
Added test case "testLinkedColumnDisabledCase" in src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java

### DIFF
--- a/src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
+++ b/src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
@@ -446,6 +446,45 @@ public class SampleTypeLinkToStudyTest extends BaseWebDriverTest
         checker().verifyEquals("No data should have been imported", 0, table.getDataRowCount());
     }
 
+    /*
+        Test coverage for : Issue 42937: Assay results grid loading performance can degrade with a large number of "copied to study" columns
+     */
+    @Test
+    public void testLinkedColumnDisabledCase()
+    {
+        log("Creating 2 more studies");
+        _containerHelper.createProject("Study 1", "Study");
+        _studyHelper.startCreateStudy().createStudy();
+
+        _containerHelper.createProject("Study 2", "Study");
+        _studyHelper.startCreateStudy().createStudy();
+
+        log("Linking one row from sample type to all the studies");
+        goToProjectHome(SAMPLE_TYPE_PROJECT);
+        linkToStudy(DATE_BASED_STUDY, SAMPLE_TYPE1, 1);
+
+        goToProjectHome(SAMPLE_TYPE_PROJECT);
+        linkToStudy(VISIT_BASED_STUDY, SAMPLE_TYPE1, 1);
+
+        goToProjectHome(SAMPLE_TYPE_PROJECT);
+        linkToStudy("Study 1", SAMPLE_TYPE1, 1);
+
+        goToProjectHome(SAMPLE_TYPE_PROJECT);
+        linkToStudy("Study 2", SAMPLE_TYPE1, 1);
+
+        goToProjectHome(SAMPLE_TYPE_PROJECT);
+        clickAndWait(Locator.linkWithText(SAMPLE_TYPE1));
+        DataRegionTable samplesTable = DataRegionTable.DataRegion(getDriver()).withName("Material").waitFor();
+        checker().verifyFalse("Linked column for Visit based study should not be present",
+                samplesTable.getColumnNames().contains("linked_to_Visit_Based_Study_Test_Project_Study"));
+        checker().verifyFalse("Missing linked column for Date based study",
+                samplesTable.getColumnNames().contains("linked_to_Date_Based_Study_Test_Project_Study"));
+        checker().verifyFalse("Missing linked column for Date based study",
+                samplesTable.getColumnNames().contains("linked_to_Study_1_Study"));
+        checker().verifyFalse("Missing linked column for Date based study",
+                samplesTable.getColumnNames().contains("linked_to_Study_2_Study"));
+    }
+
     @Before
     public void preTest() throws Exception
     {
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
```diff
@@ -446,6 +448,46 @@ public void testFolderImportExport()
...
+        Test coverage for : Issue 42937: Assay results grid loading performance can degrade with a large number of "copied to study" columns
+     */
+    @Test
+    public void testLinkedColumnDisabledCase()
+    {
+        log("Creating 2 more studies");
+        _containerHelper.createProject("Study 1", "Study");
+        _studyHelper.startCreateStudy().createStudy();
+
+        _containerHelper.createProject("Study 2", "Study");
```
cnathe: similar comment here about using more distinct project names for the studies or using subfolders of the test project

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
```diff
@@ -446,6 +448,46 @@ public void testFolderImportExport()
         checker().verifyEquals("No data should have been imported", 0, table.getDataRowCount());
     }
 
+    /*
+        Test coverage for : Issue 42937: Assay results grid loading performance can degrade with a large number of "copied to study" columns
+     */
+    @Test
+    public void testLinkedColumnDisabledCase()
```
cnathe: this test case name is a little confusing to me, what is it that is "disabled"? the other PR uses the name `testLinkedColumnNotDisplayedCase`, that one makes sense to me.

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
```diff
@@ -446,6 +448,46 @@ public void testFolderImportExport()
...
+
+        log("Linking one row from sample type to all the studies");
+        goToProjectHome(SAMPLE_TYPE_PROJECT);
+        linkToStudy(DATE_BASED_STUDY, SAMPLE_TYPE1, 1);
+
+        goToProjectHome(SAMPLE_TYPE_PROJECT);
+        linkToStudy(VISIT_BASED_STUDY, SAMPLE_TYPE1, 1);
+
+        goToProjectHome(SAMPLE_TYPE_PROJECT);
+        linkToStudy("Study 1", SAMPLE_TYPE1, 1);
```
cnathe: I didn't mention this for the other PR, but it might be good to do a quick check of the samples table grid default view after this 3rd linking to make sure that these first three columns ARE added by default before linking the 4th.

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
```diff
@@ -446,6 +448,46 @@ public void testFolderImportExport()
...
+        clickAndWait(Locator.linkWithText(SAMPLE_TYPE1));
+        DataRegionTable samplesTable = DataRegionTable.DataRegion(getDriver()).withName("Material").waitFor();
+        checker().verifyFalse("Linked column for Visit based study should not be present",
+                samplesTable.getColumnNames().contains("linked_to_Visit_Based_Study_Test_Project_Study"));
+        checker().verifyFalse("Linked column for Date based study should not be present",
+                samplesTable.getColumnNames().contains("linked_to_Date_Based_Study_Test_Project_Study"));
+        checker().verifyFalse("Linked column for Study 1 should not be present",
+                samplesTable.getColumnNames().contains("linked_to_Study_1_Study"));
+        checker().verifyFalse("Linked column for Study 2 should not be present",
+                samplesTable.getColumnNames().contains("linked_to_Study_2_Study"));
```
cnathe: as mentioned in the other PR, could add a check after these to make sure the 4 linked study columns can be added to via customize view.

cnathe (commented):
@labkey-sweta similar comments as with other PR, these changes look good just a few additional suggestions

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
labkey-sweta: testSampleTypeLinkedToMultipleStudy test in the same class does that check for two studies, U want additional verification for 3 studies?

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
labkey-sweta: ok will change, i was confused for name while writing the test.

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
labkey-sweta: Good point.

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
labkey-sweta: Does the behavior change for subfolder ? I am curious about the distinct name, any particular reason?

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
cnathe: Ah ok great. I wasn't aware of that other test case verifying that scenario. Thanks. No need for adding it here.

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
cnathe: Nope, the behavior won't change for projects vs subfolders. I was just suggesting subfolders as a way to put the study folders under the current test project so that it is one less project to need to cleanup on test completion.

labkey-sweta: Changed the name by prefixing the project name.

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
labkey-sweta: done

cnathe (approved):

PR #785, opened 2021-07-01
Fix for Test Automation Issue 41038

labkey-danield:
#### Rationale
Fix for Test Automation Issue 41038 (Add a waitForPageLoad in org.labkey.test.pages.experiment.UpdateSampleTypePage). There are random failures in LKS because the page is not completely loaded before the test uses it. Added a simple check in the constructor that should help with this.

#### Related Pull Requests
* None.

#### Changes
* Add a check in the constructor for the page being loaded to some degree.


labkey-sweta (approved):
Should this be going to 21.7?

PR #788, opened 2021-07-06
Issue 43516: "Resolve lookup by value" experimental feature flag does not work for the "Derive Samples" insert form

cnathe:
#### Rationale
See related PR for rationale. This PR fixes up an automated test which was setting/unsetting an experimental feature flag which is now being removed.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2429

#### Changes
* Remove usages of the "resolve-lookups-by-value" experimental feature flag 
* Add test case for updating a row and resolving the lookup input by display value and key value
* Add test case for Derive Samples form to resolve lookup input by display value and key value


src/org/labkey/test/tests/SampleTypeLimitsTest.java
```diff
@@ -67,83 +71,155 @@ private void doSetup()
...
+        setFormElement(Locator.name("quf_label"), rowValue);
+        clickButton("Submit");
+    }
+
+    @Test
+    public void testDeriveSamplesLookupFields() throws IOException, CommandException
+    {
+        goToProjectHome();
+
+        log("Create sample type with lookup field to SAMPLE_TYPE_NAME");
```
labkey-danield: I think rather than use SAMPLE_TYPE_NAME you may want to say: 
`log("Create sample type with lookup field to " + SAMPLE_TYPE_NAME);`
or something similar that gets the value of SAMPLE_TYPE_NAME.

labkey-danield (approved):
Just one trivial comment.

src/org/labkey/test/tests/SampleTypeLimitsTest.java
cnathe: Yup, good catch. Fixed.

PR #791, opened 2021-07-08
Link crawler improvements

labkey-tchad:
#### Rationale
Recent crawler improvements to allow it to crawl download URLs causes occasional false-positives. After crawling a URL, the crawler checks the destination for errors. The crawler also attempts script injections against the pages it crawls but is less strict about finding errors on such pages. When the crawler checks a download URL immediately after hitting an error page from an injection attempt, it will fail because it still performs the error checks that it does after crawling a non-download URL. For a similar reason, we are not attempting script injection against download URLs.
This change skips response checks after crawling a download URL and continues to attempt script injection against them.
There is also a new test to ensure the injection checker continues to function as well as a new test module to facilitate this test.

#### Related Pull Requests
* #749 

#### Changes
* Fix crawler to not inspect the browser after hitting a download URL
* [Issue 39581: Add test to ensure injection check continues to work](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=39581)

### DIFF
--- a/modules/crawlerTest/src/org/labkey/crawlertest/CrawlerTestController.java
+++ b/modules/crawlerTest/src/org/labkey/crawlertest/CrawlerTestController.java
@@ -16,14 +16,12 @@
 
 package org.labkey.crawlertest;
 
-import org.labkey.api.action.LabKeyError;
 import org.labkey.api.action.SimpleErrorView;
 import org.labkey.api.action.SimpleViewAction;
 import org.labkey.api.action.SpringActionController;
-import org.labkey.api.module.ModuleLoader;
 import org.labkey.api.security.RequiresPermission;
 import org.labkey.api.security.permissions.ReadPermission;
-import org.labkey.api.util.ErrorView;
+import org.labkey.api.settings.AppProps;
 import org.labkey.api.view.JspView;
 import org.labkey.api.view.NavTree;
 import org.springframework.validation.BindException;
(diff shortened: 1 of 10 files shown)


modules/crawlerTest/src/org/labkey/crawlertest/CrawlerTestController.java
```diff
@@ -0,0 +1,78 @@
...
+        setActionResolver(_actionResolver);
+    }
+
+    @RequiresPermission(ReadPermission.class)
+    public class InjectJspAction extends SimpleViewAction<InjectForm>
+    {
+        @Override
+        public ModelAndView getView(InjectForm form, BindException errors)
+        {
+            if (!getViewContext().getContainer().getActiveModules().contains(ModuleLoader.getInstance().getModule(CrawlerTestModule.class)))
```
labkey-jeckels: I like trying to be conservative here. Perhaps even safer and easier for the test code would be to have a short allowable list of script injection parameter values that the test uses. Everything else could get ignored, and we wouldn't need to worry about the module's enabled status and would have zero worries about it causing problems on a production server.

labkey-jeckels (approved):
An optional suggestion to make the test action even safer from leaking into the real world.

modules/crawlerTest/src/org/labkey/crawlertest/CrawlerTestController.java
labkey-tchad: I updated the action to require dev mode and to only respond to a specific test user. I'd like to be able to experiment with other injection strings without having to rebuild the module.

labkey-jeckels: OK, sounds good. Restricting by user should make it impossible to "exploit" even if this module somehow got deployed to a production server.

PR #794, opened 2021-07-09
Remove domain editor cancel workaround

labkey-tchad:
#### Rationale
Field selection state for export was included in the domain editor dirty state check. That behavior seems to have been fixed.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2428

#### Changes
* Update `FieldEditorRowSelectionActionTest.testUserProperties` to not expect a confirmation dialog.


cnathe (approved):
Cool. Thanks. Yes, I did fix that field editor issue where the row selection was inadvertantely changing the page dirty state with my set of changes from last week.

PR #807, opened 2021-07-19
Update ReactSelect

labkey-nicka:
#### Rationale
Test component and test updates associated with the update of our `react-select` dependency in `@labkey/components`.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/584
* https://github.com/LabKey/biologics/pull/945
* https://github.com/LabKey/sampleManagement/pull/632
* https://github.com/LabKey/inventory/pull/280
* https://github.com/LabKey/testAutomation/pull/807
* https://github.com/LabKey/platform/pull/2473
* https://github.com/LabKey/commonAssays/pull/351

#### Changes
* BaseReactSelect
	* Update all locators to match the DOM structure of `react-select@4.3.1`.
	* Introduce new locators `withContainerClass` and `withInputClass` to allow for easy association with configuration of the component in React.
	* Add a `createValue()` method which will attempt to create new values in the associated `react-select`. This supports when the scenarios where the component is set to `creatable={true}` (e.g. the "Alias" field in Biologics).
	* Update `getSelections()` to only return selections if they are actually made. The presence of placeholder text is no longer considered a selection/value.
	* Added `getPlaceholderText()` and `isPlaceholderVisible()` utilities to assist with processing the placeholder.
* FilteringReactSelect:
	* Update `FilteringReactSelect.typeAheadSelect` to no longer accept arbitrary elements but rather enforce selection based on the component's expected state.
	* Remove superfluous sleeps/waits statements.
* Update test locators and tests.

### DIFF
--- a/src/org/labkey/test/components/react/BaseReactSelect.java
+++ b/src/org/labkey/test/components/react/BaseReactSelect.java
@@ -367,7 +367,7 @@ public abstract class BaseReactSelect<T extends BaseReactSelect> extends WebDriv
         elementCache().input.sendKeys(value);
         elementCache().input.sendKeys(Keys.ENTER);
 
-        waitFor(() -> Locators.multiValueLabels.withText(value).existsIn(getComponentElement()), "Failed to create value \"" + value + "\".", 1_000);
+        waitFor(() -> getValueLabelLocator().withText(value).existsIn(getComponentElement()), "Failed to create value \"" + value + "\".", 1_000);
 
         return (T) this;
     }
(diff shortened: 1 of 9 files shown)


src/org/labkey/test/components/react/BaseReactSelect.java
```diff
@@ -60,89 +64,118 @@ public boolean isExpanded()
...
         catch (NoSuchElementException nse)
         {
             return false;
         }
-        return  foundElement != null;
+        return foundElement != null;
     }
 
-    /* waits until the currently selected 'value' (which can include the placeholder) equals or contains the specified string */
+    /* waits until the currently selected 'value' equals or contains the specified string */
```
labkey-chrisj, line 166: good call, differentiating placeholder from value

src/org/labkey/test/components/react/BaseReactSelect.java
```diff
@@ -339,85 +394,64 @@ protected ElementCache newElementCache()
...
         WebElement findOption(String option)
         {
-            Locator loc = Locators.options.withText(option);
-            return loc.findElement(selectMenu);
+            return Locators.options.withText(option).findElement(selectMenu);
         }
     }
 
     public static abstract class Locators
     {
```
labkey-chrisj, line 394: as an informational note, we intend to discontinue the pattern of having static Locators classes in our components- where it's practicable, it's probably best to put locators in the appropriate elementCache or make them class members if there's a need to reference them on a static basis

src/org/labkey/test/components/ui/entities/ParentEntityEditPanel.java
```diff
@@ -169,76 +156,39 @@ public boolean isAddButtonEnabled()
...
-    {
-        return getAllTypeCombo().get(index);
     }
 
     /**
      * Get the type combos by its label. This is the label to the left of the control.
      *
      * @return A combo beside a label with the given text.
      */
-    protected ReactSelect getTypeCombo(String labelText)
```
labkey-chrisj, line 214: If there's a way to preserve the ability to specify which one by label text, that would be preferable to using index (which is more fragile and obtuse) if that's the tradeoff being made.  If we're using index, usually it's because other (more deterministic) means of identifying the target weren't clear or available

labkey-chrisj (approved):

PR #812, opened 2021-07-24
Test components for Bio/SM Job overview updates

labkey-chrisj:
#### Rationale
This adds new components in testAutomation to support new UI in the Workflow Overview updates

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/939
https://github.com/LabKey/sampleManagement/pull/627

#### Changes
added wrappers for file attachment and inline edit (we will want to use these wrappers in other places)


labkey-chrisj:
(I initially created the branch with the wrong name, this is its replacement)
Refactoring tests and page classes to align with new product-side changes showed that we've got several components in use on the product/client side for which there aren't test wrappers.

labkey-tchad (approved):

PR #813, opened 2021-07-26
Response Server UI

RosalineP:
#### Rationale
Test changes supporting consolidating settings and removing dependency on Folder Management Module Properties feature, improving usability by offering instead a more clear form with added validation and guidance.

#### Related Pull Requests
* https://github.com/FDA-MyStudies/Response/pull/28
* https://github.com/LabKey/platform/pull/2492

#### Changes
* Add helper to reach Response Server Admin UI page
* Add helper for setting Response Server Configurations

### DIFF
--- a/src/org/labkey/test/BaseWebDriverTest.java
+++ b/src/org/labkey/test/BaseWebDriverTest.java
@@ -103,7 +103,6 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
--- a/src/org/labkey/test/BaseWebDriverTest.java
+++ b/src/org/labkey/test/BaseWebDriverTest.java
@@ -1573,7 +1573,7 @@ public abstract class BaseWebDriverTest extends LabKeySiteWrapper implements Cle
             else
                 setFormElement(Locator.name(prop), val);
         }
-        clickButton("Save and Finish", 0);
+        clickButton("Save and Finish");
 
         assertTextNotPresent("WCP Base URL must begin with 'http://' or 'https://'");
         assertTextNotPresent("WCP Base URL must end with '/StudyMetaData'");
--- a/src/org/labkey/test/BaseWebDriverTest.java
+++ b/src/org/labkey/test/BaseWebDriverTest.java
@@ -1574,6 +1574,12 @@ public abstract class BaseWebDriverTest extends LabKeySiteWrapper implements Cle
                 setFormElement(Locator.name(prop), val);
         }
         clickButton("Save and Finish", 0);
+
+        assertTextNotPresent("WCP Base URL must begin with 'http://' or 'https://'");
+        assertTextNotPresent("WCP Base URL must end with '/StudyMetaData'");
+        assertTextNotPresent("WCP username must not be blank");
+        assertTextNotPresent("WCP password must not be blank");
+        assertTextNotPresent("Metadata Directory path is invalid");
     }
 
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/BaseWebDriverTest.java
```diff
@@ -1552,6 +1558,30 @@ public String getModulePropertyValue(ModulePropertyValue property)
         return (String)ref.getValue();
     }
 
+    public void setResponseServerConfigurations(LinkedHashMap<String, String> props)
```
labkey-adam: I'll let @labkey-tchad weigh in on this, but it doesn't seem right that `BaseWebDriverTest` should know anything about the response server config details. It knows about module properties, but that's a very generic mechanism.

src/org/labkey/test/pages/core/admin/ShowAdminPage.java
```diff
@@ -141,6 +141,12 @@ public LookAndFeelSettingsPage clickLookAndFeelSettings()
         return new LookAndFeelSettingsPage(getDriver());
     }
 
+    public void clickResponseServerConfiguration()
```
labkey-adam, line 144: I guess there's precedent for `ShowAdminPage` knowing about all the module-specific links...

src/org/labkey/test/BaseWebDriverTest.java
```diff
@@ -1552,6 +1558,30 @@ public String getModulePropertyValue(ModulePropertyValue property)
...
+            else
+                setFormElement(Locator.name(prop), val);
+        }
+        clickButton("Save and Finish", 0);
+
+        assertTextNotPresent("WCP Base URL must begin with 'http://' or 'https://'");
+        assertTextNotPresent("WCP Base URL must end with '/StudyMetaData'");
+        assertTextNotPresent("WCP username must not be blank");
+        assertTextNotPresent("WCP password must not be blank");
+        assertTextNotPresent("Metadata Directory path is invalid");
```
labkey-tchad: Checking for the non-existence of text is not reliable (somebody might change the wording of an error).
You should be able to do something like `assertElementNotPresent(Locators.labkeyError)` to catch any of these.

src/org/labkey/test/BaseWebDriverTest.java
```diff
@@ -1552,6 +1558,30 @@ public String getModulePropertyValue(ModulePropertyValue property)
...
+        {
+            String val = props.get(prop);
+            log("setting property: " + prop + " to value: " + val);
+
+            if (prop.equals("metadataLoadLocation"))
+                checkRadioButton(Locator.radioButtonByNameAndValue("metadataLoadLocation", val));
+            else
+                setFormElement(Locator.name(prop), val);
+        }
+        clickButton("Save and Finish", 0);
```
labkey-tchad: Does "Save and Finish" not navigate? Failing to wait for page loads that occur is a frequent source of intermittent failures.
```suggestion
        clickButton("Save and Finish");
```

labkey-tchad (changes requested):
Adam is correct. Except for the most core modules; module-specific test functionality should live in that module. I left some feedback here but these changes should go in the Response repository in a helper class or a page class for `ServerConfigurationAction`

src/org/labkey/test/BaseWebDriverTest.java
RosalineP: Ahhh makes sense, done!

labkey-tchad (changes requested):
Adam is correct. Except for the most core modules; module-specific test functionality should live in that module. These changes should go in the Response repository in a helper class or a page class for ServerConfigurationAction.
This branch can be deleted.

RosalineP:
Changes moved to `Response` branch!

PR #828, opened 2021-08-10
Use consistent toggle locator for BootstrapMenu

labkey-tchad:
#### Rationale
'dropdown-toggle' is the standard css class for the 'react-bootstrap' <Dropdown> component menu toggle. `data-toggle="dropdown"` is common but not universal.
One of the early influences on the `BootstrapMenu` test component was data region header menu buttons. They don't have the 'dropdown-toggle' class but they aren't actually 'react-bootstrap' components. A related platform change adds the class to data region buttons.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2532

#### Changes
* Update toggle locator to make setting custom locator unnecessary.

### DIFF
--- a/src/org/labkey/test/components/html/BootstrapMenu.java
+++ b/src/org/labkey/test/components/html/BootstrapMenu.java
@@ -17,6 +17,7 @@ package org.labkey.test.components.html;
 
 import org.labkey.test.Locator;
 import org.labkey.test.WebDriverWrapper;
+import org.labkey.test.components.react.BaseBootstrapMenu;
 import org.labkey.test.util.LogMethod;
 import org.labkey.test.util.LoggedParam;
 import org.openqa.selenium.WebDriver;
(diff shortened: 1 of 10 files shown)


src/org/labkey/test/components/html/BootstrapMenu.java
```diff
@@ -165,9 +155,9 @@ protected WebElement findVisibleMenuItemOrNull(String text)
 
     static public class Locators
     {
-        public static Locator.XPathLocator toggleAnchor()
+        public static Locator.XPathLocator dropdownToggle()
         {
-            return Locator.tagWithAttribute("*", "data-toggle", "dropdown");
+            return Locator.byClass("dropdown-toggle");
         }
```
labkey-danield, line 161: Why not make the locator  a public static member variable and not a function? Just curious.

labkey-danield (approved):

src/org/labkey/test/components/html/BootstrapMenu.java
labkey-tchad: No particular reason. Just trying to minimize the changes but I suppose I'm already changing the method name so it wouldn't be much more of a change to make it a contant.

labkey-chrisj (approved):

PR #831, opened 2021-08-11
Fix intermittent NAb graph check

labkey-tchad:
#### Rationale
The neutralization graph takes a moment to be visible. We get intermittent failures verifying the various graph sizes:
```
java.lang.AssertionError: Graphs aren't the correct size (Small) expected:<300> but was:<0>
  at org.junit.Assert.failNotEquals(Assert.java:835)
  at org.junit.Assert.assertEquals(Assert.java:120)
  at org.labkey.test.tests.nab.NabAssayTest.verifyRunDetails(NabAssayTest.java:783)
  at org.labkey.test.tests.nab.NabAssayTest.runUITests(NabAssayTest.java:345)
```

#### Changes
* Wait for NAb graphs to be visible before checking their size


src/org/labkey/test/tests/nab/NabAssayTest.java
```diff
@@ -803,6 +804,12 @@ private void verifyRunDetails()
         assayHelper.verifyDataIdentifiers(AssayImportOptions.VisitResolverType.SpecimenIDParticipantVisit, "D");
     }
 
+    private WebElement waitForNabGraph()
+    {
+        return shortWait().until(ExpectedConditions
+                .visibilityOfElementLocated(Locator.tagWithAttribute("img", "alt", "Neutralization Graph")));
```
labkey-danield, line 810: I'm guessing that since this is in the NabAssayTest the alt value will always be "Neutralization Graph". If not would it make sense to pass in the value for the alt value as a parameter?

labkey-danield (approved):

PR #835, opened 2021-08-11
Add test coverage for sample and source file field

XingY:
#### Rationale
Pages and utils for handle file field on Details component and query grid.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/656

#### Changes
* added FileUploadField for AttachmentCard
* added ImageFileViewDialog
* support File column for GridRow
* DetailTableEdit and DetailDataPanel to handle files

### DIFF
--- a/src/org/labkey/test/components/ui/files/FileUploadField.java
+++ b/src/org/labkey/test/components/ui/files/FileUploadField.java
@@ -11,6 +11,7 @@ import org.openqa.selenium.WebElement;
 import java.io.File;
 
 // AttachmentCard.tsx
+// TODO: reconcile with AttachmentCard.java
 public class FileUploadField extends WebDriverComponent<FileUploadField.ElementCache>
 {
     final WebElement _el;
--- a/src/org/labkey/test/components/ui/files/FileUploadField.java
+++ b/src/org/labkey/test/components/ui/files/FileUploadField.java
@@ -1,0 +1,187 @@
+package org.labkey.test.components.ui.files;
+
+import org.labkey.test.Locator;
+import org.labkey.test.WebDriverWrapper;
+import org.labkey.test.components.Component;
+import org.labkey.test.components.WebDriverComponent;
+import org.labkey.test.components.react.MultiMenu;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+import java.io.File;
+
+// AttachmentCard.tsx
+public class FileUploadField extends WebDriverComponent<FileUploadField.ElementCache>
+{
+    final WebElement _el;
+    final WebDriver _driver;
+
+    public FileUploadField(WebElement element, WebDriver driver)
+    {
+        _el = element;
+        _driver = driver;
+    }
+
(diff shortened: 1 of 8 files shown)


src/org/labkey/test/components/ui/files/ImageFileViewDialog.java
```diff
@@ -0,0 +1,37 @@
...
+public class ImageFileViewDialog extends ModalDialog
+{
+    public ImageFileViewDialog(WebDriver driver, String filename)
+    {
+        super(new ModalDialogFinder(driver).withTitle(filename).waitFor().getComponentElement(), driver);
+    }
+
+    public boolean isImageRendered()
+    {
+        return elementCache().img.isDisplayed();
```
labkey-danield, line 17: Would there be value in passing in a string value that is the image file name, or some part of it, and then validating that not only is the element displayed but the correct file was used?

src/org/labkey/test/components/ui/files/FileUploadField.java
```diff
@@ -0,0 +1,187 @@
...
+import org.labkey.test.components.Component;
+import org.labkey.test.components.WebDriverComponent;
+import org.labkey.test.components.react.MultiMenu;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+import java.io.File;
+
+// AttachmentCard.tsx
+public class FileUploadField extends WebDriverComponent<FileUploadField.ElementCache>
```
labkey-danield, line 15: It looks like Trey is working on something similar:
https://github.com/LabKey/testAutomation/blob/00b6448b52fbfdeaa649befbbadde1f3d0f9002e/src/org/labkey/test/components/ui/files/AttachmentCard.java

I think it might make sense to include some of his public methods in this as well.

Because of the duplicate effort I'm going to hold off adding comments at this time until the three of us can resolve this.

src/org/labkey/test/components/ui/grids/DetailDataPanel.java
```diff
@@ -121,6 +124,42 @@ public DetailTable getAliquotDetailMetaTable()
...
+    {
+        return getFileField(fieldCaption)
+                .getAttachedFileName();
+    }
+
+    public boolean isFileFieldBlank(String fieldCaption)
+    {
+        return !getFileField(fieldCaption)
+                .hasAttachedFile();
+    }
```
labkey-danield, line 145: Do you need this method? Won't the method above, getFileName(), return an empty string if there is no file? 
I'm not opposed to this method just curious.

labkey-danield (changes requested):
We should reconcile this with Trey's change for the AttachmentCard test component.

labkey-danield (approved):

src/org/labkey/test/components/ui/files/ImageFileViewDialog.java
XingY: I think the constructor would fail if the filename doesn't match
`ModalDialogFinder(driver).withTitle(filename)`

PR #837, opened 2021-08-12
Add test for finding samples by ids.

labkey-susanh:
#### Rationale
We added functionality for finding samples by Ids in all of our premium applications.  Now we are adding tests for that functionality and we'll locate the dialog that is common across application in testAutomation.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2543
* https://github.com/LabKey/labkey-ui-components/pull/606
* https://github.com/LabKey/biologics/pull/965
* https://github.com/LabKey/inventory/pull/291
* https://github.com/LabKey/sampleManagement/pull/659

#### Changes
* add `FindByIdsDialog`
* Update `NavBar` to be able to activate the `FindByIdsDialog`.

### DIFF
--- a/src/org/labkey/test/components/ui/navigation/FindByIdsDialog.java
+++ b/src/org/labkey/test/components/ui/navigation/FindByIdsDialog.java
@@ -1,0 +1,82 @@
+package org.labkey.test.components.ui.navigation;
+
+import org.apache.commons.lang3.StringUtils;
+import org.labkey.test.Locator;
+import org.labkey.test.components.bootstrap.ModalDialog;
+import org.labkey.test.components.html.RadioButton;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+import java.util.List;
+
+public class FindByIdsDialog  extends ModalDialog
+{
+    public static final String TITLE = "Find Samples";
+
+    public FindByIdsDialog(WebDriver driver)
+    {
+        super(new ModalDialog.ModalDialogFinder(driver).withTitle(TITLE));
+    }
+
+    public boolean isBarcodeChecked()
+    {
+        return elementCache().barcodeRadio.isChecked();
+    }
+
+    public boolean isSampleIdChecked()
+    {
+        return elementCache().sampleIDsRadio.isChecked();
+    }
+
+    public FindByIdsDialog chooseBarcodes()
+    {
+        elementCache().barcodeRadio.check();
+        return this;
+    }
+
+    public FindByIdsDialog chooseSampleIDs()
+    {
+        elementCache().sampleIDsRadio.check();
+        return this;
+    }
+
+    public FindByIdsDialog addIds(List<String> ids)
+    {
+        elementCache().idTextArea.sendKeys(StringUtils.join(ids, "\n"));
+        return this;
+    }
+
+    public void clickCancel()
+    {
+        elementCache().cancelButton.click();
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/components/ui/navigation/FindByIdsDialog.java
```diff
@@ -0,0 +1,82 @@
...
+    protected class ElementCache extends ModalDialog.ElementCache
+    {
+        RadioButton barcodeRadio = RadioButton.RadioButton(Locator.radioButtonByName("uniqueIds")).findWhenNeeded(getComponentElement());
+        RadioButton sampleIDsRadio = RadioButton.RadioButton(Locator.radioButtonByName("sampleIds")).findWhenNeeded(getComponentElement());
+        WebElement idTextArea = Locator.tag("textarea").findWhenNeeded(getComponentElement());
+
+        WebElement cancelButton = Locator.tagWithText("button", "Cancel")
+                .findWhenNeeded(getComponentElement());
+        WebElement findSamplesButton = Locator.tagWithText("button", "Find Samples")
+                .findWhenNeeded(getComponentElement());
```
labkey-danield: General best practice is to mark elements in the ElementCache as final (I don't always follow best practices).

src/org/labkey/test/components/ui/navigation/NavBar.java
```diff
@@ -48,6 +48,13 @@ public HasSearchResults searchFor(String searchString)
         return null;
     }
 
+    public FindByIdsDialog findByIds()
+    {
+        elementCache().findAndSearchMenuButton.click();
+        elementCache().findSamplesOption.click();
```
labkey-danield, line 56: Menu items usually show up right away so this should be fine (and the test are ok with it on TC).

Adding this:
` waitFor(()->elementCache().findSamplesOption.isDisplayed(), "Find samples menu did not show up.", 500);`
might help make the test more resilient to any random rendering slowness.

src/org/labkey/test/components/ui/navigation/NavBar.java
```diff
@@ -85,5 +92,8 @@ public WebDriver getDriver()
         public WebElement userIcon = Locator.tagWithAttribute("img", "alt", "User Avatar").findWhenNeeded(this);
         public WebElement projectNameDisplay = Locator.tagWithClass("span", "project-name").findWhenNeeded(this);
         public Input searchBox = Input.Input(Locator.tagWithClass("input", "navbar__search-input"), getDriver()).findWhenNeeded(this);
+        public WebElement searchForm = Locator.tagWithClass("form", "navbar__search-form").findWhenNeeded(this);
+        public WebElement findAndSearchMenuButton = Locator.tagWithId("button", "find-and-search-menu").findWhenNeeded(searchForm);
+        public WebElement findSamplesOption = Locator.linkContainingText("Find Samples").findWhenNeeded(searchForm);
```
labkey-danield, line 99: Just a FYI...
Using findWhenNeeded will wait to find the element on the first time it is referenced, but it only finds the element once. Using refindWhenNeeded will refind the element each time it is referenced. This has obvious perf issues, but if the container is invalidated and refreshed using refind will protect against a staleElement exception.
 
With that said it is unlikely the NavBar will be refreshed without an entire page refresh. So this is fine.

labkey-danield (approved):
Looks fine. Just minor observations.

PR #838, opened 2021-08-12
Create test suite to quickly verify modules without suites of their own

labkey-tchad:
#### Rationale
We need a short suite to verify changes to modules that don't have a defined suite.

#### Changes
* Trim down BasicTest to just `testCredits` and `testScripts`
* Create new test, `BasicAdminTest`, for the removed test cases
* Create new `Smoke` suite containing `BasictTest` server-side "smoke" tests


src/org/labkey/test/tests/JUnitTest.java
```diff
@@ -457,4 +467,35 @@ static void log(String str, PrintStream printStream)
         String d = new SimpleDateFormat("HH:mm:ss,SSS").format(new Date());      // Include time with log entry.  Use format that matches labkey log.
         printStream.println(d + " " + str);
     }
+
+    public static class BaseJUnitTestWrapper extends BaseWebDriverTest
```
labkey-chrisj, line 471: this makes sense right now, but could probably use some commenting for our future selves to remember why this was needed

labkey-chrisj (approved):
This makes sense right now, but as a general note this code-space could benefit from context-providing comments and judicious javadocs here and there

src/org/labkey/test/tests/JUnitTest.java
labkey-tchad: Yeah. There's a backlog item for general doc improvements for the test runner.

PR #847, opened 2021-08-25
Make test helpers more reliable when using S3

labkey-tchad:
#### Rationale
Some test helpers assume particular pages will run quickly. The aren't as reliable when particular API calls have to talk to S3 before a page is ready.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2541

#### Changes
* Wait for checkboxes to appear on `StartImportPage`
* Wait for file browser to be fully loaded


src/org/labkey/test/tests/AdvancedImportOptionsTest.java
```diff
@@ -76,7 +76,7 @@
     @Override
     protected String getProjectName()
     {
-        return "AdvancedImportOptions";
+        return null;
```
labkey-chrisj, line 80: interesting, this deviates from the usual pattern- maybe helpful to leave a comment/note why?

labkey-chrisj (approved):

Sigmonia (approved):

PR #852, opened 2021-08-31
Trying to address some issues when using the OmniBox.

labkey-danield:
#### Rationale
Minor fix to address some oddness I saw with the OmniBox. Locally I was seeing a failure where the <return> was not getting processed after the text was entered in the OmniBox. Not really sure if this will help or hurt.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/682

#### Changes
* Put the `Keys.ENTER` in the same `sendKeys` call as the filter.


labkey-danield:
Backing out the change to the OmniBox, it did not help.

PR #856, opened 2021-09-07
Automation test for Dataset Category Assignment during Link to Study

labkey-sweta:
#### Rationale
Automation test for Dataset Category Assignment during Link to Study

#### Related Pull Requests
* 

#### Changes
1. Modified the linkTostudy method to handle dataset category
2. Updated the sample type designer page to set linked dataset category
3. Updated the testAutoLinkToStudy() test to handle the uncategorize case.
4. New test cases added testManualDatasetCategoryLink() and testOverWritingDatasetCategory().

### DIFF
--- a/src/org/labkey/test/params/experiment/SampleTypeDefinition.java
+++ b/src/org/labkey/test/params/experiment/SampleTypeDefinition.java
@@ -26,6 +26,7 @@ public class SampleTypeDefinition extends DomainProps
     private String _nameExpression;
     private String _description;
     private String _autoLinkDataToStudy;
+    private String _autoLinkedDatasetCategory;
     private List<FieldDefinition> _fields = new ArrayList<>();
     private Map<String, String> _parentAliases = new HashMap<>();
     // Indicates which parent aliases reference 'exp.dataInputs' instead of 'exp.materialInputs'
@@ -80,6 +81,17 @@ public class SampleTypeDefinition extends DomainProps
     public SampleTypeDefinition setAutoLinkDataToStudy(String value)
     {
         _autoLinkDataToStudy = value;
+        return this;
+    }
+
+    public String getLinkedDatasetCategory()
+    {
+        return _autoLinkedDatasetCategory;
+    }
+
+    public SampleTypeDefinition setLinkedDatasetCategory(String value)
+    {
+        _autoLinkedDatasetCategory = value;
         return this;
     }
 
@@ -210,6 +222,10 @@ public class SampleTypeDefinition extends DomainProps
         if(getAutoLinkDataToStudy() != null)
         {
            options.put("autoLinkTargetContainerId", getAutoLinkDataToStudy());
+        }
+        if(getLinkedDatasetCategory() != null){
+
+            options.put("autoLinkCategory", getLinkedDatasetCategory());
         }
         if (getInventoryMetricUnit() != null)
         {
--- a/src/org/labkey/test/params/experiment/SampleTypeDefinition.java
+++ b/src/org/labkey/test/params/experiment/SampleTypeDefinition.java
@@ -73,6 +73,7 @@ public class SampleTypeDefinition extends DomainProps
         _description = description;
         return this;
     }
+
     public String getAutoLinkDataToStudy()
     {
         return _autoLinkDataToStudy;
@@ -219,11 +220,12 @@ public class SampleTypeDefinition extends DomainProps
             }
             options.put("importAliases", importAliases);
         }
-        if(getAutoLinkDataToStudy() != null)
+        if (getAutoLinkDataToStudy() != null)
         {
-           options.put("autoLinkTargetContainerId", getAutoLinkDataToStudy());
+            options.put("autoLinkTargetContainerId", getAutoLinkDataToStudy());
         }
-        if(getLinkedDatasetCategory() != null){
+        if (getLinkedDatasetCategory() != null)
+        {
 
             options.put("autoLinkCategory", getLinkedDatasetCategory());
         }
--- a/src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
+++ b/src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
@@ -17,13 +17,16 @@ import org.labkey.test.components.html.BootstrapMenu;
 import org.labkey.test.pages.ReactAssayDesignerPage;
 import org.labkey.test.pages.admin.ExportFolderPage;
 import org.labkey.test.pages.admin.ImportFolderPage;
+import org.labkey.test.pages.query.ExecuteQueryPage;
 import org.labkey.test.pages.study.ManageStudyPage;
 import org.labkey.test.params.FieldDefinition;
 import org.labkey.test.params.experiment.SampleTypeDefinition;
 import org.labkey.test.util.DataRegionTable;
+import org.labkey.test.util.Ext4Helper;
 import org.labkey.test.util.PortalHelper;
 import org.labkey.test.util.SampleTypeHelper;
 import org.labkey.test.util.StudyHelper;
+import org.openqa.selenium.WebElement;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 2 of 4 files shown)


src/org/labkey/test/params/experiment/SampleTypeDefinition.java
```diff
@@ -211,6 +223,10 @@ protected String getKind()
         {
            options.put("autoLinkTargetContainerId", getAutoLinkDataToStudy());
         }
+        if(getLinkedDatasetCategory() != null){
```
RosalineP: Tiny nitpick: Looks like both `if(getAutoLinkDataToStudy() != null)` and `if(getLinkedDatasetCategory() != null)` need a extra space, so they look like `if (getAutoLinkDataToStudy() != null)` and `if (getLinkedDatasetCategory() != null)`

src/org/labkey/test/util/SampleTypeHelper.java
```diff
@@ -78,6 +78,10 @@ public void createSampleType(SampleTypeDefinition props)
         {
             createPage.setAutoLinkDataToStudy(props.getAutoLinkDataToStudy());
         }
+        if(props.getLinkedDatasetCategory() != null)
```
RosalineP: I think for `props.getAutoLinkDataToStudy()` and `props.getLinkedDatasetCategory()` the if should be updated with a space so it looks like `if (<condition>)` instead of `if(<condition>)`

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
```diff
@@ -496,17 +507,53 @@ public void testLinkedColumnNotDisplayedCase()
...
+
+        log("Linking the sample type to new dataset category");
+        goToProjectHome();
+        linkToStudy(DATE_BASED_STUDY, SAMPLE_TYPE2, 1, "CAT2");
+
+        checker().verifyEquals("Incorrect category for the dataset(Category exists case)", categoryName,
+                getCategory(VISIT_BASED_STUDY, SAMPLE_TYPE1));
+        checker().verifyEquals("Incorrect category for the dataset(New category case)", "CAT2",
+                getCategory(DATE_BASED_STUDY, SAMPLE_TYPE2));
+    }
```
RosalineP: For consistency here I'd either in-line the `"CAT1"` variable or pull `"CAT2"` into a `String categoryName2 = "CAT2";` variable. Just so the calls to verifyEquals() are parallel and a little more readable.

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
```diff
@@ -517,11 +564,52 @@ private void linkToStudy(String targetStudy, String sampleName, int numOfRowsToB
...
         log("Link to study: Choose target");
         selectOptionByText(Locator.id("targetStudy"), "/" + targetStudy + " (" + targetStudy + " Study)");
+        if (categoryName != null)
+            setFormElement(Locator.name("autoLinkCategory"), categoryName);
         clickButton("Next");
 
         new DataRegionTable("query", getDriver()).clickHeaderButtonAndWait("Link to Study");
     }
 
+    private void createDatasetCategory(String projectName, String name)
```
RosalineP, line 582: I'm a little surprised there's not a pre-existing helper that makes dataset categories! If you think it'd be easy and useful, I'd consider pulling this into a shareable place so others could reuse the dataset category creation function. No big deal if you think that that wouldn't be worth it, though.

RosalineP (approved):
Had some small comments, but the test cases look good!

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
```diff
@@ -45,6 +48,8 @@
     final static String SAMPLE_TYPE1 = "Sample type 1";
     final static String SAMPLE_TYPE2 = "Sample type 2";
 
+    private static int cnt = 0; // to keep count of rows which are already linked.
```
labkey-danield, line 52: Thr is nthng wrng wth vwls. :)

However, it doesn't look like this variable is ever used.

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
```diff
@@ -125,6 +130,7 @@ public void testLinkToStudy()
 
         checker().verifyEquals("Incorrect number of rows linked", 2, table.getDataRowCount());
         checker().verifyEquals("Incorrect Participant ID's", Arrays.asList("P3", "P4"), table.getColumnDataAsText("ParticipantId"));
+        checker().verifyEquals("Incorrect category for the dataset(Uncategorized case)", " ", getCategory(VISIT_BASED_STUDY, SAMPLE_TYPE1));
```
labkey-danield, line 134: Do you think it might be more robust to trim the return value from `getCategory` and compare to an empty string?
If the code for `DataRegionTable.getDataAsText` ever changes to return an empty string if nothing is there this will fail and it could be a bit of a pain to try and understand why. Removing the dependency on a space being returned might help.

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
labkey-danield: Could you use the `org.labkey.remoteapi.reports.Category` class?
It looks like it might do what you need.

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
```diff
@@ -496,17 +507,53 @@ public void testLinkedColumnNotDisplayedCase()
...
+        String categoryName = "CAT1";
+        createDatasetCategory(VISIT_BASED_STUDY, categoryName);
+        goToProjectHome();
+
+        log("Linking the sample type to preexisting dataset category");
+        linkToStudy(VISIT_BASED_STUDY, SAMPLE_TYPE1, 1, categoryName);
+
+        log("Linking the sample type to new dataset category");
+        goToProjectHome();
+        linkToStudy(DATE_BASED_STUDY, SAMPLE_TYPE2, 1, "CAT2");
```
labkey-danield: It looks like every time `linkToStudy` is called a call to `goToProjectHome` is done first. Why not pass in the project name and call `goToProjectHome` in `linkToStudy`?

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
```diff
@@ -496,17 +507,53 @@ public void testLinkedColumnNotDisplayedCase()
...
     {
-        //Cleaning up the links.
-        recallDataset(DATE_BASED_STUDY, SAMPLE_TYPE1);
-        recallDataset(DATE_BASED_STUDY, SAMPLE_TYPE2);
-        recallDataset(VISIT_BASED_STUDY, SAMPLE_TYPE1);
-        recallDataset(VISIT_BASED_STUDY, SAMPLE_TYPE2);
+        //deleting the datasets from study folders.
+        deleteDatasets(DATE_BASED_STUDY);
+        deleteDatasets(VISIT_BASED_STUDY);
+        cnt = 0; //Resetting the counter between the tests.
```
labkey-danield, line 561: This is the only place where the variable `cnt` is used.

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
```diff
@@ -517,11 +564,52 @@ private void linkToStudy(String targetStudy, String sampleName, int numOfRowsToB
...
+        goToProjectHome(projectName);
+        _studyHelper.goToManageDatasets().clickDeleteMultipleDatasets();
+        if (Locator.xpath("//table/tbody/tr").findElements(getDriver()).size() > 1)
+        {
+            checkCheckbox(Locator.xpath("//table/tbody/tr[1]/td[1]/input"));
+            clickButton("Delete Selected", 0);
+            acceptAlert().contains("Are you sure you want to delete the selected datasets? This action cannot be undone.");
+        }
+        waitFor(() -> 1 == Locator.xpath("//table/tbody/tr").findElements(getDriver()).size(), WAIT_FOR_JAVASCRIPT);
+    }
```
labkey-danield: Could you use TestDataGenerator for this?
Example:
`TestDataGenerator.deleteDomain(projectName, "study", "Sample type 2");`

labkey-danield (approved):
Just a few comments. There might be some existing API helpers that you could use (so you don't have to do everything through the UI).

src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
labkey-sweta: The variable is used in linktostudy method and beforeTest

PR #860, opened 2021-09-10
Wait for rootMenu to contain at least any menuItems

labkey-chrisj:
#### Rationale
This updates the DropdownButtonGroup component wrapper to wait for its root menuItem to contain at least 1 menuItem, hopefully to prevent tests from racing ahead of it being populated

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/993


labkey-tchad (approved):
I'd run this through SM and Biologics but it looks good to me.

PR #866, opened 2021-09-20
new lineageGrid test component

labkey-chrisj:
#### Rationale
This introduces 2 new test component wrappers: 1 for the LineageGrid, another to encapsulate the rows in the lineageGrid

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/999 <-- contains the updated tests, pages that are currently the only usages of these new components

#### Changes
Eliminates the deprecated responsiveGrid in Biologics, replaces all usages with the new components

New classes: LineageGrid, which contains the panel (including the show parents, show children buttons), the seed-status message up top, the ResponsiveGrid wrapping the table: 
image
Uses the current ResponsiveGrid class as a subcomponent to surface rowMap data, and to find specific rows.
Provides getters to find specific lineageGridRows

LineageGridRow 
image
Contains checks to replace locator usage that was in tests; e.g. find rows by type: 
seed, duplicate, firstParent, secondParent, firstChild, secondChild

### DIFF
--- a/src/org/labkey/test/components/ui/grids/LineageGrid.java
+++ b/src/org/labkey/test/components/ui/grids/LineageGrid.java
@@ -18,6 +18,9 @@ import java.util.stream.Collectors;
 
 import static org.labkey.test.components.html.Input.Input;
 
+/**
+ * Wraps the component described in ui-components internal\components\lineage\grid\LineageGridDisplay.tsx
+ */
 public class LineageGrid extends WebDriverComponent<LineageGrid.ElementCache>
 {
     private final WebElement _el;
--- a/src/org/labkey/test/components/ui/grids/LineageGrid.java
+++ b/src/org/labkey/test/components/ui/grids/LineageGrid.java
@@ -1,0 +1,207 @@
+package org.labkey.test.components.ui.grids;
+
+import org.labkey.test.Locator;
+import org.labkey.test.WebDriverWrapper;
+import org.labkey.test.components.Component;
+import org.labkey.test.components.UpdatingComponent;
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/components/ui/grids/LineageGrid.java
```diff
@@ -0,0 +1,207 @@
...
+import org.openqa.selenium.WebElement;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static org.labkey.test.components.html.Input.Input;
+
+public class LineageGrid extends WebDriverComponent<LineageGrid.ElementCache>
```
labkey-tchad: Add explicit reference to the product component that this wraps.

labkey-tchad (approved):
I'm a little concerned about performance with all of the methods that return filtered sets of `LineageGridRow`s. With the lack of caching, that's a lot of calls to `getText`.
Given the very limited use of this component, however, it probably isn't worth the headache of setting up caching of all those text values and making sure the cache is cleared correctly to avoid getters returning inaccurate values.

labkey-chrisj:
Some of these getText() calls could be avoided by implementing a finder for the `LineageGridRow `(so if we're just out to find the ones with a given lineageName that would be resolved in the xpath), but I was a little daunted by the prospect of implementing the finder as extending `GridRow.GridRowFinder`.  I considered doing that, but then I remembered we don't use this that often.

I also observed that the pager (which optionally appears) is implemented differently than the other pagers we already support in the `Pager ` class, ugh

PR #869, opened 2021-09-22
Fix ModeratorReviewTest pre-test cleanup

labkey-tchad:
#### Rationale
Not all tables support `TruncateTableCommand`. Also helpers like this don't really belong in `BaseWebDriverTest`

#### Related Pull Requests
* #863 

#### Changes
* Add distinct methods for deleting rows via `TruncateTable` and `DeleteRows`
* Create `QueryUtils`
* Pause search indexer while switching directory type

### DIFF
--- a/src/org/labkey/test/util/query/QueryUtils.java
+++ b/src/org/labkey/test/util/query/QueryUtils.java
@@ -1,0 +1,42 @@
+package org.labkey.test.util.query;
+
+import org.labkey.remoteapi.CommandException;
+import org.labkey.remoteapi.Connection;
+import org.labkey.remoteapi.query.DeleteRowsCommand;
+import org.labkey.remoteapi.query.SelectRowsCommand;
+import org.labkey.remoteapi.query.SelectRowsResponse;
+import org.labkey.remoteapi.query.TruncateTableCommand;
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/util/query/QueryUtils.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public static void truncateTable(String containerPath, String schema, String table) throws IOException, CommandException
+    {
+        Connection cn = WebTestHelper.getRemoteApiConnection();
+        TruncateTableCommand cmd = new TruncateTableCommand(schema, table);
+        cmd.execute(cn, containerPath);
+    }
+
+    public static void selectAndDeleteRows(String containerPath, String schema, String table) throws IOException, CommandException
```
labkey-danield: Maybe change the title to selectAndDelete**All**Rows?

Also do you  think a comment in this class might be useful to help people understand when they should call truncateTable and when they should call selectAndDelete?

labkey-danield (approved):

src/org/labkey/test/util/query/QueryUtils.java
labkey-tchad: Good idea.

PR #870, opened 2021-09-24
Create properties template and remove driver binaries

labkey-tchad:
#### Rationale
`test.properties` doesn't give any guidance regarding how to use it. It also ends up in changelists because people often edit it. This removes `test.properties` from source control and replaces it with `test.properties.dist`. The gradle plugin will generate `test.properties` if it is missing.
The chromedriver and geckodriver binaries checked in are out of date. It is bad practice to commit binaries to source control anyway so we will remove them and require folks to install them manually to run tests.

#### Related Pull Requests
* https://github.com/LabKey/gradlePlugin/pull/133

#### Changes
* Remove chromedriver/geckodriver binaries
* Remove `test.properties` from source control and add to .gitignore
* Add annotated `test.properties.template`
* Generate `test.properties` from `test.properties.template`


labkey-sweta (approved):

labkey-chrisj:
A little manual testing on my part says the only missing thing is documentation calling out the existence of the gradle command to create test.properties from its template:  `gradlew :server:testAutomation:initProperties`

PR #871, opened 2021-09-27
Item 9192: Initial modifications to support sample status

labkey-susanh:
#### Rationale
We are developing support for adding a status field to samples, which will determine which actions can be taken on the samples.  Here we add a few tests for this experimental feature that validate behavior when linking and recalling samples from a study dataset and when deleting a sample type containing locked samples.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2636

#### Changes
* Add test case in `SampleTypeLinkToStudyTest` for linking and recalling locked samples
* Add test case in `SampleTypeTest` for deleting a sample type with locked samples
* Add some helper methods in SampleTypeHelper for creating statuses and the like.

### DIFF
--- a/src/org/labkey/test/tests/SampleStatusTest.java
+++ b/src/org/labkey/test/tests/SampleStatusTest.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright (c) 2011-2019 LabKey Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.labkey.test.tests;
+
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.labkey.test.BaseWebDriverTest;
+import org.labkey.test.Locator;
+import org.labkey.test.categories.Daily;
+import org.labkey.test.params.experiment.SampleTypeDefinition;
+import org.labkey.test.util.DataRegionTable;
+import org.labkey.test.util.PortalHelper;
+import org.labkey.test.util.SampleTypeHelper;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+@Category({Daily.class})
+public class SampleStatusTest extends BaseWebDriverTest
+{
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/tests/SampleStatusTest.java
```diff
@@ -0,0 +1,123 @@
...
+        // If you are debugging tests change this function to do nothing.
+        // It can make re-running faster but you need to valid the integrity of the test data on your own.
+//        log("Do nothing.");
+    }
+
+    @Test
+    public void testDeleteSampleTypeWithLockedSamples()
+    {
+        SampleTypeHelper sampleTypeHelper = new SampleTypeHelper(this);
+        previousSampleStatusFlag = SampleTypeHelper.setSampleStatusEnabled(true);
```
labkey-danield: If I understand the intent of previousSampleStatusFlag, would it be better to put this in doSetup? That way if other tests are added in the future you don't have to worry about resetting the state.

src/org/labkey/test/tests/SampleTypeTest.java
```diff
@@ -123,12 +125,47 @@ private void doSetup()
...
         // If you are debugging tests change this function to do nothing.
         // It can make re-running faster but you need to valid the integrity of the test data on your own.
 //        log("Do nothing.");
     }
 
+    @Test
+    public void testDeleteSampleTypeWithLockedSamples()
+    {
+        SampleTypeHelper sampleTypeHelper = new SampleTypeHelper(this);
+        previousSampleStatusFlag = SampleTypeHelper.setSampleStatusEnabled(true);
```
labkey-danield: Same comment as before, maybe put in a more general doSetup function.

src/org/labkey/test/util/SampleTypeHelper.java
```diff
@@ -309,4 +310,34 @@ public String getDetailsFieldValue(String label)
...
+    }
+
+    public void addSampleState(String label, @Nullable String stateType)
+    {
+        setFormElement(Locator.name("quf_Label"), label);
+        if (stateType != null)
+        {
+            setFormElement(Locator.name("quf_stateType"), stateType);
+        }
+        clickButton("Submit");
```
labkey-danield, line 348: I cannot remember exact flow... does this cause a page load? If so a clickAndWait might be safer here.

labkey-danield (approved):

src/org/labkey/test/tests/SampleTypeTest.java
labkey-susanh: Hah!  Well, this one was supposed to go away since I moved it into the other class.

src/org/labkey/test/util/SampleTypeHelper.java
labkey-susanh: `clickButton` already has a default wait built in, which seems to do the trick here.

src/org/labkey/test/tests/SampleStatusTest.java
labkey-susanh: Ah yes.  Moved.  Thanks.

PR #878, opened 2021-10-08
Add regression coverage for issue 43572

labkey-chrisj:
#### Rationale
Adds a new test to verify that LabKey respects fKDisplayColumn settings in metadata

#### Related Pull Requests
replaces https://github.com/LabKey/biologics/pull/1017

#### Changes
just the new test

### DIFF
--- a/src/org/labkey/test/tests/SampleTypeLookupDisplayColumnTest.java
+++ b/src/org/labkey/test/tests/SampleTypeLookupDisplayColumnTest.java
@@ -1,0 +1,184 @@
+package org.labkey.test.tests;
+
+import org.junit.BeforeClass;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.labkey.test.BaseWebDriverTest;
+import org.labkey.test.Locator;
+import org.labkey.test.TestTimeoutException;
+import org.labkey.test.categories.InDevelopment;
+import org.labkey.test.pages.InsertPage;
+import org.labkey.test.pages.query.SourceQueryPage;
+import org.labkey.test.params.FieldDefinition;
+import org.labkey.test.params.experiment.SampleTypeDefinition;
+import org.labkey.test.util.DataRegionTable;
+import org.labkey.test.util.PortalHelper;
+import org.labkey.test.util.TestDataGenerator;
+import org.labkey.test.util.exp.SampleTypeAPIHelper;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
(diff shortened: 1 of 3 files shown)


src/org/labkey/test/tests/SampleTypeLookupDisplayColumnTest.java
```diff
@@ -0,0 +1,184 @@
+package org.labkey.test.tests;
```
labkey-tchad: Should go in a module-specific package.
```suggestion
package org.labkey.test.tests.experiment;
```

src/org/labkey/test/tests/SampleTypeLookupDisplayColumnTest.java
```diff
@@ -0,0 +1,184 @@
...
+    // it would be nice if our
+    private void setFieldValues(String name, String comment, String amount, String ingredient)
+    {
+        if (name != null)   // for update, name field is disabled
+            setFormElement(Locator.name("quf_Name"), name);
+        setFormElement(Locator.name("quf_comment"), comment);
+        setFormElement(Locator.name("quf_amount"), amount);
+        setFormElement(Locator.name("quf_ingredient"), ingredient);
+        clickAndWait(Locator.lkButton("Submit"));
+    }
```
labkey-tchad: `UpdateQueryRowPage` (returned by `DataRegionTable.clickEditRow`) might work for inserting new rows also. If you try it and it works, consider updating `DataRegionTable. clickInsertNewRow` to return one.
```suggestion
    private void setFieldValues(String name, String comment, String amount, String ingredient)
    {
        var insertPage = new UpdateQueryRowPage(this);
        if (name != null)   // for update, name field is disabled
            insertPage.setField("Name", name);
        insertPage.setField("comment", comment);
        insertPage.setField("amount", amount);
        insertPage.setField("ingredient", ingredient);
        insertPage.submit();
    }
```

labkey-tchad (approved):
Tests should, generally, go in a module-specific package. Also, a suggestion for an improvement to `setFieldValues`

src/org/labkey/test/tests/SampleTypeLookupDisplayColumnTest.java
labkey-chrisj: I considered creating a page class for an `InsertQueryRowPage` and updating `DataRegionTable.clickInsertNewRow()` to return that, but I figured that would be seen as too much feature creep.  Thanks for suggesting!
It turns out that UpdateQueryRowPage works just fine for the insert scenario!

PR #880, opened 2021-10-12
Test AspectJ 1.9.8, Java 17-compatible version

labkey-adam:
#### Rationale
Upgrade AspectJ 1.9.7 -> 1.9.8.RC3 to enable Java 16 & 17 compatibility

#### Related Pull Requests
* https://github.com/LabKey/server/pull/153


labkey-jeckels (commented):
Seems OK to me and the tests look much happier, but I'll defer to @labkey-tchad for the actual approval since he has to know more about our AspectJ usage than I do. We should presumably file an issue to switch to a non-RC, Java 16 compatible version when it's available.

labkey-tchad (approved):
We are on 1.9.7, which claims to support Java 15 and 16: https://htmlpreview.github.io/?https://github.com/eclipse/org.aspectj/blob/master/docs/dist/doc/README-197.html

labkey-adam:
I guess, but clearly TeamCity was very unhappy with Java 16 byte codes until this AspectJ upgrade 🤷

PR #888, opened 2021-10-22
Item 9584: ManageSampleStatusesPanel for sample statuses CRUD within a given container

cnathe:
#### Rationale
See related PRs for further details. We removed the state type field from the core.DataStates insert and update form so this PR changes some existing tests to use the Java API to insert sample status rows instead of the UI.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/653
* https://github.com/LabKey/platform/pull/2698
* https://github.com/LabKey/sampleManagement/pull/726
* https://github.com/LabKey/biologics/pull/1026
* https://github.com/LabKey/inventory/pull/323

#### Changes
* SampleTypeHelper update to insertSampleState() to use InsertRowsCommand


src/org/labkey/test/util/SampleTypeHelper.java
```diff
@@ -322,29 +327,31 @@ public static Boolean setSampleStatusEnabled(boolean enabled)
...
         clickAndWait(Locator.linkContainingText("view data"));
-        DataRegionTable drt = new DataRegionTable("query", this);
         for (Map.Entry<String, StatusType> statePair : states.entrySet())
         {
+            DataRegionTable drt = new DataRegionTable("query", this);
             if (drt.getRowIndex("Label", statePair.getKey()) < 0)
             {
-                drt.clickInsertNewRow();
-                addSampleState(statePair.getKey(), statePair.getValue().name());
+                insertSampleState(folderPath, statePair.getKey(), statePair.getValue().name());
```
labkey-susanh: If this is just using the API, is there a need to continue navigating to the data region and all that? The `insertSampleState` helper could also do all the rows in one insert command.

labkey-susanh (approved):
Seems functional, but might be more efficient.

src/org/labkey/test/util/SampleTypeHelper.java
cnathe: Good catch. Fixed / refactored accordingly.

PR #903, opened 2021-11-05
Improve performance of domain designer helpers

labkey-tchad:
#### Rationale
There have been complaints that tests are too slow when they have to use the UI domain designer.

#### Changes
* Refactor tests to set advanced field options in one go
* Make `DomainFieldRow` complain if the advanced options dialog is opened more than once
* Remove some (hopefully unnecessary) retries that might be slowing things down


labkey-chrisj (approved):
Nice!

PR #904, opened 2021-11-08
Various test reliability fixes

labkey-tchad:
#### Rationale
Trying to clear up some intermittent TeamCity failures.

#### Changes
* Add delay when waiting for pending requests to complete
* Wait for correct values in `setFormElement`
* Update column chart helpers to wait for plots to appear
* Make Ext4 window component find the frontmost window
* Make grid components agree on presence of checker column
* Set up lists in ClientAPITest via API
* Wait for page load in `SampleTypeLinkToStudyTest`
* Add helper for viewing reports from `ManageViewsPage`
* Don't use `UIContainerHelper` in `EmbeddedWebPartTest`
* Ignore more GWT alerts triggered by crawler
* Delete JUnit test folder via API


labkey-sweta (approved):

src/org/labkey/test/tests/ClientAPITest.java
```diff
@@ -101,34 +106,25 @@
...
+        new FieldDefinition("LastName", FieldDefinition.ColumnType.String).setLabel("Last Name").setDescription("The last name").setRequired(true),
+        new FieldDefinition("Age", FieldDefinition.ColumnType.Integer).setLabel("Age").setDescription("The age")
+    );
+
+    public final static String TEST_DATA =
+        "Key\tFirstName\tLastName\tAge\n" +
+        "1\tBill\tBillson\t34\n" +
+        "2\tJane\tJaneson\t42\n" +
+        "3\tJohn\tJohnson\t17\n" +
+        "4\tMandy\tMandy;son\t32\n" +
```
labkey-danield, line 124: Is the  ';' suppose to be there? Is this just a typo carried forward?

src/org/labkey/test/tests/ClientAPITest.java
```diff
@@ -101,34 +106,25 @@
...
+        "5\tNorbert\tNorbertson\t28\n" +
+        "6\tPenny\tPennyson\t38\n" +
+        "7\tYak\tYakson\t88\n";
```
labkey-danield, line 127: Since this is only used in one place, why leave it as public? Why not move it to where it is used and create and add to the list using Map.of(...)?

labkey-danield (approved):

src/org/labkey/test/tests/ClientAPITest.java
labkey-tchad: Not sure about the ';' but I'm always ok with more special characters.

PR #907, opened 2021-11-08
Item 9575: Update locator for new conditional labeling of "QC States Setting"

labkey-susanh:
#### Rationale
The label for "QC State Settings" has been updated to "Sample Status and QC State Settings" when the SampleManager module is available. We update the locator accordingly.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2760

#### Changes
* Update locator for QC State Settings checkbox.


XingY (approved):

src/org/labkey/test/components/ManageSampleStatusesPanel.java
```diff
@@ -0,0 +1,188 @@
...
+        public String description;
+        public SampleTypeHelper.StatusType statusType;
+    }
+
+    public ManageSampleStatusesPanel(WebElement element, WebDriver driver)
+    {
+        _driver = driver;
+        _componentElement = element;
+    }
+
```
labkey-danield, line 32: For some of the panel/components I've added a public `isPanelLoaded` method. This is an example from the LineagePanel:
```
    public boolean isPanelLoaded()
    {
        getWrapper().waitForElementToDisappear(Locators.loadingTag);

        boolean isVisible;

        try
        {
            isVisible = Locators.lineageName.findElement(elementCache().detailContainer()).isDisplayed();

        }
        catch(NoSuchElementException  | StaleElementReferenceException nse)
        {
            return false;
        }

        // The container has to be there and either the "no lineage" message or a lineage link element has to be there.
        isVisible = isVisible && (getWrapper().isElementPresent(Locators.noLineageMsg) ||
                getWrapper().isElementPresent(Locators.lineageDetailsListItem));

        return isVisible;

    }

```

This allows the consumer to wait for the panel to indicate it is ready/loaded before doing anything. Example:
```
        waitFor(()-> elementCache().lineagePanel.isPanelLoaded(),
                "The 'Sample Lineage' page did not load in time.",
                5_000);

```
Not all panels need this, and since you are writing the consumers of this panel you would have a better idea if this is needed.

src/org/labkey/test/components/ManageSampleStatusesPanel.java
```diff
@@ -0,0 +1,188 @@
...
+    {
+        elementCache().statusTypeSelect.select(statusType.toString());
+        return this;
+    }
+
+    public ManageSampleStatusesPanel clickSave()
+    {
+        elementCache().saveButton.click();
+        return this;
+    }
```
labkey-danield, line 129: Depending upon the panels behavior when an error happens you may want to add a clickSaveExpectError method. 

Example:
```
    public String clickSaveExpectError()
    {
        elementCache().saveButton.click();
        waitFor(()->
                isVisible(BootstrapLocators.errorBanner),
                "Expected a error message, but none was shown.",
                10_000);

        return errorBannerText();
    }

```
Again this may not be needed or relevant for this panel, but is just general feedback.

src/org/labkey/test/components/ManageSampleStatusesPanel.java
```diff
@@ -0,0 +1,188 @@
...
+        Locator statusItems = Locator.tagWithClass("button", "list-group-item");
+
+        WebElement addStatusButton = Locator.tagWithText("span", "Add New Status")
+                .findWhenNeeded(getComponentElement());
+        Input labelField = Input.Input(Locator.inputByNameContaining("label"), getDriver()).findWhenNeeded(this);
+        WebElement descriptionField = Locator.textarea("description").findWhenNeeded(this);
+        ReactSelect statusTypeSelect = ReactSelect.finder(getDriver()).findWhenNeeded(this);
+        WebElement saveButton = Locator.tagWithText("button", "Save").findWhenNeeded(this);
+        WebElement deleteButton = Locator.tagContainingText("button", "Delete").findWhenNeeded(this);
+    }
```
labkey-danield, line 174: Elements in here should be final.

src/org/labkey/test/components/ManageSampleStatusesPanel.java
```diff
@@ -0,0 +1,188 @@
...
+            if (statusType == null || statusType.toString().equals(name))
+                return Locator.tagWithClass("button", "list-group-item").containing(name).findElement(getComponentElement());
+            else
+                return Locator.tagWithClass("button", "list-group-item").containing(name +  statusType).findElement(getComponentElement());
+        }
+
+        Locator statusItems = Locator.tagWithClass("button", "list-group-item");
+
+        WebElement addStatusButton = Locator.tagWithText("span", "Add New Status")
+                .findWhenNeeded(getComponentElement());
```
labkey-danield, line 168: I _believe_  `this` and `getComponentElement` are equivalent.

labkey-danield (approved):

src/org/labkey/test/components/ManageSampleStatusesPanel.java
labkey-susanh: Added, though I'm not exercising this functionality here, so it'll be untested.

src/org/labkey/test/components/ManageSampleStatusesPanel.java
labkey-susanh: Done.

src/org/labkey/test/components/ManageSampleStatusesPanel.java
labkey-susanh: Leaving as is since it seems to work.

src/org/labkey/test/components/ManageSampleStatusesPanel.java
labkey-susanh: I'm going to leave this as is for now.  It might be wanted at some point, but what's currently there seems to do the trick.

PR #909, opened 2021-11-10
add dumbster page for app usage

labkey-chrisj:
#### Rationale
this adds a page class (and with it, a beginAt helper for that page) for the dumbster's begin view

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/1050

#### Changes
just the new page class

### DIFF
--- a/src/org/labkey/test/pages/dumbster/DumbsterPage.java
+++ b/src/org/labkey/test/pages/dumbster/DumbsterPage.java
@@ -1,0 +1,52 @@
+package org.labkey.test.pages.dumbster;
+
+import org.labkey.test.Locator;
+import org.labkey.test.WebDriverWrapper;
+import org.labkey.test.WebTestHelper;
+import org.labkey.test.components.dumbster.EmailRecordTable;
+import org.labkey.test.pages.LabKeyPage;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+public class DumbsterPage extends LabKeyPage<DumbsterPage.ElementCache>
+{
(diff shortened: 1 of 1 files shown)


labkey-tchad (changes requested):
Not sure what additional functionality this proves over the existing `EmailRecordTable`. Maybe just add a `beginAt` method to `EmailRecordTable`. Also, containerPath is unnecessary; the email recorder is site-wide, not tied to a particular container.

labkey-chrisj:
It adds the beginAt and the link to click back to the container, and the waitForPage (which might not be necessary)
only reason I added this as a page was for consistency; do we hang beginAts on components anywhere else?

I only plumbed the containerPath in there because you get that coming from goToModule from wherever you are; I also added the link back to the container b/c it's right there in the page

labkey-tchad:
[`PipelineStatusTable`](https://github.com/LabKey/testAutomation/blob/develop/src/org/labkey/test/util/PipelineStatusTable.java#L58) has something similar to `beginAt`. Maybe a different method name would be good to let `beginAt` have a consistent meaning. Something like `goToDumbster`.

labkey-chrisj:
That explains why we have a PipelineStatusDetailsPage but no PipelineStatusPage- we had a way to get to the pipelineStatusPage already.
I've been here for years and didn't know that until I spent some time trying to figure out why there wasn't a page for it
Usually when I write a new page class for an old component, it's because I had to spend time figuring out how to get the right component and didn't know the component by name off the top of my head, and I figured it would be worthwhile to save the next guy the effort/make the pieces more discoverable

PR #915, opened 2021-11-20
Get rid of a bunch of deprecated stuff in 'ListHelper'

labkey-tchad:
#### Rationale
These `ListHelper` classes have been deprecated for a long time. Time to cull some of the easy to refactor ones.

#### Changes
* Convert many uses of `ListColumn` to `FieldDefinition`
* Get rid of some nested ListHelper.ListColumn property classes


src/org/labkey/test/components/domain/DomainFormPanel.java
```diff
@@ -175,7 +175,7 @@ public DomainFieldRow manuallyDefineFields(String name)
     public DomainFieldRow manuallyDefineFields(FieldDefinition fieldDefinition)
```
labkey-danield, line 182: I'm sure this is beyond the scope of what you wanted to change in this PR but it feels like the name of this function, including the overloads, should be changed to the singular (`manuallyDefineField`) because there is only one field definition, or name, being passed in.

src/org/labkey/test/pages/list/EditListDefinitionPage.java
```diff
@@ -100,27 +99,27 @@ public AdvancedListSettingsDialog openAdvancedListSettings()
 
     // List Fields
 
-    public DomainFormPanel setKeyField(ListHelper.ListColumnType listKeyType, String listKeyName)
+    public DomainFormPanel manuallyDefineFieldsWithKey(FieldDefinition keyField)
```
labkey-danield, line 102: Same kind of comment as in the DomainFormPanel... I think it would be better to change the name and make it singular (`manuallyDefineFieldWithKey`)

src/org/labkey/test/pages/list/EditListDefinitionPage.java
```diff
@@ -100,27 +99,27 @@ public AdvancedListSettingsDialog openAdvancedListSettings()
...
-            selectKeyField(listKeyName);
-        }
+        fieldsPanel.manuallyDefineFields(keyField);
+        selectKeyField(keyField.getName());
 
         return fieldsPanel;
     }
 
+    @NotNull
+    public DomainFormPanel manuallyDefineFieldsWithAutoIncrementingKey(String listKeyName)
```
labkey-danield, line 112: Ditto on the name.

src/org/labkey/test/pages/list/EditListDefinitionPage.java
```diff
@@ -100,27 +99,27 @@ public AdvancedListSettingsDialog openAdvancedListSettings()
...
+    {
+        DomainFormPanel fieldsPanel = getFieldsPanel();
+        fieldsPanel.manuallyDefineFields("REMOVE_ME");
```
labkey-danield: Maybe add a comment as to why you need to add this temp "REMOVE_ME" field. Looking at the code it's not clear to me why it couldn't be implemented as:
```
DomainFormPanel fieldsPanel = getFieldsPanel();
selectAutoIncrementKeyField();
sleep(500);
fieldsPanel.getField(0).setName(listKeyName);
return fieldsPanel;
```

labkey-chrisj (approved):
Oh, I've been wanting this to happen for a while!  Thanks for doing it

labkey-danield (approved):

src/org/labkey/test/components/domain/DomainFormPanel.java
labkey-tchad: These methods are named thus because they start by clicking "define fields manually" (as opposed to uploading a file to define the fields). They are only valid for a new domain. I'll add some documentation.

src/org/labkey/test/pages/list/EditListDefinitionPage.java
labkey-tchad: Agreed that it isn't clear. `manuallyDefineFields` changes the panel mode and the "REMOVE_ME" field is a workaround for the field that gets added automatically when switching modes

PR #917, opened 2021-11-24
Update text for EntityInsertPanel mode

labkey-chrisj:
#### Rationale
The `EntityInsertPanel `component uses different text for its mode-select toggles (to select between file upload, edit grid, copy/paste), this change attempts to make the component be able to correctly find these mode-select elements for assay upload as well as for sample upload 

assay upload:
image
sample upload:
image

#### Related Pull Requests
https://github.com/LabKey/testAutomation/pull/917  <-- contains new tests for job-related sample auto-population of assay samples; calls for the ability to check whether `EntityInsertPanel `is in file-upload vs grid mode

#### Changes
This changes the text by which the mode-select list-items are found, hopefully to work as well for assay upload as for samples

### DIFF
--- a/src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
+++ b/src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
@@ -257,21 +257,24 @@ public class EntityInsertPanel extends WebDriverComponent<EntityInsertPanel.Elem
 
     public boolean isBulkUpdateVisible()
     {
-        return modeSelectListItem("from Grid").withClass("active").findOptionalElement(this).isPresent() &&
+        // for sample upload, text is 'Create Samples from Grid', for assay it is 'Enter Data into Grid'
+        return modeSelectListItem(" Grid").withClass("active").findOptionalElement(this).isPresent() &&
                 elementCache().bulkUpdateBtnLoc.existsIn(this) &&
                 isElementVisible(elementCache().bulkUpdateBtn);
     }
 
     public boolean isDeleteRowsVisible()
     {
-        return modeSelectListItem("from Grid").withClass("active").findOptionalElement(this).isPresent() &&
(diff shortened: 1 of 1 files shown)


labkey-tchad (commented):
Should this component support assay import? That seems like a misuse of this component. Assay runs aren't "entities" as far as I understand. The import pages have similar designs but I think its a mistake to shoehorn them into the same test component.

labkey-chrisj:
I'm pretty sure this component supports assay import.  I mean, look at EntityInsertPanel.java, lines 26-28:
```
/**
 * This class automates the UI component defined in <a href="https://github.com/LabKey/labkey-ui-components/blob/master/packages/components/src/components/entities/EntityInsertPanel.tsx">components/entities/EntityInsertPanel.tsx</a>
 * **This is the same component (collection of atomic elements) used in insertAssay and createSamples.**
 */
```

It's also used for assay import in SM as well as biologics, it does assay upload, full stop.
If you're stuck on the idea that the component can have different text for its buttons and how can we do this... maybe these methods shouldn't worry about the text of the mode-select, maybe they should just return whether or not the grid/file upload/text input is shown, or maybe the component should put an ID or class on those buttons if the text is going to vary

cnathe:
@labkey-chrisj @labkey-tchad 
I checked in the LKB, LKSM, and ui-components code and the assay import case does not use the EntityInsertPanel.tsx component. That component is used for the sample insert/import and the LKSM sources insert/import but the assay import does not use it. It uses very similar underlying components for the file attachment / import form and the editable grid panel.

labkey-chrisj:
Huh. 
So I guess I'm wondering if that means we should have an assay-specific component?
I can refactor my current story to do these checks in the page

It will also take some work to update the SM and Bio assay upload pages (and tests) to stop using the entityInsertPanel and use something specific to assay instead

PR #920, opened 2021-12-01
Have editableGrid wait for dropdown to populate

labkey-danield:
#### Rationale
Values in the dropdown, under a cell, in the editable grid can come from a query that may take a moment to load. This adds a better wait for the dropdown to populate before trying to select a value from it.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/765

#### Changes
* Change the waitFor to check is "Loading..." is no longer displayed.


labkey-tchad (approved):

labkey-chrisj (approved):
Adding a wait for the dialog to appear before searching for one seems extra-cautious; no disapproval here tho

PR #923, opened 2021-12-03
update unused setSelectValue/clearSelectValue methods

labkey-chrisj:
#### Rationale
this re-implements `setSelectValue `and `clearSelectValue` on DetailTableEdit (which are currently not in use) to find the intended reactSelect by searching in its scope, instead of searching for it by way of it following a label.  

This seemed like a reasonable idea when I wanted to edit Run Details in Biologics (see related PR) and noticed that these methods 1) did not work, and 2) were not in use.
My best guess is these methods worked at one point, but the components changed (the label isn't in a label tag) and when it didn't cause any tests to fail, there wasn't need to update these

#### Related Pull Requests
TBD

#### Changes
Find the select in a search context by its data-caption attribute (which is also the text on the label)
image

### DIFF
--- a/src/org/labkey/test/components/ui/grids/DetailTableEdit.java
+++ b/src/org/labkey/test/components/ui/grids/DetailTableEdit.java
@@ -295,7 +295,8 @@ public class DetailTableEdit extends WebDriverComponent<DetailTableEdit.ElementC
      **/
     public DetailTableEdit setSelectValue(String fieldCaption, List<String> selectValues)
     {
-        FilteringReactSelect reactSelect =  FilteringReactSelect.finder(_driver).followingLabelWithSpan(fieldCaption).find();
+        WebElement container = Locator.tag("td").withAttribute("data-caption", fieldCaption).findElement(this);
+        FilteringReactSelect reactSelect =  FilteringReactSelect.finder(_driver).find(container);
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/components/ui/grids/DetailTableEdit.java
```diff
@@ -295,7 +295,8 @@ public DetailTableEdit setSelectValue(String fieldCaption, String selectValue)
      **/
     public DetailTableEdit setSelectValue(String fieldCaption, List<String> selectValues)
     {
-        FilteringReactSelect reactSelect =  FilteringReactSelect.finder(_driver).followingLabelWithSpan(fieldCaption).find();
+        WebElement container = Locator.tag("td").withAttribute("data-caption", fieldCaption).findElement(this);
+        FilteringReactSelect reactSelect =  FilteringReactSelect.finder(_driver).find(container);
```
labkey-tchad: Consider adding a `FilteringReactSelect findSelect(String caption)` method to the element cache that the set and clear methods can share.

labkey-tchad (approved):

labkey-danield (approved):
I would suggest that when you make a change to these components you run tests other than the BVTs and VerifyTestBuild. Several test use this component and it might be worth it to run some tests from other modules.

PR #926, opened 2021-12-09
Fb notebook editor test

labkey-chrisj:
#### Rationale
This adds two new locator functions to BootstrapMenu's finder to support:

1. The menu with a partial toggleID (in this use case, the ID has a number that changes followed by an identifier that doesn't change from execution to execution)
2. The menu with a title attribute on the component element
image

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/1087 <-- the notebook editor's toolbar uses menus that have title attributes on their componentElement, and toggleIDs that start with generated numbers but end with invariant strings

#### Changes
Adds new finder methods on the BootstrapMenuFinder class

### DIFF
--- a/src/org/labkey/test/components/html/BootstrapMenu.java
+++ b/src/org/labkey/test/components/html/BootstrapMenu.java
@@ -198,6 +198,23 @@ public class BootstrapMenu extends BaseBootstrapMenu
             return this;
         }
 
+        /**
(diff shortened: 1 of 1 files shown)


labkey-danield (approved):

labkey-tchad (commented):
I'm hesitant to add a bunch of little one-off finder methods every time some dev finds a new variation for defining a menu. You could just use `new BootstrapMenuFinder(..).locatedBy(..)`.

PR #937, opened 2021-12-21
Avoid leaving geckodriver processes running

labkey-tchad:
#### Rationale
We run WebDriver in "detach"ed mode in order to leave browser windows open after tests. To avoid leaving driver driver processes running after tests have completed, we need to shut down the driver explicitly. This was done from [chromedriver](https://github.com/LabKey/testAutomation/commit/91243100c7acf3be2acef6c593bab531cccd32f9) but FirefoxDriver didn't support this at that time.

#### Changes
* Explicitly instantiate `GeckoDriverService` to avoid leaving zombie processes


labkey-danield (approved):
I tried to do a little homework on this with the hope of being able to better understand how this will help. I'm not sure that I did get a better understanding so I am just going to trust you on this.

PR #949, opened 2022-01-03
Various Test Fixes for Editable Grid with React Select, Assay Helper, and Editable Entity Panel

labkey-danield:
#### Rationale
Some various test fixes:

- Stamping out a few lingering issues with editable grids and the ReactSelect integration.
- In the AssayHelper, replaced the `_test.clickButton("Confirm Delete")` with a 'wait for button then click' code.
- Added a check to make sure that the entity panel is in view when editing.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/793

#### Changes
* Updated check for loading check (of ReactSelect drop down list).
* Added a check that only one domain panel is in edit mode at a time.
* Remove check in editable grid that returns an empty list if no dropdown is present. If the cell does not have a drop down the check fails.

### DIFF
--- a/src/org/labkey/test/components/ui/entities/ParentEntityEditPanel.java
+++ b/src/org/labkey/test/components/ui/entities/ParentEntityEditPanel.java
@@ -101,6 +101,14 @@ public class ParentEntityEditPanel extends WebDriverComponent<ParentEntityEditPa
         Locator infoPanel = Locator.tagWithClass("div", "panel-info");
         int infoCount = infoPanel.findElements(getDriver()).size();
 
+        if(infoCount > 1)
+        {
+            getWrapper().log("Whoa, there appears to be more than one panel in edit mode. That is odd.");
+        }
+
+        // Shouldn't need to do this, but when tests fail, because the panel did not exit edit mode, the button is not in view.
+        getWrapper().scrollIntoView(button);
+
         button.click();
 
         WebDriverWrapper.waitFor(()->
--- a/src/org/labkey/test/components/ui/entities/ParentEntityEditPanel.java
+++ b/src/org/labkey/test/components/ui/entities/ParentEntityEditPanel.java
@@ -1,5 +1,6 @@
 package org.labkey.test.components.ui.entities;
 
+import org.junit.Assert;
 import org.labkey.test.BootstrapLocators;
 import org.labkey.test.Locator;
 import org.labkey.test.WebDriverWrapper;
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/components/ui/entities/ParentEntityEditPanel.java
```diff
@@ -101,6 +101,14 @@ private void clickButtonWaitForPanel(WebElement button, int wait)
         Locator infoPanel = Locator.tagWithClass("div", "panel-info");
         int infoCount = infoPanel.findElements(getDriver()).size();
 
+        if(infoCount > 1)
+        {
+            getWrapper().log("Whoa, there appears to be more than one panel in edit mode. That is odd.");
+        }
```
labkey-tchad: Consider using the `TestLogger` to log a warning.
```suggestion
        if(infoCount > 1)
        {
            TestLogger.warn("Whoa, there appears to be more than one panel in edit mode. That is odd.");
        }
```
Has this ever happened? Should it just be an error?

src/org/labkey/test/components/ui/grids/EditableGrid.java
```diff
@@ -458,34 +458,24 @@ public EditableGrid dismissDropdownList()
...
-
-        if (div != null)
-        {
-            selectCell(gridCell);
-            waitFor(()->isCellSelected(gridCell), "Grid cell is not selected, cannot get dropdown list.", 500);
-
-            // Double click to make the ReactSelect active. This may expand the list.
-            getWrapper().doubleClick(gridCell);
+        selectCell(gridCell);
+        waitFor(()->isCellSelected(gridCell), "Grid cell is not selected, cannot get dropdown list.", 500);
```
labkey-tchad: Why doesn't `selectCell` do this wait?

labkey-tchad (approved):

src/org/labkey/test/components/ui/entities/ParentEntityEditPanel.java
labkey-danield: Long ago it was possible to have multiple panels open for edit at the same time, however that has changed.
I've updated the code to assert that there is only one panel in edit mode.

src/org/labkey/test/components/ui/grids/EditableGrid.java
labkey-danield: It does. Removed the waitFor.

PR #951, opened 2022-01-04
When calling waitForLoaded, first ensure component is present

labkey-chrisj:
#### Rationale
Recently some tests have begun to fail in the call to waitForLoaded on a reactSelect in ways that suggest that the component itself is not yet present (causing the check for the presence of the "loading..." mask or the spinner to fail because the search context isn't there yet).  
Finding components lazily can result in the situation where the componentElement isn't there yet, which can be defended in the calling context by adding `.timeout(nnn)` to the finder's call- but this change causes waitForLoaded to first check to see if the componentElement is present (if it isn't it'll fail like this) before checking to see if any of the` isLoading() `conditions are true

[here](https://teamcity.labkey.org/viewLog.html?buildId=1659818&tab=buildResultsDiv&buildTypeId=LabKey_Trunk_Premium_ProductSuites_Biologics_BiologicsBPostgres&branch_LabKey_Trunk_Premium_ProductSuites_Biologics=%3Cdefault%3E#testNameId-2172338325796289313) is an example of such a failure

#### Related Pull Requests
n/a

#### Changes
don't ask if isLoading() is true until after confirming the componentElement is displayed

### DIFF
--- a/src/org/labkey/test/components/react/BaseReactSelect.java
+++ b/src/org/labkey/test/components/react/BaseReactSelect.java
@@ -111,7 +111,7 @@ public abstract class BaseReactSelect<T extends BaseReactSelect<T>> extends WebD
 
     protected T waitForLoaded()
     {
-        waitFor(() -> !isLoading(),
+        waitFor(() -> getComponentElement().isDisplayed() && !isLoading(),
                 "Took too long for to become loaded", WAIT_FOR_JAVASCRIPT);
         return getThis();
     }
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/components/react/BaseReactSelect.java
```diff
@@ -477,7 +477,7 @@ private Locators()
     {
         private Locator.XPathLocator _locator;
         private boolean _mustBeEnabled = false;
-        private boolean _findParent = true;
+        private final boolean _findParent = true;
```
labkey-tchad: Just remove this (and the `if` that references it).
The if is a no-op:
```
if (_findParent)
    tmpLoc.parent();
```

labkey-tchad (approved):

src/org/labkey/test/components/react/BaseReactSelect.java
labkey-chrisj: OK, I'll make this change and push, then monitor results before merging

PR #954, opened 2022-01-05
Add new method for finding a pattern within a file

labkey-tchad:
#### Rationale
Need more than just the line number for some scenarios.

#### Related Pull Requests
* https://github.com/FDA-MyStudies/Response/pull/58
* https://github.com/FDA-MyStudies/UserReg-WS/pull/45

#### Changes
* Use a regex to extract matching text from a file

### DIFF
--- a/src/org/labkey/test/io/Grep.java
+++ b/src/org/labkey/test/io/Grep.java
@@ -60,7 +60,7 @@ public class Grep
     }
 
     /**
-     * Find all matches to the given pattern. Also includes line numbers for the given pattern.
+     * Find all matches to the given pattern. Also includes line numbers for each match.
      */
     public static List<Pair<String, Integer>> findMatches(File file, Pattern pattern, boolean firstMatch) throws IOException
     {
--- a/src/org/labkey/test/io/Grep.java
+++ b/src/org/labkey/test/io/Grep.java
@@ -15,12 +15,15 @@
  */
 package org.labkey.test.io;
 
+import org.apache.commons.lang3.tuple.Pair;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/io/Grep.java
```diff
@@ -51,6 +54,18 @@
 
     private static int grep(File file, Pattern pattern) throws IOException
     {
+        final List<Pair<String, Integer>> matches = findMatches(file, pattern, true);
+
+        return matches.stream().findFirst().map(Pair::getRight).orElse(-1);
+    }
+
+    /**
+     * Find all matches to the given pattern. Also includes line numbers for the given pattern.
```
labkey-adam:
```suggestion
     * Find all matches to the given pattern. Also includes line numbers for each match.
```

src/org/labkey/test/io/Grep.java
```diff
@@ -51,6 +54,18 @@
...
+     */
+    public static List<Pair<String, Integer>> findMatches(File file, Pattern pattern, boolean firstMatch) throws IOException
+    {
+        List<Pair<String, Integer>> matches = new ArrayList<>();
+
         try (BufferedReader is = Readers.getReader(file))
```
labkey-adam, line 69: Pre-existing code, but should this be `br` intead of `is`? Also, there is `BufferedReader.lines()`, which returns a `Stream<String>`. Maybe not as handy in this specific case because of the short-circuit and line numbers. And the fact that what you have works fine. 😄

labkey-adam (approved):

PR #957, opened 2022-01-11
Migrate study archives to folder archives

labkey-adam:
#### Rationale
Study archives are going away


labkey-jeckels (approved):
Do we want to keep these as .zip files? It's valuable to have test coverage of unzipping archives, but they're a much bigger pain to maintain than exploded directories.

labkey-adam:
Probably a question for @labkey-tchad . Personally, I use 7zip and WinZip, which allow editing zipped content in place (without unzipping/rezipping), so it doesn't make much difference to me. We don't have full history for these, but I suspect most of these are being touched for the first time in about 10 years.

PR #964, opened 2022-01-14
Pre-check URLs for downloads before attempting navigation

labkey-tchad:
#### Rationale
The Firefox `browser.helperApps.neverAsk.saveToDisk` parameter isn't functioning correctly for some XML files under Firefox 91. A browser download dialog appears, which WebDriver is unable to interact with, causing the crawler fail. Navigating to downloads isn't necessary for crawler coverage so we can just check the "Content-Disposition" of the HTTP response to skip URLs that would trigger a download.
This change to the crawler also makes unnecessary the `beginAt` that permits unexpected downloads.

#### Related Pull Requests
* #749 

#### Changes
* Use direct HTTP request to identify and skip download URLs
* Remove `beginAt` that permits downloads

### DIFF
--- a/src/org/labkey/test/WebDriverWrapper.java
+++ b/src/org/labkey/test/WebDriverWrapper.java
@@ -19,10 +19,8 @@ import org.apache.commons.collections4.MultiValuedMap;
 import org.apache.commons.collections4.multimap.HashSetValuedHashMap;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.SystemUtils;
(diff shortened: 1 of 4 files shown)


labkey-danield (approved):

labkey-tchad:
Daily E likes this change: https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailySuites_DailyDPostgres/1704040?showRootCauses=false&expandBuildChangesSection=true&expandBuildTestsSection=true

PR #966, opened 2022-01-17
Wait for spinners to disappear before returning from save button click.

labkey-danield:
#### Rationale
Updating details of a sample can cause other panels on a sample's overview page to reload. This checks and waits for any spinners that may be present (evidence of a panel reloading) before returning from clicking the save button.

#### Related Pull Requests
* None

#### Changes
* Added a waitFor spinners check in the clickSave method.


labkey-chrisj (approved):
There may be a race condition in which the checker looks for the spinner before it appears and returns true to unblock the thread (which will elude this wait), so the worst that can happen is we don't wait (which is what the code currently does).

labkey-danield:
Thanks for reviewing @labkey-chrisj . That was my thinking as well (if it doesn't work it is no worse than the current behavior).

PR #968, opened 2022-01-18
Add WaitForLoaded After Changing Filtering/Ordering in a Table

labkey-danield:
#### Rationale
Changing things like filtering or ordering can cause a table content to refresh. After calling a method that will modify the filter or ordering on a table the method should wait for the table to refresh before returning.

#### Related Pull Requests
* None

#### Changes
* Added a call to waitForLoaded to search, sort and filter methods on the table.


labkey-tchad (commented):
Not sure this will make any difference; `doAndWaitForUpdate` already calls `waitForLoaded` (via `waitForUpdate`)

PR #972, opened 2022-01-19
Fix crawler's identification of download URLs

labkey-tchad:
#### Rationale
The crawler now attempts to identify and skip download URLs without involving the browser. It does so by checking the HTTP response header of the URL. There have been two sorts of failures from this crawler change.
1. Some URLs have illegal characters in their query (looks like spaces are the problem). Browsers tend to automatically encode various parts of the URL when needed; the crawler needs to do similarly (`' '` -> `'+'`).
2. Some downloads use the "inline" content disposition instead of "attachment"; the crawler needs to check for these cases.

#### Related Pull Requests
* #964 

#### Changes
* Properly encode URLs before checking HTTP response
* Stop tracking and cleaning up crawler downloads

### DIFF
--- a/src/org/labkey/test/util/Crawler.java
+++ b/src/org/labkey/test/util/Crawler.java
@@ -856,6 +856,12 @@ public class Crawler
         {
             url = WebTestHelper.getBaseURL() + url;
         }
+        final String[] splitUrl = url.split("\\?", 2);
+        if (splitUrl.length > 1)
+        {
+            // Properly encode spaces in the URL query
+            url = splitUrl[0] + "?" + splitUrl[1].replace(" ", "+");
+        }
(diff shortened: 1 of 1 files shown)


labkey-chrisj (approved):
Thanks for doing this!

labkey-tchad:
Not going to make the same mistake again. Waiting for [TeamCity to verify](https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailySuites_DailyDPostgres/1677609?showRootCauses=false) before merging.

labkey-tchad:
Not quite right. [Trying again](https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailySuites_DailyDPostgres/1677914?showRootCauses=false&expandBuildChangesSection=true).

PR #985, opened 2022-01-26
API test support for Issues

labkey-chrisj:
#### Rationale
This adds support for tests to interact with issues-issues.api and issues-getIssue.api, an API helper, and a couple of new tests

#### Related Pull Requests
platform


src/org/labkey/remoteapi/issues/IssueModel.java
```diff
@@ -0,0 +1,213 @@
...
+    private final String ISSUE_DEF_NAME = "issueDefName";
+    private final String ASSIGNED_TO = "assignedto";
+    private final String ACTION = "action";
+    private final String TYPE = "type";
+    private final String PRIORITY = "priority";
+    private final String COMMENT = "comment";
+    private final String NOTIFY_LIST = "notifyList";
+    private final String RESOLUTION = "resolution";
+    private final String RESOLVED = "resolved";
+    private final String STATUS = "status";
```
labkey-klum: sometimes `enums` are easier to use than `String` constants

src/org/labkey/remoteapi/issues/IssueModel.java
```diff
@@ -0,0 +1,213 @@
...
+
+        IssueAction(String value)
+        {
+            _value = value;
+        }
+        public String getValue()
+        {
+            return _value;
+        }
+        private final String _value;
```
labkey-klum: What's the value of exposing a `getValue` method, couldn't the `enum` just be the action literals? `Enums` also override `toString` so they can be used directly in places where a `String` is expected.

src/org/labkey/remoteapi/issues/IssueResponse.java
```diff
@@ -0,0 +1,36 @@
...
+     * @param contentType   The response content type
+     * @param json          The parsed JSONObject (or null if JSON was not returned).
+     * @param sourceCommand A copy of the command that created this response
+     */
+    public IssueResponse(String text, int statusCode, String contentType, JSONObject json, IssueCommand sourceCommand)
+    {
+        super(text, statusCode, contentType, json, sourceCommand);
+
+        JSONArray issuesArray = (JSONArray)json.get("issues");
+        _issueId = ((Long)issuesArray.get(0));   // just take the first
```
labkey-klum: are we guaranteed to get back at least one issueId?

src/org/labkey/remoteapi/issues/IssueResponseModel.java
```diff
@@ -0,0 +1,103 @@
...
+    public String getType()
+    {
+        return (String) _serverProps.get(TYPE);
+    }
+
+    public List<IssueComment> getComments()
+    {
+        return _issueComments;
+    }
+}
```
labkey-klum, line 130: Maybe expose a method : `getProperty(String name)` to retrieve adhoc properties off of the response model.

src/org/labkey/remoteapi/issues/GetIssueCommand.java
```diff
@@ -0,0 +1,29 @@
+package org.labkey.remoteapi.issues;
+
+import org.json.simple.JSONObject;
+import org.labkey.remoteapi.PostCommand;
+
+public class GetIssueCommand extends PostCommand<GetIssueResponse>
```
labkey-tchad, line 6: This doesn't need to be a `PostCommand`. Just a regular `Command` will work. The 'issueId' can be a URL parameter; it doesn't need to be passed in as JSON.

src/org/labkey/test/tests/issues/IssueAPITest.java
```diff
@@ -0,0 +1,280 @@
...
+@Category({InDevelopment.class})
+public class IssueAPITest extends BaseWebDriverTest
+{
+    IssuesHelper _issuesHelper = new IssuesHelper(this);
+    APIUserHelper _userHelper = new APIUserHelper(this);
+    static String ISSUES = "issues";
+    static Integer TEST_USER_ID;
+    static String TEST_USER_NAME;
+    static String TEST_USER_DISPLAY_NAME;
+
```
labkey-klum, line 45: Why are these class variables static?

src/org/labkey/test/tests/issues/IssueAPITest.java
```diff
@@ -0,0 +1,280 @@
...
+        String title = "Insert Issue Test Issue";
+        IssueModel testIssue = basicIssueModel(title, comment);
+
+        IssueCommand cmd = new IssueCommand();
+        cmd.setIssue(testIssue);
+        IssueResponse response = cmd.execute(createDefaultConnection(), getProjectName());
+        assertNotNull("expect our issue response to tell us an issue ID", response.getIssueId());
+
+        var issuePage = DetailsPage.beginAt(this, getProjectName(), response.getIssueId().toString());
+        var issueComment = issuePage.getComments().get(0);
```
labkey-klum: Should we be using the UI to validate the inserted issue? Could we instead request the new issue using your new API command and compare the response with expected?

src/org/labkey/remoteapi/issues/IssueModel.java
```diff
@@ -0,0 +1,213 @@
...
+    {
+        _properties.put(TITLE, title);
+        return this;
+    }
+
+    // per https://www.labkey.org/Documentation/wiki-page.view?name=sampleJSscripts#issues, for a single notify user
+    // you can supply an email address as a string; for multiple notifies, userIDs are necessary
+    public String getNotify()
+    {
+        return (String) _properties.get("notifyList");
```
labkey-tchad: Should this use the constant used by the setter?
```suggestion
        return (String) _properties.get(NOTIFY_LIST);
```

src/org/labkey/test/tests/issues/ApiIssuesTest.java
```diff
@@ -0,0 +1,21 @@
...
+import org.labkey.test.categories.Data;
+import org.labkey.test.categories.Issues;
+import org.labkey.test.util.IssuesApiHelper;
+
+@Category({Issues.class, Daily.class, Data.class})
+@BaseWebDriverTest.ClassTimeout(minutes = 20)
+/**
+ * A version of the issues test that uses the API-based version of the helper.
+ */
+public class ApiIssuesTest extends IssuesTest
```
labkey-tchad, line 15: I don't have a better name but having both `ApiIssuesTest` and `IssuesAPITest` seems confusing

src/org/labkey/test/util/IssuesApiHelper.java
```diff
@@ -0,0 +1,69 @@
...
+    private final Connection _connection;
+    private final APIUserHelper _userHelper;
+    private final Long _defaultPriority = 3L;
+
+    public IssuesApiHelper(WrapsDriver driverWrapper)
+    {
+        super(driverWrapper);
+        _connection = WebTestHelper.getRemoteApiConnection(true);
+        _userHelper = new APIUserHelper(this);
+    }
```
labkey-tchad: There's no reason to cache the connection, `createDefaultConnection` is available from this class.
Caching the connection will just cause confusing errors if a test logs out at some point.
```suggestion
    private final APIUserHelper _userHelper;
    private final Long _defaultPriority = 3L;

    public IssuesApiHelper(WrapsDriver driverWrapper)
    {
        super(driverWrapper);
        _userHelper = new APIUserHelper(this);
    }
```

src/org/labkey/remoteapi/issues/IssueCommand.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.http.entity.mime.HttpMultipartMode;
+import org.apache.http.entity.mime.MultipartEntityBuilder;
+import org.json.JSONArray;
+import org.json.simple.JSONObject;
+import org.labkey.remoteapi.PostCommand;
+
+import java.io.File;
+import java.net.URI;
+
+public class IssueCommand extends PostCommand<IssueResponse>
```
labkey-tchad: Command name should match the action name.
```suggestion
public class IssuesCommand extends PostCommand<IssueResponse>
```

labkey-tchad (changes requested):
The command classes should match the actual endpoint. Don't obfuscate the fact that they operate on multiple issues at once. `IssuesCommand` should take a list of issue models and the response should expose a list of issue IDs.
The API helper would be a more appropriate place to pare it down to single issue operations.

src/org/labkey/remoteapi/issues/IssueResponse.java
labkey-chrisj: That's my assumption; this could instead serve a list of issueId that might be 0 length

src/org/labkey/remoteapi/issues/IssueModel.java
labkey-chrisj: I put the getter there not thinking that .toString() would be good form; if it's preferable I don't have opinions either way

src/org/labkey/remoteapi/issues/IssueModel.java
labkey-chrisj: I'm happy to put these into an enum!

src/org/labkey/test/tests/issues/IssueAPITest.java
labkey-chrisj: so I didn't have to reference the instance while setting up the class, no other reason

src/org/labkey/test/tests/issues/IssueAPITest.java
labkey-chrisj: switched completely to using the API now

src/org/labkey/remoteapi/issues/IssueResponseModel.java
labkey-chrisj: done

src/org/labkey/remoteapi/issues/GetIssueCommand.java
labkey-chrisj: If having it not be a PostCommand would have been less work, that's cool- but the work is done and changing it now would add work for (I'm not sure how much?) benefit.  Also I'm less familiar with `Command `than I am `PostCommand`

src/org/labkey/remoteapi/issues/IssueModel.java
labkey-chrisj: Yes, it should- these keys now use enum values tho

src/org/labkey/test/tests/issues/ApiIssuesTest.java
labkey-chrisj: You're not wrong

src/org/labkey/test/util/IssuesApiHelper.java
labkey-chrisj: done, this doesn't cache connections any more

src/org/labkey/remoteapi/issues/IssueCommand.java
labkey-chrisj: done

labkey-chrisj:
I thought it was interesting that I couldn't insert an issue and assign it to a user in the same operation; only if I'd inserted it and then acted on it would it work to assign- I'm assuming the UI creates the issue in the background before updating it somehow

labkey-tchad:
> I thought it was interesting that I couldn't insert an issue and assign it to a user in the same operation; only if I'd inserted it and then acted on it would it work to assign- I'm assuming the UI creates the issue in the background before updating it somehow

That sounds like a bug. Creating an issue through the UI is definitely a single-step operation. @labkey-klum ?

The API does respect the assignedTo field during an insert, the user needs to have `UpdatePermission` on the container else you should receive a validation error.

src/org/labkey/remoteapi/issues/IssueModel.java
labkey-tchad: Get rid of the `_value` fields in both of these enums. They aren't even being used anywhere.

src/org/labkey/remoteapi/issues/IssueResponseModel.java
labkey-tchad: Get rid of unused `_value` field.

src/org/labkey/remoteapi/issues/IssuesCommand.java
```diff
@@ -0,0 +1,72 @@
...
+            MultipartEntityBuilder builder = MultipartEntityBuilder.create();
+
+            builder.setMode(HttpMultipartMode.BROWSER_COMPATIBLE);
+            builder.addTextBody("issues", issuesArray.toString(), ContentType.APPLICATION_JSON);
+            for (File attachment : issue.getAttachments())
+            {
+                builder.addBinaryBody(attachment.getName(), attachment, ContentType.APPLICATION_OCTET_STREAM, attachment.getName());
+            }
+            request.setEntity(builder.build());
+        }
```
labkey-tchad: Need to attach all issue models simultaneously. Trying to add them one at a time will include the only final issue model as each subsequent issue will clobber the previous.
```suggestion
        JSONArray issuesArray = new JSONArray();
        for (IssueModel issue: _issues)
        {
            issuesArray.add(issue.toJSON());
        }

        MultipartEntityBuilder builder = MultipartEntityBuilder.create();
        builder.setMode(HttpMultipartMode.BROWSER_COMPATIBLE);
        builder.addTextBody("issues", issuesArray.toString(), ContentType.APPLICATION_JSON);
        for (IssueModel issue: _issues)
        {
            for (File attachment : issue.getAttachments())
            {
                builder.addBinaryBody(attachment.getName(), attachment, ContentType.APPLICATION_OCTET_STREAM, attachment.getName());
            }
        }
        request.setEntity(builder.build());
```
Not sure how this would react if you tried to attach multiple files with the same name or the same file to multiple issues. (Probably not a high priority)

src/org/labkey/test/tests/issues/IssueAPITest.java
```diff
@@ -0,0 +1,377 @@
...
+import java.util.List;
+
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.labkey.test.util.PasswordUtil.getUsername;
+
+@Category({InDevelopment.class})
```
labkey-tchad: Remove 'InDevelopment' category
```suggestion
@Category({})
```

src/org/labkey/test/tests/issues/IssueAPITest.java
```diff
@@ -0,0 +1,377 @@
...
+
+    static Long TEST_BUDDY_ID;
+    static String TEST_BUDDY_NAME = "testbuddy@issues.test";
+    static String TEST_BUDDY_DISPLAY_NAME;
+
+    @Override
+    protected void doCleanup(boolean afterTest) throws TestTimeoutException
+    {
+        super.doCleanup(afterTest);
+    }
```
labkey-tchad: Remove empty cleanup method
```suggestion
```

src/org/labkey/test/tests/issues/IssueAPITest.java
```diff
@@ -0,0 +1,377 @@
...
+    private void doSetup()
+    {
+        _containerHelper.createProject(getProjectName(), null);
+        _issuesHelper.createNewIssuesList(ISSUES, getContainerHelper());
+
+        TEST_USER_NAME = getUsername();
+        TEST_USER_ID = Long.valueOf(_userHelper.getUserId(TEST_USER_NAME));
+        TEST_USER_DISPLAY_NAME = _userHelper.getDisplayNameForEmail(TEST_USER_NAME);
+        _userHelper.createUser(TEST_BUDDY_NAME);
+        TEST_BUDDY_ID = Long.valueOf(_userHelper.getUserId(TEST_BUDDY_NAME));
```
labkey-tchad: `createUser` returns an object containing the user ID.
```suggestion
        TEST_BUDDY_ID = _userHelper.createUser(TEST_BUDDY_NAME).getUserId().longValue();
```

src/org/labkey/test/tests/issues/IssueAPITest.java
labkey-tchad: We make these sort of things static so that multiple test cases can share them. JUnit creates a new class instance for each test case.

src/org/labkey/test/tests/issues/IssueAPITest.java
```diff
@@ -0,0 +1,377 @@
...
+        issues.add(secondIssue);
+        issues.add(thirdIssue);
+
+        IssuesCommand insertCmd = new IssuesCommand();
+        insertCmd.setIssues(issues);
+        var insertResponse = insertCmd.execute(createDefaultConnection(), getProjectName());
+        List<Long> issueIds =  insertResponse.getIssueIds();
+        for (Long issueId : issueIds)
+        {
+            var issueresponseModel = getIssueResponse(issueId);
```
labkey-tchad: Blindly validating in a loop like this is **BAD**!
The insert command might return nothing, and you validate nothing.
It might return the same issue ID three times, or a dozen times, and you just validate the same repeatedly.

I bet you this command is only inserting a single issue due to a bug in the `IssuesCommand` and the test is totally missing it.

labkey-tchad (changes requested):
Could probably use some tests that exercise the issues action's error handling.
Missing required properties, invalid property values, resolving a closed issue (not actually sure that this sort of thing would throw an error)

src/org/labkey/remoteapi/issues/IssuesCommand.java
labkey-chrisj: That doesn't seem to be the observed behavior- I added testUpdateMultipleIssues() to make sure that individual issues are treated according to expectation- that they each get the expected attachments and whatnot etc.  But, I'll gladly try what you've suggested, on the expectation that you understand how MultipartEntityBuilder works much better than I do

src/org/labkey/test/tests/issues/IssueAPITest.java
labkey-chrisj: dang, I forget that so often/need facepalm emoji here

labkey-tchad: I'm just going to remove that from the template file. It is a relic of our pre-git days. InDevelopment tests now live in feature branches; they don't need a special category.

src/org/labkey/test/tests/issues/IssueAPITest.java
labkey-chrisj: very cool!  I like that a lot

src/org/labkey/test/tests/issues/IssueAPITest.java
labkey-chrisj: Well, when I debugged this, it hit every case.  I can add validation that there were 3 issues in the response to catch the empty response case tho

src/org/labkey/remoteapi/issues/IssuesCommand.java
labkey-tchad: Oh. I see. It does add all of the issue models to the request. Each time through the loop sets a bigger array of issues to the request. It probably misses the attachments from earlier issues though

src/org/labkey/test/tests/issues/IssueAPITest.java
labkey-chrisj: Well, this was a good catch- it's returning the same IssueId, 3 times.  I've added a check for distinctness of returned Ids that does catch it.  Now the question is how to actually send all 3 issues

labkey-chrisj: It turns out the bug was in the deserializing- the API was sending back 3 IDs, IssueResponse was writing down the first one 3 times (again, need the facepalm emoji something fierce here)

src/org/labkey/remoteapi/issues/IssueModel.java
```diff
@@ -0,0 +1,217 @@
...
+    {
+        return setProp(ModelKeys.action, action);
+    }
+
+    public Long getIssueId()
+    {
+        if (getProp(ModelKeys.issueid) != null)
+            return (Long) getProp(ModelKeys.issueid);
+        else
+            return null;
```
labkey-tchad: This `null` check isn't necessary. Java lets you cast a `null` to any non-primitive type you please.

src/org/labkey/remoteapi/issues/IssueModel.java
```diff
@@ -0,0 +1,217 @@
...
+        comment(),
+        notifyList(),
+        resolution(),
+        resolved(),
+        status();
+
+        ModelKeys()
+        {
+
+        }
```
labkey-tchad: Don't need a constructor for simple enums.
```suggestion
        title,
        issueid,
        issueDefName,
        assignedto,
        action,
        type,
        priority,
        comment,
        notifyList,
        resolution,
        resolved,
        status,
```

src/org/labkey/remoteapi/issues/IssueModel.java
```diff
@@ -0,0 +1,217 @@
...
+        update("update"),
+        resolve("resolve"),
+        close("close"),
+        reopen("reopen");
+
+        IssueAction(String value)
+        {
+            _value = value;
+        }
+        private final String _value;
```
labkey-tchad:
```suggestion
        insert,
        update,
        resolve,
        close,
        reopen,
```

src/org/labkey/remoteapi/issues/IssueResponseModel.java
```diff
@@ -0,0 +1,132 @@
...
+        resolved(),
+        ResolvedBy(),
+        Closed(),
+        ClosedBy(),
+        Status(),
+        Modified(),
+        ModifiedBy();
+        
+        ResponseKeys(){}
+    }
```
labkey-tchad:
```suggestion
    public enum ResponseKeys{
        Title,
        IssueId,
        issueDefName,
        AssignedTo,
        type,
        priority,
        comment,
        notifyList,
        resolution,
        resolved,
        ResolvedBy,
        Closed,
        ClosedBy,
        Status,
        Modified,
        ModifiedBy,
    }
```

src/org/labkey/test/tests/issues/IssueAPITest.java
```diff
@@ -0,0 +1,492 @@
...
+                .setAction(IssueModel.IssueAction.insert)
+                .setType("Defect");
+
+        try{
+            doIssueAction(issue);
+            fail("expect error if command is executed without required property priority");
+        }catch(CommandException success)
+        {
+            assertThat(success.getMessage(), containsString("Data does not contain required field: priority"));
+            resetErrors();
```
labkey-tchad, line 443: I wouldn't expect a missing field to log a server error. This seems like a bug.
@labkey-klum ?

labkey-tchad (approved):
Just some nits

src/org/labkey/test/tests/issues/IssueAPITest.java
labkey-klum: Correct, the server should respond with a 400 not throw. It looks like it's not catching this in validation but later during the commit. Probably has something to do with the difference between a UI and API post.

If you posted this instead : `priority : null` you would probably see a different error.

@labkey-chrisj I would open an issue and we can remove this code once we shore up validation.

labkey-chrisj: OK, I've opened https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44784
I'll add a note in the test code along with other feedback

PR #986, opened 2022-01-26
Ignore crawler 404 from bad GETs to form actions

labkey-tchad:
#### Rationale
`StudyDemoModeTest` has been hitting crawler failures from a url located in `<form>`. If you go directly to the target of that form without selecting appropriate options, the page returns a 404. These sorts of 404s should be ignored. In order to avoid ignoring 404s from references to non-existent actions, we can use an unsupported request type (`HEAD`) to check whether the endpoint exists at all and fail if the form references a non-existent action.

```
java.lang.AssertionError: /StudyDemoMode%20Project/My%20Study/experiment-deriveSamples.view?&rowIds=3915&inputRole0=&customRole0=&outputCount=1&targetSampleTypeId=
produced response code 404
Originating page: http://localhost:8111/labkey/StudyDemoMode%20Project/My%20Study/experiment-deriveSamplesChooseTarget.view?rowIds=3915&targetSampleTypeId=8
  at org.labkey.test.util.Crawler.crawlLink(Crawler.java:970)
  at org.labkey.test.util.Crawler.crawl(Crawler.java:817)
  at org.labkey.test.util.Crawler.crawlAllLinks(Crawler.java:780)
  at org.labkey.test.tests.study.StudyDemoModeTest.doVerifySteps(StudyDemoModeTest.java:71)
```

[Verifying on TeamCity](https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailySuites_DailyDPostgres/1686158?showRootCauses=false&expandBuildChangesSection=true) to make sure it doesn't a'splode.

#### Changes
* Ignore crawler 404 from bad GETs to form actions


labkey-tchad:
Seems to have cleared up most of the [failures on TC](https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailySqlServer/1687367?showRootCauses=false&expandBuildProblemsSection=true&expandBuildTestsSection=true&expandBuildChangesSection=true).

labkey-chrisj (approved):
let's hope this works

PR #992, opened 2022-02-02
Name Expression Preview Test Automation

labkey-danield:
#### Rationale
Added test automation to test the preview functionality of the sample name expression (naming pattern).

#### Related Pull Requests
* None.

#### Changes
* Added preview component to page.
* Added test to verify the preview functionality of naming patters for samples.


src/org/labkey/test/tests/SampleTypeNameExpressionTest.java
```diff
@@ -232,4 +286,298 @@ private void verifyNames(String sampleTypeName, String header, String nameExpres
...
+     * <p>
+     *     This test will:
+     *     <ul>
+     *         <li>Create a derived sample using the UI and the name expression to name it.</li>
+     *     </ul>
+     * </p>
+     * @throws Exception Can be thrown by test helper.
+     */
+    @Test
+    public void testCreateDerivedThroughUI() throws Exception
```
XingY: Perhaps rename the test and also modify comments to mention "lineage lookup field". I don't think issue 44760 has to do with creating samples from UI, but rather, it has to do with the latest lineage lookup name expression feature.

src/org/labkey/test/tests/SampleTypeNameExpressionTest.java
```diff
@@ -232,4 +286,298 @@ private void verifyNames(String sampleTypeName, String header, String nameExpres
...
+    {
+
+        // This test is for Issue 44760.
+        goToProjectHome();
+
+        SampleTypeHelper sampleHelper = new SampleTypeHelper(this);
+
+        final String sampleType = "DerivedUI_SampleType";
+        final String nameExpression = String.format("DUI_${genId}_${materialInputs/%s/Str}", PARENT_SAMPLE_TYPE);
```
XingY, line 313: Once the fix is in, can also add a test to use parent alias here.

src/org/labkey/test/tests/SampleTypeNameExpressionTest.java
```diff
@@ -232,4 +286,298 @@ private void verifyNames(String sampleTypeName, String header, String nameExpres
...
+        if(expectedPreview != null)
+        {
+            if(!error)
+            {
+                expectedToolTip.append("Example of name that will be generated from the current pattern: ");
+            }
+
+            expectedToolTip.append(expectedPreview);
+            expectedToolTip.append("\n");
+        }
```
XingY, line 380: if there is error, wouldn't there be "Unable to generate example name from the current pattern. Check for syntax errors." msg?

labkey-danield: Yes it does.
In that case the error message is the value sent in. Example:
`String expectedMsg = generateExpectedToolTip("Unable to generate example name from the current pattern. Check for syntax errors.", true);`
The testErrorsAndWarning test has more examples.
I will update the comment to clarify.

src/org/labkey/test/tests/SampleTypeNameExpressionTest.java
labkey-danield: I added a TODO comment and another link to the issue.

src/org/labkey/test/tests/SampleTypeNameExpressionTest.java
labkey-danield: I updated the function name to `testDeriveSampleFromSampleDetailsPage`.
As far as I can tell all the other tests in this file use import to create derived samples, this is the only one (in this test class) that uses the UI. Because of that I wanted to make sure the test name conveys that info.
I have added clarify text to the issue comments.

src/org/labkey/test/tests/SampleTypeNameExpressionTest.java
XingY: hmm, is there any reason to not have the function check for the error string without having to pass as param? I think this is a static msg that doesn't change based on error reason?

XingY (approved):

labkey-chrisj (approved):

PR #995, opened 2022-02-04
Updated Comment

labkey-danield:
#### Rationale
Update the comment for the SecurityApiTest. Reworded the fix hint and tried a bit more.

#### Related Pull Requests
* None

#### Changes
* Updated comment.

### DIFF
--- a/src/org/labkey/test/tests/SecurityApiTest.java
+++ b/src/org/labkey/test/tests/SecurityApiTest.java
@@ -111,23 +111,33 @@ public class SecurityApiTest extends BaseWebDriverTest
      *
      * The failed output from the test can be difficult to read. Especially if the failure was between expected and actual
      * roles and permissions. If that is the type of failure you are seeing the failure message is a rough diff between
-     * the expected JSON and the JSON that was returned. If you run the test locally from the command line you might be
-     * better able to understand what the differences are. After running locally you should see a message at the end of
-     * the run output that looks like this:
+     * the expected JSON and the JSON that was returned. Unfortunately the code that checks for equality and the code
+     * that generates the diff-error output are two different functions. The comparison code flags the two JSON objects as
+     * not being equal. That failed check then calls the error output code which converts the JSON objects to strings.
+     * It compares these two string line by line, and shows all the differences, including differences in order,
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/SecurityApiTest.java
```diff
@@ -102,30 +102,34 @@ protected void doCleanup(boolean afterTest) throws TestTimeoutException
...
+     * A little ways down is a comment "Actual:" with the actual output (go figure). It might be easier to compare the
+     * expected vs. actual results from this log output.
      *
-     * The expected output comes from the security-api.xml file. Odds are you are going to have to change that file in
-     * some way. You didn't hear it from me but the easiest fix could be to replace the contents of the security-api.xml
-     * file with the actual output from the log. Obviously the smaller the change the better, but as a last resort you
-     * could try that as a fix.
+     * The expected output comes from the security-api.xml file. If this is a role and/or permission error odds are you
+     * will need to make a change to that file. Try to make the smallest change possible. If you are changing permissions,
+     * either adding or removing, you will need to update the value(s) in the effectivePermissions collection for a given role.
```
labkey-susanh, line 133: Perhaps add a note here that the test, despite indications from the diff output, the order of the strings in the effectivePermissions collection does not matter.

labkey-susanh (approved):

src/org/labkey/test/tests/SecurityApiTest.java
labkey-danield: Good call. I added a comment in the paragraph above. That paragraph now reads as:

The failed output from the test can be difficult to read. Especially if the failure was between expected and actual
roles and permissions. If that is the type of failure you are seeing the failure message is a rough diff between
the expected JSON and the JSON that was returned. Unfortunately the code that checks for equality and the code
that generates the diff-error output are two different functions. The comparison code flags the two JSON objects as
not being equal. That failed check then calls the error output code which converts the JSON objects to strings. 
It compares these two string line by line, and shows all the differences, including differences in order, 
WHICH DOESN'T MATTER. This adds a lot of "noise" to the error output in the log that makes it hard to find the 
actual inconsistency you are looking for.

labkey-susanh: You're writing a novel here. :-)

PR #996, opened 2022-02-07
SimpleModuleTest: update workbookTest for cross-folder lists

labkey-nicka:
#### Rationale
Support for cross-folder data in Lists was recently introduced. With this the server-side handling of row-level specification of a container is also supported. This updates the test cases run at `simpletest-workbookTest.view` to account for this behavior.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3001

#### Changes
- Update the expected error for writing a row to a list in a separate folder/container.
- Add `cleanUp()` step to test process so test can be run multiple times.

### DIFF
--- a/modules/simpletest/resources/views/workbookTest.html
+++ b/modules/simpletest/resources/views/workbookTest.html
@@ -10,10 +10,13 @@ target folder, that should be respected.  The same is true for both updateRows a
 <script type='text/javascript'>
 
     function testContainerPath() {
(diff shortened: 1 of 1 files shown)


cnathe (approved):

labkey-nicka:
`SimpleModuleTest` passing with https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyPostgres/1700300?showRootCauses=true&expandBuildChangesSection=true

PR #998, opened 2022-02-08
AssayExportImportTest fix by adding some time between assay run creation

cnathe:
#### Rationale
As mentioned in the TeamCity triage note for this failure, it looks like if "Save and Import Another Run" is used by the tests for multiple assay run creation, it can go so fast the the temp file names used for the assay data aren't unique and some runs are stomping on each other. This PR isn't perfect, but it adds a 1s sleep between assay run creation since it is unlikely that actual users could click this fast to create assay runs in sequence.

### DIFF
--- a/src/org/labkey/test/tests/AssayExportImportTest.java
+++ b/src/org/labkey/test/tests/AssayExportImportTest.java
@@ -202,6 +202,7 @@ public class AssayExportImportTest extends BaseWebDriverTest
         int fileIndex = 0;
         for(Map<String, String> runProperty : runProperties)
         {
+            sleep(1000); // give it a second to finish the import of the last run in the "Save and Import Another Run" case
             runProperty.keySet().forEach((property)->setFormElement(Locator.name(property), runProperty.get(property)));
 
             if(!useFilesWebPart)


labkey-tchad (approved):
Comment should mention the file name collision.
Seems fine but are we sure we don't need a product fix. Could some sort of automated/bulk assay import script trigger this?

labkey-danield (approved):

cnathe:
> Comment should mention the file name collision. Seems fine but are we sure we don't need a product fix. Could some sort of automated/bulk assay import script trigger this?

Good point, Trey. You mentioned something along those lines in the TC test failure note. I'll leave this PR open for a bit while I look into where and how the product is generating those file names to see if we can do something on that end to prevent the collision.

cnathe:
Created a new PR with a platform change to add milliseconds to the assay run temp file name. If that PR gets approved / merged, I'll close out this one:
https://github.com/LabKey/platform/pull/3028

cnathe:
Closing PR since https://github.com/LabKey/platform/pull/3028 fixes things on the product side.

PR #1000, opened 2022-02-09
Add property to force local API dependency on TeamCity

labkey-tchad:
#### Rationale
We pull API schemas from Artifactory to avoid building the API module before each test suite. Sometimes we might need to use the cutting edge version of the labkey schemas. 

#### Related Pull Requests
* https://github.com/LabKey/server/pull/204

#### Changes
* Add test dependency override mechanism
* Update test's dependency on the LabKey API module

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -8,4 +8,4 @@ jettyRepackagedVersion=9.4.12.v20180830
 seleniumVersion=4.1.1
 mockserverNettyVersion=5.5.1
 
-labkeySchemasTestVersion=21.7.0
+labkeySchemasTestVersion=22.3-SNAPSHOT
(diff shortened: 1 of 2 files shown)


labkey-adam (approved):

gradle.properties
```diff
@@ -8,4 +8,4 @@ jettyRepackagedVersion=9.4.12.v20180830
 seleniumVersion=4.1.1
 mockserverNettyVersion=5.5.1
 
-labkeySchemasTestVersion=21.7.0
+labkeySchemasTestVersion=22.3-SNAPSHOT
```
labkey-susanh, line 11: This should really be 22.3.0-SNAPSHOT, shouldn't it?

labkey-susanh: I guess it's 22.3-SNAPSHOT in develop.

PR #1008, opened 2022-02-11
Reimplement old crawler download handling

labkey-tchad:
#### Rationale
We updated the crawler to handle file downloads but it stopped working for some XML files when we updated TeamCity to run tests on Firefox 91. An attempted fix to skip download URLs by checking the HTTP response header caused more problems than it fixed because web browsers handle improperly formatted URLs better than the Apache HTTP client.
The XML download URLs were triggering a browser save dialog, which WebDriver has no awareness of. Clearing out the `browser.download.viewableInternally.enabledTypes` browser property (which includes 'xml' by default) seems to fix the problem. I've added that property to our firefox launch options and reimplement the old, more effective crawler download handling.

#### Related Pull Requests
* Initial: #749 
* Workaround: #964 

#### Changes
* Make `Crawler.beginAt` handle download URLs

### DIFF
--- a/src/org/labkey/test/util/Crawler.java
+++ b/src/org/labkey/test/util/Crawler.java
@@ -20,10 +20,8 @@ import org.apache.commons.collections4.MultiValuedMap;
 import org.apache.commons.collections4.map.CaseInsensitiveMap;
 import org.apache.commons.collections4.multimap.HashSetValuedHashMap;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.http.Header;
 import org.apache.http.HttpResponse;
 import org.apache.http.HttpStatus;
-import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpHead;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.protocol.HttpContext;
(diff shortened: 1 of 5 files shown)


labkey-tchad:
Daily E likes this change: https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailySuites_DailyDPostgres/1704040?showRootCauses=false&expandBuildChangesSection=true&expandBuildTestsSection=true

src/org/labkey/test/util/Crawler.java
```diff
@@ -835,92 +841,103 @@ public void validatePage(@LoggedParam String url)
...
+            final File downloadDir = BaseWebDriverTest.getDownloadDir();
+            final File[] existingDownloads = downloadDir.listFiles();
+
+            long elapsedTime = _test.doAndMaybeWaitForPageToLoad(WebDriverWrapper.WAIT_FOR_PAGE, () -> {
+                final WebElement mightGoStale = Locators.documentRoot.findElement(_test.getDriver());
+                ExpectedCondition<Boolean> stalenessOf = ExpectedConditions.stalenessOf(mightGoStale);
+                // 'getDriver().navigate().to(fullURL)' assumes navigation and fails for file downloads
+                _test.executeScript("document.location = arguments[0]", fullURL);
+                //noinspection ResultOfMethodCallIgnored
+                if (!WebDriverWrapper.waitFor(() -> {
```
labkey-danield, line 886: Ohhh, a nested lambda.

labkey-danield (approved):
Some tricky code there. It looks ok to me, hopefully this will take care of the "wait" failures in the crawler.

PR #1012, opened 2022-02-15
Remove unnecessary xmlbeansVersion reference - just pull in POI's version

labkey-adam:
#### Rationale
Pull in POI's version of XMLBeans as we do in platform

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3051


labkey-adam:
Nevermind... we're going to leave the XMLBeans `implementation` line in place since it's a direct dependency

PR #1014, opened 2022-02-16
Add better support for test user configuration

labkey-chrisj:
#### Rationale
This adds a component intended to make test user creation, deletion, and configuration a little more self-contained than 

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/1157   <-- branch in biologics uses this

#### Changes
new class, TestUser - tersely contains enough information to create a user account, reset its initial password, and give it permissions in particular contexts- the following expression creates a user with working credentials, read permission in the project folder but folder admin in a subfolder of  the project:
```
new TestUser(this, USER_B)
                .setPassword(USER_B_PWD)
                .addPermission("Reader", getProjectName())
                .addPermission("Folder Administrator", getSubfolderContainerPath("subB"))
```

New class, `Card `(implements /internal/components/base/Cards.tsx), for more general use than `SampleTypeCard`
modified class `SampleTypeCard `to derive from `Card`
add method to `FilteringReactSelect`, to get the options available when `setFilter()` is called


labkey-tchad (commented):
This iteration of TestUser is way too inconsistent.
Some operations will trigger user creation, others will just return values that were passed in to setters; worst of all, some might cause the browser to navigate.
I think creating and configuring new users lazily is going to be a bad idea because we can't guarantee that the test isn't currently impersonating or on a page where the user list has already been loaded. To avoid those sorts of problems tests will end up having to explicitly trigger user creation anyway. Better to not pretend that is a valid capability.

labkey-chrisj:
OK. I'll make user creation a constructor-time event, make resetting the password happen immediately, same with applying roles

labkey-chrisj:
OK, user creation is immediate now, setting password occurs immediately, and permissions are applied immediately- and because you pointed out how impersonation can be done over API (which is AWESOME) I took the liberty of implementing `.impersonate() `and `.stopImpersonating()`  on the `TestUser `class

src/org/labkey/test/util/TestUser.java
```diff
@@ -0,0 +1,119 @@
...
+
+            _test.setFormElement(Locator.id("password"), password);
+            _test.setFormElement(Locator.id("password2"), password);
+
+            _test.clickButton("Set Password");
+        }
+        else
+        {
+            log("Warning: user " +_email+ "has already selected a password.");
+        }
```
labkey-tchad: This should throw an error
```suggestion
        else
        {
            throw new IllegalStateException("User " +_email+ "has already selected a password.");
        }
```

src/org/labkey/test/util/TestUser.java
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * Creates the user immediately via the API, and stores the createUserResponse to hang on to data like
+     *
+     * @return
+     */
+    private TestUser create()
+    {
+        _createUserResponse = _apiUserHelper.createUser(_email);
+        return this;
+    }
```
labkey-tchad: Make `create` public and take the `WebDriverWrapper` instead of doing that with the constructor. That way `TestUser` can be declared `final static` in tests.

```suggestion
    public TestUser(String email)
    {
        _email = email;
    }

    /**
     * Creates the user immediately via the API, and stores the createUserResponse to hang on to data like
     */
    public TestUser create(WebDriverWrapper test)
    {
        _test = test;
        _apiUserHelper = new APIUserHelper(_test);
        _createUserResponse = _apiUserHelper.createUser(_email);
        return this;
    }
```

src/org/labkey/test/util/TestUser.java
```diff
@@ -0,0 +1,119 @@
...
+    public void deleteUser()
+    {
+        _apiUserHelper.deleteUsers(false, _email);
+    }
+
+    public Long getUserId()
+    {
+        return (Long) _createUserResponse.getUserId();
+    }
+
```
labkey-tchad: Consider adding a method to get the user's display name.
```suggestion

    public Long getUserId()
    {
        return (Long) _createUserResponse.getUserId();
    }

    public String getDisplayName()
    {
        return _apiUserHelper.getDisplayNameForEmail(getEmail());
    }

```

labkey-tchad (changes requested):

src/org/labkey/test/util/TestUser.java
labkey-chrisj: I like this!
But with create being public (and the `WebDriverWrapper `and `CreateUserResponse `being dependencies for other methods) I think I should wrap references to `_apiUserHelper `and `_createUserResponse `in a getter that checks it for null and throws an appropriate `IllegalStateException`

labkey-tchad (approved):

PR #1019, opened 2022-02-16
Check for expected error case-insensitively in IssueAPITest

labkey-chrisj:
#### Rationale
IssueAPITest.testIssueWithoutPri() has been failing consistently on SQL but passing on PG, because it checks for expected text in an API response and SQL capitalizes one of the words ("Priority") in the expected string, while PG does not.

#### Related Pull Requests
n/a

#### Changes
Single-line change: convert the error to lower case, then ensure the expected error message in lower case


labkey-jeckels (approved):
Change looks fine. Let's wait for the suite to run on TC before merging.

https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_ModuleSuites_IssuesSqlserver/1712725?showRootCauses=false&expandBuildChangesSection=true

PR #1021, opened 2022-02-18
Tidy up timing and scope of `QueryGrid`

labkey-tchad:
#### Rationale
QueryGrid has some crufty support for switching grid tabs. `TabbedGridPanel` is more appropriately scoped and should be used instead.
The tab switching is a bit unreliable in `TabbedGridPanel` though; it needs a better way to ensure that the grid has switched. Waiting for the old grid to go stale then waiting for the new one to be loaded seems to work well.

#### Related Pull Requests
* https://github.com/LabKey/biologics/pull/1163
* https://github.com/LabKey/sampleManagement/pull/844
* https://github.com/LabKey/inventory/pull/386

#### Changes
* Improve timing of switching tabs on `TabbedGridPanel`
* Remove redundant `GridTabBar` component
* Add new ways to append Locators and expose `ComponentFinder` locators


src/org/labkey/test/components/Component.java
```diff
@@ -153,7 +153,7 @@ protected F getThis()
             return (F) this;
         }
 
-        protected final Locator buildLocator()
+        public final Locator buildLocator()
```
labkey-chrisj, line 156: oh, this is nice.

labkey-chrisj (approved):
I'm curious to learn more about the whys and use cases for Caching vs. Immutable locators, but my curiosity on that doesn't hold these branches up

src/org/labkey/test/components/ui/OmniBox.java
```diff
@@ -100,28 +85,23 @@ public OmniBox tabOut()
...
+        return new OmniBoxValueFinder(getDriver()).findAll(this);
     }
 
     public OmniBoxValue getValue(String expected)
     {
-        return new OmniBoxValue.OmniBoxValueFinder(getDriver()).withText(expected).find(this);
+        return new OmniBoxValueFinder(getDriver()).withText(expected).find(this);
     }
 
     public Optional<OmniBoxValue> getOptionalValue(String expected)
```
labkey-danield, line 96: Is this used any more?

src/org/labkey/test/components/ui/OmniBox.java
```diff
@@ -140,32 +119,34 @@ private boolean isOpen()
...
-        getWrapper().waitFor(()-> !isOpen(), 500);
+        if (isOpen())
+        {
+            elementCache().input.sendKeys(Keys.ESCAPE);
+            WebDriverWrapper.waitFor(() -> !isOpen(), "Didn't close omnibox popup", 500);
+        }
         return this;
     }
 
     public boolean isEditing()  // there is always an input; 'editing' means it currently has a value
```
labkey-danield, line 130: Does this need to be public?

src/org/labkey/test/components/ui/OmniBox.java
```diff
@@ -140,32 +119,34 @@ private boolean isOpen()
 
     private OmniBox close()
```
labkey-danield, line 120: It doesn't look like this is used.

src/org/labkey/test/components/ui/OmniBox.java
```diff
@@ -140,32 +119,34 @@ private boolean isOpen()
...
 
     public boolean isEditing()  // there is always an input; 'editing' means it currently has a value
     {
-        return !Locators.editingValue.findElement(this)
-                .getAttribute("value").isEmpty();
+        return !elementCache().input.getAttribute("value").isEmpty();
     }
 
-    public OmniBox stopEditing()    // click just outside the omnibox, above/center
+    public OmniBox stopEditing()
```
labkey-danield, line 135: I know this is a legacy method, but it just feels a little strange to have as a public method. 
Wouldn't a better design be to have an appendFilter or addFilter method that would add to the existing filter and that method takes care of editing "life cycle" of the OmniBox?

src/org/labkey/test/components/ui/Pager.java
```diff
@@ -160,7 +151,7 @@ public int total()
      * if the pager does not show a total, it indicates that there are no more to the set than are shown on the current page.
      * use this to understand whether or not there are more pages of information (which
      * could be the whole set, or a filtered range of it, or both).
-     * @return
+     * @return <code>true</code> if pager shows a "total"
      */
     public boolean hasTotal()
```
labkey-danield, line 156: I'm going to make an argument to remove this method. It doesn't look like it is used (that should be reason enough), and second elementCache().counts has a time out of 4 seconds. If I understand the code correctly that means to get a false from this method you have to wait 4 seconds.

labkey-danield (approved):
You may already be aware of this but there are changes being planned for the OmniBox. I think around filtering and sorting. Not sure if they would impact the test component.

labkey-danield (commented):
Please retarget these changes to 22.3.

PR #1027, opened 2022-02-22
Update OntologyTreeSearch component for use in insert/update

labkey-chrisj:
#### Rationale
Until now, the only product use of the OntologyTreeSearch component was in the ConceptPickerDialog, or in the Ontology page- in which cases, the search input's name was always "concept-search".  When in use in insert and update forms, the name is something different, but it is always the only input.form-control in the component.

#### Related Pull Requests
tbd- <- ontology test changes 

#### Changes
Update the locator for the search input

### DIFF
--- a/src/org/labkey/test/components/ui/ontology/OntologyTreeSearch.java
+++ b/src/org/labkey/test/components/ui/ontology/OntologyTreeSearch.java
@@ -139,7 +139,7 @@ public class OntologyTreeSearch extends WebDriverComponent<OntologyTreeSearch.El
 
     protected class ElementCache extends Component<?>.ElementCache
     {
-        final Input searchInput = Input.Input(Locator.input("concept-search"), getDriver())
+        final Input searchInput = Input.Input(Locator.tagWithClass("input", "form-control"), getDriver())
                 .timeout(2000).findWhenNeeded(this);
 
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/components/ui/ontology/OntologyTreeSearch.java
```diff
@@ -159,7 +159,7 @@ WebElement resultContainer()
     public static class OntologyTreeSearchFinder extends WebDriverComponentFinder<OntologyTreeSearch, OntologyTreeSearchFinder>
     {
         private final Locator.XPathLocator _baseLocator = Locator.tagWithClass("div", "concept-search-container");
-        private String _title = null;
+        private final String _title = null;
```
Sigmonia: Looks like this isn't being used, I think it can be removed

Sigmonia (approved):

src/org/labkey/test/components/ui/ontology/OntologyTreeSearch.java
labkey-chrisj: it's part of the template, we're not using it at the moment (but it might be useful if we want to use the annotation field's input name attribute to find this component in the future)
For now it's dead code, I'll pull it out

PR #1028, opened 2022-02-23
Update `ExportFolderPage`

labkey-tchad:
#### Rationale
Numerous tests and helpers export studies and folders without using `ExportFolderPage` as much as the could. Also, there is some page functionality in `BaseWebDriverTest` that belongs in the export page.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3082
* https://github.com/LabKey/compliance/pull/149
* https://github.com/LabKey/cloud/pull/96

#### Changes
* Move PHI and export options to `ExportFolderPage`
* Refactor study export helpers to use `ExportFolderPage`

### DIFF
--- a/src/org/labkey/test/components/ext4/RadioButton.java
+++ b/src/org/labkey/test/components/ext4/RadioButton.java
@@ -15,6 +15,7 @@
  */
 package org.labkey.test.components.ext4;
 
+import org.jetbrains.annotations.NotNull;
 import org.junit.Assert;
 import org.labkey.test.Locator;
 import org.labkey.test.util.Ext4Helper;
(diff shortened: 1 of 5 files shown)


src/org/labkey/test/components/ext4/RadioButton.java
```diff
@@ -32,6 +33,16 @@ public static RadioButtonFinder RadioButton()
         return new RadioButtonFinder();
     }
 
+    @Override
+    public void set(@NotNull Boolean checked)
+    {
+        if (!checked)
```
labkey-chrisj, line 39: seems this is an argument against radiobutton extending checkbox, but this is a good idea- fail fast and hard rather than silently

labkey-chrisj (approved):

PR #1038, opened 2022-03-11
Sample Finder test components

labkey-tchad:
#### Rationale
Need test components for testing the Sample Finder [[Product Spec]](https://docs.google.com/document/d/1AhHrWcvegLUyXku50FXXcxXkFMGL3-A62nXy-wttKxQ/edit#heading=h.weqfca5zq4ni). They are not yet in use and will likely need further refinement to make them reliable.

These wrap `labkey-ui-components` indicated in their class documentation:
 * `SampleFinder` is the top-level component
 * `EntityFieldFilterModal` represents the dialog that opens when adding/editing search criteria.
 * `FilterExpressionPanel` controls filter expressions (e.g. "Less Than 5") in the `EntityFieldFilterModal`
 * `FilterFacetedPanel` controls faceted filtering in the `EntityFieldFilterModal`

The Filter*Panel product components might be shared by a future OmniBox iteration. In anticipation, the associated test components are intended to be able to be used in other contexts if needed.

#### Changes
* Add test components for Sample Finder
* Update `BootstrapLocators.button` to work with more buttons
* Add `Locator.isDisplayed`
* Simplify `Component.waitForReady`
* Update `NavBar` to support Sample Finder

### DIFF
--- a/src/org/labkey/test/components/ui/search/EntityFieldFilterModal.java
+++ b/src/org/labkey/test/components/ui/search/EntityFieldFilterModal.java
@@ -33,9 +33,14 @@ public class EntityFieldFilterModal extends ModalDialog
                 .visibilityOfNestedElementsLocatedBy(elementCache().querySelectionPanel, listItem));
     }
 
-    public EntityFieldFilterModal selectParentQuery(String query)
+    /**
+     * Select parent
+     * @param parentName name of parent type
+     * @return this component
+     */
+    public EntityFieldFilterModal selectParent(String parentName)
     {
-        WebElement queryItem = listItem.withText(query).findElement(elementCache().querySelectionPanel);
+        WebElement queryItem = listItem.withText(parentName).findElement(elementCache().querySelectionPanel);
         getWrapper().doAndWaitForElementToRefresh(queryItem::click,
                 () -> listItem.findElement(elementCache().fieldsSelectionPanel), getWrapper().shortWait());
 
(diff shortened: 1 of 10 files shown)


src/org/labkey/test/components/ui/search/EntityFieldFilterModal.java
```diff
@@ -0,0 +1,173 @@
...
+        return elementCache().filterExpressionPanel;
+    }
+
+    /**
+     * Select the facet tab for the current field. Will throw <code>NoSuchElementException</code> if tab isn't present.
+     * @return panel for configuring faceted filter
+     */
+    public FilterFacetedPanel selectFacetTab()
+    {
+        elementCache().filterTabs.selectTab("Choose values");
```
labkey-danield, line 127: I know it is beyond the scope of this PR, but I wonder if there would be any value in add a `selectTabByID` method to the Tabs class.

src/org/labkey/test/components/ui/search/SampleFinder.java
```diff
@@ -0,0 +1,221 @@
...
+                    Locator.byClass("primary-text").withText(queryName)).findElement(filterCardsSection));
+        }
+
+        final TabbedGridPanel resultsGrid = new TabbedGridPanel.TabbedGridPanelFinder(getDriver()).findWhenNeeded(this);
+    }
+
+    /**
+     * Represents a single filter card. Don't expose outside this class because its lifecycle is confusing
+     */
+    private class FilterCard extends Component<Component<?>.ElementCache>
```
labkey-danield, line 192: You should add a getter that returns the filters that have been applied in the given card:

Screen Shot 2022-03-14 at 5 37 33 PM

labkey-danield (approved):
Nice and clean.
I think there might be some functionality missing in `SampleFinder.FilterCard` class.

PR #1045, opened 2022-03-17
Delete test files for long-deleted microarray assay functionality

labkey-jeckels:
#### Rationale
We got rid of our MageML-based microarray assay years ago. It's time to let go of the associated test data files.

#### Changes
* Remove orphaned microarray test data


labkey-sweta (approved):
We can delete the folder Microarray too.

labkey-jeckels:
@labkey-sweta the expressionMatrix subdirectory is still in use

labkey-tchad (changes requested):
`NonStudyReportsTest`, `ReportAndDatasetNotificationTest`, and `ReportThumbnailTest` use these files. They don't need these files in particular, but will need to be updated to use some other image files.

labkey-jeckels:
@labkey-tchad thanks for the pointer. We can leave these

PR #1049, opened 2022-03-21
Trigger heap dump when folder delete is blocked

labkey-tchad:
#### Rationale
On Windows, sometimes tests are unable to clean up files between tests because the server is holding onto a file. The test should trigger a heap dump for investigation purposes.

#### Changes
* Trigger a heap dump after `DirectoryNotEmptyException`

### DIFF
--- a/src/org/labkey/test/BaseWebDriverTest.java
+++ b/src/org/labkey/test/BaseWebDriverTest.java
@@ -844,18 +844,6 @@ public abstract class BaseWebDriverTest extends LabKeySiteWrapper implements Cle
             return;
         }
 
-        if (error instanceof DirectoryNotEmptyException notEmpty)
-        {
-            try
-            {
-                Path path = Paths.get(notEmpty.getFile());
-                List<String> subdirs = Files.walk(path).map(p -> path.relativize(p).toString())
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/BaseWebDriverTest.java
```diff
@@ -986,6 +988,20 @@ private void handleFailure(Throwable error, @LoggedParam String testName)
...
+            {
+                try
+                {
+                    Path path = Paths.get(notEmpty.getFile());
+                    List<String> subdirs = Files.walk(path).map(p -> path.relativize(p).toString())
+                            .filter(s -> !s.isEmpty()).collect(Collectors.toList());
+                    TestLogger.error("Remaining files after attempting to delete: " + path + "\n\t" + String.join("\t\n", subdirs), notEmpty);
+                    getArtifactCollector().dumpHeap();
+                }
+                catch (IOException ignore) { }
```
labkey-danield: Any value in logging the IOException if it occurs?

labkey-danield (approved):

src/org/labkey/test/BaseWebDriverTest.java
labkey-tchad: Now that you mention it, there's not really any reason to not log it.

PR #1052, opened 2022-03-22
Issue 45087: Remove slf4j dependency and add configuration for commons logging to turn off extraneous logs

labkey-susanh:
#### Rationale
We want to remove our dependencies on the old version of log4j that may have security issues.  The inclusion of the dependency on `slf4j-log4j12` here was added only to silence some logging from other libraries that happens when running tests in IntelliJ.  Including the dependency on the `slf4j-api` will, according to the manual: "If no binding is found on the class path, then SLF4J will default to a no-operation implementation".  This is better than spurious logs that are not useful.

#### Related Pull Requests
* https://github.com/LabKey/server/pull/223

#### Changes
* Use `org.slf4j:slf4j-api` instead of `org.slf4j:slf4j-log4j12`


labkey-tchad (approved):
Getting a lot of log spam from the remote API. Added a configuration file to disable logging from the remoteapi.

PR #1053, opened 2022-03-23
Simplify ParentEntityEditPanel

labkey-danield:
#### Rationale
Simplify the ParentEntityEditPanel. When this panel was first implemented it was possible to have multiple edit panels open on a page, that is no longer possible. This simplifies the way the edit panel is found, and the methods that are needed.

Example of Adding a Parent:
Screen Shot 2022-03-22 at 9 25 22 PM

Example of editing with multiple parent types
Screen Shot 2022-03-22 at 9 22 18 PM

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/887

#### Changes
* Removed the need to pass in a type enum to find the edit panel.
* Added a finder.
* Updated calls to the reactSelect finders to use the hidden text box (the name of the type is in the name of the input).
* Clarified method names to differentiate between the select for the type (sample or source type name) and the parent (the specific sample or source to be used as the parent).
* Updated comments.

### DIFF
--- a/src/org/labkey/test/components/ui/entities/ParentEntityEditPanel.java
+++ b/src/org/labkey/test/components/ui/entities/ParentEntityEditPanel.java
@@ -8,6 +8,8 @@ import org.labkey.test.components.Component;
 import org.labkey.test.components.WebDriverComponent;
 import org.labkey.test.components.react.FilteringReactSelect;
 import org.labkey.test.components.react.ReactSelect;
+import org.openqa.selenium.NoSuchElementException;
+import org.openqa.selenium.StaleElementReferenceException;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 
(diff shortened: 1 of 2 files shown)


labkey-tchad (changes requested):
This component should be able to wait for itself to be ready. The difference between the initial state with zero parents defined and editing existing parents is that there is no "Parent IDs" select when the panel loads with no existing parents.
We should be able to wait for at least one parent type select to be present and for any "Parent ID" selects to have at least one selection loaded.

labkey-danield:
I believe that I have the form able to identify when it is ready.

src/org/labkey/test/components/ui/entities/ParentEntityEditPanel.java
```diff
@@ -8,80 +8,114 @@
...
+                        .withNamedInput("entityType0")
+                        .findAll(this);
+
+                Assert.assertFalse("It looks like there are multiple new parent ReactSelect controls present, that should never happen.",
+                        firstParentSelector.size() > 1);
+
+                if(!firstParentSelector.isEmpty())
+                {
+                    return firstParentSelector.get(0).isInteractive();
+                }
```
labkey-tchad: I think you want to check the state of the entityType selector second. `firstParentSelector` should never never be empty, so this will always return before checking for any selected source/parent IDs.

labkey-tchad (changes requested):

src/org/labkey/test/components/ui/entities/ParentEntityEditPanel.java
labkey-danield: Ok, hopefully third time is the charm.

src/org/labkey/test/components/ui/entities/ParentEntityEditPanel.java
```diff
@@ -8,80 +8,114 @@
...
+                    // If there are existing parents then all the FilteringReactSelect controls (i.e. the parents) should
+                    // be interactive. When this panel is in a dialog waitForReady can be called after the parent entity type is selected.
+                    // In that scenario the parent selector will be empty, but it should be interactive.
+                    List<FilteringReactSelect> filteringReactSelects = FilteringReactSelect.finder(getDriver()).findAll(this);
+
+                    if(filteringReactSelects.size() > 1)
+                    {
+                        for (FilteringReactSelect select : filteringReactSelects)
+                        {
+                            if (!select.isInteractive())
```
labkey-tchad, line 85: Should wait for selections as well.
```suggestion
                            if (!select.isInteractive() || !select.hasSelection())
```

labkey-tchad (changes requested):

labkey-tchad (approved):
Approved pending previous suggestion.

PR #1054, opened 2022-03-25
give DetailTable a waitForReady()

labkey-chrisj:
#### Rationale
Occasionally some tests in Biologics (such as [this one](https://teamcity.labkey.org/viewLog.html?buildId=1755004&buildTypeId=LabKey_Trunk_Premium_ProductSuites_Biologics_BiologicsBPostgres&fromExperimentalUI=true#testNameId-8112779387230223933))  will fail because an expected value wasn't in the value extracted from a cell in a DetailTable- but in after screenshots, the value is present (which suggests that the test got the value before the app populated the table with it).  
This approach waits once, when getting the `DetailTable`'s `elementCache()`, so it's lazy-initialization-friendly.
It also moves finder methods into the elementCache() if they weren't already.

This approach may make a lot of calls to DetailTable.isLoaded() in page class waitForPage() methods redundant, (so might call for making isLoaded() private and not calling it from page classes, or perhaps we don't need this change and every page or test using this component should be responsible for knowing when the table is ready for interaction)

#### Related Pull Requests
n/a

#### Changes
* overrides and implements `waitForReady()` on `DetailTable`
* moves table element finders that weren't already in elementCache() there, to ensure synchronization

### DIFF
--- a/src/org/labkey/test/components/ui/grids/DetailTable.java
+++ b/src/org/labkey/test/components/ui/grids/DetailTable.java
@@ -1,6 +1,7 @@
 package org.labkey.test.components.ui.grids;
 
 import org.labkey.test.Locator;
+import org.labkey.test.WebDriverWrapper;
 import org.labkey.test.components.Component;
 import org.labkey.test.components.WebDriverComponent;
 import org.openqa.selenium.By;
@@ -41,6 +42,12 @@ public class DetailTable extends WebDriverComponent<DetailTable.ElementCache>
     public WebElement getComponentElement()
     {
         return _tableElement;
+    }
+
+    @Override
+    protected void waitForReady()
+    {
+        getWrapper().shortWait().withMessage("waiting for detailTable to load").until(wd -> isLoaded());
     }
 
     public Boolean isLoaded()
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/components/ui/grids/DetailTable.java
```diff
@@ -156,6 +159,7 @@ public void clickField(String fieldCaption)
     @Override
     protected ElementCache newElementCache()
     {
+        waitForReady();
```
labkey-tchad: This call is unnecessary. `waitForReady` gets called in `Component.elementCache` before it calls `newElementCache`.
```suggestion
```

labkey-tchad (changes requested):

src/org/labkey/test/components/ui/grids/DetailTable.java
labkey-chrisj: ah, OK.
I notice that Component.waitForReady(EC ec) is marked as deprecated as of 22.4, I figured that was a signal to find another way (and was, apparently, wrong 😄)

labkey-tchad (approved):

PR #1055, opened 2022-03-28
Site Level Project Creator role: Automation test

labkey-sweta:
#### Rationale
Automation test for Site Level Project Creator role 

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45046

#### Related Pull Requests
* 

#### Changes
src/org/labkey/test/tests/ProjectCreatorUserTest.java

### DIFF
--- a/src/org/labkey/test/tests/ProjectCreatorUserTest.java
+++ b/src/org/labkey/test/tests/ProjectCreatorUserTest.java
@@ -31,8 +31,8 @@ public class ProjectCreatorUserTest extends BaseWebDriverTest
 {
     private static final String PROJECT_CREATOR_USER = "project_creator@permission.test";
     private static final String READER = "reader@permission.test";
-    private static String PROJECT_NAME_PC = "Folder by Project Creator";
-    private static String TEMPLATE_PROJECT = "Template project";
+    private static final String PROJECT_NAME_PC = "Folder by Project Creator";
+    private static final String TEMPLATE_PROJECT = "Template project";
 
     @BeforeClass
     public static void setup()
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/ProjectCreatorUserTest.java
```diff
@@ -0,0 +1,138 @@
...
+        _userHelper.createUser(PROJECT_CREATOR_USER, true, true);
+        _userHelper.createUser(READER, true, true);
+        new ApiPermissionsHelper(this).addMemberToRole(PROJECT_CREATOR_USER, "Project Creator", PermissionsHelper.MemberType.user, "/");
+    }
+
+    @Override
+    protected void doCleanup(boolean afterTest) throws TestTimeoutException
+    {
+        super.doCleanup(afterTest);
+        _containerHelper.deleteProject(PROJECT_NAME_PC, afterTest, 1000000);
```
cnathe: should this wait time use WAIT_FOR_PAGE like line 79?

src/org/labkey/test/tests/ProjectCreatorUserTest.java
```diff
@@ -0,0 +1,138 @@
...
+        log("Verifying the permissions");
+        goToProjectHome(PROJECT_NAME_PC);
+        navBar().goToPermissionsPage().assertNoPermission(PROJECT_CREATOR_USER, "No Permissions");
+    }
+
+    @Test
+    public void testCreateProjectByReader() throws IOException, CommandException
+    {
+        log("Cleanup : Deleting the project");
+        _containerHelper.deleteProject(PROJECT_NAME_PC, false, WAIT_FOR_PAGE);
```
cnathe: should this be done in a @Before method since so that if more test cases are added in the future, they don't have to worry about the order of execution?

src/org/labkey/test/tests/ProjectCreatorUserTest.java
```diff
@@ -0,0 +1,167 @@
...
+        impersonate(PROJECT_CREATOR_USER);
+        Map<String, Object> params = new HashMap<>();
+        params.put("name", PROJECT_NAME_PC);
+        params.put("assignProjectAdmin", "true");
+        params.put("folderType", "Template");
+        params.put("templateSourceId", containerId);
+        createProject(params);
+        stopImpersonating();
+
+        assertTrue(projectMenu().projectLinkExists(PROJECT_NAME_PC));
```
cnathe, line 146: is it worth a few more checks here? maybe adding in the "project creator has permissions to the new project" check (just to be sure the template part doesn't changes that). Probably don't need to do a full check of the template results, but a quick check like "is the folder of the expected type (i.e. study)" and/or "does it have the expected number of study datasets".

cnathe (approved):

PR #1056, opened 2022-03-28
Automation test for Issue 45049

labkey-sweta:
#### Rationale
Automation test for Issue 45049: Stable URL fails to render when using value-less URL filter

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45049

#### Related Pull Requests
* 

#### Changes
Updated the AbstractExportTest.java to verify Stable URL option.

### DIFF
--- a/src/org/labkey/test/tests/AbstractExportTest.java
+++ b/src/org/labkey/test/tests/AbstractExportTest.java
@@ -337,6 +337,18 @@ public abstract class AbstractExportTest extends BaseWebDriverTest
         });
     }
 
+    @Test
+    public void testStableURLWithValueLessURLFilter()
+    {
+        String nonBlankColumn = getDataRegionColumnName();
+        dataRegion.setFilter(nonBlankColumn, "Is Not Blank");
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/AbstractExportTest.java
```diff
@@ -499,6 +511,12 @@ protected final void assertPerlScriptContents(String perlScript, String filterCo
             assertTrue("Script is missing filterArray property", perlScript.contains("['" + filterColumn + "', eq, ''foo'']"));
     }
 
+    protected final void assertStableURLContents(String stableURL)
+    {
+        assertTrue("Script is missing title", stableURL.contains("The following URL can be used to reload the query, preserving any filters, sorts, or custom sets of columns:"));
+        assertTrue("Script is missing URL", stableURL.contains("query-executeQuery.view?"));
```
labkey-jeckels, line 517: Let's validate more of the URL. I think everything after the query-executeQuery.view? should be stable and consistent too

labkey-jeckels (approved):
One followup request but no need for further review

labkey-tchad (commented):
We discussed alternate method of testing this. Re-request a review once that's ready

PR #1057, opened 2022-03-29
Automation test for Wiki aliases

labkey-sweta:
#### Rationale
Automation test for Wiki aliases 

#### Related Pull Requests
https://github.com/LabKey/platform/pull/3165

#### Changes
1. Manage configuration page : ManageWikiConfigurationPage.java
2. New test class : WikiAliasesTest.java

### DIFF
--- a/src/org/labkey/test/tests/wiki/WikiAliasesTest.java
+++ b/src/org/labkey/test/tests/wiki/WikiAliasesTest.java
@@ -24,6 +24,7 @@ import java.util.Map;
 public class WikiAliasesTest extends BaseWebDriverTest
 {
     private static String wikiName = "Sample Wiki for testing aliases";
+    private static String wikiName2 = "Sample Wiki for testing aliases - Wiki 2";
     private static String wikiTitle = "Title for " + wikiName;
     private static String wikiBody = "Wiki body for " + wikiName;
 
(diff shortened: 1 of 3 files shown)


src/org/labkey/test/tests/wiki/WikiAliasesTest.java
```diff
@@ -0,0 +1,113 @@
...
+        log("Verifying  all the aliases added works with wiki-page.view");
+        manageWikiConfigurationPage = wikiHelper.manageWikiConfiguration();
+        String listOfAliases[] = manageWikiConfigurationPage.getAliases().split("\n");
+        for (int i = 0; i < listOfAliases.length; i++)
+        {
+            beginAt(WebTestHelper.buildURL("wiki", getProjectName(), "page", Map.of("name", listOfAliases[i])));
+            Assert.assertEquals("Incorrect wiki body for alias " + listOfAliases[i], wikiBody,
+                    Locator.tagWithClass("div", "labkey-wiki").findElement(getDriver()).getText());
+        }
+
```
labkey-adam, line 97: Would be good to test that using aliases is case-insensitive. Here's one way:

`String listOfAliases[] = (manageWikiConfigurationPage.getAliases() + "\n" + manageWikiConfigurationPage.getAliases().toLowerCase()).split("\n");`

src/org/labkey/test/tests/wiki/WikiAliasesTest.java
```diff
@@ -0,0 +1,113 @@
...
+        String duplicateAlias = "Alias5";
+        String aliases = "Alias1\n" + "Alias2\n" + "Alias3\n" + "Alias4\n" + duplicateAlias;
+        wikiHelper.createWikiPage(wikiName, null, wikiTitle, wikiBody, true, null, false);
+
+        log("Verifying adding new alias with rename button");
+        ManageWikiConfigurationPage manageWikiConfigurationPage = wikiHelper.manageWikiConfiguration();
+        manageWikiConfigurationPage.rename(duplicateAlias, true)
+                .setAliases(aliases)
+                .save();
+
```
labkey-adam, line 87: There's nothing that verifies that an alias was automatically created for `wikiName`. I would first do a rename **without** setting aliases and verify that `wikiName` still works. Then add more aliases and verify those.

src/org/labkey/test/tests/wiki/WikiAliasesTest.java
```diff
@@ -0,0 +1,113 @@
...
+            Assert.assertEquals("Incorrect wiki body for alias " + listOfAliases[i], wikiBody,
+                    Locator.tagWithClass("div", "labkey-wiki").findElement(getDriver()).getText());
+        }
+
+        log("Verifying same alias name is not allowed");
+        manageWikiConfigurationPage = wikiHelper.manageWikiConfiguration();
+        String errMsg = manageWikiConfigurationPage.setAliases( "\n" + duplicateAlias)
+                .saveExpectingErrors();
+        Assert.assertEquals("Incorrect error message", "Warning: Alias '" + duplicateAlias + "' already exists in this folder.",errMsg);
+
```
labkey-adam, line 103: Probably most interesting to verify that adding a duplicate alias on a different wiki fails to add that alias and produces an error

src/org/labkey/test/tests/wiki/WikiAliasesTest.java
```diff
@@ -0,0 +1,113 @@
...
+        log("Testing case insensitive");
+        manageWikiConfigurationPage = wikiHelper.manageWikiConfiguration();
+        errMsg = manageWikiConfigurationPage.setAliases("\n" + duplicateAlias.toLowerCase()).saveExpectingErrors();
+        Assert.assertEquals("Incorrect error message", "Warning: Alias '" + duplicateAlias.toLowerCase() + "' already exists in this folder.",errMsg);
+
```
labkey-adam, line 120: A good additional test would be to create a new wiki with a name that matches one of the aliases (you could use `wikiName`) and verify that page.view resolves to the wiki with that name, not the wiki that has that alias.

labkey-adam (changes requested):
A few suggestions

src/org/labkey/test/tests/wiki/WikiAliasesTest.java
labkey-adam, line 87: Should also verify that the rename worked (before adding aliases)

src/org/labkey/test/tests/wiki/WikiAliasesTest.java
labkey-sweta: Done

src/org/labkey/test/tests/wiki/WikiAliasesTest.java
labkey-sweta: Done

src/org/labkey/test/tests/wiki/WikiAliasesTest.java
labkey-sweta: Done

src/org/labkey/test/tests/wiki/WikiAliasesTest.java
labkey-sweta: Good suggestion

labkey-adam (approved):

PR #1069, opened 2022-04-08
new component for assay run data 

labkey-chrisj:
#### Rationale
Entity Insert Panel seems to use different text for the mode-select buttons in different contexts
In Biologics, the assay upload panel looks like this:

image
for sample upload, like this:
image

Currently, the showGrid() method assumes the words 'from Grid' will be in the button to show the grid; in the assay case the text includes 'into Grid'- this change makes the helper work in both cases

#### Related Pull Requests
cross-folder assay automation

#### Changes
Just changes the matching text for the show grid button


labkey-tchad (approved):

labkey-tchad (changes requested):
Correction. It looks like the assay upload is a different component ([RunDataPanel.tsx](https://github.com/LabKey/labkey-ui-components/blob/develop/packages/components/src/internal/components/assay/RunDataPanel.tsx)).

labkey-chrisj:
Is this a call for a separate test component?  I can do that

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
```diff
@@ -304,7 +304,7 @@ public EntityInsertPanel showGrid()
     {
         if (!isGridVisible())
         {
-            modeSelectListItem("from Grid")
+            modeSelectListItem(" Grid")
```
labkey-danield: `modeSelectListItem("from Grid")` is used in several places in this file, why not be consistent and change them all?

labkey-tchad:
> Is this a call for a separate test component? I can do that

If it's not too much trouble but don't feel like you need to blow up the scope of this story.

labkey-tchad (approved):
New component isn't a blocker. Maybe add a TODO if it would be too much trouble

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
labkey-chrisj: I decided to not change EntityInsertPanel- I've restored the original text.
The new component (RunDataPanel) uses an enum to match the mode-button texts, hopefully that's not over-doing it

PR #1076, opened 2022-04-12
Item 10253: Remove experimental feature to officially switch to new Grid Filtering UX changes

cnathe:
#### Rationale
The app GridPanel OmniBox has served us very well for many reason, but it is time for a change. The related PRs remove the experimental feature flag to switch over to the new grid filtering UX (filter modal from sample finder, grid column header menu options for add/remove filter, grid filter status display in header, search input in grid button bar, etc.). With that change, this PR updates the test components/pages so that the filtering for an app grid is done via the GridFilterModal instead of the OmniBox (along with any grid search and sorts).

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/804
* https://github.com/LabKey/testAutomation/pull/1076
* https://github.com/LabKey/platform/pull/3247
* https://github.com/LabKey/biologics/pull/1245
* https://github.com/LabKey/inventory/pull/418
* https://github.com/LabKey/sampleManagement/pull/918
* https://github.com/LabKey/provenance/pull/103

#### Changes
* Remove references to QueryGridPanel in test locators
* update css selectors in FilterFacetedPanel
* test helper Grid updates for removal of OmniBox in favor of FilterStatus display
* Factor out GridFilterModal and change EntityFieldFilterModal to extend it


src/org/labkey/test/components/ui/FilterStatus.java
```diff
@@ -0,0 +1,79 @@
...
+import org.labkey.test.Locator;
+import org.labkey.test.components.Component;
+import org.labkey.test.components.WebDriverComponent;
+import org.labkey.test.components.ui.FilterStatusValue.FilterStatusValueFinder;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+import java.util.List;
+
+public class FilterStatus extends WebDriverComponent<FilterStatus.ElementCache>
```
labkey-tchad: I don't think this needs to be a component. It can be a plain `WebElement` in the QueryGrid element cache.
```
final WebElement filterStatusPanel = Locator.css("div.grid-panel__filter-status").findWhenNeeded(this);
```
Then move the `FilterStatusValue` getters into that element cache as well.

src/org/labkey/test/components/ui/grids/GridFilterModal.java
```diff
@@ -0,0 +1,139 @@
...
+    {
+        super(new ModalDialogFinder(driver));
+        _linkedComponent = linkedComponent;
+    }
+
+    @Override
+    protected void waitForReady(ModalDialog.ElementCache ec)
+    {
+        getWrapper().shortWait().until(ExpectedConditions.visibilityOf(listItem.findElementOrNull(elementCache().fieldsSelectionPanel)));
+    }
```
labkey-tchad: `visibilityOf` will NPE if `listItem.findElementOrNull` returns null. `findWhenNeeded` is what you want; it will play nicely with `visibilityOf` when the element isn't present.
```suggestion
    protected void waitForReady(ModalDialog.ElementCache ec)
    {
        getWrapper().shortWait().until(ExpectedConditions.visibilityOf(listItem.findWhenNeeded(elementCache().fieldsSelectionPanel)));
    }
```

src/org/labkey/test/components/ui/grids/GridFilterModal.java
```diff
@@ -0,0 +1,139 @@
...
+            getWrapper().shortWait().until(ExpectedConditions.visibilityOf(elementCache().errorAlert));
+        }
+        else
+        {
+            _linkedComponent.doAndWaitForUpdate(() -> {
+                elementCache().submitButton.click();
+                waitForClose();
+            });
+        }
+    }
```
labkey-tchad: The preferred pattern is to have separate methods for distinct behavior instead of using a flag. In this case it would allow the confirm with error to return the error.
```suggestion
    /**
     * Save current changes to the search criteria.
     * Throw <code>IllegalStateException</code> if the save button is disabled because no changes have been made.
     */
    public void confirm()
    {
        _linkedComponent.doAndWaitForUpdate(() -> {
            clickConfirm();
            waitForClose();
        });
    }

    public String confirmExpectingError()
    {
        clickConfirm();
        return getWrapper().shortWait().until(ExpectedConditions.visibilityOf(elementCache().errorAlert)).getText();
    }

    protected void clickConfirm()
    {
        if (!elementCache().submitButton.isEnabled())
        {
            throw new IllegalStateException("Confirmation button is not enabled.");
        }

        elementCache().submitButton.click();
    }
```

src/org/labkey/test/components/ui/grids/QueryGrid.java
```diff
@@ -185,43 +188,45 @@ public void doAndWaitForUpdate(Runnable func)
...
+        sortColumn(column, direction);
         return this;
     }
 
     /**
-     * clears search, sort, and filter expressions via the omnibox
+     * clears grid filter expressions added by the user (i.e. from the filter modal)
      */
-    public QueryGrid clearSortsAndFilters()
+    public QueryGrid clearUserDefinedFilters()
```
labkey-tchad: Why the distinction of "user defined". What other filters might there be?
I would prefer the brevity of just `clearFilters` if possible. And just have the JavaDoc make any necessary clarifications.

src/org/labkey/test/components/ui/grids/QueryGrid.java
```diff
@@ -185,43 +188,45 @@ public void doAndWaitForUpdate(Runnable func)
...
-     * @return The QueryGrid after the filter has been applied.
+     * Applies a sort to the grid via the column header menu and waits for the grid to refresh
      */
-    public QueryGrid filterOn(String columnName, OmniBox.FilterOperator operator, String value)
+    public QueryGrid sortOn(String column, SortDirection direction)
     {
-        getOmniBox().setFilter(columnName, operator, value);
+        sortColumn(column, direction);
         return this;
     }
```
labkey-tchad: Can we get rid of `sortOn` so that `QueryGrid` doesn't have redundant sorting methods?

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
```diff
@@ -107,9 +109,78 @@ public T sortColumnDescending(String columnLabel)
...
+        filterModal.confirm(true);
+        String errorMsg = filterModal.getErrorMsg();
+        filterModal.cancel();
+        return errorMsg;
+    }
+
+    private GridFilterModal initFilterColumn(String columnLabel, Filter.Operator operator, Object value)
+    {
+        clickColumnMenuItem(columnLabel, "Filter...", false);
+        GridFilterModal filterModal = new GridFilterModal(getDriver(), getThis()::doAndWaitForUpdate);
```
labkey-tchad: ResponsiveGrid implements `UpdatingComponent` interface that the constructor takes. This shouldn't need to use a method reference.
```suggestion
        GridFilterModal filterModal = new GridFilterModal(getDriver(), this);
```

src/org/labkey/test/components/ui/search/EntityFieldFilterModal.java
```diff
@@ -4,32 +4,25 @@
...
+        super(driver, linkedComponent);
     }
 
     @Override
     protected void waitForReady(ModalDialog.ElementCache ec)
     {
-        super.waitForReady(ec);
-        getWrapper().shortWait().until(ExpectedConditions
-                .visibilityOf(listItem.findWhenNeeded(elementCache().querySelectionPanel)));
+        getWrapper().shortWait().until(ExpectedConditions.visibilityOf(listItem.findElementOrNull(elementCache().querySelectionPanel)));
```
labkey-tchad: Need to use `findWhenNeeded` with `visibilityOf`
```suggestion
        getWrapper().shortWait().until(ExpectedConditions.visibilityOf(listItem.findWhenNeeded(elementCache().querySelectionPanel)));
```

src/org/labkey/test/tests/component/GridPanelTest.java
```diff
@@ -221,86 +221,51 @@ public void testOmniboxFilterInMultiPageSelectionSet() throws Exception
...
-
     @Test
     public void testEmptyNotEmptyFilter() throws Exception
     {
         // create a sampleType domain to use in this case
         SampleTypeDefinition props = new SampleTypeDefinition("empty_filter_test_set").setFields(standardTestSampleFields());
         TestDataGenerator sampleSetDataGenerator = SampleTypeAPIHelper.createEmptySampleType(getProjectName(), props);
-        sampleSetDataGenerator.generateRows(30);
-        for (int i=0; i < 5; i++)  // add some rows with a description we can filter on
+        sampleSetDataGenerator.generateRows(30, List.of("descColumn"));
```
labkey-tchad: Consider setting a custom data generator for 'descColumn' instead of overloading `generateRows`.
```suggestion
        sampleSetDataGenerator.addDataSupplier("descColumn", () -> null);
        sampleSetDataGenerator.generateRows(30);
```

labkey-tchad (changes requested):

src/org/labkey/test/components/ui/FilterStatus.java
cnathe: Done. This was originally the `OmniBox` test component that I renamed to `FilterStatus`. You are right, after removing several things, there isn't much left.

src/org/labkey/test/components/ui/grids/GridFilterModal.java
cnathe: Done

src/org/labkey/test/components/ui/grids/GridFilterModal.java
cnathe: Sounds good. Thanks.

src/org/labkey/test/components/ui/grids/QueryGrid.java
cnathe: The grid could be filtered by the context (i.e. baseFilters applied via code directly like in the "view samples for a given source" grid) and the grid can be filtered by the saved view. Those two cannot be cleared/removed, so I renamed this method accordingly. This likely isn't meaningful from the test side, so I'll rename to `clearFilters`.

src/org/labkey/test/components/ui/grids/QueryGrid.java
cnathe: Yes. I have removed `sortOn` and changed callers to just call `sortColumn` directly.

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
cnathe: Yup, good catch. Done

src/org/labkey/test/components/ui/search/EntityFieldFilterModal.java
cnathe: Done

src/org/labkey/test/tests/component/GridPanelTest.java
cnathe: Perfect. I wasn't aware of that. Changes made.

labkey-tchad (approved):

PR #1077, opened 2022-04-12
testLinkToStudy fix

labkey-klum:
#### Rationale
The `SampleTypeLinkToStudyTest.testLinkToStudy` test began to fail because the order of the audit records in the sample timeline audit events appeared in a different order from what the test expected: 

https://teamcity.labkey.org/test/-376634581250720409?currentProjectId=LabKey_223Release_Community_DailySuites&expandTestHistoryInvestigationsSection=true&expandTestHistoryChartSection=true

In general when assay or sample data is linked to a study we create a single audit event and note the number of records created. For samples there is an additional step for sample timeline to create additional events, one per each sample copied, see `StudyPublishManager.logPublishEvent`. The test expects a specific order but the code doesn't guarantee a specific ordering of individual samples other than they are part of the same batch.  At line 626 of `logPublishEvent` we ask the `ExperimentService` for the list of `ExpMaterials` given a list of rowIds, this results in a query to the table using an `IN CLAUSE` which makes no guarantees on the order of the rows returned. I think we were just getting lucky with the order of the records being created.

Looking closer at the audit logging code, it doesn't seem that timeline events have the expectation of ordering so it didn't make sense to try to enforce ordering to get the test to pass again.

#### Changes
- Update the test to compare using a `Set` instead of a `List`.

### DIFF
--- a/src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
+++ b/src/org/labkey/test/tests/SampleTypeLinkToStudyTest.java
@@ -34,10 +34,11 @@ import java.io.File;
 import java.io.IOException;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
-import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 @Category({Daily.class})
 @BaseWebDriverTest.ClassTimeout(minutes = 12)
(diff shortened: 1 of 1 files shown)


labkey-klum:
Makes sense for @labkey-sweta to do the test code review, but looping in @RosalineP in case you were curious about the failure.

labkey-sweta (commented):
Curious to know why test started failing recently.

labkey-klum:
It's just based on database internals, as I mentioned you might get the same ordering for a while but unless you use an `ORDER BY` clause, you can't rely on record ordering.

labkey-sweta (approved):

PR #1081, opened 2022-04-13
Move RemoteConnectionTest to dataintegration module

labkey-tchad:
#### Rationale
The button to configure Remote Connections is now hidden if the dataintegration module is not enabled. The RemoteConnectionTest should live in the dataintegration module.
During the move I noticed that `RemoteConnectionHelper` no-ops in some cases where it should not, invalidating some existing tests. I've updated the helper and test.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3240

#### Changes
* Remove `RemoteConnectionTest` (moving to dataintegration module)
* Clean up `RemoteConnectionHelper`


labkey-adam (approved):
Should `RemoteConnectionHelper` move to dataintegration as well?

PR #1085, opened 2022-04-14
TextChoice Test Components and Automation

labkey-danield:
#### Rationale
Recently there was a new field type, TextChoice, added to LKS. The field is similar to a LookUp field, but the difference is that the allowed values are defined in the field and not in some other table. The way this is done is that the allowed values are implemented as a validator on the field (the field is basically a text type). The fact that a validator is used is hidden from the user. For a TextChoice field a user can only define the list of values, they do not get to give the validator a name, description or error message. No other validator can be applied to a TextChoice field.

A TextChoice field looks like this:
Screen Shot 2022-04-14 at 11 08 59 AM

A user provides the list of allowed values by clicking the 'Add Values' button and then entering the values into the dialog that is shown:
Screen Shot 2022-04-14 at 11 08 05 AM

If a value has been used by an entry the value is locked (notice locked icon on top image) and cannot be deleted. However a locked field can be updated:
Screen Shot 2022-04-14 at 11 37 14 AM

If a field value is updated/changed to a value that already exists an error is shown (see top image).

If creating a sample type with a TextChoice field through the APIs you will need to create a FieldDefinition object and then pass the list of values into the setTextChoiceValues method:
```
        SampleTypeDefinition sampleTypeDefinition = new SampleTypeDefinition(sampleType);

        List<FieldDefinition> fields = new ArrayList<>();

        FieldDefinition textChoiceField = new FieldDefinition("Text_Choice", FieldDefinition.ColumnType.TextChoice);
        textChoiceField.setTextChoiceValues(List.of("A", "B", "C", "\u00DC"));
        fields.add(textChoiceField);

        sampleTypeDefinition.setFields(fields);

        SampleTypeAPIHelper.createEmptySampleType(getCurrentContainerPath(), sampleTypeDefinition);

```
TextChoice fields can appear in sample types and assay designs.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/939
* https://github.com/LabKey/biologics/pull/1276

#### Changes
* Add TextChoice UI components parts.
* Add a TextChoiceValidator to FieldDefinitions.
* Add test cases for sample types. Validates the TextChoice field in the designer.
* Add test cases for assays. Validates locking and updating.
* Added test cases for issues and lists. Just sanity test to validate TextChoice in other domain types.
* Added test case for exporting and then importing a folder with domains that have TextChoice fields.


src/org/labkey/test/components/domain/DomainFieldRow.java
```diff
@@ -600,6 +602,164 @@ public DomainFieldRow clickRemoveOntologyConcept()
...
+     * @return This field row.
+     */
+    public DomainFieldRow deleteTextChoiceValue(String value)
+    {
+        selectTextChoiceValue(value);
+        WebElement button = Locator.button("Delete").findElement(this);
+        WebDriverWrapper.waitFor(button::isEnabled, "'Delete' button is not enabled.", 1_000);
+        button.click();
+        return this;
+    }
```
labkey-tchad: Should this wait for the option to go away? Maybe wait for the Delete button to be disabled after clicking.

src/org/labkey/test/components/domain/DomainFormPanel.java
```diff
@@ -120,6 +121,15 @@ else if (validator instanceof FieldDefinition.RangeValidator)
...
                 }
+                else if (validator instanceof FieldDefinition.TextChoiceValidator)
+                {
+                    // TextChoice is a field type and not an attribute on a field, any other validators that are applied
+                    // will be ignored (TextChoice field cannot have validators).
+                    textChoiceValidator = (FieldDefinition.TextChoiceValidator) validator;
+                    regexValidators = new ArrayList<>();
+                    rangeValidators = new ArrayList<>();
+                    break;
+                }
```
labkey-tchad: Should just error if there are multiple validators instead of ignoring them silently.
```suggestion
                else if (validator instanceof FieldDefinition.TextChoiceValidator textChoiceValidator)
                {
                    // TextChoice is a field type; implemented using a special validator. TextChoice field cannot have other validators.
                    if (validators.size() > 1)
                    {
                        throw new IllegalArgumentException("TextChoice fields cannot have additional validators.");
                    }
                    fieldRow.setTextChoiceValues(textChoiceValidator.getValues());
                }
```

src/org/labkey/test/params/FieldDefinition.java
```diff
@@ -384,6 +385,13 @@ public FieldDefinition setPrincipalConceptSearchExpression(String ontologyName,
         return this;
     }
 
+    public FieldDefinition setTextChoiceValues(List<String> values)
+    {
+        Assert.isTrue(getType().equals(ColumnType.TextChoice), "Type is not set to 'ColumnType.TextChoice'.");
+        setValidators(List.of(new FieldDefinition.TextChoiceValidator(values)));
+        return this;
+    }
```
labkey-tchad: Non-standard assert.
```suggestion
    public FieldDefinition setTextChoiceValues(List<String> values)
    {
        Assert.assertEquals("Invalid field type for text choice values.", ColumnType.TextChoice, getType());
        setValidators(List.of(new FieldDefinition.TextChoiceValidator(values)));
        return this;
    }
```

src/org/labkey/test/params/FieldDefinition.java
```diff
@@ -836,4 +846,47 @@ public String getSecondRange()
...
+    {
+        private List<String> _values;
+
+        public TextChoiceValidator(List<String> values)
+        {
+            // The TextChoice validator only has a name and no description or message.
+            // And the name is generated (not user defined).
+            super("Text Choice Validator", "", "");
+            _values = values;
+        }
```
labkey-tchad: Consider locking down the collection of values.
```suggestion
        private final List<String> _values;

        public TextChoiceValidator(List<String> values)
        {
            // The TextChoice validator only has a name and no description or message.
            // And the name is generated (not user defined).
            super("Text Choice Validator", "", "");
            _values = Collections.unmodifiableList(values);
        }
```

src/org/labkey/test/components/domain/DomainFieldRow.java
labkey-danield: Should be instantaneous, but to be safe I added a wait for the delete button to disappear (won't be disabled because no value will be selected after the selected value is deleted).

src/org/labkey/test/components/domain/DomainFormPanel.java
labkey-danield: Fair enough. Suggested change applied.

src/org/labkey/test/params/FieldDefinition.java
labkey-danield: Oops! Updated.

src/org/labkey/test/params/FieldDefinition.java
labkey-danield: Done.

src/org/labkey/test/tests/TextChoiceImportExportAndOtherDomainsTest.java
```diff
@@ -0,0 +1,376 @@
...
+        portalHelper.addWebPart("Issue Definitions");
+        portalHelper.addWebPart("Issues Summary");
+        portalHelper.exitAdminMode();
+
+    }
+
+    /**
+     * Create a run for the assay.
+     */
+    protected void createAssayRun()
```
cnathe: how similar is this to the `createAndValidateAssayRun` method in the TextChoiceAssayTest? any reason for looking to share the assay run import parts?

src/org/labkey/test/tests/TextChoiceSampleTypeTest.java
```diff
@@ -0,0 +1,653 @@
...
+
+        List<String> lockedValues = fieldRow.getLockedTextChoiceValues();
+
+        // Sort the result from the UI, at this point more concerned about the values and not the order.
+        Collections.sort(lockedValues);
+        Collections.sort(valuesUsed);
+
+        checker()
+                .withScreenshot("ST_Designer_Locked_Values_Error")
+                .verifyEquals("Locked values not as expected.", valuesUsed, lockedValues);
```
labkey-tchad: Could use the new `assertEqualsSorted` method here (and below):
```suggestion

        checker()
                .withScreenshot("ST_Designer_Locked_Values_Error")
                .wrapAssertion(() -> LabKeyAssert.assertEqualsSorted("Locked values not as expected.", valuesUsed, lockedValues));
```

Might even be worth creating `DeferredErrorCollector.verifyEqualsSorted` instead of using `wrapAssertion`.

src/org/labkey/test/tests/TextChoiceSampleTypeTest.java
```diff
@@ -0,0 +1,653 @@
...
+    @Test
+    public void testUpdatingAndDeletingValuesInSampleType() throws IOException, CommandException
+    {
+        final String sampleTypeName = "TC_Value_Updates";
+        final String textChoiceFieldName = "TextChoice_Field_1";
+        final String namePrefix = "TCE_";
+
+        // Some TextChoice values.
+        List<String> expectedUnLockedValues = new ArrayList<>();
+        expectedUnLockedValues.add("ÅÅ");
```
labkey-tchad: Use `\u` syntax for special characters in java Strings.

src/org/labkey/test/tests/TextChoiceSampleTypeTest.java
```diff
@@ -0,0 +1,653 @@
...
+        expectedUnLockedValues.add("BB");
+        expectedUnLockedValues.add("CC");
+        expectedUnLockedValues.add("DD");
+        expectedUnLockedValues.add("E E E");
+        expectedUnLockedValues.add("ƒƒ");
```
labkey-tchad: `\u` encode these.

labkey-tchad (approved):
Only big problem I see is the unicode characters not using `\u` encoding in some strings.

src/org/labkey/test/tests/TextChoiceSampleTypeTest.java
labkey-danield: Yep, missed it.

src/org/labkey/test/tests/TextChoiceSampleTypeTest.java
labkey-danield: And missed it here as well.

labkey-sweta (approved):

src/org/labkey/test/tests/TextChoiceImportExportAndOtherDomainsTest.java
labkey-danield: Yep, had thought about that. Moved the create run code out to a shared method.

PR #1090, opened 2022-04-20
Item 10230: Entity Grid Action Updates to factor out menu/buttons from create and manage

cnathe:
#### Rationale
See related PRs for rationale. This PR is just for some test helper related updates to support the App grid button/menu updates.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/810
* https://github.com/LabKey/testAutomation/pull/1090
* https://github.com/LabKey/sampleManagement/pull/924
* https://github.com/LabKey/biologics/pull/1258
* https://github.com/LabKey/inventory/pull/420

#### Changes
* GridBar.clickButton update to account for "a" or "button" element
* Add MultiMenu helpers to get menu list and items


src/org/labkey/test/components/react/MultiMenu.java
```diff
@@ -27,6 +27,22 @@ protected MultiMenu(WebElement element, WebDriver driver)
...
+
+    public WebElement getMenuItem(String text)
+    {
+        return Locator.tag("li").withChild(Locator.tagWithAttribute("a", "role", "menuitem")
+                .withText(text)).findElement(getMenuList());
+    }
+
+    public boolean isMenuItemDisabled(String text)
+    {
+        return getMenuItem(text).getAttribute("class").toLowerCase().contains("disabled");
```
labkey-danield, line 43: You might be able to call:
`return !getMenuItem(text).isEnabled();`

src/org/labkey/test/components/ui/grids/GridBar.java
```diff
@@ -217,7 +217,11 @@ public GridBar clearAllSelections()
      */
     public void clickButton(String buttonCaption)
     {
-        Locator.buttonContainingText(buttonCaption).waitForElement(this, 5_000).click();
+        Locator.XPathLocator btnAsLink = Locator.tagWithClass("a", "btn").containing(buttonCaption);
+        if (btnAsLink.isDisplayed(this))
+            btnAsLink.findElement(this).click();
+        else
+            Locator.buttonContainingText(buttonCaption).waitForElement(this, 5_000).click();
```
labkey-danield: Just a style comment... 
Alternatively you could try this:
```
        WebElement button = Locator.tagWithClass("a", "btn").containing(buttonCaption).findWhenNeeded(this);
        waitFor(button::isDisplayed, String.format("Button with caption '%s' is not visible.", buttonCaption), 5_000);
        button.click();

```

labkey-danield (approved):

src/org/labkey/test/components/ui/grids/GridBar.java
cnathe: This change wasn't because I had to wait for the button to display. It is because sometimes the grid bar buttons are `<button>` elements and sometimes they are `<a class="btn">` elements.

src/org/labkey/test/components/react/MultiMenu.java
cnathe: Unfortunately that didn't work when I tried it.

src/org/labkey/test/components/ui/grids/GridBar.java
labkey-tchad: Would `BootstrapLocators.button` cover both kinds of button?

cnathe: Yes, looks like that should work. Thanks

PR #1093, opened 2022-04-20
Issue 45033: Missing / can't set folder descriptions (comments)

XingY:
#### Rationale
Adding test coverage for setting folder description in file browser.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3272

#### Changes
*

### DIFF
--- a/src/org/labkey/test/util/FileBrowserHelper.java
+++ b/src/org/labkey/test/util/FileBrowserHelper.java
@@ -374,6 +374,33 @@ public class FileBrowserHelper extends WebDriverWrapper
         waitForElement(fileGridCell.withText(folderName));
     }
 
+    public void setDescription(String fileName, String description)
+    {
+        Window propWindow = editProperty(fileName);
+        setFormElement(Locator.name("Flag/Comment"), description);
+        propWindow.clickButton("Save", true);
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/util/FileBrowserHelper.java
```diff
@@ -374,6 +374,33 @@ public void createFolder(String folderName)
...
+    {
+        List<String> fileList = getFileList();
+        int fileInd = fileList.indexOf(fileName);
+        if (fileInd > -1)
+        {
+            return getTexts(Locators.gridRow().childTag("td").position(7).findElements(getDriver())).get(fileInd);
+        }
+
+        return null;
+    }
```
labkey-tchad: Does the `gridRow` locator not work for directories? Something like this would be more efficient:
```suggestion
    public String getFileDescription(String fileName)
    {
        return Locators.gridRow(fileName).childTag("td").position(7).findElement(getDriver()).getText();
    }
```
`WebElement.getText` is a relatively expensive operation and unnecessary calls should be avoided.  (`getTexts` and `getFileList` call it numerous times)

labkey-tchad (approved):

PR #1094, opened 2022-04-22
44784: IssueAPI should handle missing priority param in validation

labkey-klum:
#### Rationale
Validation is handled earlier so we can modify the tests to not deal with server side errors but instead receive a normal validation error message from the service.

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44784

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/3284

### DIFF
--- a/src/org/labkey/test/tests/issues/IssueAPITest.java
+++ b/src/org/labkey/test/tests/issues/IssueAPITest.java
@@ -426,7 +426,7 @@ public class IssueAPITest extends BaseWebDriverTest
         }
         catch (CommandException ce)
         {
-            assertEquals("Missing value for required property: assignedto", ce.getMessage());
+            assertEquals("missing value for required property: assignedto", ce.getMessage().toLowerCase());
         }
     }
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/issues/IssueAPITest.java
```diff
@@ -419,28 +419,35 @@ public void testIssueWithoutAssignedTo() throws Exception
                 .setPriority(3L)
                 .setType("Defect");
 
-        var issueResponse= getIssueResponse( doIssueAction(issue));
-        assertEquals("should be assigned to guest", Long.valueOf(0), issueResponse.getAssignedTo());
-        assertEquals("should be open", "open", issueResponse.getStatus());
+        try
```
labkey-klum, line 422: This should have resulted in a validation error, since assigned to is a required field. Updating the test to reflect that.

labkey-chrisj (approved):

PR #1095, opened 2022-04-25
Test Issue: Filters not being removed from grid.

labkey-danield:
#### Rationale
Some tests failures are happening because the filter "pill" is not being removed from the grid. For some tests it is intermittent for others it is more consistent.
Screen Shot 2022-04-25 at 10 45 46 AM
Looking at the failures, like the screen shot above, the pill is not active which means the mouse is not over it. 
I'm not sure that this is the best fix, but it does appear to be working. If there is a more appropriate change please let me know.

#### Related Pull Requests
* None

#### Changes
* Explicitly move the mouse over the pill before clicking the remove icon.

### DIFF
--- a/src/org/labkey/test/components/ui/FilterStatusValue.java
+++ b/src/org/labkey/test/components/ui/FilterStatusValue.java
@@ -54,6 +54,7 @@ public class FilterStatusValue extends WebDriverComponent<FilterStatusValue.Elem
     public void remove()
     {
         String originalText = getText();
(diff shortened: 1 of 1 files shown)


labkey-tchad (approved):
Looks reasonable to me. I think the original implementation was just mousing over the element. I saw this error in a branch I was on and changed it to mouse over the icon. Hopefully mousing over both will do the trick.

PR #1102, opened 2022-05-04
Update grid search to conditionally wait for grid update

labkey-chrisj:
#### Rationale
Yesterday, I updated QueryGrid to wait for the grid to update when a test calls its `search()` method, which solved one sort of problem (that another test was failing because it didn't wait for the grid to update) but created another:
Several tests call clearSearch() to protect against the possibility that a searchTerm is in the search box, but when the expression is already empty and clearSearch() is called, it no-ops and the grid doesn't refresh- and the QueryGrid helper fails while waiting for the grid refresh that never happens.

This change modifies helper behavior as follows:

-` clearSearch()` doesn't wait for refresh unless there was a search expression in the searchBox to clear
-` search()` doesn't wait for refresh unless the supplied search expression is different from the contents of the searchBox

#### Related Pull Requests
n/a

### DIFF
--- a/src/org/labkey/test/components/ui/grids/QueryGrid.java
+++ b/src/org/labkey/test/components/ui/grids/QueryGrid.java
@@ -175,10 +175,15 @@ public class QueryGrid extends ResponsiveGrid<QueryGrid>
 
     /**
      * searches the grid, from the GridBar search input and waits for the grid to refresh
+     * if the searchTerm is different from the current searchTerm
      */
     public QueryGrid search(String searchTerm)
     {
-        doAndWaitForUpdate(()-> getGridBar().searchFor(searchTerm));
+        String currentSearchExp = getGridBar().getSearchExpression();
+        if (searchTerm.equals(currentSearchExp))
+            getGridBar().searchFor(searchTerm);
+        else
+            doAndWaitForUpdate(()-> getGridBar().searchFor(searchTerm));
         return this;
     }
 
@@ -187,7 +192,11 @@ public class QueryGrid extends ResponsiveGrid<QueryGrid>
      */
     public QueryGrid clearSearch()
     {
-        doAndWaitForUpdate(()-> getGridBar().clearSearch());
+        String currentSearchExp = getGridBar().getSearchExpression();
+        if ("".equals(currentSearchExp))
+            getGridBar().clearSearch();
+        else
+            doAndWaitForUpdate(()-> getGridBar().clearSearch());
         return this;
     }
 
--- a/src/org/labkey/test/components/ui/grids/GridBar.java
+++ b/src/org/labkey/test/components/ui/grids/GridBar.java
@@ -416,6 +416,11 @@ public class GridBar extends WebDriverComponent<GridBar.ElementCache>
         return this;
     }
 
+    public String getSearchExpression()
+    {
+        return elementCache().searchBox.get();
+    }
+
     @Override
     protected ElementCache newElementCache()
     {
@@ -492,6 +497,6 @@ public class GridBar extends WebDriverComponent<GridBar.ElementCache>
     {
         ALL,
         SAMPLES,
-        ALIQUOTS;
+        ALIQUOTS
     }
 }


labkey-chrisj:
Develop isn't building now, the test build check failure fails in the same way 😞 

Locally I was able to verify this fix against the following tests:

- SMSampleDeleteSimpleTest.testSimpleDeleteFromDetailPage
- SMSampleUpdateFromGridTest.testBatchUpdateThenToGrid
- SMSampleUpdateLineageFromGridTest.testSourceHappyPath
- SMSampleDeleteSimpleTest.testSimpleDeleteFromGrid
- SMSampleUpdateFromGridTest.testBatchUpdateDisabledFieldDoesNotUpdate
- SMSampleUpdateFromGridTest.testBatchUpdateFieldsAreNotEnabledByDefault
- SMSampleUpdateLineageFromGridTest.testDerivedSamples
- SMSampleUpdateLineageFromGridTest.testAliquotsHaveReadOnlyParents
- SMSampleUpdateFromGridTest.testGridUpdateWithOutChanges
- SMSampleUpdateFromGridTest.testBatchUpdatePopulatesWithData
- SMSampleUpdateLineageFromGridTest.testLineageDetailsUI
- SMSampleUpdateFromGridTest.testGridUpdateCancel
- SMSampleUpdateLineageFromGridTest.testParentHappyPath
- SMSampleUpdateFromGridTest.testBatchUpdateCancel
- SMSampleUpdateFromGridTest.testBatchUpdateOnMultiplePages
- SMSampleUpdateFromGridTest.testGridUpdateSelectionOnMultiplePages
- SMSampleUpdateFromGridTest.testMenuNotEnabledIfNothingSelected

cnathe (approved):

labkey-danield (approved):
Thanks for running the tests locally Chris. It looks like the build break in tests have been fixed. I've kicked some TC runs just to be sure.

src/org/labkey/test/components/ui/grids/QueryGrid.java
```diff
@@ -175,10 +175,15 @@ public void doAndWaitForUpdate(Runnable func)
...
+     * if the searchTerm is different from the current searchTerm
      */
     public QueryGrid search(String searchTerm)
     {
-        doAndWaitForUpdate(()-> getGridBar().searchFor(searchTerm));
+        String currentSearchExp = getGridBar().getSearchExpression();
+        if (searchTerm.equals(currentSearchExp))
+            getGridBar().searchFor(searchTerm);
+        else
+            doAndWaitForUpdate(()-> getGridBar().searchFor(searchTerm));
```
labkey-tchad, line 186: Why not just no-op if there's nothing to do?
```suggestion
        if (!searchTerm.equals(currentSearchExp))
            doAndWaitForUpdate(()-> getGridBar().searchFor(searchTerm));
```

src/org/labkey/test/components/ui/grids/QueryGrid.java
```diff
@@ -187,7 +192,11 @@ public QueryGrid search(String searchTerm)
      */
     public QueryGrid clearSearch()
     {
-        doAndWaitForUpdate(()-> getGridBar().clearSearch());
+        String currentSearchExp = getGridBar().getSearchExpression();
+        if ("".equals(currentSearchExp))
+            getGridBar().clearSearch();
+        else
+            doAndWaitForUpdate(()-> getGridBar().clearSearch());
```
labkey-tchad, line 199: No-op?
```suggestion
        if (!currentSearchExp.isBlank())
            doAndWaitForUpdate(()-> getGridBar().clearSearch());
```

labkey-chrisj:
I needed to merge Ian's fix in develop to my branch; my branch had his problems.  I've restarted those runs in SM and will watch them before merging

src/org/labkey/test/components/ui/grids/QueryGrid.java
labkey-chrisj: Oh no I just saw this after merging 😞

PR #1104, opened 2022-05-05
Automation test for Importing groups during create from template can result in unauthorized exceptions 

labkey-sweta:
#### Rationale
Automation test for Importing groups during create from template can result in unauthorized exceptions 

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/src/org/labkey/test/tests/ProjectCreatorUserTest.java
+++ b/src/org/labkey/test/tests/ProjectCreatorUserTest.java
@@ -9,6 +9,7 @@ import org.labkey.remoteapi.CommandException;
 import org.labkey.remoteapi.CommandResponse;
 import org.labkey.remoteapi.PostCommand;
 import org.labkey.test.BaseWebDriverTest;
+import org.labkey.test.Locator;
 import org.labkey.test.TestFileUtils;
 import org.labkey.test.TestTimeoutException;
 import org.labkey.test.categories.Daily;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/ProjectCreatorUserTest.java
```diff
@@ -151,6 +164,32 @@ public void testCreateProjectByTemplate() throws IOException, CommandException
...
+        String containerId = ((APIContainerHelper) _containerHelper).getContainerId(TEMPLATE_PROJECT);
+        goToHome();
+        impersonate(PROJECT_CREATOR_USER);
+        Map<String, Object> params = new HashMap<>();
+        params.put("name", PROJECT_NAME_PC);
+        params.put("assignProjectAdmin", "true");
+        params.put("folderType", "Template");
+        params.put("templateSourceId", containerId);
+        params.put("templateIncludeSubfolders", "true");
+        params.put("templateWriterTypes", "Project-level groups and members,Role assignments for users and groups");
```
labkey-adam: Add a `List<String>` here

labkey-adam (changes requested):

labkey-adam (changes requested):
Forget my earlier code review comment: I hadn't realized that the "groups" template writer was getting added "manually." While not ideal, this does test the functionality, which is good enough for now.

labkey-adam:
You're currently posting all the properties as parameters. It might be easier to simply convert your map to a JSONObject and call setJSONObject. This will encode and post everything as JSON which is closer to the client scenario.

labkey-adam (approved):

PR #1112, opened 2022-05-20
Reconcile AttachmentCard and FileUploadField components

labkey-tchad:
#### Rationale
`FileUploadField` wraps functionality for several components, including `AttachmentCard`. Some components were using `FileUploadField` where `AttachmentCard` would be more appropriate.
I've moved some `FileUploadField` functions that `AttachmentCard` was missing and updated `DetailDataPanel` to use `AttachmentCard`.

#### Related Pull Requests
* https://github.com/LabKey/biologics/pull/1322
* https://github.com/LabKey/sampleManagement/pull/961

#### Changes
* Reconcile `AttachmentCard` and `FileUploadField` components
* Un-deprecate `Component.findOrNull`
* Remove log-spam from `GridRow`
* Add `truncateTable` functionality to `QueryAPIHelper`

### DIFF
--- a/src/org/labkey/test/components/Component.java
+++ b/src/org/labkey/test/components/Component.java
@@ -242,10 +242,6 @@ public abstract class Component<EC extends Component.ElementCache> implements Se
             return component;
         }
 
-        /**
-         * @deprecated Use {@link #findOptional(SearchContext)}
-         */
-        @Deprecated
         public C findOrNull(S context)
         {
             return findOptional(context).orElse(null);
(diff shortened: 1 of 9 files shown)


src/org/labkey/test/components/Component.java
```diff
@@ -242,10 +242,6 @@ public C refindWhenNeeded(S context)
             return component;
         }
 
-        /**
-         * @deprecated Use {@link #findOptional(SearchContext)}
-         */
-        @Deprecated
         public C findOrNull(S context)
```
labkey-danield, line 245: Is this just clean up?

labkey-danield (approved):

src/org/labkey/test/components/Component.java
labkey-tchad: Yeah. I deprecated this and `Locator.findElementOrNull` for no particular reason at some point. I undid the deprecation of `findElementOrNull` a while back (#181) and this change matches that.

labkey-chrisj (approved):

PR #1113, opened 2022-05-23
Changes for grid header-menu interactions

labkey-chrisj:
#### Rationale
This works around a change in scroll behavior, perhaps(?) to bring the header cell into view when sorting on a column menu.  It also stops wrapping the column sort in a doAndWaitForUpdate() call when column sort already does that, to avoid having wait-within-a-wait situations.

This was found when investigating [BiologicsAssayQCTest.testUpdateQCStateFromRunDetailsPage](https://teamcity.labkey.org/viewLog.html?buildId=1829901&tab=buildResultsDiv&buildTypeId=LabKey_Trunk_Premium_GitExtraPostgres&branch_LabKey_Trunk_Premium=%3Cdefault%3E#testNameId6310373606914595483), which interacts with a table at the bottom of the screen (which needs to scroll into view) - and this showed two things:
1) the menu fly-up behavior for column header sort menus is weird- as in, the menu for the QC History Grid's Date column disappears or appears elsewhere (as in this image:) 
image

The code in ResponsiveGrid.clickColumnMenuItem attempts to scroll the header cell into view, and then attempts to scroll by another 250px, but this does not happen reliably at regular run-time.  This means the menu toggle is clicked at a time where the header cell is at the bottom of the viewport, forcing the menu to fly up (and do whatever it is that it's doing in that screenshot above).

It turns out that using the scrollToTop argument in scrollIntoView() (which attempts to scroll the target element to the top of the viewport) does reliably bring the header cell up into the viewport enough to allow that menu to drop down.
The weird fly-up behavior may be a product or U/X issue worth its own investigation  

#### Related Pull Requests
n/a>

#### Changes
remove redundant doandWaitFor -> on a call to a method already doing and waiting for update
scroll more aggressively

### DIFF
--- a/src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
+++ b/src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
@@ -182,7 +182,7 @@ public class ResponsiveGrid<T extends ResponsiveGrid> extends WebDriverComponent
 
         // scroll the header cell into view plus some extra vertical scroll to make sure the menu is visible
         getWrapper().scrollIntoView(headerCell);
-        getWrapper().shortWait().until(ExpectedConditions.elementToBeClickable(headerCell));    //
+        getWrapper().scrollBy(0, 1); // First scroll fails sometimes
         getWrapper().scrollBy(0, 250);
 
         WebElement toggle = Locator.tagWithClass("span", "fa-chevron-circle-down")
--- a/src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
+++ b/src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
@@ -92,8 +92,7 @@ public class ResponsiveGrid<T extends ResponsiveGrid> extends WebDriverComponent
      */
     public T sortColumnAscending(String columnLabel)
     {
-        doAndWaitForUpdate(()->
-            sortColumn(columnLabel, SortDirection.ASC));
+        sortColumn(columnLabel, SortDirection.ASC);
         return getThis();
     }
 
@@ -104,8 +103,7 @@ public class ResponsiveGrid<T extends ResponsiveGrid> extends WebDriverComponent
      */
     public T sortColumnDescending(String columnLabel)
     {
-        doAndWaitForUpdate(()->
-            sortColumn(columnLabel, SortDirection.DESC));
+        sortColumn(columnLabel, SortDirection.DESC);
         return getThis();
     }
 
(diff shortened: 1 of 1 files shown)


cnathe (approved):

labkey-danield (dismissed):
This branch is failing several Inventory and LKSM tests with the error:
`Element <a href="#"> could not be scrolled into view`

Example Runs:
[Inventory Postgres](https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_LabkeyPremiumTrunk_GitModules_InventoryPostgres/1831741?hideProblemsFromDependencies=false&expandBuildTestsSection=true&hideTestsFromDependencies=false&expandBuildChangesSection=true)
[Sample Manager A [Postgres]](https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_SampleManager_SampleManagerAPostgres/1831745?hideProblemsFromDependencies=false&expandBuildTestsSection=true&hideTestsFromDependencies=false&expandBuildChangesSection=true)

Please don't check this change in until those test are passing.

labkey-chrisj:
Thanks; I ran this locally against a bunch of SM/FM tests and noticed the same failure there I was trying to fix in Biologics.
Trying to figure out how this change causes the same failure in other situations

labkey-chrisj:
I think I figured it out.
The fix in the initial failing case (scroll to top instead of just into view + 250) pushed the header bar up behind the submenu on pages with enough grid beneath the header bar.

It turns out that the initial 'fix' was compensating for that the +250 scroll right after the first one wasn't happening.
adding a wait for the headerCell to become clickable seems to make that second scroll happen reliably again

labkey-chrisj:
Looks like those problem runs cleaned up with the latest commit:

[Inventory](https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_LabkeyPremiumTrunk_GitModules_InventoryPostgres?branch=gridHeaderMenu_update&buildTypeTab=overview&mode=builds)
[Sample Manager A](https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_SampleManager_SampleManagerAPostgres?branch=gridHeaderMenu_update&buildTypeTab=overview&mode=builds)

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
```diff
@@ -184,11 +182,13 @@ protected void clickColumnMenuItem(String columnLabel, String menuText, boolean
 
         // scroll the header cell into view plus some extra vertical scroll to make sure the menu is visible
         getWrapper().scrollIntoView(headerCell);
+        getWrapper().shortWait().until(ExpectedConditions.elementToBeClickable(headerCell));    //
         getWrapper().scrollBy(0, 250);
```
labkey-tchad: I was able to repro this locally. I don't know that the problem is with `ResponsiveGrid` or more of an issue with `scrollIntoView` not triggering some event. The behavior is just odd.
Scrolling the header into view works (`scrollIntoView(headerCell)`) but the extra scroll does not (`scrollBy(0, 250)`). It doesn't quite no-op though, it makes the subnav bar pop into view (but doesn't scroll) and calling `scrollBy` a second time actually scrolls down far enough to make the pop-up menu visible.
Adding a little extra scroll seems to do the trick.
```suggestion
        getWrapper().scrollIntoView(headerCell);
        getWrapper().scrollBy(0, 1); // First scroll fails sometimes
        getWrapper().scrollBy(0, 250);
```

Additionally, I wouldn't expect waiting for `elementToBeClickable` to do much more than add a slight delay here. "clickable" as far as that method is concerned mean enabled (`WebElement.isEnabled`) and not hidden(`WebElement.isDisplayed`). It doesn't know anything about whether an element is scrolled into view or blocked by another element.

labkey-tchad (dismissed):

labkey-chrisj:
Yes, this seems more like it's an issue with scrolling than with this component in particular.
I hope your fix works!
I've kicked off runs in SM and Inventory to see if it does

PR #1114, opened 2022-05-24
New Grid Filter UX Text Automation.

labkey-danield:
#### Rationale
This updates the test cases in GridPan

Screen Shot 2022-05-23 at 9 14 27 PM
elTest to test more functionality in the new grid filter dialog.

Screen Shot 2022-05-23 at 9 14 40 PM

The [New Grid Filter UX spec](https://docs.google.com/document/d/1PyPw34qpInoicg-hDx0Qzw19PmTrTi6A2s-c2FwMhAI/edit#) identified several test scenarios the needed to be covered. In addition several other tests were added as well to cover other gaps. As a result there are a lot of changes to the GridPanelTest, and it might be easier to view the file as if it were a new file and not as a diff. One of the biggest changes to the test was to use a shared sample type. Previously each test was creating its own sample type, populating and then deleting it at the end of the test. Several of the  older tests were consolidated into other tests.

For the most part only minor changes were made to the test components. Most of the changes involve how to access the filter dialog (e.g. clicking a filter 'pill' above the grid now returns the dialog). Other changes involved fixes to test components.

#### Related Pull Requests
* TBD

#### Changes
* Updated and consolidate tests in GridPanelTest.
* Removed calls to deprecated methods from GridPanelTest.
* Updated Pager component to use MultiMenu (removed references to deprecated DropDownButtonGroup)

### DIFF
--- a/src/org/labkey/test/components/ui/grids/GridBar.java
+++ b/src/org/labkey/test/components/ui/grids/GridBar.java
@@ -416,6 +416,9 @@ public class GridBar extends WebDriverComponent<GridBar.ElementCache>
 
     public GridBar searchFor(String searchStr)
     {
+
+        clearSearch();
+
         _queryGrid.doAndWaitForUpdate(()->
         {
             elementCache().searchBox.set(searchStr);
@@ -442,7 +445,7 @@ public class GridBar extends WebDriverComponent<GridBar.ElementCache>
         return elementCache().searchBox.get();
     }
 
-    public GridFilterModal getFilterDialog()
+    public GridFilterModal openFilterDialog()
     {
         clickButton("Filters");
         return new GridFilterModal(getDriver(), _queryGrid);
--- a/src/org/labkey/test/components/ui/grids/GridBar.java
+++ b/src/org/labkey/test/components/ui/grids/GridBar.java
@@ -433,6 +433,12 @@ public class GridBar extends WebDriverComponent<GridBar.ElementCache>
         return elementCache().searchBox.get();
     }
 
+    public GridFilterModal getFilterDialog()
+    {
+        clickButton("Filters");
+        return new GridFilterModal(getDriver(), _queryGrid);
+    }
+
     @Override
     protected ElementCache newElementCache()
     {
--- a/src/org/labkey/test/components/ui/grids/GridBar.java
+++ b/src/org/labkey/test/components/ui/grids/GridBar.java
@@ -416,15 +416,24 @@ public class GridBar extends WebDriverComponent<GridBar.ElementCache>
 
     public GridBar searchFor(String searchStr)
     {
-        elementCache().searchBox.set(searchStr);
-        elementCache().searchBox.getComponentElement().sendKeys(Keys.ENTER);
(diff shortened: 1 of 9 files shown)


src/org/labkey/test/tests/component/GridPanelTest.java
```diff
@@ -1,385 +1,1749 @@
...
+        return grid;
+    }
+
+    /**
+     * Helper to see if the paging menu is visible.
+     * Basically this is to check issue 45451.
+     *
+     * @return True if the dropdown page menu is visible, false otherwise.
+     */
+    public boolean isPagingMenuVisible()
```
cnathe: should this helper be on the gird component itself?

cnathe (commented):
Test cases in the refactored GridPanelTest look good! I'll defer the rest of the code review to the test team. Thanks Dan.

src/org/labkey/test/tests/component/GridPanelTest.java
labkey-danield: I wasn't sure if it would be useful outside this test, but I can see that it would make sense to be in a component. I've moved it to the Pager test component.

src/org/labkey/test/components/ui/grids/GridBar.java
```diff
@@ -416,15 +416,24 @@ else if(onSamplePage)
 
     public GridBar searchFor(String searchStr)
     {
-        elementCache().searchBox.set(searchStr);
-        elementCache().searchBox.getComponentElement().sendKeys(Keys.ENTER);
+        _queryGrid.doAndWaitForUpdate(()->
+        {
+            elementCache().searchBox.set(searchStr);
+            elementCache().searchBox.getComponentElement().sendKeys(Keys.ENTER);
+        });
```
labkey-tchad, line 426: I think there was a similar change to `searchFor` in the last update to this component and it had to get rolled back. The grid wont update if the search box doesn't actually change (I think this needs something similar to the `if` in `clearSearch`)

src/org/labkey/test/components/ui/grids/GridBar.java
```diff
@@ -433,6 +442,12 @@ public String getSearchExpression()
         return elementCache().searchBox.get();
     }
 
+    public GridFilterModal getFilterDialog()
+    {
+        clickButton("Filters");
+        return new GridFilterModal(getDriver(), _queryGrid);
+    }
+
```
labkey-tchad: `get` methods generally shouldn't change the page state. Maybe `openFilterDialog` instead.

labkey-tchad (approved):
This might be reintroducing a defect in `searchFor`

src/org/labkey/test/components/ui/grids/GridBar.java
labkey-danield: Changed.

src/org/labkey/test/components/ui/grids/GridBar.java
labkey-danield: As discussed, added call to clearSearch before setting the new value.

PR #1123, opened 2022-06-02
Use PostCommand for all testing of CreateProject API

labkey-adam:
#### Rationale
Using the same `PostCommand` approach for all `CreateProject` tests provides consistency and removes the `templateWriterTypes` hack. Also, posting form data is closer to what our clients are doing.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/1104

### DIFF
--- a/src/org/labkey/test/tests/ProjectCreatorUserTest.java
+++ b/src/org/labkey/test/tests/ProjectCreatorUserTest.java
@@ -1,6 +1,9 @@
 package org.labkey.test.tests;
 
-import org.apache.http.HttpStatus;
+import org.apache.http.client.entity.UrlEncodedFormEntity;
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.client.methods.HttpUriRequest;
+import org.apache.http.message.BasicNameValuePair;
 import org.jetbrains.annotations.Nullable;
 import org.junit.Before;
 import org.junit.BeforeClass;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/ProjectCreatorUserTest.java
```diff
@@ -197,8 +201,35 @@ public void testCreateProjectByTemplateWithSubfolderAndPermission() throws Comma
 
     private String createProject(Map<String, Object> params) throws IOException
     {
-        PostCommand<CommandResponse> command = new PostCommand<>("admin", "createProject");
-        command.setParameters(params);
+        // CreateProjectAction is not a real API action, so we POST form data not JSON
```
labkey-chrisj, line 204: Is this intended only for this test, or might it also be useful for tests to generally use when creating projects?

labkey-chrisj (approved):

labkey-sweta (approved):
Thank you for taking a look at that.

src/org/labkey/test/tests/ProjectCreatorUserTest.java
labkey-adam: Good call @labkey-chrisj ... yes, I thought about moving this to the Java API because it could be useful. Unfortunately, current implementation requires modern Java (streams and all that), but the Java API still supports Java 8, so I didn't take the leap right now. We are talking about loosening the requirements on the Java API, but that won't happen immediately.

PR #1128, opened 2022-06-08
Active User Limitations Automation Test

### LINKED ISSUE #1124: Limiting Active Users - selenium tests
#### Rationale
We want the option for LabKey Server deployments to limit the number of active users. See the spec for details: https://docs.google.com/document/d/15TenwfucW9XNO8OhHB6ibEIIyxPycbB9SqcXtoimcgE/edit#

This story adds automation for this new feature.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/3334

**Tasks**

- [x] Add tasks (0.25) @labkey-sweta
- [x] Create LimitActiveUserPage (3) @labkey-sweta
- [x] ActiveUserLimitationTest.testUserWarning - 3 @labkey-sweta
- [x] ActiveUserLimitationTest.testUserLimits - 3 @labkey-sweta
- [x] ActiveUserLimitationTest.testSystemUserLimitation - 3 @labkey-sweta
- [x] ActiveUserLimitationTest.testActivateAndDeactivateUserWithLimits - 3 @labkey-sweta
- [x] Add test case in LDAPTest -3 @labkey-sweta
- [x] Code review changes - 3 @labkey-sweta 
- [x] code review -2 @labkey-adam 
- [ ] merge - 1 @labkey-sweta
**Total Hours -21.25hrs-

labkey-sweta:
#### Rationale
Active User Limitations Automation Test
1. testUserWarning()
2. testUserLimits()
3. testSystemUserLimitation()
4. testActivateAndDeactivateUserWithLimits()

#### Related Pull Requests
https://github.com/LabKey/ldap/pull/57

#### Changes
1. New test class ActiveUserLimitationTest added to Daily 
2. New page class added for LimitActiveUserPage

### DIFF
--- a/src/org/labkey/test/pages/core/admin/LimitActiveUserPage.java
+++ b/src/org/labkey/test/pages/core/admin/LimitActiveUserPage.java
@@ -23,15 +23,17 @@ public class LimitActiveUserPage extends LabKeyPage<LimitActiveUserPage.ElementC
         return new LimitActiveUserPage(wrapper.getDriver());
     }
 
-    public LimitActiveUserPage userWarning(String value)
+    public LimitActiveUserPage userWarning(boolean value)
     {
-        elementCache().userWarning.selectByVisibleText(value);
+        String setValue = value ? "Yes" : "No";
+        elementCache().userWarning.selectByVisibleText(setValue);
         return this;
     }
 
-    public LimitActiveUserPage limitActiveUsers(String value)
+    public LimitActiveUserPage limitActiveUsers(boolean value)
     {
-        elementCache().limitActiveUsers.selectByVisibleText(value);
+        String setValue = value ? "Yes" : "No";
+        elementCache().limitActiveUsers.selectByVisibleText(setValue);
         return this;
     }
 
--- a/src/org/labkey/test/pages/core/admin/LimitActiveUserPage.java
+++ b/src/org/labkey/test/pages/core/admin/LimitActiveUserPage.java
@@ -85,6 +85,7 @@ public class LimitActiveUserPage extends LabKeyPage<LimitActiveUserPage.ElementC
     {
         return elementCache().errorMsg.getText();
     }
+
     public LimitActiveUserPage saveExpectingErrors()
     {
         elementCache().saveBtn.click();
@@ -122,6 +123,6 @@ public class LimitActiveUserPage extends LabKeyPage<LimitActiveUserPage.ElementC
         protected final WebElement saveBtn = Locator.lkButton("Save").findWhenNeeded(this);
         protected final WebElement cancelBtn = Locator.lkButton("Cancel").findWhenNeeded(this);
 
-        protected final WebElement errorMsg = Locator.tagWithClass("div","labkey-error").findWhenNeeded(this);
+        protected final WebElement errorMsg = Locator.tagWithClass("div", "labkey-error").findWhenNeeded(this);
     }
 }
--- a/src/org/labkey/test/pages/core/admin/LimitActiveUserPage.java
+++ b/src/org/labkey/test/pages/core/admin/LimitActiveUserPage.java
@@ -1,0 +1,115 @@
+package org.labkey.test.pages.core.admin;
+
+import org.labkey.test.Locator;
+import org.labkey.test.WebDriverWrapper;
+import org.labkey.test.WebTestHelper;
+import org.labkey.test.components.html.Input;
+import org.labkey.test.pages.LabKeyPage;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.ui.Select;
+
+public class LimitActiveUserPage extends LabKeyPage<LimitActiveUserPage.ElementCache>
+{
+    public LimitActiveUserPage(WebDriver driver)
+    {
+        super(driver);
+        waitForPage();
+    }
+
+    public static LimitActiveUserPage beginAt(WebDriverWrapper wrapper)
+    {
+        wrapper.beginAt(WebTestHelper.buildURL("user", "limitActiveUsers"));
+        return new LimitActiveUserPage(wrapper.getDriver());
+    }
+
+    public LimitActiveUserPage userWarning(String value)
+    {
+        elementCache().userWarning.selectByVisibleText(value);
+        return this;
+    }
+
+    public LimitActiveUserPage limitActiveUsers(String value)
+    {
+        elementCache().limitActiveUsers.selectByVisibleText(value);
+        return this;
+    }
+
+    public String getUserWarningLevel()
+    {
+        return elementCache().userWarningLevel.get();
+    }
+
+    public LimitActiveUserPage setUserWarningLevel(String value)
+    {
+        elementCache().userWarningLevel.set(value);
+        return this;
+    }
+
+    public String getUserLimitLevel()
+    {
+        return elementCache().userLimitLevel.get();
+    }
+
+    public LimitActiveUserPage setUserLimitLevel(String value)
+    {
+        elementCache().userLimitLevel.set(value);
+        return this;
+    }
+
+    public String getUserWarningMessage()
+    {
+        return elementCache().userWarningMessage.getText();
+    }
+
+    public LimitActiveUserPage setUserWarningMessage(String value)
+    {
+        elementCache().userWarningMessage.clear();
+        elementCache().userWarningMessage.sendKeys(value);
+        return this;
+    }
+
+    public String getUserLimitMessage()
+    {
+        return elementCache().userLimitMessage.getText();
+    }
+
+    public LimitActiveUserPage setUserLimitMessage(String value)
+    {
+        elementCache().userLimitMessage.clear();
+        elementCache().userLimitMessage.sendKeys(value);
+        return this;
+    }
+
+    public ShowAdminPage save()
+    {
+        clickAndWait(elementCache().saveBtn);
+        return new ShowAdminPage(getDriver());
+    }
+
+    public ShowAdminPage cancel()
+    {
+        clickAndWait(elementCache().cancelBtn);
+        return new ShowAdminPage(getDriver());
+    }
+
+    @Override
+    protected LimitActiveUserPage.ElementCache newElementCache()
+    {
+        return new LimitActiveUserPage.ElementCache();
+    }
+
+    protected class ElementCache extends LabKeyPage.ElementCache
+    {
+        protected final Select userWarning = new Select(Locator.id("userWarning").findWhenNeeded(this));
+        protected final Input userWarningLevel = Input.Input(Locator.id("userWarningLevel"), getDriver()).findWhenNeeded(this);
+        protected final WebElement userWarningMessage = Locator.id("userWarningMessage").findWhenNeeded(this);
+
+        protected final Select limitActiveUsers = new Select(Locator.id("limitActiveUsers").findWhenNeeded(this));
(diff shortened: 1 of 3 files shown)


src/org/labkey/test/tests/ActiveUserLimitationTest.java
```diff
@@ -0,0 +1,213 @@
...
+        limitActiveUserPage.save();
+
+        log("User limit is not reached yet");
+        _userHelper.createUser(USER1);
+        goToSiteUsers();
+        assertTrue("Banner should be present", isBannerPresent());
+        assertTrue("User should have been created with banner message", isTextPresent(USER1));
+
+        log("Verifying the message when user limit is reached");
+        _userHelper.createUser(USER2, false, false);
```
labkey-adam: Should check the response for the correct error message.

src/org/labkey/test/tests/ActiveUserLimitationTest.java
```diff
@@ -0,0 +1,213 @@
...
+        waitAndClickAndWait(Locator.linkWithText(USER1_DISPLAY_NAME));
+        clickButton("Edit");
+        checkCheckbox(Locator.name("quf_System"));
+        clickButton("Submit");
+
+        log("Creation of one more user should be be allowed");
+        _userHelper.createUser(USER2);
+        goToSiteUsers();
+        assertTrue("Banner should be present", isBannerPresent());
+        assertTrue("User should have been created with banner message", isTextPresent(USER2));
```
labkey-adam, line 173: At this point, test could attempt to unmark that system user... go to user details for that user and edit. There should be warning and then if the "System" box is unchecked followed by submit there should be an error and refuse to save changes.

src/org/labkey/test/tests/ActiveUserLimitationTest.java
```diff
@@ -0,0 +1,213 @@
...
+        log("Reactivate the user");
+        usersPage = goToSiteUsers();
+        usersPage.includeInactiveUsers();
+        usersPage.getUsersTable().setFilter("Email", "Equals One Of (example usage: a;b;c)", USER1 + ";" + USER2);
+        usersPage.getUsersTable().checkCheckbox(0);
+        usersPage.getUsersTable().checkCheckbox(1);
+        usersPage.getUsersTable().clickHeaderButton("Reactivate");
+        waitForElement(Locator.tagWithText("span", "Reactivate"));
+        clickButton("Reactivate");
+        assertTrue("Missing error message", isTextPresent("Failed to activate user2: User limit has been reached so no more users can be reactivated on this deployment."));
```
labkey-adam, line 222: Verify that USER1 got reactivated?

labkey-adam (changes requested):
A few suggestions

labkey-adam:
Also, since you know the `verifySuccess` assert is wrong in `APIUserHelper.createUser()`, now would be a good time to fix that so others aren't misled in the future

src/org/labkey/test/pages/core/admin/LimitActiveUserPage.java
```diff
@@ -0,0 +1,115 @@
...
+    {
+        wrapper.beginAt(WebTestHelper.buildURL("user", "limitActiveUsers"));
+        return new LimitActiveUserPage(wrapper.getDriver());
+    }
+
+    public LimitActiveUserPage userWarning(String value)
+    {
+        elementCache().userWarning.selectByVisibleText(value);
+        return this;
+    }
```
labkey-danield: I would change this parameter to a boolean because you are turning the property on or off, the string parameter doesn't really get you anything. Then set the text to select in the function.

```
public LimitActiveUserPage userWarning(boolean value)
{
    String setValue = boolean ? "Yes" : "No";
    elementCache().userWarning.selectByVisibleText(setValue);
    return this;
}
```

src/org/labkey/test/pages/core/admin/LimitActiveUserPage.java
```diff
@@ -0,0 +1,115 @@
...
+
+    public LimitActiveUserPage limitActiveUsers(String value)
+    {
+        elementCache().limitActiveUsers.selectByVisibleText(value);
+        return this;
+    }
```
labkey-danield: Same comment as above. You could probably even merge these into one method:

```
private void setActiveUsers(Select control, boolean value)
{
    String setValue = value ? "Yes" : "No";
    control.selectByVisibleText(setValue);
}

public LimitActiveUserPage userWarning(boolean value)
{
    setActiveUsers(elementCache().userWarning, value);
    return this;
}

public LimitActiveUserPage limitActiveUsers(boolean value)
{
    setActiveUsers(elementCache().limitActiveUsers, value);
    return this;
}
```

src/org/labkey/test/pages/core/admin/LimitActiveUserPage.java
```diff
@@ -0,0 +1,115 @@
...
+        clickAndWait(elementCache().saveBtn);
+        return new ShowAdminPage(getDriver());
+    }
+
+    public ShowAdminPage cancel()
+    {
+        clickAndWait(elementCache().cancelBtn);
+        return new ShowAdminPage(getDriver());
+    }
+
```
labkey-danield, line 108: It doesn't look like you have any getters for the table at the bottom of the page that shows the current count of active users, remaining users and level values:
Screen Shot 2022-06-08 at 10 18 03 PM

src/org/labkey/test/pages/core/admin/LimitActiveUserPage.java
```diff
@@ -0,0 +1,115 @@
...
+        elementCache().userLimitMessage.clear();
+        elementCache().userLimitMessage.sendKeys(value);
+        return this;
+    }
+
+    public ShowAdminPage save()
+    {
+        clickAndWait(elementCache().saveBtn);
+        return new ShowAdminPage(getDriver());
+    }
```
labkey-danield, line 101: You may want to add a `saveExpectingError` method. Since the limit values are just text boxes they can be set to anything including invalid values. _Hint: you may want to test for that._

And since they can be set to invalid value you may need to add a getter for any error messages that are shown.

src/org/labkey/test/pages/core/admin/LimitActiveUserPage.java
```diff
@@ -0,0 +1,115 @@
...
+        protected final Select userWarning = new Select(Locator.id("userWarning").findWhenNeeded(this));
+        protected final Input userWarningLevel = Input.Input(Locator.id("userWarningLevel"), getDriver()).findWhenNeeded(this);
+        protected final WebElement userWarningMessage = Locator.id("userWarningMessage").findWhenNeeded(this);
+
+        protected final Select limitActiveUsers = new Select(Locator.id("limitActiveUsers").findWhenNeeded(this));
+        protected final Input userLimitLevel = Input.Input(Locator.id("userLimitLevel"), getDriver()).findWhenNeeded(this);
+        protected final WebElement userLimitMessage = Locator.id("userLimitMessage").findWhenNeeded(this);
+
+        WebElement saveBtn = Locator.lkButton("Save").findWhenNeeded(this);
+        WebElement cancelBtn = Locator.lkButton("Cancel").findWhenNeeded(this);
```
labkey-danield: Any reason these aren't `protected final`?

src/org/labkey/test/tests/ActiveUserLimitationTest.java
```diff
@@ -0,0 +1,213 @@
...
+    }
+
```
labkey-danield, line 250: I think there might be a couple of other scenarios that would be worth checking. 

- Validate ActiveUserPage when invalid values are used (see comment in page file).
- Note: An invalid case would also be limit number set to a value less than warning number.
- Validate grid on ActiveUserPage (see comment in page file).
- Adding multiple users on the Add Users page (security-addUsers.view?) Add enough users so that some would be over the limit.
- What about having warning and limit both turned on and then add more users. Does the banner change?
- This maybe beyond the scope, but keep in mind the message can be html, so having something like `<script>alert("You have been <b>warned</b>! The system is limited to ${LimitLevel} users.")</script>` would be valid in the message.
- More realistically might be something like a link in the message. For example I could see this message having a link to an email, or other point of contact, to help resolve the limit issue.

I'll defer to @labkey-adam as to the value of the scenarios I described. Some of the them may already be covered in unit tests and do not need to be covered here.

labkey-danield (changes requested):
I think there are some missing test cases that may need to be covered.

src/org/labkey/test/pages/core/admin/LimitActiveUserPage.java
labkey-sweta: Good catch.

src/org/labkey/test/pages/core/admin/LimitActiveUserPage.java
labkey-sweta: Done

src/org/labkey/test/pages/core/admin/LimitActiveUserPage.java
labkey-sweta: Just curious as to why does it has to be boolean?

src/org/labkey/test/pages/core/admin/LimitActiveUserPage.java
labkey-sweta: Writing getter methods seemed not necessary, since the values(activeUser) is DB specific. 
Validating all the values are correctly populated in the warning message.

src/org/labkey/test/tests/ActiveUserLimitationTest.java
labkey-sweta: Done

src/org/labkey/test/tests/ActiveUserLimitationTest.java
labkey-sweta: Done

src/org/labkey/test/pages/core/admin/LimitActiveUserPage.java
labkey-danield: It doesn't _have to_ be a boolean, I think it works better as a boolean. Heres why:

- Boolean makes it easier for someone who is not familiar with the feature to turn the feature on (true) or off (false).
- A string gives no indication of the possible values that can be sent in. Boolean removes ambiguity.
- A string allows for typos ('yes' vs 'Yes').
- A boolean insulates more from product changes. For example if the value changes from 'Yes'/'No' to 'On'/'Off' only the method on the page has to change not the test cases.

labkey-sweta: Done..! Changed

src/org/labkey/test/pages/core/admin/LimitActiveUserPage.java
labkey-sweta: changed

src/org/labkey/test/tests/ActiveUserLimitationTest.java
labkey-sweta: Most of them have been added as part of code review changes

src/org/labkey/test/tests/ActiveUserLimitationTest.java
labkey-sweta: Done

src/org/labkey/test/pages/core/admin/LimitActiveUserPage.java
labkey-sweta: Done

labkey-adam (approved):

labkey-danield (approved):

PR #1129, opened 2022-06-08
Verify external schema events generate audit trail

labkey-chrisj:
#### Rationale
This provides test coverage to ensure that external schema create/update/delete events are audited, per[ issue 43317](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=43317)

#### Related Pull Requests
n/a

#### Changes
Allows the test to ensure that for a given container, there are audit events for the given external schema

### DIFF
--- a/src/org/labkey/test/tests/ExternalSchemaTest.java
+++ b/src/org/labkey/test/tests/ExternalSchemaTest.java
@@ -18,6 +18,7 @@ package org.labkey.test.tests;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.HttpStatus;
+import org.junit.Assert;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
@@ -40,6 +41,7 @@ import org.labkey.test.TestTimeoutException;
 import org.labkey.test.WebTestHelper;
 import org.labkey.test.categories.Daily;
 import org.labkey.test.categories.Data;
+import org.labkey.test.pages.core.admin.ShowAuditLogPage;
 import org.labkey.test.pages.query.InsertExternalSchemaPage;
 import org.labkey.test.util.DataRegionTable;
 import org.labkey.test.util.SchemaHelper;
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/tests/ExternalSchemaTest.java
```diff
@@ -270,6 +277,15 @@ public void testExternalSchemaDelete()
...
+        assertTrue("expect audit event for schema deletion",
+                getMatchingContainerAuditEvents(containerPath, "External schema 'Test' deleted"));
+    }
+
+    boolean getMatchingContainerAuditEvents(String containerPath, String matchingComment)
+    {
+        var rows = executeSelectRowCommand("auditLog", "ContainerAuditEvent",
+                ContainerFilter.Current, containerPath, null).getRows();
+        return rows.stream().anyMatch(a-> a.get("_labkeyurl_container").toString().contains(containerPath) && a.get("comment").toString().equals(matchingComment));
     }
```
labkey-tchad: Consider navigating to the audit page if the check fails. This will be painful to investigate if it fails.
```suggestion
    void assertContainerAuditEvents(String containerPath, String matchingComment)
    {
        var rows = executeSelectRowCommand("auditLog", "ContainerAuditEvent",
                ContainerFilter.Current, containerPath, null).getRows();
        boolean foundEvent = rows.stream().anyMatch(a-> a.get("_labkeyurl_container").toString().contains(containerPath) && a.get("comment").toString().equals(matchingComment));
        if (!foundEvent)
        {
            beginAt(WebTestHelper.buidUrl("audit", containerPath, "showAuditLog", Map.of("view", "ContainerAuditEvent")); // TODO: Add a new `ShowAuditLogPage.beginAt` method that takes the audit event type
            Assert.fail("Event not found: " + matchingComment);
        }
    }
```

labkey-chrisj: Good idea.  Thanks!
I took the liberty of wiring up that ShowAuditLogPage.beginAt method, and using it here

labkey-tchad (approved):

PR #1130, opened 2022-06-09
verify ~me~ filter for conditional format

labkey-chrisj:
#### Rationale
This adds regression coverage for [issue 40362](https://www.labkey.org/home/Developer/issues/issues-update.view?issueId=40362), the fix for which ensures conditional formats of a User column in a sampletype.

This test creates a sampleType with a user lookup column, sets a conditional filter on "Equals", "~me~", and inserts 1 row that is 'me' and one that isn't

#### Related Pull Requests
n/a

#### Changes
New test method

### DIFF
--- a/src/org/labkey/test/tests/SampleTypeTest.java
+++ b/src/org/labkey/test/tests/SampleTypeTest.java
@@ -26,6 +26,7 @@ import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.labkey.remoteapi.CommandException;
 import org.labkey.remoteapi.Connection;
+import org.labkey.remoteapi.query.Filter;
 import org.labkey.remoteapi.query.SelectRowsCommand;
 import org.labkey.remoteapi.query.SelectRowsResponse;
 import org.labkey.test.BaseWebDriverTest;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/SampleTypeTest.java
```diff
@@ -163,6 +172,50 @@ public void testCreateSampleTypeNoExpression()
...
+                .setField("size", 3)
+                .setField("user", OptionSelect.SelectOption.textOption(USER_FOR_FILTERTEST.getUserDisplayName()))
+                .submit();
+
+        var meCell = Locator.tag("td").withChild(Locator.tagWithText("a", getDisplayName()))
+                .waitForElement(getDriver(), WAIT_FOR_JAVASCRIPT);
+        var notMeCell = Locator.tag("td").withChild(Locator.tagWithText("a", USER_FOR_FILTERTEST.getUserDisplayName()))
+                .findElement(getDriver());
+        assertThat("expect custom format for me filter",
+                meCell.getAttribute("style"), containsString("background-color: rgb(244, 78, 59)"));
```
labkey-tchad: FYI, `WebElement.getCssValue` will parse the style attribute for you and handle when the style attribute isn't set, preventing an NPE (`containsString` doesn't like comparing to null values if I remember correctly).
```suggestion
        assertEquals("expect custom format for me filter", "rgb(244, 78, 59)", meCell.getCssValue("background-color"));
```

labkey-tchad (approved):
Looks good.

src/org/labkey/test/tests/SampleTypeTest.java
labkey-chrisj: cool!  thanks

PR #1131, opened 2022-06-09
Attempt to resolve header menu interaction issues by scrolling to middle

labkey-chrisj:
#### Rationale
Recently some tests (like [this one](https://teamcity.labkey.org/viewLog.html?buildId=1851564&buildTypeId=LabKey_Trunk_LabkeyPremiumTrunk_GitModules_BiologicsPostgres95&fromExperimentalUI=true#testNameId6310373606914595483)) have been failing when they need to scroll-then-scroll-some in order to interact with grid header menus.  It turns out that if we use the scrollIntoView method and tell it to scroll to the top, sometimes that puts the header cell behind the navBar, causing failures.  If we just scrollIntoView when the target cell is at the bottom or below the viewport, the component's menu fly-up behavior seems to evade the automation.

This attempts to find a middle path (literally) between these two cases, by scrolling the target cell vertically (on the y-axis) only.

#### Related Pull Requests
n/a

#### Changes
adds a new 'scrollToMiddle' method to WebDriverWrapper
updates ResponsiveGrid to use it for clicking columnMenus

### DIFF
--- a/src/org/labkey/test/WebDriverWrapper.java
+++ b/src/org/labkey/test/WebDriverWrapper.java
@@ -105,6 +105,7 @@ import java.io.UnsupportedEncodingException;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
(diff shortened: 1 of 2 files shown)


cnathe (commented):
This looks like a good approach. I'll defer the actual CR to Trey.

labkey-chrisj:
[Biologics A](https://teamcity.labkey.org/viewLog.html?buildId=1852378&buildTypeId=LabKey_Trunk_LabkeyPremiumTrunk_GitModules_BiologicsPostgres95) with this fb seems to work out
So does [SM B](https://teamcity.labkey.org/viewLog.html?buildId=1852380&tab=buildResultsDiv&buildTypeId=LabKey_Trunk_Premium_SampleManager_SampleManagerBPostgres&branch_LabKey_Trunk_Premium_SampleManager=scrollToMiddleHopefully)

labkey-tchad (approved):
Seems like a good solution and a useful method.
Thanks for the warning cleanups 👍

PR #1132, opened 2022-06-10
also scroll header cells horizontally to bring them into view

labkey-chrisj:
#### Rationale
merge https://github.com/LabKey/testAutomation/pull/1131 addressed one problem (that header cells weren't being reliably scrolled into an interactable position on the y axis), but inadvertently broke tests that relied on scrollIntoView to horizontally scroll column headers when they're off to the right or left of the viewport.
This fix worked locally, hopefully it also does on TC  (but we've all seen cases where the second script-execution doesn't work, so maybe we will have to update scrollToMiddle such that it also scrolls by x)

#### Changes
call scrollIntoView first if the target isn't displayed, which handles any necessary x-scrolling
     if scrolling into view first, wait 750msec before centering the target
call scrollToMiddle, which approximately centers the target vertically in the viewport

### DIFF
--- a/src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
+++ b/src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
@@ -32,6 +32,7 @@ import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.Assert.assertTrue;
 import static org.labkey.test.BaseWebDriverTest.WAIT_FOR_JAVASCRIPT;
+import static org.labkey.test.WebDriverWrapper.sleep;
 import static org.labkey.test.WebDriverWrapper.waitFor;
 
 public class ResponsiveGrid<T extends ResponsiveGrid> extends WebDriverComponent<ResponsiveGrid<T>.ElementCache> implements UpdatingComponent
@@ -182,7 +183,11 @@ public class ResponsiveGrid<T extends ResponsiveGrid> extends WebDriverComponent
 
         // scroll the header cell as much to the center of the viewport as possible; if it is at the bottom the menu
         // fly-up can be problematic to automate, if to the top sometimes that puts it behind the navbar
-        getWrapper().scrollIntoView(headerCell);    // for cells to the right or left of the viewport, scroll X to visible if visible
+        if (!headerCell.isDisplayed())
+        {
+            getWrapper().scrollIntoView(headerCell);    // for cells to the right or left of the viewport, scroll X to visible if visible
+            sleep(750);     // give script a chance to complete before executing another
+        }
         getWrapper().scrollToMiddle(headerCell);    // scroll Y to middle of the viewport to prevent burying behind navbar
 
         WebElement toggle = Locator.tagWithClass("span", "fa-chevron-circle-down")
--- a/src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
+++ b/src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
@@ -182,10 +182,8 @@ public class ResponsiveGrid<T extends ResponsiveGrid> extends WebDriverComponent
     {
         WebElement headerCell = elementCache().getColumnHeaderCell(columnLabel);
 
-        // scrollIntoView handles horizontal scrolling for cases where the header is non-interactable
-        getWrapper().scrollIntoView(headerCell);    // for cells to the right or left of the viewport, scrollIntoView handles horizontal scroll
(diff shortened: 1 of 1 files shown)


labkey-chrisj:
Unfortunately, webElement.isDisplayed() returns true when the element is otherwise not interactable, so attempting to only scrollIntoView() if the element wasn't visible resulted in not scrolling horizontally when really that was needed.

Until we can contrive a smart way to only scroll by X if needed, it seems the best strategy is to always scrollIntoView, sleep briefly to make it more likely that the next scroll script execution will actually happen, then scrollYToCenter.  (ugh, I don't like the sleep but if it makes the scroll-to-center happen reliably it's what we've got to work with)

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
```diff
@@ -180,13 +182,15 @@ protected void clickColumnMenuItem(String columnLabel, String menuText, boolean
     {
         WebElement headerCell = elementCache().getColumnHeaderCell(columnLabel);
 
-        // scroll the header cell as much to the center of the viewport as possible; if it is at the bottom the menu
-        // fly-up can be problematic to automate, if to the top sometimes that puts it behind the navbar
-        getWrapper().scrollToMiddle(headerCell);
+        // scrollIntoView handles horizontal scrolling for cases where the header is non-interactable
+        getWrapper().scrollIntoView(headerCell);    // for cells to the right or left of the viewport, scrollIntoView handles horizontal scroll
+        sleep(500);  // todo: find a way to test for whether or not x-scroll is needed, and only x-scroll if necessary
+                         //  sleep here to give scrollToMiddle call below a better chance of firing
```
labkey-tchad: Not sure if it would work any better but have you tried scrolling the entire grid to the middle instead of the header cell? The scrolling methods might be getting confounded by the nested scrolling areas (scrolling the grid vs scrolling the whole page).
```suggestion
        // make sure grid headers are in the middle of the page
        getWrapper().scrollToMiddle(getComponentElement());
```

labkey-danield (commented):
I agree with the comment @labkey-tchad left. I think scrolling the grid element first before trying to scroll the header element. There are several pages where the grid is rendered well below the the bottom of the browser. So it might be worth a try.

On the positive side TC does appear to be happy with this change.

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
labkey-chrisj: I'm happy to try- but my experience so far suggests that for tests that interact with (for example) the Storage Status column, scrollToMiddle leaves those columns scrolled out of view to the right, causing the failures we're seeing in develop right now. 

What I haven't been able to figure out yet is a way to check from java whether or not the target cell is inside the bounds of the component's rectangle- maybe this needs to be done in JS or there's something else I haven't checked or noticed

labkey-tchad: The horizontal scrolling doesn't seem to be a problem on develop. This failure is trying to filter the "Status" column, which is visible:
image

labkey-chrisj: I just tried your suggestion (scrolling the component, instead of the target cell) and unfortunately it wasn't successful.  I'm willing to push to TC and see if it works any better there but I doubt it will

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
```diff
@@ -198,6 +200,14 @@ protected void clickColumnMenuItem(String columnLabel, String menuText, boolean
...
     }
 
+    private boolean isElementWithinComponentRectangle(WebElement element)
+    {
+        var containerRect = getComponentElement().getRect();
+        var elementRect = element.getRect();
+        return elementRect.getX() > containerRect.getX();
+    }
+
+
```
labkey-tchad: Is this unused?
```suggestion
```

labkey-tchad (approved):
Thanks for trying.
Remove the unused method, and the sleep, if possible.

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
labkey-chrisj: yes, it's unused. this was a crack at testing to see if I could be smart about whether or not the x-scroll is necessary
Abandoned when I couldn't tell if the assumptions (is the component rect what I see, or does it include the scrolled-out-of-view parts?) were true

labkey-danield (approved):

PR #1143, opened 2022-06-21
Scroll grid into view

labkey-tchad:
#### Rationale
Need to make sure the grid header is in view before clicking. Not sure why `scrollToMiddle` is ineffective.
[Issue 45553: App Grid: Grid column menu does not scroll when page is scrolled.](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45553)

#### Changes
* Scroll grid instead of header cell before opening menu


labkey-tchad:
This change broke some other interactions.
Closing for now; some product change seems to have cleared up the problem anyway.

PR #1149, opened 2022-07-02
Additions to EntityInsertPanel

labkey-martyp:
#### Rationale
The Biologics registry entity insert panel usage has some new cases. Built-in registry types may or may not have the tab to insert via a grid. Some built-in registry types have different allowable formats, and some do not allow the update existing data on import option. 

#### Related Pull Requests
* https://github.com/LabKey/biologics/pull/1419

#### Changes
* Add some helpers to EntityInsertPanel test component to verify the tabs available, allowable formats and update option.

### DIFF
--- a/src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
+++ b/src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
@@ -52,7 +52,16 @@ public class EntityInsertPanel extends WebDriverComponent<EntityInsertPanel.Elem
 
     public ReactSelect targetEntityTypeSelect()
     {
-        return ReactSelect.finder(getDriver()).withName("targetEntityType").waitFor();
+        // Which tabs are available and selected can vary so try finding the visible react select
+        ReactSelect firstTabSelect = ReactSelect.finder(getDriver()).withName("targetEntityType").index(0).waitFor();
+        if (firstTabSelect.getComponentElement().isDisplayed())
+        {
+            return firstTabSelect;
+        }
+        else
+        {
+            return ReactSelect.finder(getDriver()).withName("targetEntityType").index(1).waitFor();
+        }
     }
 
     private ReactSelect parentEntityTypeSelect(String label)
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
```diff
@@ -185,6 +194,11 @@ public EntityInsertPanel setUpdateDataForFileUpload(boolean checked)
         return this;
     }
 
+    public boolean hasUpdateDataOption()
+    {
+        return elementCache().updateDataCheckbox.isDisplayed();
```
labkey-chrisj: using isDisplayed() here suggests that it will always be present, but only sometimes not hidden via style- is that the case?  If it's possible that it might not be there at all, it might be advisable to use dataCheckBoxLocator.existsIn(context) as a strategy to see if that elements exists

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
```diff
@@ -52,7 +52,16 @@ public WebDriver getDriver()
 
     public ReactSelect targetEntityTypeSelect()
     {
-        return ReactSelect.finder(getDriver()).withName("targetEntityType").waitFor();
+        // Which tabs are available and selected can vary so try finding the visible react select
+        ReactSelect firstTabSelect = ReactSelect.finder(getDriver()).withName("targetEntityType").index(0).waitFor();
```
labkey-chrisj, line 56: rather than finding by index and testing for .isDisplayed(), perhaps .notHidden() will work?
Unfortunately, searching by index without limiting scope can return you a reactSelect that might not be within the bounds of the component.  Definitely consider supplying the searchContext in the .waitFor(getComponentElement()) call

labkey-chrisj (approved):

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
labkey-martyp: Cool thanks updated per your recommendation

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
labkey-martyp: I wasn't able to get that to work. Couldn't get this to work without checking the first two indexes (one for the two possible tabs)

PR #1159, opened 2022-07-11
Issue 44518: Use date picker for date fields in the editable grid

XingY:
#### Rationale
Add util to set day and time using datetime picker instead of keyboard input

#### Related Pull Requests
*

### DIFF
--- a/src/org/labkey/test/components/react/ReactDatePicker.java
+++ b/src/org/labkey/test/components/react/ReactDatePicker.java
@@ -1,5 +1,6 @@
 package org.labkey.test.components.react;
 
+import org.apache.commons.lang3.StringUtils;
 import org.labkey.test.Locator;
 import org.labkey.test.WebDriverWrapper;
 import org.labkey.test.components.Component;
@@ -49,6 +50,18 @@ public class ReactDatePicker extends WebDriverComponent<ReactDatePicker.ElementC
         set(value, true);
     }
 
+    // use keyboard input to set year, month. then use picker to select day and time
+    // value must be of "yyyy-MM-dddd" or "yyyy-MM-dddd hh:ss" format
+    public void select(String value)
+    {
+        set("", false);
+        String[] dateParts = getParsedDateParts(value);
+        set(dateParts[0], false); // use keyboard input to set year and month
+        elementCache().datePickerDateCell(dateParts[1]).click(); // use calendar ui to select day
+        if (!StringUtils.isEmpty(dateParts[2])) // use timepicker to select time
+            clickTime(dateParts[2]);
+    }
+
     public void clear()
     {
         set("");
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/components/react/ReactDatePicker.java
```diff
@@ -37,10 +37,16 @@ public String get()
...
         elementCache().input.set(value);
-        elementCache().input.getComponentElement().sendKeys(Keys.ENTER); // Dismiss date picker
+        if (close)
+            elementCache().input.getComponentElement().sendKeys(Keys.ENTER); // Dismiss date picker
+    }
+
+    public void set(String value)
+    {
+        set(value, true);
     }
```
labkey-tchad: Create a new method for setting via the UI
```suggestion
        elementCache().input.getComponentElement().sendKeys(Keys.ENTER); // Dismiss date picker
    }

    public void setWithPopup(String yearMonth, String day, String time)
    {
        expand();
        elementCache().input.set(value); // use keyboard input to set year and month
        elementCache().datePickerDateCell(day).click(); // use calendar ui to select day
        if (!StringUtils.isEmpty(time)) // use timepicker to select time
            clickTime(time); // Clicking a time dismisses the date picker
        else
            elementCache().input.getComponentElement().sendKeys(Keys.ENTER); // Dismiss date picker
    }
```

src/org/labkey/test/components/ui/grids/EditableGrid.java
```diff
@@ -366,6 +396,34 @@ public void setCellValue(int row, String columnName, Object value)
         }
     }
 
+    /**
+     * Parse "2022-07-11 08:30" to ["2022-07", "11", "8:30 AM"]
+     * @param fullDateStr
+     * @return
+     */
+    private String[] getParsedDateParts(String fullDateStr)
```
labkey-danield: Doesn't this assume dates will always be in the format yyyy-mm-dd? Could that assumption be avoided by using Java's Date object to parse the string and get the parts from it?

src/org/labkey/test/components/ui/grids/EditableGrid.java
```diff
@@ -351,10 +365,26 @@ public void setCellValue(int row, String columnName, Object value)
...
+                ReactDatePicker datePicker = elementCache().datePicker();
+                datePicker.set("", false);
+                String[] dateParts = getParsedDateParts((String) value);
+                datePicker.set(dateParts[0], false); // use keyboard input to set year and month
+                elementCache().datePickerDateCell(dateParts[1]).click(); // use calendar ui to select day
+                if (!StringUtils.isEmpty(dateParts[2])) // use timepicker to select time
+                    datePicker.clickTime(dateParts[2]);
+
+                WebElement inputCell = elementCache().inputCell();
+                inputCell.sendKeys(Keys.RETURN);
```
labkey-tchad: Move the calendar and day pickers into `ReactDatePicker`:
```suggestion
                ReactDatePicker datePicker = elementCache().datePicker();
                String[] dateParts = getParsedDateParts((String) value);
                datePicker.setWithPopup(dateParts[0], dateParts[1], dateParts[2]);
```

src/org/labkey/test/components/ui/grids/EditableGrid.java
```diff
@@ -351,10 +365,26 @@ public void setCellValue(int row, String columnName, Object value)
             // Treat the object being sent in as a string.
             // Get the inputCell enter the text and then make the inputCell go away (hit RETURN).
 
-            WebElement inputCell = elementCache().inputCell();
-            inputCell.clear();
+            if (useDatePicker)
```
labkey-tchad, line 368: Consider checking `value`s type instead of adding a special flag for using the date picker
```suggestion
            if (value instanceof Date)
```

labkey-tchad (commented):
The date picker interactions should go in the date picker component.

src/org/labkey/test/components/ui/grids/EditableGrid.java
XingY: The idea is that both keyboard input and picker selection are supported, and we don't want to switch all date field test path to use picker. But will add a format check to confirm dateStr is of yyyy-MM-dd format.

labkey-danield (approved):
I only had one comment to add (other than the ones @labkey-tchad added).
May want to be careful about assuming specific date formats.

src/org/labkey/test/components/ui/grids/EditableGrid.java
trey: Hi! 👋🏻 Wrong Trey.

XingY:
@labkey-tchad I had to merge this branch before you re-review since I accidentally merged sampleManagement PR which has a test dependency on this branch. Free feel review post merge and I can make additional changes in a new pr. thanks.

PR #1161, opened 2022-07-12
Add component support for ParentEntityEditPanel

labkey-chrisj:
#### Rationale
This change adds new test components to wrap the lineage/parent/source details panels seen on sample overview pages as instrumented in ParentEntityEditPanel, using a view/edit pattern similar to DetailDataPanel and DetailTableEdit.

This component is used to show (and make editable) lineage relationships for Samples or other entities, often seen as 'Parent Details' or 'Source Details' panels looking like this:
image
Note that this contains a detail table containing the name of the target type, and a grid showing records for instances of that type designated as parent or source for the current record, _for each target type_

In edit mode, the component shows ReactSelects in pairs; one allowing the user to select a type, the other allowing the user to select one or more records of that type.  (also shown: the ability to add or remove parent/source types, set and clear parent/source IDs)
image

#### Related Pull Requests
biologics PR containing relevant tests and page changes https://github.com/LabKey/biologics/pull/1439

#### Changes

- adds new components to wrap the ParentEntityEditPanel
- fixes an old copy/paste error to correct a wrong error message


src/org/labkey/test/components/ui/grids/ParentDetailPanel.java
```diff
@@ -0,0 +1,177 @@
...
+        return getWrapper().getTexts(elementCache().gridParentTypesLinks());
+    }
+
+    /*
+        gets the details table for the given parent/source type
+     */
+    public DetailTable detailTableFor(String type)
+    {
+        return elementCache().detailTableFor(type);
+    }
```
labkey-danield: This method name feels odd. The data it is getting, basically the name of the parent sample type, is a table only for formatting purposes and there isn't any detail other than the name. Not sure that it adds any more useful info than the getParentTypes() method?

src/org/labkey/test/components/ui/grids/ParentDetailPanel.java
```diff
@@ -0,0 +1,177 @@
...
+
+    /*
+        gets the grid containing parents/sources for the given parent/source type
+     */
+    public ResponsiveGrid getParentsGridFor(String type)
+    {
+        return elementCache().responsiveGridFor(type);
+    }
```
labkey-danield, line 95: Nice! This is the key to the panel.

src/org/labkey/test/components/ui/grids/ParentDetailPanel.java
```diff
@@ -0,0 +1,177 @@
...
+
+    /*
+        gets the details table for the given parent/source type
+     */
+    public DetailTable detailTableFor(String type)
+    {
+        return elementCache().detailTableFor(type);
+    }
+
```
labkey-danield, line 88: You may want to consider adding a case when there are no parents. For example you may want to add a hasParents() method (although getParentTypes can tell you that by returning an empty list).

When there are no parents returning a DetailTable might be useful, or you could just return the text of the panel.

src/org/labkey/test/components/ui/grids/ParentDetailPanel.java
```diff
@@ -0,0 +1,177 @@
...
+{
+    private final WebElement _el;
+    private final WebDriver _driver;
+
+    protected ParentDetailPanel(WebElement element, WebDriver driver)
+    {
+        _el = element;
+        _driver = driver;
+    }
+
```
labkey-danield, line 35: You should consider adding a something like a waitForReady method (or it's equivalent). If a sample has several parents across multiple types it could take a moment to load them. You should check that there are no spinners and/or that a responsive table is present or that the "No parent type has been set" text is present.

src/org/labkey/test/components/ui/grids/ParentDetailPanelEdit.java
```diff
@@ -0,0 +1,164 @@
...
+import org.labkey.test.components.react.ReactSelect;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.ui.ExpectedConditions;
+
+/*
+    implements internal/components/ParentEntityEditPanel, in its edit mode.
+
+ */
+public class ParentDetailPanelEdit extends WebDriverComponent<ParentDetailPanelEdit.ElementCache>
```
labkey-danield: Note:
There already is a org.labkey.test.components.ui.entities.ParentEntityEditPanel I don't think this class is needed.

labkey-danield (changes requested):
There already is a ParentEntityEditPanel under labkey.test.components.ui.entities, I don't think the ParentDetailPanelEdit is needed.

src/org/labkey/test/components/ui/grids/ParentDetailPanelEdit.java
labkey-chrisj: Thanks, I didn't spot that.  Refactoring now to use it instead

labkey-danield (approved):

PR #1165, opened 2022-07-15
Sample Finder Saving Queries - selenium tests

XingY:
#### Rationale
This PR adds test components for managing and saving Sample Finder views.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/901
* https://github.com/LabKey/sampleManagement/pull/1106
* https://github.com/LabKey/testAutomation/pull/1165

#### Changes
* Added Sample Finder test components
   * ManageSampleFinderViewsModal
   * SavedSearchesMenu
   * SaveSampleFinderViewModal

### DIFF
--- a/src/org/labkey/test/components/ui/search/ManageSampleFinderViewsModal.java
+++ b/src/org/labkey/test/components/ui/search/ManageSampleFinderViewsModal.java
@@ -29,8 +29,6 @@ public class ManageSampleFinderViewsModal extends ModalDialog
         {
             editIcon.click();
         }
-        clearElementCache();
-        elementCache(); // waitForReady
     }
 
     public void deleteView(String viewName)
@@ -40,8 +38,6 @@ public class ManageSampleFinderViewsModal extends ModalDialog
         {
             deleteIcon.click();
         }
-        clearElementCache();
-        elementCache(); // waitForReady
     }
 
     public void deleteAllViews()
@@ -124,8 +120,7 @@ public class ManageSampleFinderViewsModal extends ModalDialog
 
     public void clickDone()
     {
-        elementCache().doneBtn.click();
-        waitForClose();
+        dismiss("Done editing");
     }
 
     public String getErrorMsg()
@@ -147,12 +142,9 @@ public class ManageSampleFinderViewsModal extends ModalDialog
 
     protected class ElementCache extends ModalDialog.ElementCache
     {
-        WebElement errorMsg = Locator.tagWithClassContaining("div", "alert-danger").findWhenNeeded(getComponentElement());
+        WebElement errorMsg = Locator.tagWithClassContaining("div", "alert-danger").refindWhenNeeded(getComponentElement());
 
-        WebElement nameInput = Locator.tag("input").findWhenNeeded(getComponentElement());
+        WebElement nameInput = Locator.tag("input").refindWhenNeeded(getComponentElement());
 
-        WebElement doneBtn = Locator.tagWithClassContaining("button", "btn-default")
-                .withText("Done editing")
-                .findWhenNeeded(getComponentElement());
     }
 }
--- a/src/org/labkey/test/components/ui/search/ManageSampleFinderViewsModal.java
+++ b/src/org/labkey/test/components/ui/search/ManageSampleFinderViewsModal.java
@@ -1,0 +1,158 @@
+package org.labkey.test.components.ui.search;
+
+import org.labkey.test.Locator;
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/components/ui/search/ManageSampleFinderViewsModal.java
```diff
@@ -0,0 +1,158 @@
...
+
+    public void editViewName(String viewName)
+    {
+        WebElement editIcon = getViewEditIcon(viewName);
+        if (editIcon != null)
+        {
+            editIcon.click();
+        }
+        clearElementCache();
+        elementCache(); // waitForReady
```
labkey-danield: I'm not sure you need to do this. If the concern is that the controls will become stale/invalid you could try using a refindWhenNeeded for the controls in the element cache. The refind will find the elements every time they are referenced.

Alternatively if you are waiting for the dialog to be ready (it takes time to get and render the data) you could try implementing a waitForReady method that you explicitly call. There is an example of one in org.labkey.test.components.inventory.shared.FMBaseLocationTreeDialog

src/org/labkey/test/components/ui/search/ManageSampleFinderViewsModal.java
```diff
@@ -0,0 +1,158 @@
...
+
+    public void deleteView(String viewName)
+    {
+        WebElement deleteIcon = getViewDeleteIcon(viewName);
+        if (deleteIcon != null)
+        {
+            deleteIcon.click();
+        }
+        clearElementCache();
+        elementCache(); // waitForReady
```
labkey-danield: Same comment as above.

src/org/labkey/test/components/ui/search/ManageSampleFinderViewsModal.java
```diff
@@ -0,0 +1,158 @@
...
+        input.sendKeys(name);
+        input.sendKeys(Keys.TAB);
+        return this;
+    }
+
+    public void clickDone()
+    {
+        elementCache().doneBtn.click();
+        waitForClose();
+    }
```
labkey-danield: Because this extends ModalDialog you don't need to implement this method. The test can simple call the dismiss method with the button text. Alternatively you could change this method to:
```
    public void clickDone()
    {
        dismiss("Done editing");
    }

```
Calling the dismiss method will also call the waitForClose method.

src/org/labkey/test/components/ui/search/ManageSampleFinderViewsModal.java
```diff
@@ -0,0 +1,158 @@
...
+    protected ElementCache elementCache()
+    {
+        return (ElementCache) super.elementCache();
+    }
+
+    protected class ElementCache extends ModalDialog.ElementCache
+    {
+        WebElement errorMsg = Locator.tagWithClassContaining("div", "alert-danger").findWhenNeeded(getComponentElement());
+
+        WebElement nameInput = Locator.tag("input").findWhenNeeded(getComponentElement());
```
labkey-danield: Maybe use refindWhenNeeded for these elements.

src/org/labkey/test/components/ui/search/SaveSampleFinderViewModal.java
```diff
@@ -0,0 +1,87 @@
...
+    public String clickSave(boolean hasError)
+    {
+        elementCache().saveBtn.click();
+        if (!hasError)
+        {
+            waitForClose();
+            return null;
+        }
+        return getErrorMsg();
+    }
```
labkey-danield: I believe the pattern we are trying to make the standard in test automation is to have a method named `clickSaveExpectingError()` rather than have an overload that takes a parameter. Admittedly we are not very consistent, but I believe that is the goal. 

These two methods could become:
```
    public void clickSave()
    {
        dismiss("Save");
    }

    public String clickSaveExpectingError()
    {
        elementCache().saveBtn.click();
        return getErrorMsg();
    }

```

src/org/labkey/test/components/ui/search/SaveSampleFinderViewModal.java
```diff
@@ -0,0 +1,87 @@
...
+
+    public void clickCancel()
+    {
+        elementCache().cancelButton.click();
+        waitForClose();
+    }
```
labkey-danield: I _think_ the close() method in ModalDialog is the same as calling cancel.

labkey-danield (approved):

PR #1168, opened 2022-07-18
Improve test support for script engines and reports

labkey-tchad:
#### Rationale
Interaction with the script report editor has gone through the `RReportHelper` for a long time. The page is used for other report types and having a distinct page class will be useful.

#### Related Pull Requests
* https://github.com/LabKey/docker/pull/74

#### Changes
* Create `ScriptReportPage` test wrapper
* Add support for "Docker Report" script engine

### DIFF
--- a/src/org/labkey/test/pages/ConfigureReportsAndScriptsPage.java
+++ b/src/org/labkey/test/pages/ConfigureReportsAndScriptsPage.java
@@ -15,7 +15,7 @@
  */
 package org.labkey.test.pages;
 
-import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.text.WordUtils;
 import org.jetbrains.annotations.Nullable;
 import org.labkey.test.Locator;
 import org.labkey.test.WebDriverWrapper;
@@ -74,7 +74,7 @@ public class ConfigureReportsAndScriptsPage extends LabKeyPage
     public void setSiteDefault(String engineName)
     {
         log("Ensure " + engineName + " is the site default engine");
-        editEngine(engineName);
+        EditEngineWindow editEngineWindow = editEngine(engineName);
         Locator.XPathLocator defaultCheckbox = Locator.id("editEngine_default-inputEl");
         if (_ext4Helper.isChecked(defaultCheckbox))
         {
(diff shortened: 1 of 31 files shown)


src/org/labkey/test/pages/ConfigureReportsAndScriptsPage.java
```diff
@@ -363,33 +370,26 @@ public void setOutputFileName(String outputFileName)
...
-                        return "R Docker";
-                    }
-                };
+        R_DOCKER,
+        ;
 
         public String toString()
         {
-            return StringUtils.capitalize(name().toLowerCase());
+            return WordUtils.capitalize(name().toLowerCase().replace("_", " "));
```
labkey-danield, line 392: Are you using WordUtil because it will handle a null?

src/org/labkey/test/pages/reports/ScriptReportPage.java
```diff
@@ -0,0 +1,227 @@
...
+        }
+        else
+        {
+            windowTitle = isSaveAs ? "Save Report As" : "Save Report";
+        }
+
+        Window<?> window = new Window<>(windowTitle, getWrappedDriver());
+        WebElement nameInput = Locator.xpath("//input[contains(@class, 'x4-form-field')]").findElement(window);
+        setFormElement(nameInput, name);
+        waitForFormElementToEqual(nameInput, name); // Make sure it sticks
```
labkey-danield, line 112: Beyond the scope of this change, but I think some of our _set_ methods should check that they did indeed do what they were asked to do.

src/org/labkey/test/pages/reports/ScriptReportPage.java
```diff
@@ -0,0 +1,227 @@
...
+            return _label;
+        }
+
+        @Override
+        public boolean isCheckbox()
+        {
+            return _isCheckbox;
+        }
+
+        public String getSection()
```
labkey-danield, line 234: I think you forgot the `@Override` for this method.

labkey-danield (approved):

src/org/labkey/test/pages/ConfigureReportsAndScriptsPage.java
labkey-tchad: `WordUtils` handles multiple words. Relevant here for `DOCKER_REPORT`.
`WordUtils.capitalize("DOCKER_REPORT".toLowerCase().replace("_", " "))` will return "Docker Report". The `StringUtils` version would return "Docker report".

src/org/labkey/test/pages/reports/ScriptReportPage.java
labkey-tchad: `setFormElement` does wait for the value to be correct but it only warns if there is a problem. There are enough places would probably fail if `setFormElement` was more strict. There are some places where the text gets immediately re-rendered into a span or something and the input element is just kind of a placeholder.

labkey-chrisj (approved):

PR #1174, opened 2022-07-20
Add test automation for Issue 45861.

labkey-danield:
#### Rationale
When sample types are renamed saved views and charts should still be visible. Issue 45861 calls this out.

#### Related Pull Requests
* None

#### Changes
* Added a test to validate a saved view is still available after the sample type is renamed.
* Added a test to validate a saved (bar) chart is still available after the sample type is renamed.

### DIFF
--- a/src/org/labkey/test/tests/SampleTypeRenameTest.java
+++ b/src/org/labkey/test/tests/SampleTypeRenameTest.java
@@ -1,0 +1,267 @@
+package org.labkey.test.tests;
+
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.labkey.remoteapi.CommandException;
+import org.labkey.test.BaseWebDriverTest;
+import org.labkey.test.Locator;
+import org.labkey.test.categories.Daily;
+import org.labkey.test.components.ChartTypeDialog;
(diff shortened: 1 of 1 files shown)


XingY (approved):

src/org/labkey/test/tests/SampleTypeRenameTest.java
```diff
@@ -0,0 +1,267 @@
...
+    {
+        SampleTypeRenameTest init = (SampleTypeRenameTest) getCurrentTest();
+        init.doSetup();
+    }
+
+    private void doSetup() throws IOException, CommandException
+    {
+        PortalHelper portalHelper = new PortalHelper(this);
+        _containerHelper.createProject(PROJECT_NAME, null);
+        portalHelper.enterAdminMode();
```
labkey-sweta: You don't need to explicitly goto Page Admin mode and exit, addWebPart does it

labkey-sweta (approved):

src/org/labkey/test/tests/SampleTypeRenameTest.java
labkey-danield: Thanks for the pointer.
Updated code to:
`new PortalHelper(this).addWebPart("Sample Types");`

PR #1176, opened 2022-07-21
Fix Test Issue in LKSM

labkey-danield:
#### Rationale
There are some intermittent failure in the lineage tests in SampleManager. These failures happen because a parent sample/source was not removed when editing a child sample. I suspect that the issue is the ReactControl did not successfully remove the selected item from the control. This adds code in the ParentEntityEditPanel to check that an update to a parent, add or remove, happened. If the addition or removal is not successful an exception will be thrown. This will cause the test to fail at that time and not later on where it is more difficult to debug.

All of the Biologics, Labbook, Inventory and SampleManager tests have been run on TC using this change/branch.

#### Related Pull Requests
* TBD

#### Changes
* Added a waitFor check in the addParents method.
* Added a waitFor check in the removeParent method.

### DIFF
--- a/src/org/labkey/test/components/ui/entities/ParentEntityEditPanel.java
+++ b/src/org/labkey/test/components/ui/entities/ParentEntityEditPanel.java
@@ -424,9 +424,12 @@ public class ParentEntityEditPanel extends WebDriverComponent<ParentEntityEditPa
 
     public ParentEntityEditPanel removeParent(String parentEntity, String parentId)
     {
-        getParent(parentEntity).removeSelection(parentId);
 
-        WebDriverWrapper.waitFor(()->!getParent(parentEntity).getSelections().contains(parentId),
+        WebDriverWrapper.waitFor(()->
+                {
+                    getParent(parentEntity).removeSelection(parentId);
+                    return !getParent(parentEntity).getSelections().contains(parentId);
+                },
                 String.format("Parent %s / %s was not removed.", parentEntity, parentId), 1_000);
 
         return this;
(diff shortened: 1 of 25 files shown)


src/org/labkey/test/components/ui/entities/ParentEntityEditPanel.java
```diff
@@ -423,7 +426,23 @@ public ParentEntityEditPanel removeEntityType(String typeName)
...
     {
-        getParent(parentEntity).removeSelection(parentId);
+
+        // The ReactSelect does not always reliably remove a selected item. If it did not successfully do it on the first
+        // try, try again until it is gone or the time limit is hit.
+        WebDriverWrapper.waitFor(()->
+                {
+                    // Guard against react selector being in a 'Loading...' state.
+                    WebDriverWrapper.waitFor(()->getParent(parentEntity).getComponentElement().isDisplayed(),
+                            String.format("Selector for parents from '%s' is not visible.", parentEntity), 1_000);
```
labkey-tchad: This shouldn't be necessary. `BaseReactSelect.removeSelection` already does this wait (plus more).
```suggestion
```

labkey-tchad (commented):
Discussed moving change to `BaseReactSelect`. Ping for review once that's ready.

src/org/labkey/test/components/ui/entities/ParentEntityEditPanel.java
labkey-danield: I've removed my changes to the ParentEntityEditPanel, and now only have a check in the BaseReactSelect to fail if it does not remove a selected element in the multi select control.
This should help debug random/intermittent failures in LKSM where lineage updates are not as expected.

labkey-tchad (approved):

PR #1177, opened 2022-07-22
StudySecurityTest fix for masked WebElement with locked table column header

cnathe:
#### Rationale
Test failure after locked column headers added to study dataset security page because of the header locking masking a select input after the page is scrolled.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3541

#### Changes
* Scroll dataset security select input to middle of page before selecting from it


labkey-danield (approved):
I think this will generally work, but I don't know that it guarantees that the element will be out from the locked column header. Lacking anything better I think this is ok.

Maybe at some point in the future we could add a method specific for elements in grids that takes into account the locked column header and uses the scrollBy method.

labkey-tchad (approved):

PR #1179, opened 2022-07-23
Extract CreateProjectCommand and prepare to move to remote api

labkey-adam:
#### Rationale
Improve the usability of this command. We'll move it into labkey-client-api once the library supports Java 17.

#### Related Pull Requests
* https://github.com/LabKey/labkey-api-java/pull/31

#### Changes
* Extract a named class and use setters instead of String-based maps, like a real API

### DIFF
--- a/src/org/labkey/test/tests/CreateProjectCommand.java
+++ b/src/org/labkey/test/tests/CreateProjectCommand.java
@@ -10,33 +10,74 @@ import org.labkey.remoteapi.PostCommand;
 import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.List;
-import java.util.Map;
 
 public class CreateProjectCommand extends PostCommand<CommandResponse>
 {
-    private final Map<String, Object> _params;
+    private String _name;
+    private String _folderType;
+    private boolean _assignProjectAdmin = false;
+    private String _templateSourceId;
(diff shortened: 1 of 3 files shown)


src/org/labkey/test/tests/CreateProjectCommand.java
```diff
@@ -0,0 +1,93 @@
+package org.labkey.test.tests;
```
labkey-tchad: Could you put this with the other test commands?
```suggestion
package org.labkey.remoteapi.admin;
```

labkey-tchad:
How would this `CreateProjectCommand` relate to the existing `CreateContainerCommand` in labkey-client-api? Seems kind of redundant.

labkey-adam:
There's overlap, but, for better or worse, these commands invoke different server-side actions. `CoreController.CreateContainerAction` is a true API action that doesn't support creating folders from a template. `AdminController.CreateProjectAction` extends `CreateFolderAction`, the `FormViewAction` that powers the create folder wizard. It supports templates (because the base class does) and the special "Project Creator" role that allows a site administrator to create a project and assign themselves project admin. CreateProjectCommand is interesting only if it's going to be used by non-site admins that have been assigned the special role.

src/org/labkey/test/tests/CreateProjectCommand.java
labkey-adam: Sure

labkey-sweta (approved):

PR #1188, opened 2022-07-29
Test Components for Customizing and Saving Grid Views in Apps

labkey-danield:
#### Rationale
This is the test automation for customizing grids and saving views from app grids. It only impacts query grids, editable and and read only grids cannot be customized or save a view. There are several test component additions and changes.

This is a draft PR at this time. Test cases still need to be written, but I wanted to get a review of the components started before too many tests were written.

### Changes to QueryGrid and ResponsiveGrid.
For the grids there is now the Insert Column and Hide Column menu options. Clicking Insert Column will show a Customize Grid dialog. Clicking Hide Column immediately hides the column. The Edit Label functionality hasn't been implemented yet.

When a grid is updated a status message is shown in the panel header as well as the Undo and Save buttons. Clicking the Save button will show the Save View dialog.

Grid_Updates

If changes are made to a saved view the Save button will have a dropdown menu option that will allow the changed view to be saved with a different name:

Screen Shot 2022-08-11 at 10 51 20 AM

If a change is made to a saved view and the 'Save' button is clicked no Save View Dialog will be shown. The changes will be saved to the current view.

Changes to the default view will never show the "Save as..." option, the Save View Dialog will always be shown.

### Save View Dialog
Fairly straight forward. If the 'Make default view for all users' checkbox is checked this will become the default view for the grid, and the view name text box will be disabled.

Grid_SaveView_Dialog

### Manage Saved Views Dialog
Also fairly straight forward. This dialog will allow a user to rename or delete an exIsting view. The default view cannot be renamed or deleted. If you have changed the current (named) view the dialog will also not allow a user to edit or delete it:

Screen Shot 2022-08-11 at 11 00 06 AM

Clicking the pencil icon will allow the user to rename a view:

Screen Shot 2022-08-11 at 11 00 54 AM

Selecting a view for deletion gives a prompt and a Yes/No option:

Screen Shot 2022-08-11 at 11 01 11 AM

### Customize Grid Dialog
The Customize Grid dialog allows the addition of new fields/columns to the grid. This dialog appears when the Insert Column menu option is selected.

Fields are added from the Available Fields panel by clicking the "+" on the right hand side of the field name. Fields can be nested and are expanded\collapsed by clicking the "+" or "-" to the left of the field name.

The Shown in Grid panel shows field\column labels. The labels can be renamed or removed from this panel. Drag drop has not yet been implemented, and most likely will not be implemented.

Grid_Customize_Dialog

### View Menu on Grid
The view menu on the grid now has a "Customize Grid View", "Manage Saved Views" and a "Save Grid View" option. The "Customize" and "Save" options will show the appropriate dialog. The "Save Grid View" will always show the save view dialog regardless of the current view being a named view or not. It is equivalent to the "Save as..." button drop down menu.

Screen Shot 2022-08-11 at 11 07 20 AM

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3598
* https://github.com/LabKey/sampleManagement/pull/1161
* https://github.com/LabKey/biologics/pull/1509

#### Changes
* Updated CustomizeView.saveCustomView to allow for unchecking the shared checkbox (LKS component).
* Added CustomizeGridModalDialog, SaveViewDialog and ManageViewDialog
* Refactored GirdPanelTest. Added a base test with shared methods.
* Added methods for header bar elements in QueryGrid (alert text, Save & Undo buttons and displayed view name).
* Added hideColumn and insertColumn methods to ResponsiveGrid.
* Added overload initColumnsAndIndices method so columnNames and indexes can be recalculated if needed.
* Removed view options from existing filter grid code. Views names are no longer shown like filter options.
* Added a GridPanelViewTest.


src/org/labkey/test/components/CustomizeView.java
```diff
@@ -194,12 +194,20 @@ public void saveCustomView(String name, boolean shared, boolean inherit)
         {
             saveWindow.shareCheckbox.check();
         }
+        else
+        {
+            saveWindow.shareCheckbox.uncheck();
+        }
```
labkey-tchad: I'm worried that this will interfere with tests that depend on the current behavior (where `false` actually means to leave the default).
I'd suggest changing these flags to be nullable and update the other `saveCustomView` overloads to pass in `null` instead of false.
Then set the checkbox only if the flag isn't null.
Something like this:
```
public void saveCustomView(@Nullable String name) { saveCustomView(name, null, null); }

public void saveCustomView(String name, @Nullable Boolean shared) { saveCustomView(name, shared, null); }

public void saveCustomView(String name, @Nullable Boolean shared, @Nullable Boolean inherit)
{
    SaveWindow saveWindow = clickSave();

    if (shared != null)
    {
        saveWindow.shareCheckbox.set(shared);
    }

    if (inherit != null)
    {
        saveWindow.inheritCheckbox.set(inherit);
    }

etc.
```

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
```diff
@@ -0,0 +1,491 @@
...
+    public CustomizeGridDialog setShowAll(boolean checked)
+    {
+        if(checked)
+        {
+            elementCache().checkbox.check();
+        }
+        else
+        {
+            elementCache().checkbox.uncheck();
+        }
```
labkey-tchad: Simplify:
```suggestion
        elementCache().checkbox.set(checked);
```

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
```diff
@@ -0,0 +1,491 @@
...
+    }
+
+    /**
+     * Add a nested field to the 'Shown in Grid' list. You must provide the list of parent fields.
+     *
+     * @param parentFields The list of parent fields to first expand.
+     * @param fieldName The field to add.
+     * @return This dialog.
+     */
+    public CustomizeGridDialog addAvailableFieldToGrid(List<String> parentFields, String fieldName)
```
labkey-tchad: Combine both `addAvailableFieldToGrid` methods into one varargs method:
```suggestion
    public CustomizeGridDialog addAvailableFieldToGrid(String... fieldKeyParts)
```

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
```diff
@@ -0,0 +1,491 @@
...
+        return this;
+    }
+
+    /**
+     * Expand a hierarchy of fields.
+     *
+     * @param parentFields The list of fields to expand.
+     * @return This dialog.
+     */
+    public CustomizeGridDialog expandAvailableFieldsTo(List<String> parentFields)
```
labkey-tchad: Combine both `expandAvailableField` methods into one varargs method:
```suggestion
    public CustomizeGridDialog expandAvailableFieldsTo(String... fieldKeyParts)
```

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
```diff
@@ -0,0 +1,491 @@
...
+
+        String iconClass;
+        if(expand)
+        {
+            iconClass = "fa-plus-square";
+        }
+        else
+        {
+            iconClass = "fa-minus-square";
+        }
```
labkey-tchad: Simplify:
```suggestion
        String iconClass = expand ? "fa-plus-square" : "fa-minus-square";
```

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
```diff
@@ -0,0 +1,491 @@
...
+
+        protected WebElement availableFieldsPanel()
+        {
+            return contentPanelLocator.findElements(this).get(0);
+        }
+
+        protected WebElement shownInGridPanel()
+        {
+            return contentPanelLocator.findElements(this).get(1);
+        }
```
labkey-tchad: Are these sub-panels really that unstable that we can't cache them?
```suggestion
        protected final WebElement availableFieldsPanel = contentPanelLocator.index(0).findWhenNeeded(this);
        protected final WebElement shownInGridPanel = contentPanelLocator.index(1).findWhenNeeded(this);
```

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
```diff
@@ -0,0 +1,491 @@
...
+
+        // The data-fieldkey attribute is only present in items in the Available Fields panel.
+        // Similar value to field-name (no spaces, but casing is the same). For child fields it will contain the parent path.
+        protected WebElement getListItemElementByFieldKey(WebElement panel, String fieldKey)
+        {
+            return Locator.tagWithClass("div", "list-group-item")
+                    .withAttribute("data-fieldkey", fieldKey)
+                    .findElement(panel);
+        }
```
labkey-tchad: Why specify the panel if this is only relevant to the available fields panel?
```suggestion
        protected WebElement getListItemElementByFieldKey(WebElement panel, String fieldKey)
        {
            return Locator.tagWithClass("div", "list-group-item")
                    .withAttribute("data-fieldkey", fieldKey)
                    .findElement(availableFieldsPanel());
        }
```

src/org/labkey/test/components/ui/grids/QueryGrid.java
```diff
@@ -353,6 +402,13 @@ public List<FilterStatusValue> getFilterStatusFilterValues()
         }
 
         final WebElement removeAllFilters = Locator.tagWithClass("a", "remove-all-filters").refindWhenNeeded(this);
+
+        // The panel header element which will contain the Save and Undo buttons.
+        public WebElement panelHeader()
+        {
+            return getComponentElement().findElement(By.xpath("preceding-sibling::div[contains(@class,'panel-heading')]"));
```
labkey-tchad: Use our `Locator` class instead of the default `By`. We decorate `WebElements` with extra retry capabilities.
```suggestion
            return Locator.xpath("preceding-sibling::div[contains(@class,'panel-heading')]").findElement(this);
```

src/org/labkey/test/components/ui/grids/QueryGrid.java
```diff
@@ -391,4 +447,104 @@ protected Locator locator()
...
+            return checkbox.isChecked();
+        }
+
+        /**
+         * Save the view.
+         */
+        public void saveView()
+        {
+            dismiss("Save", 1);
+        }
```
labkey-tchad: If you made this class non-static, you could wait for some state change in the `QueryGrid` when clicking save.
```suggestion
        public void saveView()
        {
            dismiss("Save", 1);
            // Wait for save button on the grid to disappear or something like that
            // e.g.
            waitFor(() -> !Locator.buttonContainingText("Save").findWhenNeeded(QueryGrid.this.getComponentElement()).isDisplayed(), 5_000);
        }
```

src/org/labkey/test/tests/component/GridPanelBaseTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+    }
+
+    /**
+     * Make sure there are no filters or search values persisted for the given sample type.
+     *
+     * @param sampleType The sample type used to populate the grid.
+     * @return A queryGrid object.
+     */
+    protected QueryGrid initQueryGrid(String sampleType)
```
labkey-tchad: "init" usually implies configuring the server or test project. Maybe "goto" or "beginAt"?
```suggestion
    protected QueryGrid beginAtQueryGrid(String sampleType)
```

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
labkey-danield: Changed.

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
labkey-danield: Changed.

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
labkey-danield: Done.

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
labkey-danield: Done.

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
labkey-danield: No they should not be. I think this is a left over from a more generic approach I was thinking of taking but proved to be more complex that what was needed. Changed.

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
labkey-danield: Agreed. Changed.

src/org/labkey/test/components/ui/grids/QueryGrid.java
labkey-danield: Done.

src/org/labkey/test/components/ui/grids/QueryGrid.java
labkey-danield: I could but it feels like that is something that would be more appropriate to be done in ModalDialog.dismiss(String buttonText).

src/org/labkey/test/tests/component/GridPanelBaseTest.java
labkey-danield: Fair enough. beginAt it is.

src/org/labkey/test/components/CustomizeView.java
labkey-danield: Sounds reasonable. I updated the overloads and other tests that were currently passing in false for either of those fields.

labkey-sweta (approved):

labkey-chrisj (approved):
Thank you for taking on this work!

src/org/labkey/test/tests/component/GridPanelViewTest.java
```diff
@@ -0,0 +1,1389 @@
...
+    // Keep track of any custom views that may have been created.
+    private static List<String> savedViewsForDefaultSampleType = new ArrayList<>();
+
+    // Using the core-components.view adds 'GridPanel - ' to the panel header. Need to take that into account .
+    private static final String PANEL_VIEW_NAME_PREFIX = "GridPanel - %s";
+
+    // Tests that need to be written:
+    // Validate "Save As..." from the grid save button.
+    // Validate views that are locked, or in some other way, cannot be updates in the manage views dialog.
+    // ???
```
cnathe, line 85: do you have this list somewhere to fill out this ???

src/org/labkey/test/tests/component/GridPanelViewTest.java
```diff
@@ -0,0 +1,1389 @@
...
+     * <p>
+     *     Simple test to validate 'My Default' view.
+     * </p>
+     * <p>
+     *     This test will:
+     *     <ul>
+     *         <li>Modify the default view for a sample type in LabKey Server.</li>
+     *         <li>Save the default view, but do not make it public (default for all).</li>
+     *         <li>Validate in the App grid that 'My Default' is shown under the Views menu.</li>
+     *         <li>Validate that other users se no change to the default view.</li>
```
cnathe:
```suggestion
     *         <li>Validate that other users see no change to the default view.</li>
```

cnathe (commented):
The test cases here in GridPanelViewTest look great. I like that these are done via the core-components page. We talked about getting some set of these tests moved to the BVT level in some way. This would be a good one to include in that set.

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
```diff
@@ -0,0 +1,541 @@
...
+    }
+
+    @Override
+    protected void waitForReady()
+    {
+        WebDriverWrapper.waitFor(()->
+                {
+                    WebElement spinner = BootstrapLocators.loadingSpinner.findWhenNeeded(this);
+                    return !spinner.isDisplayed() &&
+                            elementCache().contentPanelLocator.findElements(this).size() == 2;
```
labkey-tchad: This `findWhenNeeded` -> `isDisplayed` pattern is pretty common so I recently added an `isDisplayed` method to `Locator`.
```suggestion
                    return !BootstrapLocators.loadingSpinner.isDisplayed(this) &&
                            elementCache().contentPanelLocator.findElements(this).size() == 2;
```

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
```diff
@@ -0,0 +1,541 @@
...
+                },
+                "Customize Grid dialog did not render in time.", 1_500);
+    }
+
+    /**
+     * Check or uncheck the 'Show all system and user-defined fields' checkbox checked?
+     *
+     * @param checked Set to true to check the box, false to uncheck it.
+     * @return True if checked false otherwise.
```
labkey-tchad: Copy/paste errors
```suggestion
     * Check or uncheck the 'Show all system and user-defined fields' checkbox.
     *
     * @param checked Set to true to check the box, false to uncheck it.
     * @return This dialog
```

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
```diff
@@ -0,0 +1,541 @@
...
+    /**
+     * Get the list of visible fields from the 'Available Fields' panel. Children of an expanded filed only have the name
+     * in the list, this des not include the path.
+     *
+     * @return The list of field names.
+     */
+    public List<String> getAvailableFields()
+    {
+        List<WebElement> listItemElements = elementCache().getListItemNameElements(elementCache().availableFieldsPanel);
+        return listItemElements.stream().map(WebElement::getText).collect(Collectors.toList());
```
labkey-tchad: IntelliJ always suggest simplifying usages of `Collectors.toList`
```suggestion
        return listItemElements.stream().map(WebElement::getText).toList();
```

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
```diff
@@ -0,0 +1,541 @@
...
+            // If this isn't the last item in the collection don't add a "/" to the expected data-fieldkey value.
+            if(iterator.hasNext())
+            {
+                fieldKey.append("/");
+            }
+
+        }
+
+        return this;
+    }
```
labkey-tchad: This iteration is duplicated in several methods (e.g. `addAvailableFieldToGrid` and `isAvailableFieldAddedToGrid`).
Consider refactoring this so that it can be shared. Also, not sure why it needs to be public.
```suggestion
    /**
     * Expand a field or a hierarchy of fields. If a single field is passed in only it will be expanded. If multiple values
     * are passed in it is assumed to be a path and all fields will be expanded to the last field.
     *
     * @param fields The list of fields to expand.
     * @return Row element the final specified field
     */
    public WebElement expandAvailableFields(String... fields)
    {
        if (fields.length == 0)
        {
            throw new IllegalArgumentException("Specify a field or fields to expand");
        }

        StringBuilder fieldKey = new StringBuilder();

        Iterator<String> iterator = Arrays.stream(fields).iterator();

        while(iterator.hasNext())
        {
            fieldKey.append(iterator.next().replace(" ", ""));

            expandOrCollapseByFieldKey(fieldKey.toString(), true);

            // If this isn't the last item in the collection don't add a "/" to the expected data-fieldkey value.
            if(iterator.hasNext())
            {
                fieldKey.append("/");
            }

        }

        return elementCache().getListItemElementByFieldKey(fieldKey.toString());
    }
```

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
```diff
@@ -0,0 +1,541 @@
...
+
+        if(expand)
+            errorMessage = String.format("There is no expand icon for field with data-fieldkey attribute '%s' in the 'Available Fields' panel.", fieldKey);
+        else
+            errorMessage = String.format("There is no collapse icon for field with data-fieldkey attribute '%s' in the 'Available Fields' panel.", fieldKey);
+
+        Assert.assertTrue(errorMessage, listItem.isDisplayed());
+
+        expandIcon.click();
+
```
labkey-tchad: This won't work correctly for fields that are already in the desired state. It should no-op for fields that are already in the expected state.
```suggestion
        if (isFieldKeyExpanded(listItem) != expand)
        {
            WebElement expandIcon = Locator.tagWithClass("div", "field-expand-icon").findElement(listItem);
            expandIcon.click();
            
            WebDriverWrapper.waitFor(() -> isFieldKeyExpanded(listItem) == expand,
                    String.format("Field, %s, did not %s.", fieldKey, expand ? "expand" : "collapse"), 5_000);
        }
```

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
```diff
@@ -0,0 +1,541 @@
...
+     */
+    private boolean isFieldKeyExpanded(String fieldKey)
+    {
+        WebElement listItem = elementCache().getListItemElementByFieldKey(fieldKey);
+
+        // As long as there is no plus/expand icon then this field is expanded.
+        return   !Locator.tagWithClass("div", "field-expand-icon")
+                .withChild(Locator.tagWithClass("i", "fa-plus-square"))
+                .findWhenNeeded(listItem).isDisplayed();
+    }
```
labkey-tchad: This should probably take a `WebElement` instead of the fieldKey. Most usages of it have already found the list item before calling this. Also, fail if the field isn't expandable.
```suggestion
    /**
     * Private helper to see if the given row is expanded or not. Will throw if the row is not expandable.
     *
     * @param listItem The available field list item to inspect.
     * @return True if row is expanded.
     */
    private boolean isFieldKeyExpanded(WebElement listItem)
    {
        WebElement expandIcon = Locator.byClass("div", "field-expand-icon").childTag("i").findElement(listItem);
        return expandIcon.getAttribute("class").contains("fa-minus-square");
    }
```

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
```diff
@@ -0,0 +1,541 @@
...
+
+        // Unfortunately using setFormElement doesn't work in this case. That method calls WebElement.clear which clears
+        // the current text but also causes the focus to the input control to be lost. When the focus is lost the input
+        // goes away. Need to do update the control using the selenium actions.
+
+        // Select the current text, type in the new value then change focus (tab) to commit the change.
+        Actions replaceCurrentText = new Actions(getDriver());
+        replaceCurrentText.sendKeys(Keys.END)
+                .keyDown(Keys.SHIFT)
+                .sendKeys(Keys.HOME)
```
labkey-tchad: Just an FYI. `END` and `HOME` can behave differently on different OSs. The default behavior on MacOS is to jump to the start/end of the whole document; Linux and Windows stay on the same line (I think IntelliJ does something to mimic this behavior).
Not a problem here, since it isn't a multi-line input.

Also, there is an existing helper for this sort of scenario, `getWrapper().actionClear` will clear an input without losing focus. Note that you still can't use `setFormElement`. (I have been tiptoeing around this oddness in my `ReactInput` refactor)

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
```diff
@@ -0,0 +1,541 @@
...
+    }
+
+    /**
+     * Is the 'Undo Edits' button enabled.
+     *
+     * @return True if enabled false otherwise.
+     */
+    public boolean isUndoEditsEnabled()
+    {
+        // It looks like .isEnabled is not accurate if disabled-action-text class attribute is used to disable a tag.
```
labkey-tchad: We should start pushing back on this.
HTML has a `disabled` attribute that Selenium recognizes; not sure why folks don't like using it.

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
```diff
@@ -0,0 +1,541 @@
...
+
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Wraps EntityFieldFilterModal.tsx in UI components.
+ */
+public class CustomizeGridDialog extends ModalDialog
```
labkey-tchad: Wrong product component (Copy/paste error I presume). Class name should match also.
```suggestion
/**
 * Wraps CustomizeGridViewModal.tsx in UI components.
 */
public class CustomizeGridViewDialog extends ModalDialog
```

src/org/labkey/test/components/ui/grids/ManageGridViewsDialog.java
```diff
@@ -0,0 +1,174 @@
...
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.interactions.Actions;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Dialog used to manage saved views.
+ */
+public class ManageGridViewsDialog extends ModalDialog
```
labkey-tchad: The name should match the product component.
```suggestion
/**
 * Dialog used to manage saved views. Wraps ManageViewsModal.tsx in UI components.
 */
public class ManageViewsDialog extends ModalDialog
```

src/org/labkey/test/components/ui/grids/ManageGridViewsDialog.java
```diff
@@ -0,0 +1,174 @@
...
+{
+    QueryGrid grid;
```
labkey-tchad: Final?
```suggestion
    private final QueryGrid grid;
```

src/org/labkey/test/components/ui/grids/ManageGridViewsDialog.java
```diff
@@ -0,0 +1,174 @@
...
+    /**
+     * Click 'No' button to not delete a button.
+     *
+     * @return This dialog.
+     */
+    public ManageGridViewsDialog clickDeleteNoButton()
+    {
+        elementCache().deleteNoButton.click();
+        return this;
+    }
```
labkey-tchad: These method names are confusing. What about confirm and cancel
```suggestion
    public ManageGridViewsDialog confirmDelete()
    {
        elementCache().deleteYesButton.click();
        return this;
    }

    /**
     * Click 'No' button to not delete a button.
     *
     * @return This dialog.
     */
    public ManageGridViewsDialog cancelDelete()
    {
        elementCache().deleteNoButton.click();
        return this;
    }
```

src/org/labkey/test/components/ui/grids/QueryGrid.java
```diff
@@ -299,13 +355,181 @@ public QueryGrid clearAllSelections()
...
+        WebElement saveButton = Locator.buttonContainingText("Save").findElement(elementCache().panelHeader());
+        saveButton.click();
+
+        if(expectSaveDialog)
+        {
+            return new SaveGridViewDialog(getDriver(), this);
+        }
+        else
+        {
+            return null;
```
labkey-tchad, line 609: Is there a way to wait for the save to complete? Maybe wait for the save button to disappear?
```suggestion
            getWrapper().shortWait().until(ExpectedConditions.stalenessOf(saveButton));
            return null;
```

src/org/labkey/test/components/ui/grids/QueryGrid.java
```diff
@@ -299,13 +355,181 @@ public QueryGrid clearAllSelections()
...
+    }
+
+    /**
+     * Is the 'Save' button visible on the grid.
+     *
+     * @return True if visible, false otherwise.
+     */
+    public boolean isSaveButtonVisible()
+    {
+        return Locator.buttonContainingText("Save").findWhenNeeded(elementCache().panelHeader()).isDisplayed();
```
labkey-tchad: `Locator.isDisplayed`
```suggestion
        return Locator.buttonContainingText("Save").isDisplayed(elementCache().panelHeader());
```

src/org/labkey/test/components/ui/grids/SaveGridViewDialog.java
```diff
@@ -0,0 +1,120 @@
...
+import org.labkey.test.components.bootstrap.ModalDialog;
+import org.labkey.test.components.html.Checkbox;
+import org.labkey.test.components.html.Input;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+/**
+ * Dialog used to save a grid view in an app.
+ */
+public class SaveGridViewDialog extends ModalDialog
```
labkey-tchad: Reference product components.
```suggestion
/**
 * Dialog used to save a grid view in an app. Wraps SaveViewModal.tsx in UI Components.
 */
public class SaveViewDialog extends ModalDialog
```

src/org/labkey/test/tests/component/GridPanelBaseTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+    protected void initProject()
+    {
+        _containerHelper.createProject(getProjectName(), null);
+
+        // Add the 'Sample Types' web part. It is easier when debugging etc...
+        PortalHelper portalHelper = new PortalHelper(this);
+        portalHelper.enterAdminMode();
+        portalHelper.addWebPart("Sample Types");
+        portalHelper.exitAdminMode();
```
labkey-tchad: `addWebPart` will enter and exit admin mode if needed.
```suggestion
        new PortalHelper(this).addWebPart("Sample Types");
```

src/org/labkey/test/tests/component/GridPanelBaseTest.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    /**
+     * Helper to remove the 'Flag' column from the default view. It just gets in the way for some tests, and is easier
+     * to remove it.
```
labkey-tchad, line 84: How does the `Flag` column interfere?

src/org/labkey/test/tests/component/GridPanelBaseTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+        SampleTypeHelper sampleHelper = new SampleTypeHelper(this);
+        DataRegionTable drtSamples = sampleHelper.getSamplesDataRegionTable();
+        drtSamples.goToView(VIEW_DEFAULT);
+
+        CustomizeView cv = drtSamples.openCustomizeGrid();
+        cv.removeColumn(REMOVED_FLAG_COLUMN);
+
+        // This will save the changes to the default view for everyone. The view menu option will remain as "Default"
+        cv.saveCustomView("", true);
```
labkey-tchad: Could probably safely make `CustomizeView.saveDefaultView` default to `shared=true`, or add a new `saveDefaultView` override that takes the shared flag.

src/org/labkey/test/tests/component/GridPanelBaseTest.java
```diff
@@ -0,0 +1,151 @@
...
+                .getGridPanel(TEST_SCHEMA, sampleType);
+
+        grid.clearAllSelections();
+
+        if(clearFilters)
+        {
+            grid.clearFilters();
+        }
+
+        grid.clearSearch();
```
labkey-tchad: Consider adding a `resetView` method to `QueryGrid` that does all of this. This method has no way to deal with a combination of custom filters and view filters.
```suggestion
        grid.resetView();
```
In `QueryGrid`:
```
    public void resetView()
    {
        clearAllSelections();

        if(isUndoButtonVisible())
        {
            clickUndoButton();
        }

       clearSearch();
    }
```

src/org/labkey/test/tests/component/GridPanelViewTest.java
```diff
@@ -0,0 +1,1389 @@
...
+                            grid.hasColumnFilterIcon(entry.getKey()));
+                }
+
+            }
+
+        }
+
+        checker().screenShotIfNewError(String.format("%s_Views_Menu_Column_Error", screenShotPrefix));
+
+        return checker().errorsSinceMark() == 0;
```
labkey-tchad: `screenshotIfNewError` resets the error mark. This will always return `true`.
I think it would make sense for `screenshotIfNewError` to return `true` if it took a screenshot.

src/org/labkey/test/tests/component/GridPanelViewTest.java
```diff
@@ -0,0 +1,1389 @@
...
+            checker().verifyFalse("'Save' button is visible on the grid, it should not be.",
+                    grid.isSaveButtonVisible());
+
+            checker().verifyFalse("'Undo' button is visible on the grid, it should not be.",
+                    grid.isUndoButtonVisible());
+        }
+
+        checker().screenShotIfNewError(String.format("%s_Grid_Header_Error", screenShotPrefix));
+
+        return checker().errorsSinceMark() == 0;
```
labkey-tchad: Always `true`

labkey-tchad (changes requested):

src/org/labkey/test/tests/component/GridPanelViewTest.java
labkey-danield: I've added a story to Samurai Puffin back log to track this (Grid Views -Selenium Test Part 2).
It includes these test and a couple of others.

src/org/labkey/test/components/ui/grids/QueryGrid.java
labkey-danield: Following up on this.
This code has changed. In the apps modifying a grid view and then saving it is different from LKS. In the app modifying the grid and saving the view are two different dialogs. Modifying the gird will result in an update to the grid, and that dialog does take a reference to the grid and waits for it to update. Saving a grid view does not result in a grid update and will not need to wait for it.

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
labkey-danield: Yes, but that makes it an immutable list. I would stumble over that in tests that would like to add/remove elements from the list to use for comparisons.

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
labkey-danield: Updated.

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
labkey-danield: Updated.

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
labkey-danield: Changed `isAvailableFieldAdded` and `addAvailableField` to call `expandAvailableFields`.
Changed `expandAvailableFields` to private. Initially was public to help some test automation that wasn't needed.

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
labkey-danield: Changed to take a WebElement. Using a WebElement made the NoOp check in expandOrCollapseByFieldKey easier.

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
labkey-danield: Changed to call  `getWrapper().actionClear(elementCache().columnLabelEdit);` before setting the text using selenium Actions.

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
labkey-danield: It would be nice to have some more consistency. Maybe there was some particular reason this wasn't done.

src/org/labkey/test/components/ui/grids/ManageGridViewsDialog.java
labkey-danield: Updated.

src/org/labkey/test/components/ui/grids/ManageGridViewsDialog.java
labkey-danield: Removed this variable it is not needed/used.

src/org/labkey/test/components/ui/grids/ManageGridViewsDialog.java
labkey-danield: Changed.

src/org/labkey/test/components/ui/grids/QueryGrid.java
labkey-danield: Changed.

src/org/labkey/test/components/ui/grids/QueryGrid.java
labkey-danield: Changed.

src/org/labkey/test/components/ui/grids/SaveGridViewDialog.java
labkey-danield: Changed.

src/org/labkey/test/tests/component/GridPanelBaseTest.java
labkey-danield: Changed.

src/org/labkey/test/tests/component/GridPanelBaseTest.java
labkey-danield: It is not shown in the Apps.

src/org/labkey/test/tests/component/GridPanelBaseTest.java
labkey-danield: Added and override `CustomizeView.saveDefaultView(boolean)`.

src/org/labkey/test/tests/component/GridPanelBaseTest.java
labkey-danield: I see your point, but can you think of a better method name? Now that views are available in this grid, and you can edit an existing view, a method named `resetView` could be misinterpreted.
Maybe something more verbose like `resetToDefaultAndClearSelections()` or maybe just `resetToDefaultState()`?

src/org/labkey/test/tests/component/GridPanelViewTest.java
labkey-danield: I've changed the DeferredErrorCollector.screenShotIfNewError to return a boolean. True if it took a screenshot, false otherwise.

src/org/labkey/test/tests/component/GridPanelViewTest.java
labkey-danield: Same comment as above.

src/org/labkey/test/components/ui/grids/CustomizeGridDialog.java
labkey-danield: I've added a NoOp check at the start that will return if the given row is already in the desired state.

src/org/labkey/test/tests/component/GridPanelBaseTest.java
labkey-tchad: Oh, certainly. I wasn't attached to the name; just thought it would be useful functionality since we frequently want to just reset a grid.

src/org/labkey/test/util/DeferredErrorCollector.java
```diff
@@ -156,17 +156,21 @@ public int errorsSinceMark()
...
+        boolean noError = true;
         if (errorsSinceMark() > 0)
         {
             final String s = artifactCollector.dumpPageSnapshot(screenshotName);
             allErrors.get(allErrors.size() - 1).setScreenshotName(s);
             artifactCollector.reportTestMetadata(s);
+            noError = false;
         }
         setErrorMark();
+        return noError;
```
labkey-tchad: The `setErrorMark` call is a no-op if there aren't any errors so the variable isn't really necessary here (not sure why I put the call outside the if).
I'd also flip the return value; if I saw `if (checker().screenShotIfNewError())`, I would expect it to return `true` if it took a screenshot.
```suggestion
    public boolean screenShotIfNewError(@NotNull String screenshotName)
    {
        if (errorsSinceMark() > 0)
        {
            final String s = artifactCollector.dumpPageSnapshot(screenshotName);
            allErrors.get(allErrors.size() - 1).setScreenshotName(s);
            artifactCollector.reportTestMetadata(s);
            setErrorMark();
            return true;
        }
        return false;
```

labkey-tchad (approved):
Just a couple of small changes

src/org/labkey/test/util/DeferredErrorCollector.java
labkey-danield: Changed to return true if a screenshot is taken. Updated comment as well.

src/org/labkey/test/tests/component/GridPanelBaseTest.java
labkey-danield: I've `added QueryGrid.resetToDefaultState(boolean revertDefaultView)`.
It will do the best it can to get the grid back to its original state. It will apply the default view. If the parameter is set to true it will revert the default view before applying it, otherwise it will apply the default view with any modifications it may have.

PR #1194, opened 2022-08-03
Make 'beginAt' handle app navigation

labkey-tchad:
#### Rationale
Navigating directly to an app page (e.g. `beginAt("home/biologics-app.view?#/")`) may or may not trigger a page load. We have numerous workarounds:
- `BiologicsHelper.beginAt` compares the target URL with the current URL to determine whether to wait.
- `SMTestUtils.beginAt` doesn't wait for a page load if navigating between app pages. The current implementation is too lax; it would miss some page loads but isn't widely used.
- Several `*Page.beginAt` methods just don't wait for a page load.

This change pulls the logic from `BiologicsHelper.beginAt` into the main `WebDriverWrapper.beginAt`.

#### Related Pull Requests
* https://internal.labkey.com/Scrumtime/Backlog/list-grid.view?listId=958&query.BranchName~eq=fb_appBeginAt

#### Changes
* Make `WebDriverWrapper.beginAt` detect when a page load should occur.


src/org/labkey/test/WebDriverWrapper.java
```diff
@@ -1130,6 +1129,30 @@ public long beginAt(String relativeURL, int millis)
...
+            String currentUrl = getDriver().getCurrentUrl();
+            Crawler.ControllerActionId current = new Crawler.ControllerActionId(currentUrl);
+            Crawler.ControllerActionId destination = new Crawler.ControllerActionId(destinationUrl);
+
+            return !destinationUrl.contains("#") ||
+                    destinationUrl.contains("?#") != currentUrl.contains("?#") ||
+                    !current.getAction().equals(appAction) ||
+                    !destination.getAction().equals(appAction) ||
+                    !current.getController().equals(destination.getController()) ||
+                    !current.getFolder().equals(destination.getFolder());
```
labkey-danield: I don't see any obvious flaws with the comparisons, I trust you to have done the homework on the validity of each.

Just a couple of questions/comments:
If you are not paying attention it is easy to confuse `current` with `currentUrl`. Maybe change the variable names of `current` and `destination` to `currentActionId` & `destinationActionId`? Or some other value to make it obvious at a glance?

For these two checks:
```
!current.getAction().equals(appAction) ||
!destination.getAction().equals(appAction) ||
```
Would it be the same as:
`!current.getAction().equals(destination.getAction())`
Or is it only true if the action is app?

labkey-danield (approved):

labkey-danield (approved):

PR #1198, opened 2022-08-08
Clear page load listeners before test postamble

labkey-tchad:
#### Rationale
There is a page load listener that collects all visited URLs. This list of URLs is used to start off the crawler.
This combination can put crawled injection URLs in the initial set of URLs for subsequent tests; this can bypass the crawler's special handling of injection URLs.

#### Related Pull Requests
* N/A

#### Changes
* Clear page load listeners before test postamble

### DIFF
--- a/src/org/labkey/test/BaseWebDriverTest.java
+++ b/src/org/labkey/test/BaseWebDriverTest.java
@@ -1063,6 +1063,8 @@ public abstract class BaseWebDriverTest extends LabKeySiteWrapper implements Cle
     {
         disablePageUnloadEvents();
 
+        clearPageLoadListeners();
+
         ensureSignedInAsPrimaryTestUser();
 
         checkQueries();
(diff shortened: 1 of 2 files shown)


labkey-jeckels (approved):
This looks like a good change, as we don't want to accumulate this listeners over time.

When reviewing the code to understand that returnUrl parameter failure, I had missed that BaseWebDriverTest.urlsSeen was static. Is that helpful? It seems like later tests would be less likely to hit their own URLs if they're competing with URLs previously seen during all other tests.

labkey-tchad:
> When reviewing the code to understand that returnUrl parameter failure, I had missed that BaseWebDriverTest.urlsSeen was static. Is that helpful? It seems like later tests would be less likely to hit their own URLs if they're competing with URLs previously seen during all other tests.

Unfortunately, JUnit forces us to use static for things that we would rather not. Each `@Test` method is run with a separate instance of the test class and `@BeforeClass` and `@AfterClass` (where the crawler is run) are run statically.

labkey-jeckels:
@labkey-tchad I see. What about clearing our urlsSeen in the setUp() method?

labkey-danield (approved):

PR #1201, opened 2022-08-11
Refactor deprecated ReactInput

labkey-tchad:
#### Rationale
The deprecated biologics `ReactInput` has lingered for too long and is resulting in some unnecessary cross-module test dependencies. This change renames the component to `ValidatingInput` and moves it to `testAutomation`. It also tries to extend the functionality of `Input` more clearly and cleanly. I was unable to merge them entirely because they have different expectations of whether the input will lose focus after setting.
We should standardize the functionality of these components and `WebDriverWrapper.setFormElement` but my experimentation made it clear that doing so would be a sizable project.

The deprecated biologics `Toggle` component was interchangeable with `ToggleButton` without much modification. Just had to pull in a different Locator from `ToggleFinder`

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/1201
* https://github.com/LabKey/labbook/pull/252
* https://github.com/LabKey/biologics/pull/1506
* https://github.com/LabKey/accounts/pull/154

#### Changes
* Move deprecated `ReactInput` from biologics and rename to `ValidatingInput`
* Clean up `WebDriverWrapper.setFormElement` a bit
* Integrate deprecated `ToggleFinder` into `ToggleButtonFinder`

### DIFF
--- a/src/org/labkey/test/WebDriverWrapper.java
+++ b/src/org/labkey/test/WebDriverWrapper.java
@@ -3292,6 +3292,9 @@ public abstract class WebDriverWrapper implements WrapsDriver
         try
         {
             fireEvent(input, SeleniumEvent.change);
+            String elementClass = input.getAttribute("class");
+            if (elementClass.contains("gwt-TextBox") || elementClass.contains("gwt-TextArea") || elementClass.contains("x-form-text"))
+                fireEvent(input, SeleniumEvent.blur); // Make GWT and ExtJS form elements behave better
         }
         catch(StaleElementReferenceException stale)
         {
@@ -3510,9 +3513,6 @@ public abstract class WebDriverWrapper implements WrapsDriver
 
     public void setDropZone(WebElement dropZone, List<File> files)
     {
-        // Remove class so that WebDriver can interact with concealed form element
-        executeScript("arguments[0].setAttribute('class', '');arguments[0].setAttribute('style', '');", dropZone);
-        shortWait().until(ExpectedConditions.elementToBeClickable(dropZone)); // Takes a moment for DOM to update
         setInput(dropZone, files);
     }
 
--- a/src/org/labkey/test/WebDriverWrapper.java
+++ b/src/org/labkey/test/WebDriverWrapper.java
@@ -3249,10 +3249,12 @@ public abstract class WebDriverWrapper implements WrapsDriver
         {
             TestLogger.warn("Please use a File object to set file input");
             setFormElement(el, new File(text));
-            return;
         }
-
-        if (isHtml5InputTypeSupported(inputType))
+        else if (el.getTagName().equals("select"))
+        {
+            selectOptionImproperly(el, text);
+        }
+        else if (isHtml5InputTypeSupported(inputType))
         {
             setHtml5Input(el, inputType, text);
         }
@@ -3268,7 +3270,7 @@ public abstract class WebDriverWrapper implements WrapsDriver
         {
             actionClear(input);
         }
-        else if (!input.getTagName().equals("select") && text.length() < 1000 && !text.contains("\n") && !text.contains("\t"))
+        else if (text.length() < 1000 && !text.contains("\n") && !text.contains("\t"))
         {
             actionClear(input); // Some inputs swallow standard 'WebElement.clear' in certain cases
             if (!waitFor(()-> getFormElement(input).length() == 0, 500))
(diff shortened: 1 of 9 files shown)


src/org/labkey/test/WebDriverWrapper.java
```diff
@@ -3482,6 +3489,20 @@ public void setFormElementJS(WebElement el, String text)
         fireEvent(el, SeleniumEvent.change);
     }
 
+    private void selectOptionImproperly(WebElement el, String text)
+    {
+        try
+        {
+            selectOptionByText(el, text);
+            log("WARNING: Use selectOptionByText(..) instead of setFormElement(..) for select elements");
```
labkey-danield: Just curious, why use `log("WARNING:` and not `TestLogger.warn(`?

src/org/labkey/test/WebDriverWrapper.java
```diff
@@ -3482,6 +3489,20 @@ public void setFormElementJS(WebElement el, String text)
...
+        }
+        catch (NoSuchElementException x)
+        {
+            selectOptionByValue(el, text);
+            log("WARNING: Use selectOptionByValue(..) instead of setFormElement(..) for select elements");
```
labkey-danield: Same question.

src/org/labkey/test/WebDriverWrapper.java
```diff
@@ -3502,9 +3523,6 @@ public void setInput(WebElement el, List<File> files)
 
     public void setDropZone(WebElement dropZone, List<File> files)
     {
-        // Remove class so that WebDriver can interact with concealed form element
-        executeScript("arguments[0].setAttribute('class', '');arguments[0].setAttribute('style', '');", dropZone);
-        shortWait().until(ExpectedConditions.elementToBeClickable(dropZone)); // Takes a moment for DOM to update
         setInput(dropZone, files);
     }
```
labkey-danield: I believe this method is only used in one place (`SignalDataUploadPage.uploadFile`).  Given this change would it make more sense to remove this method completely and replace the call in `SignalDataUploadPage.uploadFile` with:
```
    public void uploadFile(File... file)
    {
        WebElement dropFileInputEl = Locators.dropFileInput.findElement(_test.getDriver());
        _test.setInput(dropFileInputEl, Arrays.asList(file));
    }

```

src/org/labkey/test/components/html/ValidatingInput.java
```diff
@@ -0,0 +1,45 @@
...
+        super(el, driver);
+    }
+
+    @Override
+    public void set(String value)
+    {
+        set(value, true);
+    }
+
+    public void set(String value, boolean validateValue)
```
labkey-danield, line 26: I kind of see the use here, but I'm wondering why a test/page/etc... would use a ValidatingInput and then not validate? Why not just use an Input?

src/org/labkey/test/components/react/BaseReactSelect.java
```diff
@@ -147,19 +148,19 @@ public boolean hasOption(String value)
...
         open();
-        getWrapper().setFormElement(elementCache().input, value);
-        WebElement foundElement;
+        elementCache().input.clear();
+        elementCache().input.sendKeys(value);
         try
         {
             var optionElement = ReactSelect.Locators.options.containing(value);
-            foundElement = optionElement.waitForElement(elementCache().selectMenu, 4000);
+            optionElement.waitForElement(elementCache().selectMenu, 4000);
```
labkey-danield, line 156: This wait time of 4 seconds has always felt a bit arbitrary to me. Why not 5 seconds or 2.5? I'm wondering if we should be more consistent in our wait times, maybe use global constants.

src/org/labkey/test/components/react/BaseReactSelect.java
```diff
@@ -255,18 +256,29 @@ public T removeSelection(String value)
...
-        getWrapper().shortWait().until(ExpectedConditions.stalenessOf(removeBtn));
+        attemptRemove(value);
 
         // Validate that the selected item really was removed.
         WebDriverWrapper.sleep(500);
-        waitFor(()->!getSelections().contains(value), String.format("Failed to remove selection '%s'.", value), WAIT_FOR_JAVASCRIPT);
+        if (getSelections().contains(value))
+        {
+            String msg = String.format("Failed to remove selection '%s'.", value);
+            TestLogger.warn(msg + ": retrying");
```
labkey-danield, line 266:
```suggestion
TestLogger.warn(String.format("Failed to remove selection '%s' : retrying", value));
```

src/org/labkey/test/components/react/BaseReactSelect.java
```diff
@@ -255,18 +256,29 @@ public T removeSelection(String value)
...
         scrollIntoView();
 
-        WebElement removeBtn = Locators.removeMultiSelectValueButton(value).findWhenNeeded(getComponentElement());
-        removeBtn.click();
-
-        getWrapper().shortWait().until(ExpectedConditions.stalenessOf(removeBtn));
+        attemptRemove(value);
 
         // Validate that the selected item really was removed.
         WebDriverWrapper.sleep(500);
```
labkey-danield, line 262: I had put this sleep in here out of desperation. Do you think it is needed?

labkey-danield (approved):

src/org/labkey/test/WebDriverWrapper.java
labkey-tchad: Just a copy-paste error

src/org/labkey/test/WebDriverWrapper.java
labkey-tchad: Sure, why not:
https://github.com/LabKey/signalData/pull/19

src/org/labkey/test/components/react/BaseReactSelect.java
labkey-tchad: The `msg` is used sans "retrying" for the error message if the retry fails.

src/org/labkey/test/components/react/BaseReactSelect.java
labkey-tchad: It might take a moment for it to reappear (The item definitely goes stale, so it must reappear at some point). I'm not too worried about an extra half second since this has such limited use.

src/org/labkey/test/components/html/ValidatingInput.java
labkey-tchad: There are a couple of places that want to validate conditionally (some inputs get cleared in some error cases).
Mostly just pulling in the existing functionality from `ReactInput`. The threads really started to unravel when I tried to simplify things too much.

PR #1203, opened 2022-08-12
Avoid bad test handling of switching between header menus

labkey-tchad:
#### Rationale
The assertion of `projectLinkExists` isn't especially useful and causes intermittent failures.

#### Related Pull Requests
* #1055 

#### Changes
* Simplify `testCreateProjectByTemplate`

### DIFF
--- a/src/org/labkey/test/tests/ProjectCreatorUserTest.java
+++ b/src/org/labkey/test/tests/ProjectCreatorUserTest.java
@@ -18,6 +18,7 @@ import org.labkey.test.TestFileUtils;
 import org.labkey.test.TestTimeoutException;
 import org.labkey.test.categories.Daily;
 import org.labkey.test.components.list.ManageListsGrid;
+import org.labkey.test.pages.admin.PermissionsPage;
 import org.labkey.test.util.APIContainerHelper;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/ProjectCreatorUserTest.java
```diff
@@ -164,9 +165,9 @@ public void testCreateProjectByTemplate() throws IOException, CommandException
         stopImpersonating();
 
         goToProjectHome(PROJECT_NAME_PC);
-        assertTrue(projectMenu().projectLinkExists(PROJECT_NAME_PC));
```
labkey-sweta, line 167: Asserting project is present got deleted accidentally?

labkey-tchad: This was intentional. `goToProjectHome` does essentially the same validation. Opening the project menu to check for the link is what was causing the intermittent failures.

labkey-sweta: ok

labkey-sweta (approved):

PR #1205, opened 2022-08-15
Fix ContainerContextTest

labkey-adam:
#### Rationale
Fix `ContainerContextTest` to pass when server is in no-question-mark mode. Add a method that tests can use to query experimental features.


labkey-tchad (approved):

labkey-tchad (approved):
Looks good. I piggy-backed some more updates to `ContainerContextTest`

PR #1212, opened 2022-08-19
Update interaction when delete data class objects

labkey-susanh:
#### Rationale
Deletion of data class objects now uses the confirmDelete.js script, which displays an Ext4.Msg panel instead of loading a new page. For these tests for error processing, the button text has been updated and we don't expect a page load when there are errors that prevent deletion from happening.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3623

#### Changes
* Add more parameters to internal method to account for different delete interactions

### DIFF
--- a/src/org/labkey/test/tests/TriggerScriptTest.java
+++ b/src/org/labkey/test/tests/TriggerScriptTest.java
@@ -691,7 +691,7 @@ public class TriggerScriptTest extends BaseWebDriverTest
         DataRegionTable drt = new DataRegionTable(tableName, this);
         int rowId = drt.getRowIndex(columnName, columnValue);
         drt.checkCheckbox(rowId);
-        doAndMaybeWaitForPageToLoad(1000, () ->
+        doAndMaybeWaitForPageToLoad(defaultWaitForPage, () ->
         {
             drt.clickHeaderButton("Delete");
             Alert alert = getAlertIfPresent();
--- a/src/org/labkey/test/tests/TriggerScriptTest.java
+++ b/src/org/labkey/test/tests/TriggerScriptTest.java
@@ -697,10 +697,8 @@ public class TriggerScriptTest extends BaseWebDriverTest
             Alert alert = getAlertIfPresent();
             if (alert != null)
                 alert.accept();
-            else if (expectPageLoad)
-                clickButton(deleteButtonText);
             else
-                click(Locator.linkWithText(deleteButtonText));
+                clickButton(deleteButtonText, expectPageLoad ? defaultWaitForPage : 0);
             return expectPageLoad;
         });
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/TriggerScriptTest.java
```diff
@@ -679,23 +679,29 @@ private void insertSingleRowViaUI(String dataRegionName, Map<String,String> reco
...
+     * @param expectPageLoad indicates whether confirming deletion will result in a page load or not
      */
-    private void deleteSingleRowViaUI(String columnName, String columnValue, String tableName)
+    private void deleteSingleRowViaUI(String columnName, String columnValue, String tableName, String deleteButtonText, boolean expectPageLoad)
     {
         DataRegionTable drt = new DataRegionTable(tableName, this);
         int rowId = drt.getRowIndex(columnName, columnValue);
         drt.checkCheckbox(rowId);
-        doAndWaitForPageToLoad(() ->
+        doAndMaybeWaitForPageToLoad(1000, () ->
```
labkey-tchad: One second is a little short for a page load. `defaultWaitForPage` is the standard duration.

src/org/labkey/test/tests/TriggerScriptTest.java
```diff
@@ -679,23 +679,29 @@ private void insertSingleRowViaUI(String dataRegionName, Map<String,String> reco
...
         {
             drt.clickHeaderButton("Delete");
             Alert alert = getAlertIfPresent();
             if (alert != null)
                 alert.accept();
+            else if (expectPageLoad)
+                clickButton(deleteButtonText);
             else
-                clickButton("Confirm Delete");
+                click(Locator.linkWithText(deleteButtonText));
```
labkey-tchad: I don't think this needs to check the `expectPageLoad` flag. The surrounding `doAndMaybeWaitForPageToLoad` will handle the page load if necessary. You can pass a zero timeout to `clickButton` to suppress its page load expectation.
```
if (alert != null)
    alert.accept();
else
    clickButton(deleteButtonText, 0);
```

labkey-tchad (changes requested):

labkey-tchad (approved):
Pre-approving so that you don't have to wait for a re-review.

src/org/labkey/test/tests/TriggerScriptTest.java
labkey-susanh: Done.

src/org/labkey/test/tests/TriggerScriptTest.java
labkey-susanh: Done.

PR #1221, opened 2022-08-30
Update EntityInsertPanel tab select

labkey-martyp:
#### Rationale
EntityInsertPanel may remount in some scenarios when selecting tabs, which can leave the test with a stale element. After selecting file upload tab need to get the component again and wait for tab render.

#### Related Pull Requests
* https://github.com/LabKey/biologics/pull/1532

#### Changes
* Update EntityInsertPanel.showFileUpload to avoid stale element


labkey-chrisj (approved):
This makes sense, it's a bummer that it was necessary

PR #1229, opened 2022-09-02
Update Query Grid to Allow for a Locked Left Column

labkey-danield:
#### Rationale
Query grids now have the left column locked. The test component needs to take this into account. The method to scroll the grid is public only the ResponsiveGrid.clickColumnMenuItem will do it automatically. 

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/1198

#### Changes
* Added public method that will scroll the grid to the top row and the left  most column.
* Updated ResponsiveGrid.clickColumnMenuItem to check for a locked column and scroll the grid accordingly.

### DIFF
--- a/src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
+++ b/src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
@@ -87,6 +87,27 @@ public class ResponsiveGrid<T extends ResponsiveGrid> extends WebDriverComponent
     }
 
     /**
+     * Is the left column on this grid locked. I think this will always be true for this grid type.
+     *
+     * @return True if left column is locked, false otherwise.
+     */
+    public boolean hasLockedColumn()
+    {
+        return Locator.tagWithClass("div", "grid-panel__grid")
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
```diff
@@ -86,6 +86,27 @@ public Boolean hasData()
         return !Locators.emptyGrid.existsIn(this);
     }
 
+    /**
+     * Is the left column on this grid locked. I think this will always be true for this grid type.
+     *
+     * @return True if left column is locked, false otherwise.
+     */
+    public boolean hasLockedColumn()
```
labkey-tchad, line 94: Does this need to be public?

labkey-tchad (approved):

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
labkey-danield: It is used in a test to validate that the expected grid has a locked column.

PR #1160, opened 2022-07-11
Check for external links with bad 'rel' attribute

labkey-tchad:
#### Rationale
[40708: Create automated tests to look for anchor tags with link to an outside server](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=40708)

~~Currently blocked by [Issue 45852](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=45852).~~

#### Changes
* Have crawler check for `noopener noreferrer` on external links

PR #1234, opened 2022-09-09
Stop crawling deeper if redirected to an external site

labkey-tchad:
#### Rationale
Some tests create "link" reports pointing at labkey.org; this causes the crawler to get redirected to a page that it would normally not crawl. The crawler shouldn't further inspect such pages.
```
java.lang.AssertionError: Bad 'rel' attribute for link to https://www.labkey.org/wiki/Documentation/page.view?name=default. On Page: https://www.labkey.org/home/project-begin.view
Expected: (a collection containing "noopener" and a collection containing "noreferrer")
     but: a collection containing "noopener" was ""
  at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
  at org.labkey.test.util.Crawler.crawlLink(Crawler.java:1037)
  at org.labkey.test.util.Crawler.crawl(Crawler.java:824)
  at org.labkey.test.util.Crawler.crawlAllLinks(Crawler.java:787)
```

I also noticed that the crawler is skipping the [`rel` attribute validation](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=40708) for some actions.

#### Related Pull Requests
* #1160 

#### Changes
* Stop crawling deeper if redirected to an external site.
* Defer filtering of `crawlLink` results until all checks have been performed

### DIFF
--- a/src/org/labkey/test/util/Crawler.java
+++ b/src/org/labkey/test/util/Crawler.java
@@ -460,6 +460,10 @@ public class Crawler
 
         public UrlToCheck(URL origin, String urlText, int depth)
         {
+            if (depth < 0)
+            {
+                throw new IllegalArgumentException("Invalid crawl depth: " + depth);
+            }
             _origin = origin;
             _urlText = urlText;
             _depth = depth;
@@ -528,9 +532,10 @@ public class Crawler
             return _isFromForm;
         }
 
-        public void setFromForm(boolean fromForm)
+        public UrlToCheck setFromForm(boolean fromForm)
         {
             _isFromForm = fromForm;
+            return this;
         }
 
         public URL getOrigin()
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/util/Crawler.java
```diff
@@ -1039,27 +1044,31 @@ private List<UrlToCheck> crawlLink(final UrlToCheck urlToCheck)
...
+
+                        for (String url : _test.getFormAddresses())
+                        {
+                            try
+                            {
+                                newUrlsToCheck.add(new UrlToCheck(actualUrl, url, nextDepth).setFromForm(true));
+                            }
+                            catch (IllegalArgumentException ignore)
+                            {
+                                // forms might have strange target action (e.g. '../formulations')
```
labkey-danield, line 1071: I'm not sure I understand the code entirely but it looks like the constructor to UrlToCheck will throw an IllegalArgumentException if the depth is negative. Are you ok with eating that exception here?

labkey-danield (approved):
Looks ok to me, just one question with a try/catch.

src/org/labkey/test/util/Crawler.java
labkey-tchad: That shouldn't be a problem. The crawler used to have a way to indicate a `-1` depth in order to crawl a single link (I forget the specifics). I replaced that with the `validatePage` method at some point. There is not, currently, any way for the depth to be `-1`.

PR #1241, opened 2022-09-19
Fix Java transform build.gradle dependencies

labkey-adam:
#### Rationale
Need an explicit dependency on commons logging, as of labkey-client-api v3.0.0


labkey-susanh (approved):

labkey-jeckels (approved):
This looks good, but are there any other dependencies we should add?

labkey-adam:
> This looks good, but are there any other dependencies we should add?

I'll run the test and we'll find out

labkey-adam:
ProgrammaticQCTest passes on this FB: https://teamcity.labkey.org/viewLog.html?buildId=2008684&buildTypeId=bt20&tab=testsInfo&branch_LabkeyTrunk_DailySuites=1713

PR #1246, opened 2022-09-24
Test against 22.10-SNAPSHOT API

labkey-adam:
#### Rationale
API snapshot with org.json.old.* changes should be available from Artifactory

### DIFF
--- a/src/org/labkey/remoteapi/issues/IssueModel.java
+++ b/src/org/labkey/remoteapi/issues/IssueModel.java
@@ -12,14 +12,14 @@ import java.util.stream.Collectors;
 public class IssueModel
 {
     private final Map<String, Object> _properties = new CaseInsensitiveHashMap<>();
-    private final List<File> _attachments = new ArrayList();
+    private final List<File> _attachments = new ArrayList<>();
 
     // https://www.labkey.org/Documentation/wiki-page.view?name=sampleJSscripts#issues
     public IssueModel()
     {
     }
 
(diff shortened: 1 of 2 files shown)


src/org/labkey/remoteapi/issues/IssueModel.java
```diff
@@ -12,14 +12,14 @@
...
-    private final List<File> _attachments = new ArrayList();
+    private final List<File> _attachments = new ArrayList<>();
 
     // https://www.labkey.org/Documentation/wiki-page.view?name=sampleJSscripts#issues
     public IssueModel()
     {
     }
 
-    public IssueModel setProperties(Map properties)
+    public IssueModel setProperties(Map<String, String> properties)
```
labkey-tchad, line 22: `_properties` is a String-Object map (not sure if the action actually accepts any non-String properties)
```suggestion
    public IssueModel setProperties(Map<String, Object> properties)
```

labkey-tchad (approved):

src/org/labkey/remoteapi/issues/IssueModel.java
labkey-adam: The suggested change isn't legal, since the caller (and its caller) pass in `Map<String, String>`. So either we leave off the generic types (as before) or keep this `Map<String, String>`. Note that it's legal to call `putAll()` with a `Map` parameter that holds more specific types than the target.

PR #1248, opened 2022-09-27
Migrate labkey-client-api to use a modern JSON library: JSON-java

labkey-adam:
#### Rationale
labkey-client-api was using an old, unsupported JSON library


labkey-tchad (approved):
Made some fixes to `JSONHelper`

PR #1250, opened 2022-09-27
Update bulk edit and editable grid components

labkey-tchad:
#### Rationale
Some new helpers to help test storage grids.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/1220

#### Changes
* Add `WebDriverUtils.MODIFIER_KEY` to get Cmd/Ctrl key depending on OS
* Add `TestDataUtils.merge` to combine lists
* Update `ModalDialog` locators to find what they claim to (e.g. dialog title)
* Make `EntityBulkUpdateDialog` able to wait for a grid to update when saving
* Add `EditableGridForm` to link a save button to `EditableGrid`
* Make `TestUser` impersonation/creation methods less strict to allow sharing between tests
* Clean up usages of deprecated JUnit `Assert.assertThat`


src/org/labkey/test/components/bootstrap/ModalDialog.java
```diff
@@ -177,13 +177,19 @@ public ModalDialogFinder(WebDriver driver)
...
         public ModalDialogFinder withTitle(String title)
         {
-            _locator = Locators.dialog.withDescendant(Locator.tagWithClass("h4","modal-title")).containing(title);
+            _locator = Locators.dialog.withDescendant(Locator.tagWithClass("h4","modal-title").containing(title));
+            return this;
+        }
+
+        public ModalDialogFinder withTitleIgnoreCase(String title)
+        {
+            _locator = Locators.dialog.withDescendant(Locator.tagWithClass("h4","modal-title").containingIgnoreCase(title));
```
labkey-danield, line 186: Would it have been better to have the `withTitle` finder ignore case?

src/org/labkey/test/components/ui/entities/EntityBulkUpdateDialog.java
```diff
@@ -69,11 +75,7 @@ public List<String> getSelectionFieldValues(String fieldKey)
 
     public EntityBulkUpdateDialog setTextArea(String fieldKey, String text)
     {
-        setEditableState(fieldKey, true);
-        Input input = elementCache().textArea(fieldKey);
-        WebDriverWrapper.waitFor(()-> input.getComponentElement().getAttribute("disabled")==null,
-                "the input did not become enabled in time", WAIT_TIMEOUT);
-        input.set(text);
+        enableAndWait(fieldKey, elementCache().textArea(fieldKey)).set(text);
```
labkey-danield, line 78: Can this be used in the `setSelectionField` method (just to be consistent)?

```
    public EntityBulkUpdateDialog setSelectionField(String columnTitle, List<String> selectValues)
    {
        FilteringReactSelect reactSelect = enableAndWait(columnTitle, elementCache().getSelect(columnTitle));
        selectValues.forEach(reactSelect::filterSelect);
        return this;
    }

```

Also might be nice to change the parameter name to fieldKey for that method, again for consistency, especially since the parameters are used in the same way.

src/org/labkey/test/tests/WorkbookTest.java
```diff
@@ -117,28 +118,28 @@ public void testSteps()
         menu.expand();
 
         Assert.assertEquals("The current containerPath should have two folder levels", 3, getCurrentContainerPath().split("/").length);
```
labkey-danield, line 120: Why not change this one as well?

src/org/labkey/test/util/ExcelHelper.java
```diff
@@ -213,7 +213,12 @@ public static List<String> getRowData(Sheet sheet, int rowIdx)
         while (row.hasNext())
```
labkey-danield: I know this isn't part of your change, but I found the variable name `row` a little confusing, especially when looking at it in this context and seeing `row.hasNext`. Wouldn't `cells` be a more appropriate name?

labkey-danield (approved):

src/org/labkey/test/components/ui/entities/EntityBulkUpdateDialog.java
labkey-tchad: I don't think `reactSelect.getComponentElement().isEnabled()` is accurate, so the same technique wouldn't work.

src/org/labkey/test/tests/WorkbookTest.java
labkey-tchad: I was just refactoring usages of the deprecated `org.junit.Assert.assertThat`, not generally cleaning up imports.

PR #1251, opened 2022-09-29
refactor changes for group management tests and components in apps

labkey-chrisj:
#### Rationale
Adding in-app administration for groups in addition to roles calls for components to support their very-similar but slightly different implementation- this is an attempt to update our components to be able to interact with roles and groups separately but distinctly

#### Related Pull Requests
test changes + new tests in biologics - https://github.com/LabKey/biologics/pull/1634
Update in SM - https://github.com/LabKey/sampleManagement/pull/1275

#### Changes
modified RoleRow to work with the UI changes that went into adding the notion of groups- now users and groups can be represented as being in a role 
image

created GroupRow to support groups, which can have users and groups as members
image

extracted PermissionsMember (was internal to RoleRow) so RoleRow and GroupRow can both use it
Added GroupDetailsPanel, similar to UserDetailsPanel
Expanded the info UserDetailsPanel can get from the dom- in particular, groups the user is a member of

### DIFF
--- a/src/org/labkey/test/components/ui/permissions/GroupRow.java
+++ b/src/org/labkey/test/components/ui/permissions/GroupRow.java
@@ -1,7 +1,6 @@
 package org.labkey.test.components.ui.permissions;
 
 import org.labkey.test.Locator;
-import org.labkey.test.components.react.ReactSelect;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.ui.ExpectedConditions;
@@ -55,14 +54,6 @@ public class GroupRow extends PermissionsRowBase<GroupRow>
     {
         final WebElement actionContainer = Locator.tagWithClass("div", "expandable-container__action-container")
                 .findWhenNeeded(this);
-
-        public ElementCache()
-        {
-            super(
-                    g -> ReactSelect.Locators.option.withText("Group: " + g), // group options start with "Group: "
-                    null // user options show just their emails
-            );
-        }
 
         final WebElement deleteEmptyGroupBtn()
         {
--- a/src/org/labkey/test/components/ui/permissions/GroupRow.java
+++ b/src/org/labkey/test/components/ui/permissions/GroupRow.java
@@ -1,122 +1,31 @@
 package org.labkey.test.components.ui.permissions;
 
 import org.labkey.test.Locator;
-import org.labkey.test.WebDriverWrapper;
-import org.labkey.test.components.WebDriverComponent;
-import org.labkey.test.components.react.ReactSelect;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.ui.ExpectedConditions;
 
-import java.util.List;
-import java.util.stream.Collectors;
-
-public class GroupRow extends WebDriverComponent<GroupRow.ElementCache>
+public class GroupRow extends PermissionsRowBase<GroupRow>
 {
-    private final WebElement _el;
-    private final WebDriver _driver;
-
     protected GroupRow(WebElement el, WebDriver driver)
     {
-        _el = el;
-        _driver = driver;
+        super(el, driver);
     }
 
     @Override
-    public WebElement getComponentElement()
+    protected WebElement getAssignmentContainer()
     {
-        return _el;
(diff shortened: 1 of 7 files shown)


src/org/labkey/test/components/ui/permissions/GroupRow.java
```diff
@@ -0,0 +1,99 @@
...
+    @Override
+    protected GroupRow getThis()
+    {
+        return this;
+    }
+
+    protected boolean isDeleteEmptyButtonDisabled()
+    {
+        return elementCache().deleteEmptyGroupBtn().getAttribute("class")
+                .contains("disabled-button-with-tooltip");
```
labkey-tchad: Should be able to use the standard `isEnabled` method.
```suggestion
        return !elementCache().deleteEmptyGroupBtn().isEnabled();
```

src/org/labkey/test/components/ui/permissions/GroupRow.java
```diff
@@ -0,0 +1,99 @@
...
+
+        @Override
+        protected Locator locator()
+        {
+            if (_groupName != null)
+                return _baseLocator.withDescendant(Locator.byClass("permissions-title").withText(_groupName));
+            else
+                return _baseLocator;
+        }
+    }
```
labkey-tchad, line 84: Most of this can be shared as well.
```
protected static abstract class PermissionsRowFinder<T> extends WebDriverComponentFinder<T, PermissionsRowBase<T>.PermissionsRowFinder>
    {
        private final Locator.XPathLocator _baseLocator = Locator.tagWithClass("div", "container-expandable")
                .withChild(Locator.tagWithClass("div", "container-expandable-grey"));
        private String _title = null;

        protected PermissionsRowFinder(WebDriver driver)
        {
            super(driver);
        }

        // Call from group/role row finder
        protected void withTitle(String groupName)
        {
            _groupName = groupName;
        }

        @Override
        protected abstract T construct(WebElement el, WebDriver driver);

        @Override
        protected Locator locator()
        {
            if (_title != null)
                return _baseLocator.withDescendant(Locator.byClass("permissions-title").withText(_title));
            else
                return _baseLocator;
        }
    }
```
The generics might not be quite right, but something like that; then extend it for each of the subclasses.

src/org/labkey/test/components/ui/permissions/PermissionsRowBase.java
```diff
@@ -0,0 +1,165 @@
...
+    {
+        expand();
+        findGroup(name).select();
+        return getThis();
+    }
+
+    public T addGroup(String name)
+    {
+        expand();
+        elementCache().memberSelect.select(name);
```
labkey-tchad:
```suggestion
        elementCache().memberSelect.select("Group: " + name);
```

src/org/labkey/test/components/ui/permissions/GroupRow.java
labkey-chrisj: I'm sure it can be shared, and I've tried to get it to work, but no joy.

src/org/labkey/test/components/ui/permissions/PermissionsRowBase.java
```diff
@@ -150,17 +151,31 @@ protected PermissionsMember findGroup(String name)
...
-                .setOptionLocator(ReactSelect.Locators.option::endsWith);
+        private final ReactSelect groupMemberSelect;
+        private final ReactSelect userMemberSelect;
+
+        protected ElementCache(Function<String, Locator> groupOptionLocator, Function<String, Locator> userOptionLocator)
+        {
+            // Use same WebElement for both selects but with different option locators.
+            ReactSelect rawMemberSelect = ReactSelect.finder(getDriver()).findWhenNeeded(this);
+
+            groupMemberSelect = new ReactSelect(rawMemberSelect);
```
labkey-chrisj, line 168: The nice thing about centralizing/sharing the configuration of how these options will be matched in the select is that it discovers that the product isn't consistent about how it shows these options
The downside to doing it is that it doesn't work with the product as is; the assumption that the product will present role options consistently from GroupRow to RoleRow turns out to be wrong.  Tests currently fail because they're looking for an item with "Group: cheese eaters" and the option the test actually wants is just "cheese eaters"

I propose just matching by endsWith(groupName), because it worked before and it works now when I do it locally- or just letting the test fail until the product can be updated to be consistent across scenarios

labkey-tchad (approved):

PR #1253, opened 2022-10-03
Add Method 'getTabsWithCounts' to TabbedGridPanel

labkey-danield:
#### Rationale
Added method getTabsWithCounts to TabbedGridPanel. This will return a map with the tab name as the key and the count as the value. Used int tests that validate sample counts are filtered in the tabs.

#### Related Pull Requests
* https://github.com/LabKey/inventory/pull/550
* https://github.com/LabKey/biologics/pull/1618

#### Changes
* Added method getTabsWithCounts to TabbedGridPanel.

### DIFF
--- a/src/org/labkey/test/components/ui/grids/TabbedGridPanel.java
+++ b/src/org/labkey/test/components/ui/grids/TabbedGridPanel.java
@@ -9,7 +9,7 @@ import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.ui.ExpectedConditions;
 
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/components/ui/grids/TabbedGridPanel.java
```diff
@@ -56,6 +58,25 @@ public List<String> getTabsWithoutCounts()
                 .collect(Collectors.toList());
     }
 
+    public Map<String, Integer> getTabsWithCounts()
+    {
+        Map<String, Integer> counts = new HashMap<>();
```
labkey-tchad: Use a `LinkedHashMap` to retain tab order.
```suggestion
        Map<String, Integer> counts = new LinkedHashMap<>();
```

labkey-tchad (approved):

src/org/labkey/test/components/ui/grids/TabbedGridPanel.java
labkey-danield: Updated.

PR #1259, opened 2022-10-11
Automation test for View vs download options for file and attachment fields

labkey-sweta:
#### Rationale
Automation test for View vs download options for file and attachment fields

Added test case for attachment and file type fields in list and sample type respectively.

#### Related Pull Requests
* 

#### Changes
New test class
src/org/labkey/test/tests/AttachmentFieldTest.java

### DIFF
--- a/src/org/labkey/test/components/domain/DomainFieldRow.java
+++ b/src/org/labkey/test/components/domain/DomainFieldRow.java
@@ -279,6 +279,19 @@ public class DomainFieldRow extends WebDriverComponent<DomainFieldRow.ElementCac
         return this;
     }
 
+    public String getAttachmentBehavior()
+    {
+        expand();
+        return elementCache().attachmentBehavior.getFirstSelectedOption().getText();
+    }
+
+    public DomainFieldRow setAttachmentBehavior(String value)
+    {
+        expand();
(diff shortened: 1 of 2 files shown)


labkey-martyp (commented):
This looks great Sweta. I probably wasn't very clear about assays and datasets. The field type for those is actually "File" and it will have the exact same behavior as "Attachment", but there is a different implementation on the server so it would be good to have a test for one of those types.

labkey-sweta:
> Yes marty, thats why I have added one test case for sample type where the field is called File and one for lists where the field is Attachment.

You think adding for Assay or Dataset will add additional coverage?

labkey-martyp:
> > Yes marty, thats why I have added one test case for sample type where the field is called File and one for lists where the field is Attachment.
> 
> You think adding for Assay or Dataset will add additional coverage?

Got it. The test name threw me off.  Looks good.

labkey-martyp (approved):

labkey-sweta:
> > > Yes marty, thats why I have added one test case for sample type where the field is called File and one for lists where the field is Attachment.
> > 
> > 
> > You think adding for Assay or Dataset will add additional coverage?
> 
> Got it. The test name threw me off. Looks good.

Thought so..! i renamed it now. Thank you

PR #1262, opened 2022-10-13
Change list creation to use updated methods.

labkey-danield:
#### Rationale
Test was using deprecated list creation helper. Updated to test to use current method.

#### Related Pull Requests
None

#### Changes
* Replace calls to deprecated methods with current methods.


labkey-danield:
This update isn't valid anymore. The SimpleModuleTest.createPeopleListInFolde has been removed by this PR https://github.com/LabKey/testAutomation/pull/1265

PR #1266, opened 2022-10-17
Delete lists confirmation page automation

labkey-nicka:
#### Rationale
This adjusts automation to account for the confirmation page when deleting lists from the list management page.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3762

#### Changes
* Add list specific `ConfirmDeletePage`.
* Use in automation of deletion of multiple lists.

### DIFF
--- a/src/org/labkey/test/components/list/ManageListsGrid.java
+++ b/src/org/labkey/test/components/list/ManageListsGrid.java
@@ -17,6 +17,7 @@ package org.labkey.test.components.list;
 
 import org.labkey.test.pages.LabKeyPage;
 import org.labkey.test.pages.list.BeginPage;
+import org.labkey.test.pages.list.ConfirmDeletePage;
 import org.labkey.test.pages.list.EditListDefinitionPage;
 import org.labkey.test.pages.list.ImportListArchivePage;
 import org.labkey.test.util.DataRegionTable;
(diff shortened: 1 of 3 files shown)


src/org/labkey/test/pages/list/ConfirmDeletePage.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    protected ElementCache newElementCache()
+    {
+        return new ElementCache();
+    }
+
+    protected class ElementCache extends LabKeyPage.ElementCache
+    {
+        WebElement deleteButton = Locator.lkButton("Confirm Delete").findWhenNeeded(this);
```
labkey-chrisj, line 30: It might be worthwhile to either implement a waitForPage or put timeouts on elementCache elements; if there's a way to arrive at this page without waiting for a page refresh this will invite race conditions

src/org/labkey/test/components/list/ManageListsGrid.java
```diff
@@ -52,8 +53,9 @@ public File exportSelectedLists()
 
     public BeginPage deleteSelectedLists()
     {
-        deleteSelectedRows();
-        return new BeginPage(getDriver());
+        clickHeaderButtonAndWait("Delete");
```
labkey-chrisj, line 56: perhaps `clicHeaderButtonAndWait `could return a new `ConfirmDeletePage` and this method could chain those calls

labkey-chrisj (approved):

PR #1274, opened 2022-10-21
Try to fix scrolling problem in react selects

labkey-tchad:
#### Rationale
Some react selection operations are scrolling too far. Try to eliminate unnecessary scrolls.

#### Related Pull Requests
* N/A

#### Changes
* Remove some scrolling from react select components
* Remove retry from `filterSelect` (it can never trigger)


labkey-tchad:
Change is too broad. Breaks more than it fixes

PR #1284, opened 2022-10-26
Add test coverage for list improvements

labkey-chrisj:
#### Rationale
This adds some test coverage to cover new list behavior as described in [this spec](https://docs.google.com/document/d/1da08QkWFFdDATIQi9Xd8ZYbhu9GAxcqlTaN40tfEWEE/edit) and [issue 45826](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45826)

#### Related Pull Requests
n/a

#### Changes
added test cases to `CrossfolderListTest`
added method to `DataRegionTable `to add a getter for the containerfilter

### DIFF
--- a/src/org/labkey/test/tests/list/CrossFolderListTest.java
+++ b/src/org/labkey/test/tests/list/CrossFolderListTest.java
@@ -4,6 +4,7 @@ import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.labkey.test.BaseWebDriverTest;
+import org.labkey.test.Locator;
 import org.labkey.test.categories.Daily;
 import org.labkey.test.categories.Data;
 import org.labkey.test.categories.Hosting;
@@ -13,6 +14,8 @@ import org.labkey.test.params.FieldDefinition;
 import org.labkey.test.params.list.IntListDefinition;
 import org.labkey.test.params.list.ListDefinition;
 import org.labkey.test.util.DataRegionTable;
+import org.labkey.test.util.DomainUtils;
+import org.labkey.test.util.ListHelper;
 import org.labkey.test.util.query.QueryApiHelper;
 
 import java.util.Arrays;
@@ -21,21 +24,24 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
-import static org.hamcrest.CoreMatchers.hasItems;
-import static org.hamcrest.MatcherAssert.assertThat;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 
 @Category({Daily.class, Data.class, Hosting.class})
 public class CrossFolderListTest extends BaseWebDriverTest
 {
+    private static final String PROJECT_NAME = "CrossFolderListTest Project";
     private static final String SUBFOLDER_A = "subA";
-    private static String SUBFOLDER_A_PATH;
+    private static final String SUBFOLDER_A_PATH = PROJECT_NAME + "/" + SUBFOLDER_A;
+
+    private static final String SHARED_LIST = "shared_list_for_crossfolder_test";
 
     @Override
     protected void doCleanup(boolean afterTest)
     {
         _containerHelper.deleteProject(getProjectName(), afterTest);
+        DomainUtils.ensureDeleted("/Shared", "lists", SHARED_LIST);
     }
 
     @BeforeClass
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/tests/list/CrossFolderListTest.java
```diff
@@ -49,7 +55,94 @@ private void doSetup()
...
+        // click the 'stringy' link in the record we put into the shared list
+        clickAndWait(Locator.linkWithText("stringy"));  // the presence of this link confirms data in this view
+        assertThat(getURL().toString()).as("expect data url to be in project, is not in /Shared")
+                .contains("CrossFolderListTest")
+                .doesNotContain("Shared");
+        // now click on the list link, expecting that to navigate us to where it is defined, in ths /Shared folder
+        clickAndWait(Locator.linkWithText(SHARED_LIST));
+        assertThat(getURL().toString()).as("expect link to shared list to navigate to shared folder")
+                .doesNotContain("CrossFolderListTest")
+                .contains("Shared");
```
labkey-tchad: Should be able to do an exact match on the container path. `getCurrentContainerPath` includes a leading slash but that's easy enough to handle:
```suggestion
        assertThat(getCurrentContainerPath()).as("expect data url to be in project, is not in /Shared")
                . isEqualTo("/" + getProjectName());
        // now click on the list link, expecting that to navigate us to where it is defined, in the /Shared folder
        clickAndWait(Locator.linkWithText(SHARED_LIST));
        assertThat(getCurrentContainerPath()).as("expect link to shared list to navigate to shared folder")
                .isEqualTo("/Shared");
```

src/org/labkey/test/util/DataRegionTable.java
```diff
@@ -1306,6 +1306,20 @@ public void setContainerFilter(ContainerFilterType filterType)
...
 
+    public String getContainerFilter()
+    {
+        getViewsMenu().openMenuTo("Folder Filter", "Folder Filter");
+        var visibleFilterMenuItemOptions = getViewsMenu().findVisibleMenuItems();
+        for (WebElement el : visibleFilterMenuItemOptions)
+        {
+            if (Locator.tagWithClass("i", "fa-check-square-o").existsIn(el))
+            {
+                return el.getText();   // would be nice if enum.valueOf() allowed us to get the enum constant from its text but oh java you crazy
```
labkey-nicka: nit: While I appreciate the sentiment this comment isn't very useful.

labkey-nicka (commented):
Looks good. Leaving to @labkey-tchad for test review.

src/org/labkey/test/util/DataRegionTable.java
labkey-chrisj: Fair enough.  I just put it there as a way to fish for suggestions- I mean I could implement a 'fromLabel' method on that enum that streams over its possible label values and returns the constant value for the matching one (or null), but sometimes Trey will have a trick or helper somewhere that solves the problem.  I can remove the comment if we're ok with returning a string instead of an enum

src/org/labkey/test/tests/list/CrossFolderListTest.java
```diff
@@ -78,8 +171,8 @@ public void testAddDataInSubfolderToTopLevelList() throws Exception
         var displayedData = subFolderListPage.getGrid().getTableData();
         assertEquals("expect only subfolder data to be shown here by default",
                 rowsToInsert.size(), displayedData.size());
-        assertThat("expect only the data inserted here to be shown",
-                subFolderListPage.getGrid().getColumnDataAsText("String Column"), hasItems("stringy", "chewy"));
+        assertThat(subFolderListPage.getGrid().getColumnDataAsText("String Column"))
+                .as("expect only the data inserted here to be shown").contains("stringy", "chewy");
```
labkey-tchad, line 174: Should check for an exact match.
```suggestion
        assertThat(subFolderListPage.getGrid().getColumnDataAsText("String Column"))
                .as("expect only the data inserted here to be shown").containsExactly("stringy", "chewy");
```

src/org/labkey/test/util/DataRegionTable.java
labkey-tchad: No tricks for this one. Adding a `ContainerFilterType.fromLabel` method seems reasonable though. We have such methods in several test enums.

labkey-tchad (approved):
One minor fix and a couple of comments.

src/org/labkey/test/tests/list/CrossFolderListTest.java
labkey-chrisj: My concern about that is that other test cases might create lists that will appear in this view; I only wanted to ensure that these two (that I know must be present) are there

PR #1289, opened 2022-10-28
Automation test for "Assays are messed up after folder rename" issue

labkey-sweta:
#### Rationale
 Test coverage : Issue 46473: Assays are messed up after folder rename
 https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46473

1. Create an Assay project.
2. Create an assay design in that project.
3. Rename the project from Folder Management --> Rename and uncheck the Alias current name (recommended).
4. Manage Assay design --> Edit assay design should navigate to original assay design created and renamed folder.

#### Related Pull Requests
https://github.com/LabKey/platform/pull/3794

#### Changes
New test page src/org/labkey/test/pages/admin/RenameFolderPage.java

### DIFF
--- a/src/org/labkey/test/pages/admin/RenameFolderPage.java
+++ b/src/org/labkey/test/pages/admin/RenameFolderPage.java
@@ -14,10 +14,10 @@ public class RenameFolderPage extends LabKeyPage<RenameFolderPage.ElementCache>
         super(driver);
     }
 
-    public static ReorderFoldersPage beginAt(WebDriverWrapper driver, String containerPath)
+    public static RenameFolderPage beginAt(WebDriverWrapper driver, String containerPath)
     {
         driver.beginAt(WebTestHelper.buildURL("admin", containerPath, "renameFolder"));
-        return new ReorderFoldersPage(driver.getDriver());
+        return new RenameFolderPage(driver.getDriver());
     }
 
     public String getProjectName()
--- a/src/org/labkey/test/pages/admin/RenameFolderPage.java
+++ b/src/org/labkey/test/pages/admin/RenameFolderPage.java
@@ -1,0 +1,84 @@
+package org.labkey.test.pages.admin;
+
+import org.labkey.test.Locator;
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/pages/admin/RenameFolderPage.java
```diff
@@ -0,0 +1,84 @@
...
+import org.openqa.selenium.WebElement;
+
+public class RenameFolderPage extends LabKeyPage<RenameFolderPage.ElementCache>
+{
+    public RenameFolderPage(WebDriver driver)
+    {
+        super(driver);
+    }
+
+    public static ReorderFoldersPage beginAt(WebDriverWrapper driver, String containerPath)
```
labkey-danield: Why is the beginArt for the the **Rename**FolderPage returning a **Reorder**FolderPage?

src/org/labkey/test/pages/admin/RenameFolderPage.java
```diff
@@ -0,0 +1,84 @@
...
+    {
+        driver.beginAt(WebTestHelper.buildURL("admin", containerPath, "renameFolder"));
+        return new ReorderFoldersPage(driver.getDriver());
```
labkey-danield: Do you need a `waitForPageLoad` or something like to make sure the page is loaded before returning?

labkey-danield (changes requested):
Need some clarity on the return type for the beginAt method.

src/org/labkey/test/pages/admin/RenameFolderPage.java
labkey-sweta: good catch

src/org/labkey/test/pages/admin/RenameFolderPage.java
labkey-sweta: page is loading quickly

labkey-danield (approved):

PR #1290, opened 2022-10-29
Test Automation for Issue 46465 (Sample export do not respect filter when samples have been selected)

labkey-danield:
#### Rationale
Automation for [Issue 46465: LKSM/LKB: Sample export and some grid actions do not respect filter when samples have been selected](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46465)

#### Related Pull Requests
* None

#### Changes
* Added scenario to GridPanelTest.testExport


labkey-susanh (approved):
Test seems good for the export case. Once a fix is added, we'll want to add tests for other actions as well.  I'm no sure when I'll get to this fix, but it's up to you whether to mute this test now or make other adjustments in TC.

PR #1294, opened 2022-11-02
Fix PuppeteerSettings response handling

labkey-adam:
#### Rationale
NotebookExportTest is failing due to `PuppeteerSettings` having problems parsing the JSON response from `UpdateSettingsAction`. The new `JSONObject(Object)` constructor doesn't handle `Map` parameters like the old one did; a simple cast should do the trick.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/1287

### DIFF
--- a/src/org/labkey/remoteapi/puppeteer/PuppeteerSettings.java
+++ b/src/org/labkey/remoteapi/puppeteer/PuppeteerSettings.java
@@ -17,6 +17,7 @@ package org.labkey.remoteapi.puppeteer;
 
 import org.json.JSONObject;
 import org.labkey.remoteapi.CommandResponse;
+import org.labkey.test.util.TestLogger;
 
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


labkey-jeckels (approved):

src/org/labkey/remoteapi/puppeteer/PuppeteerSettings.java
```diff
@@ -50,7 +52,7 @@ public PuppeteerSettings(JSONObject json)
 
     public PuppeteerSettings(CommandResponse response)
     {
-        this(new JSONObject(response.getParsedData().get("data")));
+        this(new JSONObject((Map<String, Object>)response.getParsedData().get("data")));
```
labkey-tchad, line 48: I don't think this will change anything. I don't think there's any particular reason for this to take a JSONObject. Could make the constructor just take the map, or use `opt` methods to pull the properties from the JSONObject.

labkey-tchad (approved):

src/org/labkey/remoteapi/puppeteer/PuppeteerSettings.java
labkey-adam: This causes the correct JSONObject `Map` constructor to be called (methods bind at compile time and we really don't want that `Object` version to be called). Agreed that JSONObject isn't really needed, but was trying for minimal changes. I expect to use `opt*` but first temp logging to get TC to show me what's really optional and what's not.

PR #1299, opened 2022-11-04
Fix for module property name not the same as the label on the page

labkey-danield:
#### Rationale
Fix for failures in the Accounts test. Updates to return the new Json object is causing problems with custom module property. The existing test code was assuming that the property name would match the label on the page. That is no longer a valid assumption 100% of the time. This fix allows for the module properties test object to have a label property to help find it on the page.

#### Related Pull Requests
* None

#### Changes
* Added a label property to ModuleProperty (added a getter).
* Added overloaded constructor to the ModuleProperty to take a label value. If no label is provided use the property name as the label.
* Overloaded constructors in ModulePropertyValue object that can take a label property.
* Modify BaseWebDriverTest.getModulePropertyFieldRef to wait for property label.

### DIFF
--- a/src/org/labkey/test/ModulePropertyValue.java
+++ b/src/org/labkey/test/ModulePropertyValue.java
@@ -81,14 +81,30 @@ public class ModulePropertyValue extends ModuleProperty
         _inputType = inputType;
     }
 
+    public ModulePropertyValue(String moduleName, String containerPath, String propertyName, String propertyLabel, Object value, InputType inputType)
+    {
+        super(moduleName, containerPath, propertyName, propertyLabel, value);
+        _inputType = inputType;
+    }
+
     public ModulePropertyValue(String moduleName, String containerPath, String propertyName, String value)
     {
         this(moduleName, containerPath, propertyName, value, InputType.text);
     }
 
+    public ModulePropertyValue(String moduleName, String containerPath, String propertyName, String propertyLabel, String value)
+    {
+        this(moduleName, containerPath, propertyName, propertyLabel, value, InputType.text);
+    }
+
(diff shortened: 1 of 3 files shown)


src/org/labkey/test/ModulePropertyValue.java
```diff
@@ -81,16 +81,32 @@ public ModulePropertyValue(String moduleName, String containerPath, String prope
...
     public ModulePropertyValue(String moduleName, String containerPath, String propertyName, boolean value)
     {
         this(moduleName, containerPath, propertyName, value, InputType.checkbox);
     }
 
+    public ModulePropertyValue(String moduleName, String containerPath, String propertyName, String propertyLabel, boolean value)
+    {
+        this(moduleName, containerPath, propertyName, propertyLabel, value, InputType.checkbox);
+    }
+
```
labkey-adam, line 109: Optional: might be time to consider a builder pattern to replace these six constructors.

labkey-adam (approved):

labkey-adam:
For the historical record: the previous `waitForText(property.getPropertyName())` was, in many cases, not doing what anyone would have expected. It was not actually waiting for the ExtJS-based inputs to render because they show the property label (if provided), not the property name. Bizarrely, this check was seeing the property name inside LABKEY.moduleContext JSON, where module property values are in-lined for client use. This surely was not the intent of the test helper. A subtle change to JSONObject pointed out the erroneous check.

src/org/labkey/test/ModulePropertyValue.java
labkey-danield: Would make sense, but I'm going to defer that change to later.

PR #1306, opened 2022-11-09
Change How ReactDatePicker.clickTime Finds the List Item

labkey-danield:
#### Rationale
The SMSampleCreateTest.testCreateWithDateInputs in SampleManage has been failing with this error:
```
ElementNotInteractableException: Click failed; try clicking a child element. Firefox doesn't like clicking certain wrapping elements
Element <li class="react-datepicker__time-list-item "> could not be scrolled into view
```
This attempts to address that issue in the same way as the [fix for SampleLocationPanel in inventory](https://github.com/LabKey/inventory/commit/0989ad2d6c52ab39706a6dcb58df2fbe422f1921).

#### Related Pull Requests
* None

#### Changes
* Changed how the ReactDatePicker.clickTIme finds and clicks on the time value in the list.

### DIFF
--- a/src/org/labkey/test/components/react/ReactDatePicker.java
+++ b/src/org/labkey/test/components/react/ReactDatePicker.java
@@ -85,7 +85,10 @@ public class ReactDatePicker extends WebDriverComponent<ReactDatePicker.ElementC
     public ReactDatePicker clickTime(String time)
     {
         expand();
-        elementCache().timeListItem(time).click();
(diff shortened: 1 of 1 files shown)


labkey-tchad (approved):
In the screenshot, the element that the test is trying to click is, indeed, not visible. I think Alan's change to the scroll margins is causing some things to scroll too much.

PR #1307, opened 2022-11-11
ELN, Group Assignment test

labkey-chrisj:
#### Rationale
This updates `PermissionsRowBase `to not fail when adding a user that happens to have a suffix (such as their userDisplayName in parenthesis) added to the expected email.  (the `addMember `method waits for the list of member chips in it to have one with the added email- this updates that check to be satisfied when one starts with the email)

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/1722
https://github.com/LabKey/labbook/pull/319

#### Changes
updates `addMember `to tolerate new behavior


labkey-danield (approved):

labkey-danield:
My understanding is that if an account displays the user display name in parens it means that there are two accounts with the same display name.

PR #1315, opened 2022-11-22
add automated test coverage for user-supplied barcode fields

labkey-chrisj:
#### Rationale
This updates domain designer test classes to be able to interact with the 'scannable' checkbox, used to enable using a field as an alternate barcode field

#### Related Pull Requests
https://github.com/LabKey/sampleManagement/pull/1409

#### Changes
`DomainFieldRow `- added ability to interact with the scannable checkbox, and know whether it is present or not
'EntityTypeDesigner' - added ability to interact with stored amount display units, barcode info

### DIFF
--- a/src/org/labkey/test/components/domain/DomainFieldRow.java
+++ b/src/org/labkey/test/components/domain/DomainFieldRow.java
@@ -416,6 +416,30 @@ public class DomainFieldRow extends WebDriverComponent<DomainFieldRow.ElementCac
         return getWrapper().isElementPresent(elementCache().getCharScaleInputLocForRow(rowIndex));
     }
 
+    // barcode- scannable options
+
+    /**
+     *  Indicates that the field should be searched when
+     */
+    public boolean isFieldScannable()
+    {
+        expand();
+        return elementCache().scannableCheckbox.get();
+    }
+
+    public DomainFieldRow setFieldScannable(boolean checked)
+    {
+        expand();
+        elementCache().scannableCheckbox.set(checked);
+        return this;
+    }
+
+    public boolean isScannableCheckboxPresent()
+    {
+        expand();
+        return elementCache().scannableCheckboxLoc.existsIn(this);
+    }
+
     //
     // date field options.
 
@@ -1099,7 +1123,7 @@ public class DomainFieldRow extends WebDriverComponent<DomainFieldRow.ElementCac
     {
         private final Locator.XPathLocator _baseLocator = Locator.tagWithClassContaining("div", "domain-field-row").withoutClass("domain-floating-hdr");
         private String _title = null;
-        private DomainFormPanel _domainFormPanel;
+        private final DomainFormPanel _domainFormPanel;
 
         public DomainFieldRowFinder(DomainFormPanel panel, WebDriver driver)
         {
@@ -1185,6 +1209,11 @@ public class DomainFieldRow extends WebDriverComponent<DomainFieldRow.ElementCac
                 .refindWhenNeeded(this));
         public final Input charScaleInput = new Input(Locator.tagWithAttributeContaining("input", "id", "domainpropertiesrow-scale")
                 .refindWhenNeeded(this), getDriver());
+
+        // barcode option
+        public final Locator scannableCheckboxLoc = Locator.input("domainpropertiesrow-scannable");
+        public final Checkbox scannableCheckbox = new Checkbox(scannableCheckboxLoc.findWhenNeeded(this).withTimeout(WAIT_FOR_JAVASCRIPT));
+
         // date field options
         public final Input dateFormatInput = new Input(Locator.tagWithAttributeContaining("input", "id", "domainpropertiesrow-format")
                 .refindWhenNeeded(this), getDriver());
--- a/src/org/labkey/test/components/domain/DomainFieldRow.java
+++ b/src/org/labkey/test/components/domain/DomainFieldRow.java
@@ -25,6 +25,7 @@ import org.openqa.selenium.support.ui.FluentWait;
 import org.openqa.selenium.support.ui.Select;
 
 import java.time.Duration;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/components/domain/DomainFieldRow.java
```diff
@@ -1185,6 +1209,11 @@ protected class ElementCache extends WebDriverComponent.ElementCache
                 .refindWhenNeeded(this));
         public final Input charScaleInput = new Input(Locator.tagWithAttributeContaining("input", "id", "domainpropertiesrow-scale")
                 .refindWhenNeeded(this), getDriver());
+
+        // barcode option
+        public final Locator scannableCheckboxLoc = Locator.input("domainpropertiesrow-scannable");
+        public final Checkbox scannableCheckbox = new Checkbox(scannableCheckboxLoc.findWhenNeeded(this).withTimeout(WAIT_FOR_JAVASCRIPT));
```
labkey-danield: Shouldn't this be `refindWhenNeeded`? It's possible to change the field type and the checkbox will go away, and then change it back to a supported type and it will reappear.

src/org/labkey/test/components/domain/DomainFieldRow.java
```diff
@@ -1185,6 +1209,11 @@ protected class ElementCache extends WebDriverComponent.ElementCache
                 .refindWhenNeeded(this));
         public final Input charScaleInput = new Input(Locator.tagWithAttributeContaining("input", "id", "domainpropertiesrow-scale")
                 .refindWhenNeeded(this), getDriver());
+
+        // barcode option
+        public final Locator scannableCheckboxLoc = Locator.input("domainpropertiesrow-scannable");
```
labkey-danield: Does this need to be public?

src/org/labkey/test/components/ui/domainproperties/EntityTypeDesigner.java
```diff
@@ -230,6 +232,27 @@ public Optional<WebElement> optionalWarningAlert()
...
+    public String getStoredAmountDisplayUnits()
+    {
+        return elementCache().storedAmountDisplayUnitsSelect.getValue();
+    }
+
+    public T setStoredAmountDisplayUnits(String displayUnits)
+    {
+        elementCache().storedAmountDisplayUnitsSelect.select(displayUnits);
+        return getThis();
+    }
```
labkey-danield: These aren't needed here they are already defined in FMSampleTypeDefinition, and are not visible outside of the app (i.e. in LKS).

src/org/labkey/test/components/ui/domainproperties/EntityTypeDesigner.java
```diff
@@ -230,6 +232,27 @@ public Optional<WebElement> optionalWarningAlert()
...
+
+    public boolean isBarcodeInfoShown()
+    {
+        return elementCache().uniqueIdMsgLoc.existsIn(this);
+    }
+
+    public String getBarcodeInfoMsg()
+    {
+        return elementCache().uniqueIdMsgLoc.waitForElement(this, WAIT_FOR_JAVASCRIPT).getText();
+    }
```
labkey-danield: Same comment here, these fields aren't used and are not visible outside of the app.

src/org/labkey/test/components/ui/domainproperties/EntityTypeDesigner.java
```diff
@@ -297,6 +320,11 @@ public final WebElement helpTarget(String divLabelText)
 
         protected final WebElement resetGenIdButton = Locator.tagWithClass("button", "reset-genid-btn").findWhenNeeded(this);
 
+        protected final ReactSelect storedAmountDisplayUnitsSelect = ReactSelect.finder(getDriver())
+                .withContainerClass("sampleset-metric-unit-select-container").timeout(WAIT_FOR_JAVASCRIPT).findWhenNeeded(this);
+
+        final Locator uniqueIdMsgLoc = Locator.tagWithClass("div", "uniqueid-msg");
+
```
labkey-danield, line 307: Not needed.

labkey-danield (approved):
Changes to the EntityTypeDesigner are not needed, and don't look like they are used.

src/org/labkey/test/components/ui/domainproperties/EntityTypeDesigner.java
labkey-chrisj: OK, I only did these because I didn't know how to use an FMSampleTypeDefinition to drive the UI (does it deal with the UI? I traced into the code a bit where it's used but assumed it deals with the API when I didn't see our UI wrappers showing evidence of being aware they exist)
If I'm wrong here I'll happily change this

labkey-danield: Fair point, and I was mistaken in referring to FM SampleTypeDefinition (my bad). I knew this had already been implemented before I just forgot where. Access to the stored amount fields are already present and are exposed through SMSampleTypeDefinition. They are there because they are only required for sample types in the App.

The methods you have provided here are not used and can be removed.

An example of how you would set the stored amount using the existing code is:
```
    @Test
    public void testSearchOnAlternateBarcode() throws Exception
    {
        String sampleTypeName = "test_search_on_alternate_barcodes";
        SMSampleTypeDefinition sampleTypeDefinition = new SMSampleTypeDefinition(sampleTypeName, InventoryMetricUnit.UNIT);
        sampleTypeDefinition.setFields(getTestFields());

        var dgen = SampleTypeAPIHelper.createEmptySampleType(getProjectName(), sampleTypeDefinition);
```

The sample type definition returned can be used to create a sample type using the UI or the API.

src/org/labkey/test/components/ui/domainproperties/EntityTypeDesigner.java
labkey-chrisj: I had figured it might be a useful check for future me to be able to confirm that from outside of the app but okay, consider it gone

PR #1321, opened 2022-12-06
LKSM Starter/Pro: Split up test batches across composite suites

labkey-tchad:
#### Rationale
We currently split up some test suites to run in parallel with the syntax `BVT[a/b]` where `a` is the batch number and `b` is the total number of batches. The is inadequate in the case where we want to parallelize more complicated suite definitions (e.g. `samplemanagement,-SMStarter`). If we make the shard numbers separate parameters, we can parallelize such suites and keep the batches.

Designed to integrate with TeamCity parallel test batching: https://www.jetbrains.com/help/teamcity/parallel-tests.html#custom-tests

#### Related Pull Requests
* N/A

#### Changes
* Add parameters for splitting composite suites into batches
  * `webtest.parallelTests.currentBatch`
  * `webtest.parallelTests.totalBatches`

### DIFF
--- a/src/org/labkey/test/Runner.java
+++ b/src/org/labkey/test/Runner.java
@@ -33,6 +33,7 @@ import org.apache.commons.lang3.time.DateUtils;
 import org.apache.commons.lang3.time.FastDateFormat;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.jetbrains.annotations.NotNull;
 import org.junit.Ignore;
 import org.junit.runner.Description;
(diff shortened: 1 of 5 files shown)


labkey-tchad:
Verified with new parameters:
BVT A: https://teamcity.labkey.org/buildConfiguration/bt7/2155171
BVT B: https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_BvtBPostgres/2155273
_Note: BVT A test count is lower because it doesn't include some redundant tests from BVT B_

labkey-danield (approved):

PR #1331, opened 2022-12-20
Fix path issue in APIAssayHelper.

labkey-danield:
#### Rationale
Test started failing recently on the embedded test suite with a [IllegalArgumentException: URI path begins with multiple slashes](https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Internal_Embedded_SampleManagerEmbeddedPostgres/2180604?buildTab=tests&status=failed&name=org.labkey.test.tests.samplemanagement.aliquots.SMProAliquotRollupSourcesTest.testAssaysPage). The call stacks pointed to various methods in APIAssayHelper where they would always prepend a "/" to the path sent int. This change checks if the path sent in already has a leading "/" before appending one.

#### Related Pull Requests
* None

#### Changes
* Update various methods in APIAssayHelper to check before prepending a "/" to the path parameter.

### DIFF
--- a/src/org/labkey/test/util/APIAssayHelper.java
+++ b/src/org/labkey/test/util/APIAssayHelper.java
@@ -65,8 +65,9 @@ public class APIAssayHelper extends AbstractAssayHelper
     {
         ImportRunCommand  irc = new ImportRunCommand(assayID, file);
         irc.setBatchProperties(batchProperties);
-        irc.setTimeout(180000); // Wait 3 minutes for assay import
-        return irc.execute(_test.createDefaultConnection(), "/" + projectPath);
+        irc.setTimeout(180000); // Wait 3 minutes for
+        projectPath = projectPath.startsWith("/") ? projectPath : "/" + projectPath;// assay import
+        return irc.execute(_test.createDefaultConnection(), projectPath);
     }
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/util/APIAssayHelper.java
```diff
@@ -65,8 +65,8 @@ public ImportRunResponse importAssay(int assayID, File file, String projectPath,
     {
         ImportRunCommand  irc = new ImportRunCommand(assayID, file);
         irc.setBatchProperties(batchProperties);
-        irc.setTimeout(180000); // Wait 3 minutes for assay import
-        return irc.execute(_test.createDefaultConnection(), "/" + projectPath);
+        irc.setTimeout(180000); // Wait 3 minutes for
+        return irc.execute(_test.createDefaultConnection(), projectPath);
```
labkey-tchad: Accidentally truncated comment

labkey-tchad (approved):

src/org/labkey/test/util/APIAssayHelper.java
labkey-danield: Replaced the truncated comment.

PR #1332, opened 2022-12-21
Consolidate code for folder/project menu and getProductMenu to NavBar

labkey-danield:
#### Rationale
Since more of the app code is shared I thought it would make sense to consolidate some of the getters for the common menu items.

The productMenu component checks for the number of "columns" that are present before it confirms that it has been expanded. This doesn't feel necessary any more. However since the products menu will be changing in 23.01 I don't feel it is appropriate to change that behavior at this time. I do think this code should be revisited when the menu is changed in the next few weeks.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/1481

#### Changes
* Move the getters for the folder (i.e. project) menu to the NavBar.
* Move the getProductMenu code to the NavBar.


labkey-chrisj:
I'm usually a fan of code consolidation/elimination of redundancy, but it occurs to me that passing through getters to a base class in a separate repo might make for more painful maintenance down the line in case we decide to make the app behave one way in one app and a different way in another.
I trust that this will work, but a downside to pulling code into base classes can have the effect of hiding it or creating a 'you just have to know about it' situation that sometimes has me kicking myself for not walking up into the base class of whatever I'm looking at

labkey-danield:
It is always a possibility that UI in the apps might change, however given the effort by the devs and pms to standardize the experience between the two apps makes it less likely that the experience will diverge anytime soon. I would argue that having the shared code will make maintenance easier not harder. I've seen a few PRs where they devs made small changes, because of the standardization effort, to code like this in several places. Being able to narrow that down seems like a good thing.

src/org/labkey/test/components/ui/navigation/NavBar.java
```diff
@@ -117,5 +125,13 @@ protected abstract class ElementCache extends Component<ElementCache>.ElementCac
         public WebElement projectNameDisplay = Locator.tagWithClass("span", "project-name").findWhenNeeded(this);
         public Input searchBox = Input.Input(Locator.tagWithClass("input", "navbar__search-input"), getDriver()).findWhenNeeded(this);
         public MultiMenu searchMenu = new MultiMenu.MultiMenuFinder(getDriver()).withButtonId("find-and-search-menu").findWhenNeeded(this);
+        public final MultiMenu folderMenu = new MultiMenu.MultiMenuFinder(getDriver())
+                .withButtonId("folder-menu").refindWhenNeeded(getDriver());
+        public final ProductMenu productMenu(int columnCount)
+        {
+            return ProductMenu.finder(getDriver()).timeout(1000).findWhenNeeded(this)
+                    .setExpectedSectionCount(columnCount);
+        }
```
labkey-tchad: I'd avoid exposing a public method that forces one to specify the expected column count (`getProductMenu`, earlier in this PR). You could get the count from the SM/BioNavBar subclasses (instead of getting the menu itself like this used to do)
```suggestion
        public final ProductMenu productMenu = ProductMenu.finder(getDriver()).timeout(1000).findWhenNeeded(this)
                .setExpectedSectionCount(getExpectedProductMenuColumnCount());
```

Then add an abstract `getExpectedProductMenuColumnCount` to the `NavBar` or the `NavBar.ElementCache`:
```
public abstract int getExpectedProductMenuColumnCount();
```

labkey-danield: The count is no longer needed.

labkey-tchad (approved):

PR #1335, opened 2022-12-22
Fix clicking in lineage explorer test component

labkey-tchad:
#### Rationale
The previous implementation wasn't waiting and scrolling explicitly and would fail intermittently. This solution seems more reliable.

Testing all usages on TeamCity: https://teamcity.labkey.org/buildConfiguration/bt7/2186721?buildTab=tests

#### Related Pull Requests
* N/A

#### Changes
* Scroll to lineage node detail before clicking


labkey-chrisj (approved):
Hope this works!

PR #1338, opened 2022-12-27
Fix case-sensitivity issue in test

bbimber:


bbimber:
@labkey-chrisj: hi chris, you assigned me the ScriptValidationTest failure. Would it be possible to mirror this branch on the LK side of the fork so that TeamCity will run tests against it? It does run pick this up b/c I created this branch on my github and I dont have permissions on the LK org.

labkey-chrisj:
Hey Ben, thanks for responding.
I'm unfamiliar with the way your branches relate to LK so it's probably best for me to punt to someone who regularly interacts with you- I'm guessing that's @labkey-tchad  or @josheckels?

labkey-tchad:
I'll take a look.

src/org/labkey/test/tests/ScriptValidationTest.java
```diff
@@ -147,15 +148,16 @@ public void doTestCrossFolderSaveRows() throws Exception
                 new Object[][]{new Object[]{modelId, colors.get(0).name, 2000, 1234, new Date()}}, null);
         CommandResponse response = saveRowsCommand.execute(cn, "/home");
         Map<String, Object> row = (Map)((Map)((List)((Map)((List)response.getParsedData().get("result")).get(0)).get("rows")).get(0)).get("values");
-        Object vehicleRowId = row.get("rowid");
+        Object vehicleRowId = row.get("rowId");
```
labkey-tchad: Consider using a case-insensitive map
```suggestion
        Object vehicleRowId = new CaseInsensitiveHashMap(row).get("rowId");
```
Other references to "rowId" in this test should be ok, since they are inspecting `SelectRowsResponse`, which already uses such a map. (no harm in updating them though)

bbimber: Generally a good idea, but labkey should be deterministic in the case in which it returns columns. While our test could be case-insensitive, JS code is case sensitive and changing around the server's response is a big deal.

I think something has changed in the case in which columns are returned. I would have expected the response to mirror the case in vehicles.xml, and the schema browser shows this. So why is the SaveRowsResponse converting everything to lowerCamelCase?

labkey-tchad:
Verifying fix on TeamCity: https://teamcity.labkey.org/buildConfiguration/LabKey_2211Release_Community_BvtASqlserver/2193749

src/org/labkey/test/tests/ScriptValidationTest.java
bbimber: Yes, this is weird. We're providing SaveRowsAction with a map using the camel case values (which matches what shows as the field case in schema browser:

image


here's the actual map in java (rowsToProcess):

image

however, after the call to: 

List<Map<String, Object>> responseRows =
                        commandType.saveRows(qus, rowsToProcess, getUser(), container, configParameters, extraContext);

The case on responseRows has all been normalized to lowerCamelCase:

image

bbimber:
> Verifying fix on TeamCity: https://teamcity.labkey.org/buildConfiguration/LabKey_2211Release_Community_BvtASqlserver/2193749

So making the test either case-insensitive or correcting the case of the "RowId" string will work; however, I think something else in the product changed to trigger this failure. Something is globally changing the case on al the returned column names relative to what I would expect.

labkey-tchad:
> > Verifying fix on TeamCity: https://teamcity.labkey.org/buildConfiguration/LabKey_2211Release_Community_BvtASqlserver/2193749
> 
> So making the test either case-insensitive or correcting the case of the "RowId" string will work; however, I think something else in the product changed to trigger this failure. Something is globally changing the case on al the returned column names relative to what I would expect.

I don't think there's any recent change causing this. We've seen this sort of discrepancy between SQL Server and Postgres in the past. I can't remember the underlying problem but I think its only inconsistent for some subset of queries.
This is now failing on postgres (Looks like I was wrong about case-insensitivity for the row response): https://teamcity.labkey.org/buildConfiguration/LabKey_2211Release_Community_BvtAPostgres/2193755

src/org/labkey/test/tests/ScriptValidationTest.java
bbimber: Update: the row's case is preserved until QueryController line 3989, where it takes the updated rows and re-selects them. This re-selected map has lowerCamelCase keys.

```
                // Issue 42519: Submitter role not able to insert
                // as per the definition of submitter, should allow insert without read
                if (qus.hasPermission(user, ReadPermission.class))
                {
                    updatedRows = qus.getRows(user, container, updatedRows); 
                }
```

It maps further into DefaultQueryUpdateService._select, line 233. The original case is preserved until here. TableSelector is called using the PK, and the resulting map has the field keys using lowerCamelCase. However, I cannot figure out any change that plausibly changed how this behavior worked though.

bbimber:
> > > Verifying fix on TeamCity: https://teamcity.labkey.org/buildConfiguration/LabKey_2211Release_Community_BvtASqlserver/2193749
> > 
> > 
> > So making the test either case-insensitive or correcting the case of the "RowId" string will work; however, I think something else in the product changed to trigger this failure. Something is globally changing the case on al the returned column names relative to what I would expect.
> 
> I don't think there's any recent change causing this. We've seen this sort of discrepancy between SQL Server and Postgres in the past. I can't remember the underlying problem but I think its only inconsistent for some subset of queries. This is now failing on postgres (Looks like I was wrong about case-insensitivity for the row response): https://teamcity.labkey.org/buildConfiguration/LabKey_2211Release_Community_BvtAPostgres/2193755

The reason I'm really interested in this is that changing client-side case of fields on the LK query responses was a gigantic pain when I worked on EHR. As far as I can remember, the client-side result always matched the case defined in the schema / schema-XML. Not following that here worries me.

bbimber:
Another update: the case change happens in ResultSetRowMapFactory line 81, but this code is incredibly old. The case-changing behavior also only occurs when using SaveRows with the command insertWithKeys, and not a conventional insert. There's no good reason this test needs to be using insertWithKeys, but it's exposing inconsistency in the client APIs.

I also bet this particular issue never hit EHR since I was pretty strict about lower-casing the first letter, and that's my default when I write code (unlike simpletest module). This code calls Introspector.decapitalize(), meaning it only changes the first letter, but otherwise preserves the TableInfo column's case. Therefore I would not have encountered this otherwise.

labkey-tchad:
Please feel free to open a ticket. I'm fairly confident that our query APIs have always behaved this way but I can't find any old issue explaining why we haven't fixed it.
The Java and [JavaScript](https://labkey.github.io/labkey-api-js/classes/Query.Row.html#get) APIs both store rows in case-insensitive maps.
We've definitely had to make numerous column name workarounds in tests to make them work on both postgres and SQL Server. I found an old issue where we fixed this sort of thing for datasets: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=6036

bbimber:
@labkey-tchad: I think my last commit will pass, if you're able to merge that.

Note: the inconsistent behavior can be see on ScriptValidationTest line 145, where "RowId' returns a value from the case-sensitive HashMap, since that uses conventional InsertRows. In contrast, line 153, which uses SaveRows/insertWithKeys will hit this alternate codepath where the data are inserted, but then re-selected for some reason (QueryController line 3989). This code ends up calling Introspector.decapitalize() on the column names (ResultSetRowMapFactory line 81) for an unclear reason. The result is that the column case is respected, with the sole exception that the first letter is converted to lower case. Therefore if the column has a lowercase first letter, it's identical, but if not then the case is changed.

labkey-tchad:
Verifying `ScriptValidationTest` on TeamCity
- [Postgres](https://teamcity.labkey.org/buildConfiguration/LabKey_2211Release_Community_BvtAPostgres/2193835)
- [SQL Server](https://teamcity.labkey.org/buildConfiguration/LabKey_2211Release_Community_BvtASqlserver/2193839)

bbimber:
Thanks @labkey-tchad. I think I wrote my summary while you were replying. I also think this behavior is old. It's arguably not consistent, but I dont think I'm going to take time to fix it. I think these issues are exposed primarily when working on code that uses the JSON directly, which I think includes Ext4 stores, SaveRows directly, etc. I recall the javascript LABKEY.SelectRows being case-sensitive, but I dont need to belabor that here. If tests pass I'm ok being done with this.

labkey-tchad (changes requested):
This is failing on postgres

bbimber:
> This is failing on postgres

am i missing something or did it pass? github's UI is showing a pass.

labkey-tchad:
> > This is failing on postgres
> 
> am i missing something or did it pass? github's UI is showing a pass.

Those are the normal BVTs, which don't run this test. The runs I linked above show the failure.

bbimber:
ok, got it @labkey-tchad. I dont currently see a reason in the code to explain a difference in postgres vs sqlserver. would you be able to push the commit i just added?

bbimber:
@labkey-tchad: i'm throwing in the towel. Added the CaseInsensitiveHashMap to the test, which I think is OK here. 

after thinking, the SelectRows response is the most critical for case-sensitivity and JS code. EHR does consume the response from SaveRows/insertWithKeys, but nothing detectable has changed in this codepath in a long time.

labkey-tchad:
Verifying the latest change:
- [Postgres](https://teamcity.labkey.org/buildConfiguration/LabKey_2211Release_Community_BvtAPostgres/2195172)
- [SQL Server](https://teamcity.labkey.org/buildConfiguration/LabKey_2211Release_Community_BvtASqlserver/2195174)

labkey-tchad (approved):
Test is passing now on postgres and SQL Server

bbimber:
Thanks @labkey-tchad. Unfortunately I can't actually do the squash/merge on this repo. Any chance someone could?

PR #1344, opened 2023-01-03
Make Crawler handle anchor navigation

labkey-tchad:
#### Rationale
Some non-app navigation doesn't trigger a page load.
For example, when navigating from `query-begin.view` to `query-begin.view#sbh-ssp-coreTarget`

#### Related Pull Requests
* N/A

#### Changes
* Make Crawler handle anchor navigation.


labkey-chrisj (approved):
ooooh, nice

PR #1346, opened 2023-01-04
Make memCheck see leaks during startup

labkey-tchad:
#### Rationale
`BWDT.checkLeaks` can't catch leaks that were generated during initial startup. The previous solution to avoid failing numerous tests due to a single leak was unnecessarily elaborate. This modifies `checkLeaks` to keep track of which leaks it has already seen. This also makes it unnecessary for `JUnitHeader` and `JUnitFooter` to jump through hoops to check for leaks from server-side tests.

#### Related Pull Requests
* N/A

#### Changes
* Make memCheck see leaks during startup


labkey-danield (approved):
I don't have any comments/questions. It looks ok to me.

PR #1354, opened 2023-01-12
Add basic tests for links to and from background imports grid

labkey-susanh:
#### Rationale
In the related PRs, updates were made to our viewing and linkage of the server notification menu and page. We added some basic test coverage there as well that visits the imports page and status page to assure we stay in the application instead of linking to LKS.  Thus the pages are needed by both the LKB and LKSM tests.

#### Related Pull Requests
- https://github.com/LabKey/biologics/pull/1860
- https://github.com/LabKey/inventory/pull/689
- https://github.com/LabKey/labkey-ui-components/pull/1082
- https://github.com/LabKey/sampleManagement/pull/1543

#### Changes
* Move ImportsPage from sampleManagement to here
* Move StatusPage from sampleManagement to here
* Add `clickViewAll` to `ServerNotificationMenu`

### DIFF
--- a/src/org/labkey/test/components/ui/pipeline/ImportsPage.java
+++ b/src/org/labkey/test/components/ui/pipeline/ImportsPage.java
@@ -37,11 +37,6 @@ public class ImportsPage extends LabKeyPage<LabKeyPage<?>.ElementCache>
                 2_500);
     }
 
-    public static ImportsPage beginAt(WebDriverWrapper webDriverWrapper)
-    {
-        return beginAt(webDriverWrapper, webDriverWrapper.getCurrentContainerPath());
-    }
-
     public static ImportsPage beginAt(WebDriverWrapper webDriverWrapper, String containerPath)
     {
         if (StringUtils.isBlank(containerPath) || "home".equalsIgnoreCase(containerPath))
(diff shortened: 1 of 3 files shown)


src/org/labkey/test/components/ui/pipeline/ImportsPage.java
```diff
@@ -0,0 +1,96 @@
...
+                },
+                "The 'Background Imports' page did not load in time.",
+                2_500);
+    }
+
+    public static ImportsPage beginAt(WebDriverWrapper webDriverWrapper)
+    {
+        return beginAt(webDriverWrapper, webDriverWrapper.getCurrentContainerPath());
+    }
+
```
labkey-danield: I've gotten feedback from Trey that beginAt should not have this overloaded. We should always have the container path provided (like below).
I suspect this is the implementation in LKSM, but I have been trying to update the various page objects when I come across them.

labkey-danield (approved):

src/org/labkey/test/components/ui/pipeline/ImportsPage.java
labkey-susanh: I don't quite understand why this overload is a bad idea, but I've removed it by essentially inlining what the method does.

PR #1355, opened 2023-01-13
Wait for file browser toolbar before clicking buttons

labkey-tchad:
#### Rationale
The button toolbar seems to take a moment to be added when tied to an S3 file root. Try to wait for it to appear.

#### Related Pull Requests
* N/A

#### Changes
* Wait for file browser toolbar before clicking/finding buttons


labkey-tchad:
Retargeting to 23.1, where the test is failing.
#1356

PR #1360, opened 2023-01-18
Update PipelineTriggerWizard to use new name for radio buttons.

labkey-danield:
#### Rationale
Update PipelineTriggerWizard to use new name for radio buttons.
Opened test issue [47109: Need to add test cases for "Update" option for PipelineTrigger](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47109) to make track the automation gap for the update scenarios.

#### Related Pull Requests
* None

#### Changes
* Update PipelineTriggerWizard to use new name for radio buttons.

### DIFF
--- a/src/org/labkey/test/components/pipeline/PipelineTriggerWizard.java
+++ b/src/org/labkey/test/components/pipeline/PipelineTriggerWizard.java
@@ -150,12 +150,8 @@ public class PipelineTriggerWizard extends WebDriverComponent<PipelineTriggerWiz
 
     public PipelineTriggerWizard setAction(String action)
     {
-        if (action.equals("Merge"))
-            elementCache().action.index(0).findElement(this).click();
-        else if (action.equals("Update"))
-            elementCache().action.index(1).findElement(this).click();
-        else
-            elementCache().action.index(2).findElement(this).click(); // Append
+        Locator.radioButtonByNameAndValue("insertOption", action)
+                .findElement(this).click();
         return this;
     }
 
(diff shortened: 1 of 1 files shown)


XingY (commented):
The changes made are applicable for "Import Samples from data file" only. But for "Reload lists using data file", the old code applies. Will this test code change break lists reload?

labkey-danield:
Good point @XingY. I only ran the specific tests that failed. I'll kick off more test runs and see what happens.
If this change does not break any other tests, which would imply there is an automation gap, I would be incline to check in the fix the get the existing tests to pass. Later when new tests are added PipelineTriggerWizard.java can be updated as needed.
If this change does break other tests then I will revisit and update my fix.

labkey-danield:
@XingY , @labkey-tchad had a great suggestion on improving the finder for the radio buttons that I believe will work with the "Import Samples from data file" and "Reload lists using data file" scenarios. 
There is still a test automation gap but I think that is better addressed in a different change.

XingY:
> @XingY , @labkey-tchad had a great suggestion on improving the finder for the radio buttons that I believe will work with the "Import Samples from data file" and "Reload lists using data file" scenarios. There is still a test automation gap but I think that is better addressed in a different change.

For the radio group id, samples is updated to use "insertOption" but list continue to use "mergeOption". I don't think the change works for list.

labkey-tchad (approved):
Looks good.
Is the `value` of the radio button not case-sensitive for that Locator? The failure snapshots had all-caps values (e.g. `value="MERGE"`)

labkey-tchad:
> Looks good. Is the `value` of the radio button not case-sensitive for that Locator? The failure snapshots had all-caps values (e.g. `value="MERGE"`)

Oh, I see, you changed the callers to pass in all-caps.

PR #1369, opened 2023-02-02
Handle update and clarify merge - LKS

XingY:
#### Rationale
The bulk update page for lists, datasets, samples and dataclasses now support an explicit UPDATE option.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4089
* https://github.com/LabKey/testAutomation/pull/1369

#### Changes
* update ImportDataPage to reflect the new update/merge UI

### DIFF
--- a/src/org/labkey/test/pages/ImportDataPage.java
+++ b/src/org/labkey/test/pages/ImportDataPage.java
@@ -23,6 +23,7 @@ import org.labkey.test.Locator;
 import org.labkey.test.Locators;
 import org.labkey.test.components.ext4.Checkbox;
 import org.labkey.test.components.ext4.ComboBox;
+import org.labkey.test.components.ext4.RadioButton;
 import org.labkey.test.util.Ext4Helper;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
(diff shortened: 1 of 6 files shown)


src/org/labkey/test/pages/ImportDataPage.java
```diff
@@ -173,13 +253,21 @@ public ElementCache()
         WebElement uploadFileDiv = Locator.byClass("panel-body").findWhenNeeded(uploadPanel);
         WebElement uploadExpando = Locator.byClass("lk-import-expando").findWhenNeeded(uploadPanel);
         WebElement uploadFileFilePath = Locator.input("file").findWhenNeeded(uploadFileDiv);
+        RadioButton addRowsFile = RadioButton.RadioButton().withLabelContaining("Add").findWhenNeeded(uploadPanel);
+        RadioButton updateRowsFile = RadioButton.RadioButton().withLabelContaining("Update").findWhenNeeded(uploadPanel);
+        Checkbox mergeFile = Ext4Checkbox().withLabelContaining("Allow new").findWhenNeeded(uploadPanel);
+        Checkbox mergeFileOptional = Ext4Checkbox().withLabelContaining("Allow new").findOrNull(uploadPanel);
```
labkey-danield, line 259: It is not a big deal, and entirely up to you if you want to change it, but if you call findWhenNeeded or refindWhenNeeded you can check for visibility of the component which would then change isFileMergeOptionPresent to:

```
    public boolean isFileMergeOptionPresent()
    {
        return elementCache().mergeFileOptional.isVisible();
    }
```
One thing using findWhenNeeded does help with is if in the future the the code is changed to hide or show the control through css attributes this will still work. Using findOfNull would find a control this is present but not visible.

labkey-danield (approved):

src/org/labkey/test/pages/ImportDataPage.java
labkey-tchad: @labkey-danield I think you meant `elementCache().mergeFile.getComponentElement().isVisible()`.
The `Checkbox` component doesn't have an `isVisible` method.

labkey-danield: I keep forgetting that.
In that case I wouldn't change it from the findOrNull implementation.

PR #1374, opened 2023-02-09
Updated Test Components for Perf Tests

labkey-danield:
#### Rationale
I needed to make some minor changes to accommodate the UI Perf tests.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/1552

#### Changes
* Added a getter for the grid Pager dropdown menu. The perf test uses it to get a reference to the menu item it needs to click.
* Use a more generic class name in the finder for the success button on the EntityBulkUpdateDialog. It wasn't finding the button when doing a bulk update from the 'Add New' page.
* Added DataClassAPIHelper.getRowIdsForDataEntity. Used to get the row id of an entity in a data class, specifically the row id of a source.

### DIFF
--- a/src/org/labkey/test/util/exp/DataClassAPIHelper.java
+++ b/src/org/labkey/test/util/exp/DataClassAPIHelper.java
@@ -1,6 +1,11 @@
 package org.labkey.test.util.exp;
 
+import org.junit.Assert;
 import org.labkey.remoteapi.CommandException;
+import org.labkey.remoteapi.Connection;
+import org.labkey.remoteapi.query.Filter;
+import org.labkey.remoteapi.query.SelectRowsCommand;
+import org.labkey.remoteapi.query.SelectRowsResponse;
 import org.labkey.test.WebTestHelper;
 import org.labkey.test.params.FieldDefinition;
 import org.labkey.test.params.experiment.DataClassDefinition;
@@ -9,7 +14,11 @@ import org.labkey.test.util.TestDataGenerator;
 
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 public class DataClassAPIHelper
 {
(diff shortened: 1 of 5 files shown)


labkey-susanh (approved):

src/org/labkey/test/util/exp/DataClassAPIHelper.java
```diff
@@ -51,4 +60,31 @@ public static List<FieldDefinition> dataClassTestFields()
                 new FieldDefinition("attachmentColumn", FieldDefinition.ColumnType.Attachment));
     }
 
+    public static Map<String, Integer> getRowIdsForDataEntity(String containerPath, String dataTypeName, List<String> entityNames) throws IOException, CommandException
```
labkey-tchad: Try to use consistent terminology. "Data Class" or "Source" are the widely used terms for things in `exp.data` (we might want to do some standardization across tests, similar to the change from "Sample Set" to "Sample Type").
I'd also name this method to better match the equivalent method for sample types (`SampleTypeAPIHelper.getRowIdsForSamples`). Maybe `getRowIdsForSources`? I think "entity" is a term that can refer to samples and sources.

labkey-tchad (approved):

src/org/labkey/test/util/exp/DataClassAPIHelper.java
```diff
@@ -51,4 +60,31 @@ public static List<FieldDefinition> dataClassTestFields()
...
+        for(Map<String, Object> row : response.getRows())
+        {
+            Object name = row.get("Name");
+            Object value = row.get("RowId");
+            rowIds.put(name.toString(), Integer.parseInt(value.toString()));
+        }
+
+        // Check that the names returned from the query match the names sent in.
+        Set<String> names = new HashSet<>(entityNames);
+        Assert.assertTrue("The names of the data entities returned from the query do not match the names sent in.",
```
labkey-jeckels: Should they be an exact match? Or is one a superset of the other? If they're the same, doing an Assert.assertEquals() has the nice benefit of giving more detail about their contents when the check fails.

labkey-jeckels (approved):

src/org/labkey/test/util/exp/DataClassAPIHelper.java
labkey-danield: Updated to explicitly refer to sources and source types and changed the name to your suggestion.

src/org/labkey/test/util/exp/DataClassAPIHelper.java
labkey-danield: Good point, they should be  identical. Changed the check to assertEquals.

PR #1393, opened 2023-02-24
Issue 47280: LKSM: Trailing whitespace in Source name won't resolve when deriving samples

XingY:
#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4150
* https://github.com/LabKey/testAutomation/pull/1393

#### Changes
* add test coverage for inserting and importing samples with trailing spaces in name and other fields

### DIFF
--- a/src/org/labkey/test/tests/SampleTypeTest.java
+++ b/src/org/labkey/test/tests/SampleTypeTest.java
@@ -170,7 +170,7 @@ public class SampleTypeTest extends BaseWebDriverTest
         sampleTypeHelper.verifyDataValues(data);
     }
 
-    // Issue 47280: LKSM: Trailing whitespace in Source name won't resolve when deriving samples
+    // Issue 47280: LKSM: Trailing/Leading whitespace in Source name won't resolve when deriving samples
     @Test
     public void testImportSamplesWithTrailingSpace()
     {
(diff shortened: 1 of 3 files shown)


src/org/labkey/test/tests/SampleTypeTest.java
```diff
@@ -170,6 +170,51 @@ public void testCreateSampleTypeNoExpression()
...
+
+        log("Create a new sample type with no name expression");
+        projectMenu().navigateToFolder(PROJECT_NAME, FOLDER_NAME);
+        SampleTypeHelper sampleTypeHelper = new SampleTypeHelper(this);
+        sampleTypeHelper.createSampleType(sampleTypeDefinition);
+        sampleTypeHelper.goToSampleType(sampleTypeName);
+
+        log("Add a single row to the sample type, with trailing spaces");
+        Map<String, String> fieldMap = Map.of("Name", "S-1 ", "StringCol", "Ess ", "IntCol", "1 ");
+        sampleTypeHelper.insertRow(fieldMap);
```
labkey-susanh, line 194: Should we also test for leading spaces while we're at it?  

How about a test for renaming that adds a space?

labkey-susanh (approved):

PR #1394, opened 2023-02-24
Teach security API test about SampleWorkflowDeletePermission

labkey-tchad:
#### Rationale
The Editor role now has `SampleWorkflowDeletePermission`. The test needs to know to expect it.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4140

#### Changes
* Teach security API test about SampleWorkflowDeletePermission


cnathe (approved):
Thanks Trey. Sorry I missed that with my changes from last week.

PR #1395, opened 2023-02-24
Issue 47356: Regression coverage

labkey-nicka:
#### Rationale
This adds regression coverage for [Issue 47356](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47356).

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4151

#### Changes
* Add test case that renames a list and expects to return to a valid page.


labkey-chrisj (approved):
Nice, thanks for doing this

PR #1412, opened 2023-03-13
Make `signIn` work better on CAS-only servers

labkey-tchad:
#### Rationale
`signOutHTTP` uses the wrong session info when the server auto-redirects to CAS login, resulting in a 401 response. This only happens when the user isn't even logged in so signing out is unnecessary.

#### Related Pull Requests
* N/A

#### Changes
* Don't call `simpleSignOut` unnecessarily
* Use authenticated connection for enabling JS error checker


labkey-tchad:
This needs a fix in 22.11: https://github.com/LabKey/testAutomation/pull/1413

PR #1414, opened 2023-03-13
Update test helpers for compliance functionality

labkey-tchad:
#### Rationale
https://github.com/LabKey/platform/issues/4120

#### Related Pull Requests
* https://github.com/LabKey/compliance/pull/181
* https://github.com/LabKey/complianceActivities/pull/19
* https://github.com/LabKey/OConnorLabModules/pull/101

#### Changes
* Improve functionality of `AuditLogHelper` and `ShowAuditLogPage`
  * `containerPath` is not relevant, rowIds are global
* Unify compliance settings pages
  * Move `Accounts`, `Audit`, and `Login` tabs under `BaseComplianceSettingsPage`
  * Delete `BaseComplianceSettingsTab` and `ComplianceSettingsPage`
* Make Test API helpers able to handle PHI settings for fields
* Create test page for testing `executeSql` API results (`ExecuteSqlTestWiki.java` and `executeSql.html`)
* Make domain designer helper able to set advance field options more efficiently


src/org/labkey/test/components/domain/AdvancedFieldSetting.java
labkey-danield, line 1: Nice clean up.

src/org/labkey/test/pages/DatasetInsertPage.java
```diff
@@ -81,12 +83,17 @@ public void insert(Map<String, String> values, boolean expectSuccess, String err
...
+            assertElementNotPresent(Locators.labkeyError);
+        }
+        else if (!StringUtils.isBlank(errorMsg))
+        {
+            assertTextNotPresent(errorMsg);
+        }
+        else
+        {
+            assertElementPresent(Locators.labkeyError);
         }
```
labkey-danield, line 97: It looks like there is one call that expects a success and looks for an error message:
`AbstractStudyTimeKeyFieldTest.testCannotTurnOffExtraTimeKeyIfViolatesUnique`[185]
I'm not sure I understand how that could be. If that is still a valid test this the error message won't be validated.

Would it be worthwhile to follow the design patterns of having two methods `insert` and `insertExpectingError`? Break out most of this method to a private method that both would call, and then `insert` would look for a success message and `insertExpectingError` would look for an error message. (Maybe beyond the scope of this change).

src/org/labkey/test/pages/ExecuteSqlTestWiki.java
```diff
@@ -0,0 +1,96 @@
...
+    {
+        super(driver);
+    }
+
+    public static ExecuteSqlTestWiki beginAt(WebDriverWrapper webDriverWrapper, String containerPath)
+    {
+        webDriverWrapper.beginAt(WebTestHelper.buildURL("wiki", containerPath, "page", Map.of("name", wikiName)));
+        return new ExecuteSqlTestWiki(webDriverWrapper.getDriver());
+    }
+
```
labkey-danield, line 40: Do you need a `waitForPage`?

src/org/labkey/test/pages/compliance/ComplianceSettingsAuditPage.java
```diff
@@ -0,0 +1,87 @@
...
+    public void assertRadiosDisabled()
+    {
+        assertFalse("Radio button is enabled when it should not be", elementCache().allRadio.isEnabled());
+        assertFalse("Radio button is enabled when it should not be", elementCache().primaryRadio.isEnabled());
+    }
+
+    public void assertRadiosEnabled()
+    {
+        assertTrue("Radio button is enabled when it should not be", elementCache().allRadio.isEnabled());
+        assertTrue("Radio button is enabled when it should not be", elementCache().primaryRadio.isEnabled());
```
labkey-danield: The failure message doesn't match the method name/intent.

labkey-danield (approved):

src/org/labkey/test/pages/DatasetInsertPage.java
labkey-tchad: This change should make that error check more general. Previously, it would only fail if that exact specified error appeared; now it will fail with any error.

src/org/labkey/test/pages/ExecuteSqlTestWiki.java
labkey-tchad: This is a static HTML page. Nothing dynamic to wait for.

PR #1416, opened 2023-03-14
Wait for dropdown in the editableGrid

labkey-danield:
#### Rationale
In the editable grid the dropdown can take a moment to show up, most likely because a query is used to populate the grid. Some tests in the inventory module have started failing because the dropdown is taking a little longer to show. This adds a waitFor with a 1/2 timeout for the dropdown to show up.

#### Related Pull Requests
* TBD

#### Changes
* Add a waitFor in the editable grid to wait for the dropdown.


labkey-tchad (approved):
Seems reasonable, assuming this method doesn't get called too much.

labkey-susanh (approved):

PR #1422, opened 2023-03-16
Regression test for Issue 47504

labkey-danield:
#### Rationale
Regression test for Issue 47504

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4206

#### Changes
* Regression test for Issue 47504.

### DIFF
--- a/src/org/labkey/test/tests/SampleTypeNameExpressionTest.java
+++ b/src/org/labkey/test/tests/SampleTypeNameExpressionTest.java
@@ -37,13 +37,17 @@ import org.labkey.test.util.PortalHelper;
 import org.labkey.test.util.SampleTypeHelper;
 import org.labkey.test.util.TestDataGenerator;
 import org.labkey.test.util.exp.SampleTypeAPIHelper;
+import org.openqa.selenium.TimeoutException;
 import org.openqa.selenium.WebElement;
 
 import java.io.IOException;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/SampleTypeNameExpressionTest.java
```diff
@@ -176,6 +180,105 @@ public void testSimpleNameExpression()
...
+        var importDataPage = dataRegion.clickImportBulkData();
+        importDataPage.setCopyPasteInsertOption(false);
+        importDataPage.selectCopyPaste()
+                .setFormat(ImportDataPage.Format.TSV)
+                .setText(pasteData)
+                .submit();
+
+        // Get the date as soon after sample creation as possible.
+        LocalDateTime ldt = LocalDateTime.now();
+        String date = DateTimeFormatter.ofPattern(dateFormat, Locale.ENGLISH).format(ldt);
```
labkey-susanh, line 265: I predict some random failures at some point in the future when the test spans the midnight hour, but I hope that should be pretty rare.

labkey-susanh (approved):

src/org/labkey/test/tests/SampleTypeNameExpressionTest.java
labkey-danield: Yep, and I had thought about removing this from the test. However, there are no test that uses the 'now' option in the name expression, so I thought the coverage would be good. 
If/when it does fail because of that hopefully it will be obvious.
If it happens a lot, I'll remove it.

PR #1423, opened 2023-03-17
Remove invalid lookup from LabkeyDemoStudy

labkey-martyp:
#### Rationale
With the related change, the domain editor will throw an error and not allow saving if there is an invalid lookup. The only sample data I found with an invalid lookup that caused issues in tests is LuminexAssay dataset field AnalyteIsotype in LabkeyDemoStudy. I removed the fk from the datasets_metadata and rezipped the study.

Also added invalid domain lookup test.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/1141

#### Changes
* Update AnalyteIsotype fk and rezip sample study
* Add testInvalidDomainFieldLookup test


labkey-tchad (approved):

labkey-sweta (approved):

labkey-martyp (commented):
Looks good Sweta

PR #1424, opened 2023-03-17
Fix Name Expression Test

labkey-danield:
#### Rationale
Needed to escape unicode characters so comparison works as expected on Windows.

#### Related Pull Requests
* None

#### Changes
* Escape unicode characters.
* Add a screenshot on error.


labkey-danield:
Verified passing on TeamCity ([MSSQL/Windows](https://teamcity.labkey.org/buildConfiguration/LabKey_233Release_Community_DailySuites_DailyBSqlserver/2338803?buildTab=tests))

labkey-sweta (approved):

PR #1453, opened 2023-03-30
Make Helpers More Reliable

labkey-danield:
#### Rationale
The FindSamplesById dialog was returning immediately after clicking the dismiss button. Some tests would immediately look at the results grid before before it was populated. This adds a waitForClose which will wait for the dialog to close before returning to the test.

#### Related Pull Requests
* None

#### Changes
* Add waitForClose to the dismiss methods.


labkey-chrisj (approved):
Thanks for doing this

PR #1430, opened 2023-03-21
Make `ReclickingWebElement.revealElement` more strict

labkey-tchad:
#### Rationale
`ReclickingWebElement.revealElement` was attempting to handle too many scenarios, which made it fail in some scenarios where it should be able to succeed (e.g. tooltips).

#### Related Pull Requests
* N/A

#### Changes
* Wait for tooltips to disappear (not just stop animating)

PR #1459, opened 2023-04-03
ModalDialog dismiss() option that allows for passing through a waitSeconds param

cnathe:
#### Rationale
The AssayQCTest.testQCStateChangeRequiresComment test failure was from a recent change in the ModalDialog related to the dismiss method call to waitForClose(). This PR adds a method signature that allows for passing through the `waitSeconds` param to `dismiss()`.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/1430/files#diff-34ea1f18083c4eea07a452968d844878119d228d63a1bedec19ec2ed4f2b64beR132

#### Changes
* adds a method signature that allows for passing through the `waitSeconds` param to `dismiss()`

### DIFF
--- a/src/org/labkey/test/components/bootstrap/ModalDialog.java
+++ b/src/org/labkey/test/components/bootstrap/ModalDialog.java
@@ -100,11 +100,16 @@ public class ModalDialog extends WebDriverComponent<ModalDialog.ElementCache>
 
     public void dismiss()
     {
+        dismiss(10);
+    }
+
+    public void dismiss(Integer waitSeconds)
+    {
         WebElement button = Locator.tagWithClass("button", "close").findElement(getComponentElement());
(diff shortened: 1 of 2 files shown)


labkey-danield (approved):
I'm ok with this, but I could see an argument that it could just be a test change. The test could call the `dismiss(String buttonText, Integer waitSeconds)` method.

cnathe:
> I'm ok with this, but I could see an argument that it could just be a test change. The test could call the `dismiss(String buttonText, Integer waitSeconds)` method.

I tried using that dismiss(buttonText, waitSeconds) method, but there isn't actually a button with text on this alert dialog. There is a "x" close in the upper right hand corner, but it didn't work to pass in the "x" as the button text. This currently used version of the dismiss() uses `Locator.tagWithClass("button", "close")` to click on that close "x".

PR #1461, opened 2023-04-04
regression coverage for issue 46738

labkey-chrisj:
#### Rationale
This change adds test coverage to ensure navigation between Admin Console, Site Users and the user domain editor.  In some deployments, links to the user domain editor page show 404s, this should let us know when we break that link

#### Related Pull Requests
n/a

#### Changes

- [x] New method in ShowAdminPage, supporting navigation to user fields editor
- [x] New test method in AdminConsoleTest, exercising this navigation and ensuring the page is functional

### DIFF
--- a/src/org/labkey/test/pages/core/admin/ShowAdminPage.java
+++ b/src/org/labkey/test/pages/core/admin/ShowAdminPage.java
@@ -18,6 +18,7 @@ package org.labkey.test.pages.core.admin;
 import org.labkey.test.Locator;
 import org.labkey.test.WebDriverWrapper;
 import org.labkey.test.WebTestHelper;
+import org.labkey.test.components.DomainDesignerPage;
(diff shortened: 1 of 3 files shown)


labkey-tchad (changes requested):
Adam wanted to test on different deployments. Let's put this into a separate test class so that we can add it to other suites (like the trial suites).

labkey-chrisj:
Can you suggest particular suites to add this test class to?  Like maybe Biologics and SampleManagement?

labkey-tchad (approved):
Yeah. You can add it to the `Daily` suite, as well as the trial AMI suites for [SM](https://github.com/LabKey/sampleManagement/blob/develop/test/sampledata/suites/trialAmiSampleManager.txt) and [Biologics](https://github.com/LabKey/biologics/blob/develop/test/sampledata/suites/trialAmiBiologics.txt).

PR #1463, opened 2023-04-05
Domain designer loading state

labkey-martyp:
#### Rationale
Latest update to domain designer slowed down initialization enough that automated tests were adding fields before the component was completely initialized. The related PR updates the domain designer to have a loading state which prevents the tests from adding fields before initialization is done.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/1165

#### Changes
* When getting the field panel in DomainDesignerPage and EditMetaDataPage, wait for the Loading... spinner to go away first.

### DIFF
--- a/src/org/labkey/test/components/DomainDesignerPage.java
+++ b/src/org/labkey/test/components/DomainDesignerPage.java
@@ -9,7 +9,6 @@ import org.labkey.test.components.domain.BaseDomainDesigner;
 import org.labkey.test.components.domain.DomainFormPanel;
 import org.labkey.test.components.domain.DomainPanel;
 import org.labkey.test.components.domain.UnsavedChangesModalDialog;
-import org.labkey.test.components.html.Checkbox;
 import org.labkey.test.util.Maps;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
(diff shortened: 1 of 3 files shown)


labkey-tchad (approved):

labkey-tchad (commented):
Actually. Could you put this fix in 23.4? Not a huge deal if that would be a hassle.

labkey-martyp:
> Actually. Could you put this fix in 23.4? Not a huge deal if that would be a hassle.

@labkey-tchad It would be a little time consuming to backport the domain designer part of this fix since it is an update in @labkey/components. We'd have to backport it to a previous package version as well as the latest version. If it's not a big deal, it would be a lot less time to stick with develop. Let me know if that is going to be a problem leaving the tests flaky on 23.4 and getting the fix in develop.

labkey-tchad:
> > Actually. Could you put this fix in 23.4? Not a huge deal if that would be a hassle.
> 
> @labkey-tchad It would be a little time consuming to backport the domain designer part of this fix since it is an update in @labkey/components. We'd have to backport it to a previous package version as well as the latest version. If it's not a big deal, it would be a lot less time to stick with develop. Let me know if that is going to be a problem leaving the tests flaky on 23.4 and getting the fix in develop.

That's not a problem. I'll mute the failure in 23.4.

labkey-sweta (approved):

PR #1465, opened 2023-04-05
Export system diagnostics after running tests against a remote server

labkey-tchad:
#### Rationale
Tests don't have file-system access to remote servers, so we can't dump tomcat logs to TeamCity. We can, however, export the server's system diagnostics. The diagnostics includes server logs as well as other, possibly useful, information.
This adds a new pseudo-test (`ExportDiagnosticsPseudoTest`) to download these diagnostics after running tests against a remote server (or via a new `webtest.enable.export.diagnostics` flag).

This is the third test that we want to force to run in a particular order (in addition to `BasicTest` and `DatabaseDiagnosticsTest`). Rather than adding more special handling for the new test (it should run last), I created a `Prioritized` annotation for tests to specify that they should run before/after other tests.

#### Related Pull Requests
* N/A

#### Changes
* Export system diagnostics after running tests against a remote server
* Create new annotation control test priority
* Add helper to download directly to a file

### DIFF
--- a/src/org/labkey/test/TestSet.java
+++ b/src/org/labkey/test/TestSet.java
@@ -17,10 +17,12 @@
 package org.labkey.test;
 
 import org.jetbrains.annotations.NotNull;
+import org.labkey.test.util.Prioritized;
 
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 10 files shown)


labkey-tchad:
Here's a trial run on TeamCity: https://teamcity.labkey.org/buildConfiguration/LabKey_233Release_Internal_CloudInfrastructure_TrialLimsIntegrationTest/2375671?buildTab=artifacts#%2Fdiagnostics_2023-04-05_10-37-18.zip

labkey-tchad:
Also testing with an embedded server (which doesn't have a tomcat log directory): https://teamcity.labkey.org/buildConfiguration/LabKey_233Release_Internal_Embedded_SampleManagerEmbeddedPostgres/2376430

src/org/labkey/test/TestSet.java
```diff
@@ -90,6 +92,40 @@ public List<Class<?>> getTestList()
...
+            }
+            else
+            {
+                unprioritized.add(next);
+            }
+        }
+        firstTests.sort(Comparator.comparingDouble(c -> c.getAnnotation(Prioritized.class).priority()));
+        Collections.reverse(firstTests);
+        lastTests.sort(Comparator.comparingDouble(c -> c.getAnnotation(Prioritized.class).priority()));
+        Collections.reverse(lastTests);
```
labkey-danield, line 120: I think I understand but want to make sure. The `Collections.reverse` is used to make sure that a test with priority 2, runs before 1 because 2 is a higher priority. Is that correct? 

If that is correct, do you want the reverse on the `lastTests`? Test in that list have a priority that is negative, and in that case wouldn't you want -1 to run before -2?

labkey-danield (approved):

src/org/labkey/test/TestSet.java
labkey-tchad: Yeah. The default comparator sorts [-2, -1], which is the reverse of what we want.

PR #1473, opened 2023-04-11
prevent test from cleaning up a project it never created

labkey-chrisj:
#### Rationale
`UserDomainEditNavigationTest` is failing because it assumes no cleanup is required for a project it didn't create. Unfortunately, the default implementation of cleanup (in `BaseWebDriverTest`) attempts to clean up the project.
This overrides cleanup to not do that in this test class

#### Related Pull Requests
https://github.com/LabKey/testAutomation/pull/1461


labkey-tchad (approved):
Just having `getProjectName` return `null` should be sufficient but no harm in overriding `doCleanup`

PR #1484, opened 2023-04-14
Test coverage for metadataQuery-sourceQuery workflow

labkey-chrisj:
#### Rationale
[Issue 47495](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47495) addresses customer confusion regarding the way metadata edits to a domain can be inadvertently saved, thus producing unintended updates to domain lookups

While testing this, I found what appears to be a related issue: unless the user resets to default, subsequent metadata edits appear to carry previous changes forward
For now the tests work around this by explicitly resetting to default after running

#### Related Pull Requests
n/a

#### Changes

- [x] New test class, QueryMetadataTest
- Updates to EditMetadataPage
     - [x] implement beginAt for this page/action
     - [x] clickViewSource now returns a sourceQueryPage
- Update to SourceQueryPage
     - [x] page now surfaces a way to get queryXml

### DIFF
--- a/src/org/labkey/test/WebDriverWrapper.java
+++ b/src/org/labkey/test/WebDriverWrapper.java
@@ -3994,6 +3994,11 @@ public abstract class WebDriverWrapper implements WrapsDriver
         _extHelper.setCodeMirrorValue(id, value);
     }
 
+    protected String getCodeEditorValue(String id)
+    {
+        return _extHelper.getCodeMirrorValue(id);
+    }
+
     public void waitForElements(final Locator loc, final int count)
     {
         waitForElements(loc, count, WAIT_FOR_JAVASCRIPT);
--- a/src/org/labkey/test/tests/query/QueryMetadataTest.java
+++ b/src/org/labkey/test/tests/query/QueryMetadataTest.java
@@ -99,7 +99,8 @@ public class QueryMetadataTest extends BaseWebDriverTest
                 "  </table>\n" +
                 "</tables>";
         var queryPage = editPage.clickEditSource();
-        checker().verifyEquals("expect xml to show only the delta for value description",
+        checker().withScreenshot("xml_mismatch")
+                .verifyEquals("expect xml to show only the delta for value description",
                 expectedXml, queryPage.getMetadataXml());
 
         // clean up after
@@ -126,7 +127,8 @@ public class QueryMetadataTest extends BaseWebDriverTest
                 "</tables>";
 
         var queryPage = editPage.clickEditSource();
-        checker().verifyEquals("expect xml to show only the delta for value description",
+        checker().withScreenshot("xml_mismatch")
+                .verifyEquals("expect xml to show only the delta for value description",
                 expectedXml, queryPage.getMetadataXml());
 
         // clean up after
@@ -189,9 +191,6 @@ public class QueryMetadataTest extends BaseWebDriverTest
             .wrapAssertion(()-> assertThat(actualAfterXml)
             .as("expect xml to show only the delta for value description")
             .isEqualTo(expectedAfterXml));
-
-        // clean up after
-        EditMetadataPage.beginAt(this, getProjectName(), "lists", TEST_LIST).resetToDefault();
     }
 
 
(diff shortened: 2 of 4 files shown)


src/org/labkey/test/tests/query/QueryMetadataTest.java
```diff
@@ -0,0 +1,210 @@
...
+                "      </column>\n" +
+                "    </columns>\n" +
+                "  </table>\n" +
+                "</tables>";
+        var queryPage = editPage.clickEditSource();
+        checker().verifyEquals("expect xml to show only the delta for value description",
+                expectedXml, queryPage.getMetadataXml());
+
+        // clean up after
+        EditMetadataPage.beginAt(this, getProjectName(), "lists", TEST_LIST).resetToDefault();
```
labkey-nicka: Instead of only doing this at the end of a successful test case I recommend doing this in an `afterEach()` (if that exists in junit) to ensure the metadata is reset. Otherwise, if that is not available, then move these clean up statements to the beginning of the test cases so they defend themselves against others failing.

src/org/labkey/test/tests/query/QueryMetadataTest.java
```diff
@@ -0,0 +1,210 @@
...
+                new FieldDefinition("users", FieldDefinition.ColumnType.User),
+                new FieldDefinition("date", FieldDefinition.ColumnType.DateAndTime),
+                new FieldDefinition("samples", new FieldDefinition.IntLookup(getProjectName(),"exp.materials", TEST_SAMPLES)),
+                new FieldDefinition("selfLookup", new FieldDefinition.IntLookup(getProjectName(),"lists", TEST_LIST)));
+        var dgen = new IntListDefinition(TEST_LIST, "Key").setFields(listColumns)
+                .create(createDefaultConnection(), getProjectName());
+    }
+
+    @Test
+    public void testUpdateLookupFields()
```
labkey-nicka, line 77: It would be helpful for these test cases to link out to specific Issues they're attempting to cover for regressions. Always find those useful.

src/org/labkey/test/tests/query/QueryMetadataTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+
+@Category({})
+public class QueryMetadataTest extends BaseWebDriverTest
+{
+    static public String TEST_LIST = "queryMetadataTestList";
+    static public String TEST_SAMPLES = "queryMetadataSamples";
+
+    @Override
+    protected void doCleanup(boolean afterTest)
```
labkey-nicka, line 26: nit: `BaseWebDriverTest` will do this cleanup for you. No need to override this method.

labkey-nicka (approved):

src/org/labkey/test/WebDriverWrapper.java
```diff
@@ -3994,6 +3994,11 @@ public void setCodeEditorValue(String id, String value)
         _extHelper.setCodeMirrorValue(id, value);
     }
 
+    protected String getCodeEditorValue(String id)
+    {
+        return _extHelper.getCodeMirrorValue(id);
+    }
+
```
labkey-tchad, line 4001: Use `CodeMirrorHelper` directly. I don't know why the ext helper had this functionality.
```suggestion
    protected String getCodeEditorValue(String id)
    {
        return new CodeMirrorHelper(this, id).getCodeMirrorValue();
    }

```
(Please update `WebDriverWrapper.setCodeEditorValue` in the same way)

src/org/labkey/test/tests/query/QueryMetadataTest.java
```diff
@@ -0,0 +1,214 @@
...
+        editPage.resetToDefault();
+        queryPage = editPage.clickEditSource();
+
+        String expectedAfterXml =
+                "<tables xmlns=\"http://labkey.org/data/xml\">\n" +
+                        "  <table tableName=\"queryMetadataTestList\" tableDbType=\"NOT_IN_DB\">\n" +
+                        "    <columns>\n" +
+                        "    </columns>\n" +
+                        "  </table>\n" +
+                        "</tables>\n";
```
labkey-tchad, line 193: Use text blocks for stuff like this. IntelliJ should offer to do the refactor automatically.
```suggestion
        String expectedAfterXml = """
                <tables xmlns="http://labkey.org/data/xml">
                  <table tableName="queryMetadataTestList" tableDbType="NOT_IN_DB">
                    <columns>
                    </columns>
                  </table>
                </tables>
                """;
```

labkey-tchad (approved):
Just some nits

PR #1488, opened 2023-04-18
Automation test for Clone permissions to an existing user 

labkey-sweta:
#### Rationale
Automation test for Clone permissions to an existing user 

#### Related Pull Requests
https://github.com/LabKey/platform/pull/4292

#### Changes
*

### DIFF
--- a/src/org/labkey/test/pages/user/ClonePermissionsPage.java
+++ b/src/org/labkey/test/pages/user/ClonePermissionsPage.java
@@ -41,6 +41,12 @@ public class ClonePermissionsPage extends LabKeyPage<ClonePermissionsPage.Elemen
         return new ShowUsersPage(getDriver());
     }
 
+    public String clonePermissionExpectingError()
+    {
+        elementCache().clonePermissionBtn.click();
+        return Locator.tagWithClass("div", "labkey-error").findElement(getDriver()).getText();
+    }
+
     public ShowUsersPage cancelPermission()
     {
         clickAndWait(elementCache().cancelBtn);
--- a/src/org/labkey/test/pages/user/ClonePermissionsPage.java
+++ b/src/org/labkey/test/pages/user/ClonePermissionsPage.java
@@ -1,0 +1,64 @@
+package org.labkey.test.pages.user;
+
+import org.labkey.test.Locator;
+import org.labkey.test.WebDriverWrapper;
+import org.labkey.test.WebTestHelper;
+import org.labkey.test.components.html.Input;
+import org.labkey.test.pages.LabKeyPage;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+import java.util.Map;
+
+public class ClonePermissionsPage extends LabKeyPage<ClonePermissionsPage.ElementCache>
+{
+    public ClonePermissionsPage(WebDriver driver)
+    {
+        super(driver);
+    }
+
+    public static ClonePermissionsPage beginAt(WebDriverWrapper webDriverWrapper, Integer userId)
+    {
+        Map<String, String> params = Map.of("userId", userId.toString(), "schemaName", "core", "query.queryName", "Users", "targetUser", userId.toString());
+        webDriverWrapper.beginAt(WebTestHelper.buildURL("security", "clonePermissions", params));
+        return new ClonePermissionsPage(webDriverWrapper.getDriver());
+    }
+
+    public String getWarningMessage()
+    {
+        return elementCache().warningMsg.getText();
+    }
+
+    public ClonePermissionsPage setCloneUser(String value)
+    {
+        elementCache().cloneUser.set(value);
+        return this;
+    }
+
+    public ShowUsersPage clonePermission()
+    {
+        clickAndWait(elementCache().clonePermissionBtn);
+        return new ShowUsersPage(getDriver());
+    }
+
+    public ShowUsersPage cancelPermission()
+    {
+        clickAndWait(elementCache().cancelBtn);
+        return new ShowUsersPage(getDriver());
+    }
+
+    @Override
+    protected ClonePermissionsPage.ElementCache newElementCache()
+    {
+        return new ClonePermissionsPage.ElementCache();
+    }
+
+    protected class ElementCache extends LabKeyPage<?>.ElementCache
+    {
+        WebElement warningMsg = Locator.xpath("//table/tbody/tr[3]/td").findWhenNeeded(this);
+        Input cloneUser = new Input(Locator.name("cloneUser").findWhenNeeded(this), getDriver());
+        WebElement clonePermissionBtn = Locator.lkButton("Clone Permissions").findWhenNeeded(this);
+        WebElement cancelBtn = Locator.lkButton("Cancel").findElement(this);
+
+    }
+}
--- a/src/org/labkey/test/tests/UserClonePermissionTest.java
+++ b/src/org/labkey/test/tests/UserClonePermissionTest.java
@@ -1,0 +1,92 @@
+package org.labkey.test.tests;
+
+import org.jetbrains.annotations.Nullable;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.labkey.test.BaseWebDriverTest;
+import org.labkey.test.categories.Daily;
+import org.labkey.test.pages.user.ClonePermissionsPage;
+import org.labkey.test.util.ApiPermissionsHelper;
+import org.labkey.test.util.PermissionsHelper;
+
+import java.util.Arrays;
+import java.util.List;
+
+@Category({Daily.class})
+@BaseWebDriverTest.ClassTimeout(minutes = 6)
+public class UserClonePermissionTest extends BaseWebDriverTest
+{
+    private static final String CLONED_SOURCE_SITE_USER = "sourcesiteuser@clonepermission.test";
+    private static final String CLONED_SOURCE_APP_USER = "sourceappuser@clonepermission.test";
+    private static final String CLONED_TARGET_SITE_USER = "targetsiteuser@clonepermission.test";
+    private static final String CLONED_TARGET_APP_USER = "targetappuser@clonepermission.test";
+    private static final String TEST_PASSWORD = "123456";
+    private static final String CLONE_GROUP = "Test clone group";
+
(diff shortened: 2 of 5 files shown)


labkey-adam (approved):

src/org/labkey/test/tests/UserClonePermissionTest.java
```diff
@@ -0,0 +1,107 @@
...
+        _userHelper.createUser(CLONED_SOURCE_SITE_USER);
+        _permissionsHelper.addUserToSiteGroup(CLONED_SOURCE_SITE_USER, "Site Administrators");
+
+        _userHelper.createUser(CLONED_SOURCE_APP_USER);
+        _permissionsHelper.addUserAsAppAdmin(CLONED_SOURCE_APP_USER);
+
+        goToProjectHome();
+        _permissionsHelper.createPermissionsGroup(CLONE_GROUP);
+        _permissionsHelper.setPermissions(CLONE_GROUP, "Reader");
+        _permissionsHelper.setPermissions(CLONE_GROUP, "Electronic Signer");
```
labkey-tchad: "Electronic Signer" requires the compliance module. Just use roles that are available in community.

src/org/labkey/test/tests/UserClonePermissionTest.java
```diff
@@ -0,0 +1,107 @@
...
+        _permissionsHelper.addUserToProjGroup(CLONED_SOURCE_SITE_USER, getProjectName(), CLONE_GROUP);
+        _permissionsHelper.addMemberToRole(CLONED_SOURCE_SITE_USER, "Author", PermissionsHelper.MemberType.user);
+
+        _permissionsHelper.addUserToSiteGroup(CLONED_SOURCE_SITE_USER, "Developers");
+
+        _userHelper.createUser(CLONED_TARGET_SITE_USER);
+        _permissionsHelper.addUserToSiteGroup(CLONED_TARGET_SITE_USER, "Site Administrators");
```
labkey-tchad: I don't think the target user should have any permissions to start with.

src/org/labkey/test/tests/UserClonePermissionTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+        _userHelper.createUser(CLONED_TARGET_APP_USER);
+    }
+
+    @Test
+    public void testSiteUserClonePermission()
+    {
+        impersonate(CLONED_TARGET_SITE_USER);
```
labkey-tchad: Use a separate user to perform the actions. The primary use case is an admin cloning permissions between non-admins. This test case can just run as the primary test user without impersonation.
If we want to test a user cloning somebody's permissions onto themself, that should be a separate test case, with a separate user.

src/org/labkey/test/tests/UserClonePermissionTest.java
```diff
@@ -0,0 +1,107 @@
...
+    private static final String CLONED_TARGET_APP_USER = "targetappuser@clonepermission.test";
+    private static final String CLONE_GROUP = "Test clone group";
+
+    ApiPermissionsHelper _permissionsHelper = new ApiPermissionsHelper(this);
+
+    @Override
+    protected @Nullable String getProjectName()
+    {
+        return getClass().getSimpleName() + " Project";
+    }
```
labkey-tchad, line 35: This should return `null` if the test doesn't create any projects.
```suggestion
    @Override
    protected @Nullable String getProjectName()
    {
        return null;
    }
```

src/org/labkey/test/tests/UserClonePermissionTest.java
```diff
@@ -0,0 +1,107 @@
...
+        Assert.assertEquals("Incorrect warning message",
+                "Warning! Cloning permissions will delete all group memberships and direct role assignments for targetsiteuser (targetsiteuser@clonepermission.test) and replace them with the group memberships and direct role assignments of the user selected below.",
+                clonePermissionsPage.getWarningMessage());
+
+        clonePermissionsPage.setCloneUser(CLONED_SOURCE_SITE_USER);
+        clonePermissionsPage.clonePermission();
+        stopImpersonating();
+
+        _permissionsHelper.assertPermissionSetting(CLONED_TARGET_SITE_USER, "Reader");
+        _permissionsHelper.assertPermissionSetting(CLONED_TARGET_SITE_USER, "Developers");
```
labkey-tchad: This should test that and previously assigned roles have been removed.

src/org/labkey/test/pages/user/ClonePermissionsPage.java
```diff
@@ -0,0 +1,64 @@
...
+    {
+        elementCache().cloneUser.set(value);
+        return this;
+    }
+
+    public ShowUsersPage clonePermission()
+    {
+        clickAndWait(elementCache().clonePermissionBtn);
+        return new ShowUsersPage(getDriver());
+    }
```
labkey-tchad, line 42: This should verify that there weren't any errors. Otherwise, tests will call this then fail on subsequent checks without much information. There should also be a matching `clonePermissionExpectingError()` that verifies that there **is** an error.

src/org/labkey/test/tests/UserClonePermissionTest.java
labkey-adam: It could have some permissions... and then verify that those were deleted.

src/org/labkey/test/tests/UserClonePermissionTest.java
labkey-adam: Agreed. Not important to test cloning permissions to yourself.

src/org/labkey/test/tests/UserClonePermissionTest.java
labkey-sweta: Changed

src/org/labkey/test/tests/UserClonePermissionTest.java
labkey-sweta: Done

src/org/labkey/test/tests/UserClonePermissionTest.java
labkey-sweta: I will be using the project created.

src/org/labkey/test/tests/UserClonePermissionTest.java
```diff
@@ -0,0 +1,179 @@
...
+
+        clonePermissionsPage = new ClonePermissionsPage(getDriver());
+        clonePermissionsPage.setCloneUser(CLONED_SOURCE_SITE_USER + " (" + SOURCE_SITE_USERNAME + ")");
+        clonePermissionsPage.clonePermission();
+
+        log("Verifying project level permission");
+        _permissionsHelper = new ApiPermissionsHelper(getProjectName());
+        _permissionsHelper.assertPermissionSetting(CLONED_TARGET_SITE_USER, "Author");
+        _permissionsHelper.assertPermissionSetting(CLONED_TARGET_SITE_USER, "Editor");
+        _permissionsHelper.assertNoPermission(CLONED_TARGET_SITE_USER, "Submitter");
```
labkey-adam, line 106: You're testing the permissions associated with `CLONE_GROUP`, which is good, but should test membership in that group as well.

src/org/labkey/test/tests/UserClonePermissionTest.java
```diff
@@ -0,0 +1,179 @@
...
+@BaseWebDriverTest.ClassTimeout(minutes = 3)
+public class UserClonePermissionTest extends BaseWebDriverTest
+{
+    private static final String CLONED_SOURCE_SITE_USER = "sourcesiteuser@clonepermission.test";
+    private static final String SOURCE_SITE_USERNAME = "sourcesiteuser";
+    private static final String CLONED_SOURCE_APP_USER = "sourceappuser@clonepermission.test";
+    private static final String CLONED_TARGET_SITE_USER = "targetsiteuser@clonepermission.test";
+    private static final String TARGET_SITE_USERNAME = "targetsiteuser";
+    private static final String CLONED_TARGET_APP_USER = "targetappuser@clonepermission.test";
+    private static final String TARGET_APP_USERNAME = "targetappuser";
```
labkey-tchad: Don't hard-code the user display names. They might not match what's here (usually due to collisions with other tests). `_userHelper.getDisplayNameForEmail` will get them when they are needed.

src/org/labkey/test/tests/UserClonePermissionTest.java
```diff
@@ -0,0 +1,179 @@
...
+
+        Assert.assertEquals("Incorrect warning message",
+                "Warning! Cloning permissions will delete all group memberships and direct role assignments for " + TARGET_APP_USERNAME + " (" +
+                        CLONED_TARGET_APP_USER + ") and replace them with the group memberships and direct role assignments of the user selected below.",
+                clonePermissionsPage.getWarningMessage());
+
+        clonePermissionsPage.setCloneUser(CLONED_SOURCE_APP_USER);
+        clonePermissionsPage.clonePermission();
+        stopImpersonating();
+
```
labkey-tchad: This should verify that an app admin cannot clone permissions from a site admin.

labkey-tchad (changes requested):
Missing an important scenario.

src/org/labkey/test/tests/UserClonePermissionTest.java
```diff
@@ -0,0 +1,179 @@
...
+        log("Verifying project level permission");
+        _permissionsHelper = new ApiPermissionsHelper(getProjectName());
+        _permissionsHelper.assertPermissionSetting(CLONED_TARGET_APP_USER, "Author");
+        _permissionsHelper.assertPermissionSetting(CLONED_TARGET_APP_USER, "Editor");
```
labkey-adam, line 159: Test group membership here as well

labkey-adam (changes requested):
Adding check for project group membership would be good

src/org/labkey/test/tests/UserClonePermissionTest.java
labkey-tchad: @labkey-sweta `_permissionsHelper.isUserInGroup` can do this check.

labkey-tchad:
@labkey-adam Should we test this for project/folder admins? They can also create users.

labkey-adam:
> @labkey-adam Should we test this for project/folder admins? They can also create users.

Project/folder admins can create users, but they can't clone permissions. Only site and app admins can do that.

labkey-adam (approved):

labkey-tchad (approved):

PR #1496, opened 2023-04-25
Export list archive automation test

labkey-sweta:
#### Rationale
Automation test for Issue 47289: Export List Archive if the user is an Admin of the folders of the selected Lists, else throw Permission error
[https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47289]

#### Related Pull Requests
* 

#### Changes
New test file : src/org/labkey/test/tests/ListArchiveExportTest.java

### DIFF
--- a/src/org/labkey/test/tests/ListArchiveExportTest.java
+++ b/src/org/labkey/test/tests/ListArchiveExportTest.java
@@ -29,7 +29,6 @@ public class ListArchiveExportTest extends BaseWebDriverTest
     private final static String LIST_FOLDER_A = "LIST_FOLDER_A";
     private final static String LIST_A = "List Export A";
     private final static String LIST_B = "List Export B";
-    private final static String SHARED_LIST = "Shared list C";
     private final static String _listUser = "listuser@listarchiveexport.test";
 
     @Override
@@ -79,6 +78,10 @@ public class ListArchiveExportTest extends BaseWebDriverTest
         _listHelper.insertNewRow(rowData);
     }
 
+    /*
+        Test coverage for
+        Issue 47289: Export List Archive if the user is an Admin of the folders of the selected Lists, else throw Permission error
+     */
     @Test
     public void testExportListArchive()
     {
@@ -86,6 +89,7 @@ public class ListArchiveExportTest extends BaseWebDriverTest
         impersonate(_listUser);
         ManageListsGrid listsGrid = goToManageLists().getGrid();
         listsGrid.setContainerFilter(DataRegionTable.ContainerFilterType.ALL_FOLDERS);
+        listsGrid.setFilter("Container", "Equals One Of (example usage: a;b;c)", LIST_FOLDER_A + ";" + getProjectName());
         listsGrid.checkAllOnPage();
         listsGrid.clickHeaderButton("Export List Archive");
         waitForElement(Locators.labkeyErrorHeading);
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/ListArchiveExportTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+@Category({Daily.class, Hosting.class})
+@BaseWebDriverTest.ClassTimeout(minutes = 2)
+public class ListArchiveExportTest extends BaseWebDriverTest
+{
+    private final static String LIST_SUBFOLDER = "List subfolder";
+    private final static String LIST_FOLDER_A = "LIST_FOLDER_A";
+    private final static String LIST_A = "List Export A";
+    private final static String LIST_B = "List Export B";
+    private final static String SHARED_LIST = "Shared list C";
```
labkey-bpatel: It looks like this variable is not being used

src/org/labkey/test/tests/ListArchiveExportTest.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    @Test
+    public void testExportListArchive()
+    {
+        goToProjectHome(LIST_FOLDER_A);
+        impersonate(_listUser);
+        ManageListsGrid listsGrid = goToManageLists().getGrid();
+        listsGrid.setContainerFilter(DataRegionTable.ContainerFilterType.ALL_FOLDERS);
+        listsGrid.checkAllOnPage();
```
labkey-bpatel, line 97: I wonder if there will be other Lists created via other tests that might interfere with this causing assertEquals() at line 92 to fail if the List selected is a different list from a different folder rather than LIST_B. 
I briefly looked at other List tests which are doing the cleanup of deleting the folder, but not sure whether other (non-List) tests containing Lists are doing the cleanup or not.

src/org/labkey/test/tests/ListArchiveExportTest.java
labkey-sweta: good catch

src/org/labkey/test/tests/ListArchiveExportTest.java
labkey-sweta: Every test clears itself out after it runs, unless failed on teamcity. 

Updated the test

labkey-bpatel (approved):

src/org/labkey/test/tests/ListArchiveExportTest.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    private void createListWithData(String name, Map rowData)
+    {
+        log("Creating test list in " + getCurrentContainerPath());
+        _listHelper.createList(getCurrentContainerPath(), name, ListHelper.ListColumnType.AutoInteger, "RowId",
+                new ListHelper.ListColumn("Shape", "Shape", ListHelper.ListColumnType.String),
+                new ListHelper.ListColumn("Count", "Count", ListHelper.ListColumnType.Integer));
+        _listHelper.insertNewRow(rowData);
+    }
```
labkey-tchad: Use API helpers to avoid a bunch of unnecessary UI interaction.
```suggestion
    private void createListWithData(String name, Map rowData)
    {
        var connection = createDefaultConnection();
        var listDef = new IntListDefinition(name, "RowId").setFields(List.of(
                new FieldDefinition("Shape", FieldDefinition.ColumnType.String),
                new FieldDefinition("Count", FieldDefinition.ColumnType.Integer));
        var dataGenerator = listDef.create(connection, getCurrentContainerPath());
        dataGenerator.insertRows(connection, List.of(rowData));
    }
```
Also, `ListHelper.ListColumn` is deprecated and should never be used.

labkey-tchad (changes requested):

labkey-tchad (approved):

PR #1499, opened 2023-05-01
Improve date comparison when test runs at midnight

labkey-tchad:
#### Rationale
The date used to verify the grid should be locked in when the data being checked is imported.

#### Related Pull Requests
* N/A

#### Changes
* Improve date comparison when test runs at midnight

### DIFF
--- a/src/org/labkey/test/tests/ExpTest.java
+++ b/src/org/labkey/test/tests/ExpTest.java
@@ -16,6 +16,7 @@
 
 package org.labkey.test.tests;
 
+import org.assertj.core.api.Assertions;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.labkey.test.BaseWebDriverTest;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/ExpTest.java
```diff
@@ -148,7 +149,7 @@ public void testSteps()
         _ext4Helper.clickExt4Tab("Data");
         waitForText(WAIT_FOR_JAVASCRIPT, "editedCreated");
         dateFormat = new SimpleDateFormat("ddd MMM dd yyyy");
```
labkey-chrisj, line 152: Ah, re-using dateFormat here was briefly confusing; I looked up the page and saw its initial declaration and wondered how it could match that date string with a format expression of yyyy-MM-dd
My thought was that the reason for the failure might have been an unexpected change in server date format

src/org/labkey/test/tests/ExpTest.java
```diff
@@ -175,7 +176,7 @@ public void testSteps()
         _customizeViewsHelper.addColumn("WrappedRowId/Created");
         _customizeViewsHelper.applyCustomView();
         // Verify that it was joined and formatted correctly
-        textCount = countText(dateFormat.format(new Date()));
+        textCount = countText(dateFormat.format(importDate));
         // records for generated files experiment.xar.log and experiment.xar.xml may have been created automatically
         assertTrue("Number of records is not as expected", textCount == 5 || textCount == 7);
```
labkey-chrisj: It might be helpful to future us to log the value of `dateFormat.format(importDate)` when this fails; that plus the evidence in the screencap will make it obvious what happened

labkey-chrisj (approved):

PR #1500, opened 2023-05-02
Issue 46942: Link reports are completely broken

labkey-sweta:
#### Rationale
Automation test for https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47004

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/src/org/labkey/test/tests/LinkedReportTest.java
+++ b/src/org/labkey/test/tests/LinkedReportTest.java
@@ -16,7 +16,7 @@ import java.util.List;
 public class LinkedReportTest extends BaseWebDriverTest
 {
     private static final String REPORT_NAME = "Labkey linked report";
-    private static final String LINK_REPORT_URL = "https://www.labkey.com";
+    private static final String LINK_REPORT_URL = "https://www.labkey.com/";
 
     @Override
     protected @Nullable String getProjectName()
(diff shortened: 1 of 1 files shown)


labkey-danield (approved):

labkey-tchad (commented):
Update the existing `NonStudyReportsTest.doLinkReportTest` to use this URL.
Also, don't make the verification so specific to the current text on `labkey.com`. That stuff is fairly likely to change at some point.

labkey-sweta:
> Update the existing `NonStudyReportsTest.doLinkReportTest` to use this URL. Also, don't make the verification so specific to the current text on `labkey.com`. That stuff is fairly likely to change at some point.

Does it make sense to keep one test with url linking to external website and one with url linking to internal.

src/org/labkey/test/tests/LinkedReportTest.java
```diff
@@ -0,0 +1,58 @@
...
+        setFormElement(Locator.name("viewName"), REPORT_NAME);
+        setFormElement(Locator.name("linkUrl"), LINK_REPORT_URL);
+        clickButton("Save");
+        waitForText("Manage Views");
+
+        waitAndClick(Locator.linkWithText(REPORT_NAME));
+        switchToWindow(1);
+
+        waitForElement(Locator.linkWithText("Solutions"));
+        Assert.assertTrue("Linked report navigated to incorrect external link", isElementPresent(Locator.linkWithText("Get a Demo")));
```
labkey-tchad: Still too specific. Just checking the URL should be sufficient.

labkey-tchad (approved):
You're probably right. There's enough difference between link reports to external and internal resources that we should test them both (considering there was an issue with one but not the other).

src/org/labkey/test/tests/LinkedReportTest.java
labkey-sweta: using getURL after swtichtowindow
is throwing Bad location from selenium tester: about:blank
any work around?

PR #1502, opened 2023-05-03
Issue 47841: Project menu bar webparts are lost on folder import or create from template project/folder

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47841

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/4385

#### Changes
- add test case check in MenuBarTest for creating project from template to verify menu webpart config comes through

### DIFF
--- a/src/org/labkey/test/util/UIContainerHelper.java
+++ b/src/org/labkey/test/util/UIContainerHelper.java
@@ -44,6 +44,12 @@ public class UIContainerHelper extends AbstractContainerHelper
     @LogMethod
     protected void doCreateProject(String projectName, String folderType)
     {
+        doCreateProject(projectName, folderType, null);
+    }
+
+    @LogMethod
+    public void doCreateProject(String projectName, String folderType, String templateFolder)
+    {
         _test.log("Creating project with name " + projectName);
         _test.ensureAdminMode();
         _test.goToCreateProject();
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/util/UIContainerHelper.java
```diff
@@ -43,6 +43,12 @@ public void doCreateFolder(String parentPath, String child, String foldertype)
     @Override
     @LogMethod
     protected void doCreateProject(String projectName, String folderType)
+    {
+        doCreateProject(projectName, folderType, null);
+    }
+
+    @LogMethod
+    public void doCreateProject(String projectName, String folderType, String templateFolder)
```
labkey-tchad: Make a separate method for creating from a template, rather than having to pass in a dummy folder type.
```suggestion
    {
        doCreateProject(projectName, folderType, null);
    }

    public void createProjectFromTemplate(String projectName, String templateFolder)
    {
        doCreateProject(projectName, "Create From Template Folder", templateFolder);
    }

    @LogMethod
    protected void doCreateProject(String projectName, String folderType, String templateFolder)
```

labkey-tchad (changes requested):

src/org/labkey/test/util/UIContainerHelper.java
cnathe: Good suggestion. Done.

labkey-tchad (approved):

labkey-jeckels (approved):

PR #1504, opened 2023-05-04
Automation test : Issue 47577: Add client API for renaming folders

labkey-sweta:
#### Rationale
Create Endpoint test against java api story.

#### Related Pull Requests
- https://github.com/LabKey/platform/issues/4323
- https://github.com/LabKey/platform/pull/4400

#### Changes
*

### DIFF
--- a/src/org/labkey/test/tests/RenameFolderJavaClientApiTest.java
+++ b/src/org/labkey/test/tests/RenameFolderJavaClientApiTest.java
@@ -10,10 +10,10 @@ import org.labkey.remoteapi.Connection;
 import org.labkey.remoteapi.security.RenameContainerCommand;
 import org.labkey.remoteapi.security.RenameContainerResponse;
 import org.labkey.test.BaseWebDriverTest;
+import org.labkey.test.Locators;
 import org.labkey.test.WebTestHelper;
 import org.labkey.test.categories.Daily;
 import org.labkey.test.pages.core.admin.ShowAuditLogPage;
-import org.labkey.test.pages.samplemanagement.admin.AuditLogPage;
 import org.labkey.test.util.LogMethod;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/RenameFolderJavaClientApiTest.java
```diff
@@ -0,0 +1,153 @@
...
+
+        log("Required values missing: Verify error appears");
+        command = new RenameContainerCommand(null, null, false);
+        try
+        {
+            command.execute(cn, getProjectName());
+        }
+        catch (CommandException e)
+        {
+            Assert.assertEquals("Blank names are not allowed.", 500, e.getStatusCode());
```
labkey-tchad: @RosalineP This seems wrong. Bad client input should be some 4xx error (400 or 422 maybe?). 500 means the server did something wrong.

src/org/labkey/test/tests/RenameFolderJavaClientApiTest.java
```diff
@@ -0,0 +1,153 @@
...
+        Assert.assertEquals("Rename container api failed ", 200, response.getStatusCode());
+        Assert.assertEquals("Incorrect new name", newContainerName + "1", response.getName());
+        Assert.assertEquals("Incorrect new title", newTitle + "1", response.getProperty("title"));
+
+        log("Required values missing: Verify error appears");
+        command = new RenameContainerCommand(null, null, false);
+        try
+        {
+            command.execute(cn, getProjectName());
+        }
```
labkey-tchad: If you're expecting an error, this should fail when an error doesn't occur.
```suggestion
        try
        {
            command.execute(cn, getProjectName());
            Assert.fail("Rename should have failed");
        }
```

src/org/labkey/test/tests/RenameFolderJavaClientApiTest.java
RosalineP: Oh shoot, I was using ApiUsageExceptions, which do 500 out. I'll fix this server-side
Update: New branch made

RosalineP (approved):
Coverage looks good!

PR #1506, opened 2023-05-05
Update test page so it can find the warning message

labkey-adam:
#### Rationale
Removing error rendering from the clone permissions page confused the test page

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4292


labkey-adam:
Not needed; updated the product to be consistent with the current test instead: https://github.com/LabKey/platform/pull/4397

PR #1511, opened 2023-05-10
Configure all code-only modules to auto-uninstall

labkey-adam:
#### Rationale
`CodeOnlyModule` now sets auto-uninstall to true

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4412


src/org/labkey/serverapi/reader/MimeMap.java
```diff
@@ -148,7 +149,7 @@ public int hashCode()
     }
 
 
-    private Hashtable<String, MimeType> map = new Hashtable<>();
+    private final Hashtable<String, MimeType> map = new Hashtable<>();
```
labkey-jeckels, line 152: `Hashtable`?

labkey-jeckels (approved):

PR #1532, opened 2023-06-02
Add annotation to control test method ordering

labkey-tchad:
#### Rationale
As discussed in the team meeting yesterday. I updated the existing annotation used to control the order of test classes; now it can be used with JUnit's `OrderWith` annotation to control execution order of test methods.

#### Related Pull Requests
* N/A

#### Changes
* Add annotation to control test method ordering


src/org/labkey/test/tests/BasicTest.java
```diff
@@ -28,15 +28,15 @@
...
+import org.labkey.test.util.Order;
 
 import java.util.List;
 
 /**
  * Short test to verify installed modules are well-formed
  */
 @Category({Base.class, DRT.class, Daily.class, Git.class, Hosting.class, Smoke.class})
-@Prioritized(priority = 1)
+@Order(-1)
```
labkey-danield, line 39: I'm assuming that -1 is a special value, and this test and the DatabaseDiagnosticsTest test will not participate in the ordering. Correct?

labkey-danield (approved):
Excellent, thanks for doing this!

src/org/labkey/test/tests/BasicTest.java
labkey-tchad: Zero is the default value. Negatives will run first, higher numbers will run last.

PR #1539, opened 2023-06-13
Automation Test for Export Does Not Include Failed Lookup Values

labkey-sweta:
#### Rationale
Automation coverage for Issue 47268: Export Does Not Include Failed Lookup Values

#### Related Pull Requests
* 

#### Changes
*


src/org/labkey/test/tests/AbstractExportTest.java
```diff
@@ -51,17 +51,35 @@ public abstract class AbstractExportTest extends BaseWebDriverTest
...
     protected DataRegionExportHelper exportHelper;
 
-    /** Return true if the rows can be selected in the grid. */
+    /**
+     * Return true if the rows can be selected in the grid.
+     */
     protected abstract boolean hasSelectors();
+
+    protected abstract boolean hasBrokenLookup();
+
```
labkey-tchad: This seems redundant. Checking whether `getTestLookUpColumnHeader` is `null` would work.

src/org/labkey/test/tests/AbstractExportTest.java
```diff
@@ -51,17 +51,35 @@ public abstract class AbstractExportTest extends BaseWebDriverTest
...
     protected abstract String getTestColumnTitle();
+
+    protected abstract String getTestLookUpColumnHeader();
+
+    protected abstract int getTestLookUpColumnIndex();
```
labkey-tchad, line 63: Consider making these non-abstract so that not all subclasses need to implement them.
```suggestion
    protected String getTestLookUpColumnHeader()
    {
        return null;
    }

    protected int getTestLookUpColumnIndex()
    {
        return -1;
    }
```

labkey-tchad (approved):
A couple of suggestions. No need for a re-review if you're confident in your updates (or decide to not make any)

src/org/labkey/test/tests/AbstractExportTest.java
labkey-sweta: good point, deleting the method

src/org/labkey/test/tests/AbstractExportTest.java
labkey-sweta: plan was to edit the rest of the sample files for other grids to test the same broken lookup case.

PR #1540, opened 2023-06-13
Add Better Check to SurveyTest.testDateTimeWithExtConfig

labkey-danield:
#### Rationale
The SurveyTest.testDateTimeWithExtConfig fails about once a month because there is no value in the Date Time Field after the survey has been saved ([example](https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyCPostgres/2494246?buildTab=tests&status=failed&name=SurveyTest.testDateTimeWithExtConfig)). Not really sure what the issue is, but could be one of three possibilities:

1. Values are not actually entered into the field and disappear before the survey is saved.
2. Values are in the field but the survey does not always save the value.
3. The values are saved in the survey but they take a moment to render in the grid (not very likely).

This change adds some better error checking to tries and  capture when the data actually disappears.

#### Related Pull Requests
* None

#### Changes
* Adding some checks to validate survey fields are set before saving.
* Changing check to wait for the text to show up after saving and validating.

### DIFF
--- a/src/org/labkey/test/tests/SurveyTest.java
+++ b/src/org/labkey/test/tests/SurveyTest.java
@@ -29,6 +29,7 @@ import org.labkey.test.util.Ext4Helper;
 import org.labkey.test.util.LogMethod;
 import org.labkey.test.util.PortalHelper;
 import org.labkey.test.util.WikiHelper;
+import org.openqa.selenium.Keys;
 import org.openqa.selenium.WebElement;
 
 import java.io.File;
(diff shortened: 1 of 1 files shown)


labkey-sweta (commented):
Would it be better to retry to set the form element instead of assertEquals?

labkey-danield:
That's a fair question @labkey-sweta, however my intent here is not to fix the test but to add better error checks to understand where the failure is happening. If the issue is that setFormElement doesn't always work then it should be investigated. If I retry setting the form elements and it works the test may pass but the possible issue with setFormElement would still exist. Happy to chat with you if that doesn't make sense.

labkey-sweta:
> That's a fair question @labkey-sweta, however my intent here is not to fix the test but to add better error checks to understand where the failure is happening. If the issue is that setFormElement doesn't always work then it should be investigated. If I retry setting the form elements and it works the test may pass but the possible issue with setFormElement would still exist. Happy to chat with you if that doesn't make sense.

Makes sense.

labkey-sweta (approved):

PR #1550, opened 2023-06-29
Allow `WebDriverWrapper.waitFor` to return non-boolean values

labkey-tchad:
#### Rationale
This allows us to wait for a thing then have the thing instead of fetching it separately after the wait finishes. This mimics the behavior of Selenium's `FluentWait.until` behavior but this isn't fatal if the wait doesn't succeed.

#### Related Pull Requests
* N/A

#### Changes
* Allow `WebDriverWrapper.waitFor` to return non-boolean values

### DIFF
--- a/src/org/labkey/test/WebDriverWrapper.java
+++ b/src/org/labkey/test/WebDriverWrapper.java
@@ -2170,22 +2170,24 @@ public abstract class WebDriverWrapper implements WrapsDriver
     }
 
     /**
-     * Wait for Supplier to return true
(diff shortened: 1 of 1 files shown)


labkey-danield (approved):
This doesn't require any changes to tests?

labkey-tchad:
> This doesn't require any changes to tests?

Nope! Sometime generics work quite nicely. Anything that was using `Boolean`s before will get the behavior they expect.

PR #1558, opened 2023-07-04
expand filteringReactSelect before attempting to interact with filter input

labkey-chrisj:
#### Rationale
Recently, tests like [CrossFolderRegistryTest.testRegistryFilters](https://teamcity.labkey.org/viewLog.html?buildId=2522162&buildTypeId=LabKey_Trunk_Premium_ProductSuites_Biologics_LimsStarter_LimsStarterPostgres&fromSakuraUI=true#testNameId8049891800230736841) have begun failing because the input on a FilteringReactSelect is not accessible by keyboard.

It turns out that if the reactSelect is expanded, the issue seems not to repro.
Hopefully this is not too invasive an approach

#### Related Pull Requests
n/a

#### Changes

- [x] expand the reactSelect before trying to type a filter expression into it

### DIFF
--- a/src/org/labkey/test/components/react/FilteringReactSelect.java
+++ b/src/org/labkey/test/components/react/FilteringReactSelect.java
@@ -162,8 +162,7 @@ public class FilteringReactSelect extends BaseReactSelect<FilteringReactSelect>
 
     private List<WebElement> setFilter(String value)
     {
-        if (!isOpen())
-            open();
+        open();
         elementCache().input.sendKeys(value);
         long filterStart = System.currentTimeMillis();
         WebDriverWrapper.waitFor(()-> {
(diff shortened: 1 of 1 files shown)


labkey-danield (approved):
I think this is ok, but it might be interesting to understand why this fails some of the time. It may be a timing issue and this code slows things down just enough for it to work.

src/org/labkey/test/components/react/FilteringReactSelect.java
```diff
@@ -162,6 +162,8 @@ public FilteringReactSelect filterSelect(String value, Locator elementToWaitFor)
 
     private List<WebElement> setFilter(String value)
     {
+        if (!isOpen())
+            open();
```
labkey-tchad: I don't think this check is needed. `open` checks `isExpanded`. The real question is why are there separate `isOpen` and `isExpanded` methods and do they do different things?

labkey-tchad (approved):

src/org/labkey/test/components/react/FilteringReactSelect.java
labkey-chrisj: `isOpen `checks to see if the component is decorated with the `select-input__control--menu-is-open` class, `isExpanded` looks for the state of the selectMenu, pauses the check if it's loading, then returns whether or not the selectMenu is displayed or if `isOpen `returns true

If anything, isExpanded is the more-robust one

PR #1561, opened 2023-07-07
Test fix for AuditLogMaintenanceTest.testAuditLogDeletionByAdmin 

labkey-sweta:
#### Rationale
StaleElementReferenceException fix for below error.

https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_LabkeyPremiumTrunk_GitModules_ProfessionalSqlserver/2526335?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildTestsSection=true

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/src/org/labkey/test/pages/pipeline/PipelineStatusDetailsPage.java
+++ b/src/org/labkey/test/pages/pipeline/PipelineStatusDetailsPage.java
@@ -13,6 +13,7 @@ import org.labkey.test.util.LogMethod;
 import org.labkey.test.util.PipelineStatusTable;
 import org.labkey.test.util.TextSearcher;
 import org.openqa.selenium.By;
+import org.openqa.selenium.StaleElementReferenceException;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.ui.ExpectedConditions;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/pages/pipeline/PipelineStatusDetailsPage.java
```diff
@@ -76,7 +77,16 @@ public String getEmail()
...
+        String retVal = null;
+        try
+        {
+            retVal = elementCache().statusText.getText();
+        }
+        catch (StaleElementReferenceException e)
+        {
+            return retVal;
+        }
+        return retVal;
```
labkey-danield, line 89: I don't think this will work the way you want it to.

For example, looking at the call stack PipelineStatusDetailsPage is called which has this line:
`        waitFor(() -> FINISHED_STATES.contains(getStatus()), () -> "Pipeline job did not finish. Final state: " + getStatus(), timoutMs);`

if `getStatus` returns null then `FINISHED_STATES.contains(getStatus())` will throw a null pointer exception.

Since this function will now possibly return a null you will need to look at all the callers of this method and decide if they need to be able to deal with a null.

I think to accomplish what you want you should change PipelineStatusDetailsPage to:
`        waitFor(() -> (null != getStatus()) && FINISHED_STATES.contains(getStatus()), () -> "Pipeline job did not finish. Final state: " + getStatus(), timoutMs);`

labkey-danield (commented):
I don't know that this will accomplish what you want. Take a look at my comment, push back if you disagree.

src/org/labkey/test/pages/pipeline/PipelineStatusDetailsPage.java
labkey-tchad: That is not correct. `FINISHED_STATES` could, theoretically, contain `null`. The check is fine as is.

labkey-danield: D'oh, yep you are right @labkey-tchad  and I was wrong. I thought FINISHED_STATES was a string.

labkey-danield (approved):

PR #1579, opened 2023-07-21
Add support for source (data class) parents in our apps

labkey-susanh:
#### Rationale
We are now supporting adding data class parents to data classes in our applications, so there's an opportunity for code sharing to support testing this functionality.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-premium/pull/138
- https://github.com/LabKey/platform/pull/4597
- https://github.com/LabKey/sampleManagement/pull/1970
- https://github.com/LabKey/biologics/pull/2240
- https://github.com/LabKey/labkey-ui-components/pull/1236

#### Changes
* Move alias handling methods and element cache objects to `EntityTypeDesigner` base class.
* Add `DataClassHelper` akin to `SampleTypeHelper`
* Update `DataClassApiHelper` with utility method

### DIFF
--- a/src/org/labkey/test/components/ui/domainproperties/samples/SampleTypeDesigner.java
+++ b/src/org/labkey/test/components/ui/domainproperties/samples/SampleTypeDesigner.java
@@ -62,87 +62,6 @@ public abstract class SampleTypeDesigner<T extends SampleTypeDesigner<T>> extend
         return getThis();
     }
 
-    public int getParentAliasIndex(String parentAlias)
-    {
-        List<Input> inputs = elementCache().parentAliases();
-        for (int i = 0; i < inputs.size(); i++)
-        {
-            if (inputs.get(i).get().equals(parentAlias))
-            {
-                return i;
-            }
-        }
-        throw new NotFoundException("No such parent alias: " + parentAlias);
-    }
-
-    public List<String> getParentAliasOptions(int index)
-    {
-        expandPropertiesPanel();
-        return elementCache().parentAliasSelect(index).getOptions();
-    }
-
-    public T removeParentAlias(String parentAlias)
-    {
-        expandPropertiesPanel();
-        int aliasIndex = getParentAliasIndex(parentAlias);
-        return removeParentAlias(aliasIndex);
-    }
-
-    public T removeParentAlias(int index)
-    {
-        expandPropertiesPanel();
-        elementCache().removeParentAliasIcon(index).click();
-        return getThis();
-    }
-
-    protected T setParentAlias(int index, @Nullable String alias, @Nullable String optionDisplayText)
-    {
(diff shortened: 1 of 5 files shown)


src/org/labkey/test/components/ui/domainproperties/samples/SampleTypeDesigner.java
```diff
@@ -62,87 +62,6 @@ public T addParentAlias(String alias, @Nullable String optionDisplayText)
         return getThis();
     }
```
labkey-danield, line 60: Removing this code left some unused imports :
```
import org.labkey.test.components.html.Input;
import org.labkey.test.components.react.ReactSelect;
```
May want to clean those up.

src/org/labkey/test/util/DataClassHelper.java
```diff
@@ -0,0 +1,286 @@
...
+    {
+        startTsvImport(data, IMPORT_OPTION)
+                .submit();
+    }
+
+    public void bulkImportExpectingError(List<Map<String, String>> data, String importOption)
+    {
+        startTsvImport(data, importOption)
+                .submitExpectingError();
+    }
```
labkey-danield: I know the tests helpers aren't very consistent, so take this with a grain of salt...
Generally when I have an 'ExpectingError' method I will wait for the error banner, or some other appropriate indicator, to show up and then I return either the error message or a reference to the error element. If no error was generated I return an empty string or a null.

src/org/labkey/test/util/DataClassHelper.java
```diff
@@ -0,0 +1,286 @@
...
+    {
+        startTsvImport(data, DataClassHelper.MERGE_OPTION)
+                .submit();
+    }
+
+    public void mergeImportExpectingError(List<Map<String, String>> data)
+    {
+        startTsvImport(data, DataClassHelper.MERGE_OPTION)
+                .submitExpectingError();
+    }
```
labkey-danield: Same comment as above. 
Since these don't appear to be used you may also choose to remove these methods. Whoever writes the test cases that could use these methods can add them.

labkey-danield (approved):
This looks ok to me. Just waiting for the TeamCity build to run.

src/org/labkey/test/components/ui/domainproperties/samples/SampleTypeDesigner.java
labkey-susanh: Thanks. Done.

src/org/labkey/test/util/DataClassHelper.java
labkey-susanh: I've just removed these and a few other unused methods. There are still some unused methods left here, but they have better chance of working as written.

PR #1580, opened 2023-07-22
Issue 43420: Swap jTidy --> jSoup

labkey-jeckels:
#### Rationale
jTidy appears abandoned and doesn't support common HTML5 tag elements

[Issue #43420: Replace jTidy with jSoup...or similar](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43420)

### Related
* https://github.com/LabKey/platform/pull/4542

#### Changes
* Swapped jSoup library in place of jTidy
* Add test coverage for script detection

### DIFF
--- a/src/org/labkey/test/tests/wiki/WikiLongTest.java
+++ b/src/org/labkey/test/tests/wiki/WikiLongTest.java
@@ -34,6 +34,7 @@ import java.util.Arrays;
 import java.util.List;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 
 @Category({Daily.class, Wiki.class})
 @BaseWebDriverTest.ClassTimeout(minutes = 11)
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/tests/wiki/WikiLongTest.java
```diff
@@ -491,6 +507,28 @@ public void testSteps()
...
+        log("test fixup for unsafe _blank targets, see #33356");
+        _wikiHelper.createNewWikiPage();
+        setFormElement(Locator.name("name"), WIKI_PAGE9_NAME);
+        _wikiHelper.setWikiBody(WIKI_PAGE9_CONTENT);
+        _wikiHelper.saveWikiPage();
+        assertTextNotPresent("New Page");  // Should not be an error, so should have left the editor
+
+        String renderedPage9 = getHtmlSource();
+        assertTrue("Fixed up link not present", renderedPage9.contains("<a href=\"http://labkey.com\" rel=\"noopener noreferrer\" target=\"_blank\">Fixup</a>"));
+        assertTrue("Safe link mangled", renderedPage9.contains("<a href=\"http://labkey.com\">Safe link</a>"));
```
Sigmonia: Could these use constants to keep in sync with the expected page content? As is, this will potentially fail as a false positive, so we should see a difference if a change occurred.

Sigmonia (approved):

src/org/labkey/test/tests/wiki/WikiLongTest.java
labkey-jeckels: @labkey-ians is the commit that I just pushed along the lines of what you were thinking?

Sigmonia: yeah, that's what I was thinking.

PR #1585, opened 2023-07-26
Test Automation Fixes

labkey-danield:
#### Rationale
Addressing some test issues that appear with a slow environment.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/1985
* https://github.com/LabKey/inventory/pull/986

#### Changes
* Make sure the dialog waits for a spinner in the waitForReady method.
* Make sure "Save" button, especially for lineage, is enabled before clicking in the EntityEditor.
* Don't silently fail in ManageSampleFinderViewsModal if edit icon is not present to click.
* Wait for the name edit field in ManageSampleFinderViewsModal before setting.
* Fixed error message.
* Wait for grid filters to update the grid.


src/org/labkey/test/components/ui/DeleteConfirmationDialog.java
```diff
@@ -42,7 +42,7 @@ protected void waitForReady()
     {
         WebDriverWrapper.waitFor(()-> elementCache().body.isDisplayed() &&
                         !BootstrapLocators.loadingSpinner.existsIn(this),
-                "The 'Choose Samples to Add' dialog did not display.", 1_000);
+                "The delete confirmation dialog did not become ready.", 1_000);
     }
```
labkey-danield, line 46: I'm not sure this waitForReady adds any value with the other change in the ModalDialog.

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
```diff
@@ -60,9 +61,10 @@ public WebElement getComponentElement()
     public Boolean isLoaded()
     {
         return getComponentElement().isDisplayed() &&
-                !Locators.loadingGrid.existsIn(this) &&
+                (!Locators.loadingGrid.existsIn(this) &&
                 !Locators.spinner.existsIn(this) &&
-                Locator.tag("td").existsIn(this);
+                Locator.tag("td").existsIn(this)) ||
+                getGridEmptyMessage().isPresent();
```
labkey-chrisj, line 67: nice.

src/org/labkey/test/components/ui/search/ManageSampleFinderViewsModal.java
```diff
@@ -126,7 +151,7 @@ public void clickDone()
 
     public String getErrorMsg()
     {
-        if(WebDriverWrapper.waitFor(()->elementCache().errorMsg.isDisplayed(), 1_000))
+        if(Boolean.TRUE.equals(WebDriverWrapper.waitFor(()->elementCache().errorMsg.isDisplayed(), 1_000)))
```
labkey-chrisj, line 154: Just for my information, is there a difference between primitive 'true' and Boolean.TRUE? 
(this is my paranoia that it's subtly, terribly different talking 🤣 )

labkey-chrisj (approved):

labkey-susanh (approved):

src/org/labkey/test/components/ui/search/ManageSampleFinderViewsModal.java
labkey-tchad: There is a subtle difference.
`Boolean b = null;` is valid, whereas `boolean b = null;` is not.
This change would only matter if the `WebDriverWrapper.waitFor` could return `null`. The old implementation would throw an NPE if it did but. `WebElement.isDisplayed` can't be null though, so that isn't a concern.

PR #1591, opened 2023-08-01
Fix multi-value filter type text area locators

XingY:
#### Rationale
[StudyPublishTest.testSteps](https://teamcity.labkey.org/viewLog.html?buildId=2564741&buildTypeId=LabkeyTrunk_DailyPostgres_2&fromSakuraUI=true#testNameId-3640016114532088128), [TimeChartVisitBasedTest.testSteps](https://teamcity.labkey.org/viewLog.html?buildId=2564741&buildTypeId=LabkeyTrunk_DailyPostgres_2&fromSakuraUI=true#testNameId500862954482810598), [DataReportsTest.doQueryReportTests](https://teamcity.labkey.org/viewLog.html?buildId=2564741&buildTypeId=LabkeyTrunk_DailyPostgres_2&fromSakuraUI=true#testNameId-7096353592880289930)

#### Related Pull Requests
* 

#### Changes
* update locator

### DIFF
--- a/src/org/labkey/test/components/CustomizeView.java
+++ b/src/org/labkey/test/components/CustomizeView.java
@@ -443,6 +443,10 @@ public class CustomizeView extends WebDriverComponent<CustomizeView.Elements>
         if (!StringUtils.isBlank(filter))
         {
             WebElement valueInput = Locator.css(".item-value input").findElement(clauseRow);
+            if (filter_type.endsWith(" Of"))
+            {
+                valueInput = Locator.css(".item-value textarea").findElement(clauseRow);
+            }
             _driver.setFormElement(valueInput, filter);
             _driver.fireEvent(valueInput, BaseWebDriverTest.SeleniumEvent.blur);
         }


src/org/labkey/test/components/CustomizeView.java
```diff
@@ -443,6 +443,10 @@ public void addFilter(String[] fieldKeyParts, String column_name, String filter_
         if (!StringUtils.isBlank(filter))
         {
             WebElement valueInput = Locator.css(".item-value input").findElement(clauseRow);
+            if (filter_type.endsWith(" Of"))
```
labkey-chrisj, line 446: logic here based on wording of the filter seems fragile, somehow.
If we really just want to find a textArea if that's on offer, consider a locator that will find both?

`WebElement valueInput = Locator.XPathLocator.union(
                    Locator.tagWithClass("div","item-value").descendant("input"),
                    Locator.tagWithClass("div", "item-value").descendant("textArea")).findElement(clauseRow);`

(I haven't tested this but I expect it will work)

labkey-chrisj (approved):
I trust that this will work as expected, but I suggest using a locator that will find either textarea or input without depending upon the phrasing of the filter

src/org/labkey/test/components/CustomizeView.java
XingY: I prefer finding by filterType because it also serves as additional test to check that textarea instead of text input is used for specific filter types.

PR #1596, opened 2023-08-03
update text assert to align to product

labkey-chrisj:
#### Rationale
Recently, userTest has begun [failing ](https://teamcity.labkey.org/viewLog.html?buildId=2568694&tab=buildResultsDiv&buildTypeId=LabkeyTrunk_DailyPostgres_2#testNameId7233972401637378005)because an expected error message has changed by the omission of a second space between sentences. This updates the check for that error to verify the sentences separately.

#### Related Pull Requests
n/a

#### Changes
just the test update


labkey-chrisj:
Well it looks like someone else fixed this, guess we don't need a PR for this any more

PR #1609, opened 2023-08-18
Remove a inline event handlers

labkey-matthewb:
#### Rationale
Remove a lot of inline event handlers as part of the ongoing Content-Security-Policy work.

#### Related Pull Requests
* 

#### Changes
*


labkey-tchad (approved):

modules/miniassay/resources/assay/simple/views/upload.html
```diff
@@ -55,11 +55,10 @@
 <br/>
 <div id='runs_div'></div>
 
-<script type="text/javascript" nonce="<%=getScriptNonce()%>">
-  LABKEY.requiresScript("FileUploadField.js");
+<script type="application/javascript" nonce="<%=scriptNonce%>">
   document.getElementById("X-LABKEY-CSRF").value = LABKEY.CSRF;
 </script>
-<script type="text/javascript" nonce="<%=getScriptNonce()%>">
```
labkey-tchad: Why is this changed to be `applications/javascript`? Most other scripts use `text/javascript`, which seems to be the preferred option from a quick search.

labkey-matthewb: fixed

PR #1614, opened 2023-08-23
align table cell text helper to new UI

labkey-chrisj:
#### Rationale
Recent changes to remove inline markup elements caused the finder for a non-dataregion table to [fail ](https://teamcity.labkey.org/viewLog.html?tab=queuedBuildOverviewTab&buildId=2599345&buildTypeId=LabkeyTrunk_DailyPostgres_2&fromSakuraUI=true#testNameId-5979966853213322683) because the locator looking for it expected its dataregion title element to be wrapped in a //div/h3/a/span, but is now in a //div/h3/span/span.  I've taken the liberty to make the locator a little more robust to UI changes like this in the future

#### Related Pull Requests
n/a

#### Changes

- [x] Fixed locator

### DIFF
--- a/src/org/labkey/test/WebDriverWrapper.java
+++ b/src/org/labkey/test/WebDriverWrapper.java
@@ -159,6 +159,7 @@ public abstract class WebDriverWrapper implements WrapsDriver
     public final static int WAIT_FOR_JAVASCRIPT = 10000;
     public final static int WAIT_FOR_PAGE = 60000;
 
+    @lombok.Getter
     public int defaultWaitForPage = WAIT_FOR_PAGE;
     public int longWaitForPage = defaultWaitForPage * 5;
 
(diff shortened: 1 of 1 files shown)


labkey-chrisj:
I'm not sure what's happening here or if someone rebased testAutomation somehow, but my feature branch with fixes somehow pull in changes I haven't made and result in a non-compiling build 🙄

labkey-chrisj:
OK, attempting to manually revert the changes that weren't intentional

labkey-tchad (commented):
Are you sure this wasn't an accidental product change? `<a>` -> `<span>` seems like it might have broken something.

labkey-chrisj:
Yes, I talked to @labkey-matthewb  about it, he thought it would be right to update this helper. It was part of the work to remove inline markup, probably not surprising to see it not affecting everything else

labkey-tchad (approved):

PR #1615, opened 2023-08-24
try to prevent stale reference to textarea

labkey-chrisj:
#### Rationale
Recently,  BiologicsAdminTest.testDeleteEmptyProject began [failing ](https://teamcity.labkey.org/viewLog.html?tab=queuedBuildOverviewTab&buildId=2600117&buildTypeId=LabKey_Trunk_Premium_ProductSuites_Biologics_LimsStarter_LimsStarterPostgres&fromSakuraUI=true#testNameId-3845714979850337579)on lims_starter with an error saying that the user-comment textArea was stale.

I've recently observed modal dialog-derived components failing in the same way and it seems that awaiting the component to be ready is helpful- it's possible that the component's lifecycle requires us to wait more rigorously, or to just use refindingElements within it.

#### Related Pull Requests
n/a

#### Changes

- [x] wait for the textArea to be present
- [x] use refindingWebElement for textArea

### DIFF
--- a/src/org/labkey/test/components/ui/DeleteConfirmationDialog.java
+++ b/src/org/labkey/test/components/ui/DeleteConfirmationDialog.java
@@ -41,7 +41,6 @@ public class DeleteConfirmationDialog<SourcePage extends WebDriverWrapper, Confi
     protected void waitForReady()
     {
         WebDriverWrapper.waitFor(()-> elementCache().body.isDisplayed() &&
-                        Locator.tagWithClass("textarea", "form-control").existsIn(this) &&
                         !BootstrapLocators.loadingSpinner.existsIn(this),
                 "The 'Choose Samples to Add' dialog did not display.", 1_000);
     }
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/components/ui/DeleteConfirmationDialog.java
```diff
@@ -41,6 +41,7 @@ protected DeleteConfirmationDialog(ModalDialogFinder finder, SourcePage sourcePa
     protected void waitForReady()
     {
         WebDriverWrapper.waitFor(()-> elementCache().body.isDisplayed() &&
+                        Locator.tagWithClass("textarea", "form-control").existsIn(this) &&
```
labkey-danield: This assumes all delete dialogs will have a text are, that is not true.

labkey-danield (changes requested):
Cannot assume all delete dialogs will have a text area.
[Examples of failures in LKSM with this change.](https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_SampleManager_SampleManagerCPostgres/2609013?buildTab=tests&status=failed&name=SMProSampleDeleteLineageTest)

labkey-danield (approved):

PR #1619, opened 2023-08-29
QueryGrid: Add showChart and closeChart

labkey-alan:
#### Rationale
Add showChart and closeChart methods to QueryGrid so our tests can more easily open and close charts.

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/1619
- https://github.com/LabKey/biologics/pull/2323

#### Changes
* QueryGrid: Add showChart and closeChart

### DIFF
--- a/src/org/labkey/test/components/ui/grids/QueryGrid.java
+++ b/src/org/labkey/test/components/ui/grids/QueryGrid.java
@@ -675,7 +675,7 @@ public class QueryGrid extends ResponsiveGrid<QueryGrid>
     public WebElement showChart(String chartName)
     {
         elementCache().chartsMenu.clickSubMenu(false, chartName);
-        return elementCache().svgChart;
+        return elementCache().svgChart();
     }
 
     public void closeChart()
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/components/ui/grids/QueryGrid.java
```diff
@@ -721,6 +732,16 @@ public WebElement panelHeader()
             return Locator.xpath("preceding-sibling::div[contains(@class,'panel-heading')]").findWhenNeeded(this);
         }
 
+        final BootstrapMenu chartsMenu = new MultiMenu.MultiMenuFinder(getDriver()).withText("Charts").findWhenNeeded(this);
+
+        final WebElement chartPanel = Locator.byClass("chart-panel").refindWhenNeeded(this);
```
labkey-chrisj, line 737: FYI if this grid is going to be used to show more than a couple of charts, 'refindWhenNeeded' doesn't re-find indefinitely (I think this is to avoid endless looping re-tries).

labkey-chrisj (approved):

src/org/labkey/test/components/ui/grids/QueryGrid.java
labkey-alan: Yeah, in theory we could write a test case that does something like:

- Navigate to grid
- Click chart menu, select chart
- Verify chart output
- Click chart menu, select chart
- rinse, repeat

Which is why I went for refindWhenNeeded because I thought that was how we avoided "stale" element issues. Is there a better solution?

labkey-chrisj: To support re-use of the page to show multiple charts or reports, you could find the report new instead of caching a refindingElement.  So, instead of 
`final WebElement chartPanel = Locator.byClass("chart-panel").refindWhenNeeded(this);`

you could do this
`WebElement chartPanel()
    {
        return  Locator.byClass("chart-panel").waitForElement(2000, this);
    }`

PR #1622, opened 2023-08-31
Add awaiting method to DeferredErrorCollector

labkey-tchad:
#### Rationale
Integrating basing awaitility assertion functionality into DeferredErrorCollector.
The previous pattern of creating a bunch of subclasses doesn't work when we might want to mix functionality of several of them (awaiting + screenshot for example)

#### Related Pull Requests
* #1617 

#### Changes
* Add `checker().await(Duration)` method
* Combine `DeferredErrorCollectorWrapper` subclasses into a single class


labkey-chrisj (approved):
Thanks for doing this; I've wanted to see this happen but wasn't sure how to incorporate awaitility into DeferredErrorCollector or if that sort of change would be welcome

PR #1623, opened 2023-09-02
Update test helper to match database auth dialog change

labkey-adam:
#### Rationale
The list of database auth strength options has expanded and the dialog now uses a drop-down list instead of buttons

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4707


labkey-chrisj (approved):
This is either 'good' or 'strong' 
Thanks for doing it

labkey-tchad (approved):

PR #1626, opened 2023-09-06
Don't ping healthcheck in a tight loop during bootstrap

labkey-tchad:
#### Rationale
When bootstrapping a server via this test, the access log starts with 10MB+ of calls to check whether the server is ready yet. We can probably ease up.

#### Related Pull Requests
* N/A

#### Changes
* Add sleep to avoid hammering healthcheck during bootstrap


labkey-chrisj (approved):
Tangentially, it happens that awaitility can await a condition and poll with fixed intervals, or with fibonacci intervals if that's what you'd like to see happen

PR #1631, opened 2023-09-08
Attempt to address staleness in delete confirmation dialog

labkey-chrisj:
#### Rationale
Some tests are intermittently [failing ](https://teamcity.labkey.org/viewLog.html?buildId=2624497&buildTypeId=LabKey_Trunk_Premium_ProductSuites_Biologics_BiologicsPostgres&fromSakuraUI=true#testNameId3243660359584709410)in biologics because when trying to set the user comment in DeleteConfirmationDialog, somehow the dialog body element is stale.
DeleteConfirmationDialog is one of the few ModalDialog subclasses that usually doesn't find the dialog by specifying its title, and it's possible that while rendering the component (updating the title) React might re-draw it (?)
I've attempted to address this failure a couple of ways without much success, but this is an attempt to wait until the dialog's title text is not empty (crossing fingers that this works)

#### Related Pull Requests
n/a

#### Changes

- [x] Attempt to improve `DeleteConfirmationDialog.waitforReady`
- [x] Don't click `DetailTableEdit.clickSave` save button until it's clickable

### DIFF
--- a/src/org/labkey/test/components/ui/DeleteConfirmationDialog.java
+++ b/src/org/labkey/test/components/ui/DeleteConfirmationDialog.java
@@ -41,7 +41,8 @@ public class DeleteConfirmationDialog<SourcePage extends WebDriverWrapper, Confi
     protected void waitForReady()
     {
         WebDriverWrapper.waitFor(()-> elementCache().body.isDisplayed() &&
-                        !BootstrapLocators.loadingSpinner.existsIn(this),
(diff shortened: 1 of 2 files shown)


labkey-tchad (approved):
🤞

labkey-chrisj:
While we're thinking about this, it might be necessary to modify the default locator for ModalDialogs to have non-zero title text?  If my theory is correct that could delay finding until the dialog is ready to be found

PR #1635, opened 2023-09-12
Make GridPanelTest more robust on MSSQL/Windows runs.

labkey-danield:
#### Rationale
The [GridPanelViewTest.testEditCustomView](https://teamcity.labkey.org/buildConfiguration/LabKey_239Release_Community_DailySuites_DailyCSqlserver/2630801?buildTab=tests&status=failed&name=GridPanelViewTest.testEditCustomView) has been failing on 23.9 Windows/MSSQL because the grid.getRows() method is not getting the updated rows from the grid after the filters have been removed. This test is not failing on develop.

The test does appear to be [passing on TC](https://teamcity.labkey.org/buildConfiguration/LabKey_239Release_Community_DailySuites_DailyCSqlserver?branch=improveGridPanleViewTest&buildTypeTab=overview&mode=builds) with this fix.

#### Related Pull Requests
* None.

#### Changes
* Change the check for the rowsCount to be in a waitFor.

### DIFF
--- a/src/org/labkey/test/tests/component/GridPanelViewTest.java
+++ b/src/org/labkey/test/tests/component/GridPanelViewTest.java
@@ -27,6 +27,7 @@ import org.labkey.test.util.PermissionsHelper;
 import org.labkey.test.util.SampleTypeHelper;
 import org.labkey.test.util.TestDataGenerator;
 import org.labkey.test.util.exp.SampleTypeAPIHelper;
+import org.openqa.selenium.WebDriverException;
 
 import java.io.IOException;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/component/GridPanelViewTest.java
```diff
@@ -644,8 +644,10 @@ private void testEditView(String testName, String viewName)
 
         validateGridHeader(testName, grid, EDITED_ALERT, true);
 
-        checker().verifyTrue("Filter was removed but number of rows in grid did not increase.",
-                grid.getRows().size() > rowCount);
+        // Wait until grid.getRows().size() gets an updated count.
+        checker().wrapAssertion(()->
+                waitFor(()->grid.getRows().size() > rowCountBefore,
+                        "Filter was removed but number of rows in grid did not increase.", 5_000));
```
labkey-tchad, line 655: `waitFor` doesn't throw an exception that `checker()` knows how to deal with by default.
```suggestion
        checker().addRecordableErrorType(WebDriverException.class).wrapAssertion(()->
                waitFor(()->grid.getRows().size() > rowCountBefore,
                        "Filter was removed but number of rows in grid did not increase.", 5_000));
```

labkey-tchad (approved):

src/org/labkey/test/tests/component/GridPanelViewTest.java
labkey-danield: Had to modify slightly, addRecordableErrorType has a void return.

PR #1636, opened 2023-09-13
Wait for grid update when saving view

labkey-chrisj:
#### Rationale
BiologicsComponentTest.testGridViewAcrossFolders has been intermittently [failing ](https://teamcity.labkey.org/viewLog.html?tab=queuedBuildOverviewTab&buildId=2630679&buildTypeId=LabKey_Trunk_Premium_ProductSuites_Biologics_LimsStarter_LimsStarterPostgres&fromSakuraUI=true#testNameId6418676379570466375)because The element reference of <span class="alert-info view-edit-alert"> is stale.

In the test's context, the test has just saved a view on the queryGrid of a Samples page, and is attempting to immediately select the default view.  To select any view, the component checks to see if the selection is also the current view, and it seems likely that the call to getEditAlertText() falls in the shuffle of whatever update occurs following the saving of the prior view.

#### Related Pull Requests
n/a

#### Changes

- [x] cause saveView to wait for update (hope there isn't a condition where it shouldn't expect that)

### DIFF
--- a/src/org/labkey/test/components/ui/grids/QueryGrid.java
+++ b/src/org/labkey/test/components/ui/grids/QueryGrid.java
@@ -435,7 +435,7 @@ public class QueryGrid extends ResponsiveGrid<QueryGrid>
                 dialog.setMakeAvailable(false);
         }
 
-        dialog.saveView();
+        doAndWaitForUpdate(()-> dialog.saveView());
     }
 
     /**
--- a/src/org/labkey/test/components/ui/grids/QueryGrid.java
+++ b/src/org/labkey/test/components/ui/grids/QueryGrid.java
@@ -435,7 +435,7 @@ public class QueryGrid extends ResponsiveGrid<QueryGrid>
                 dialog.setMakeAvailable(false);
         }
 
-        doAndWaitForUpdate(()-> dialog.saveView());
+        dialog.saveView();
     }
 
     /**
--- a/src/org/labkey/test/components/ui/grids/SaveViewDialog.java
+++ b/src/org/labkey/test/components/ui/grids/SaveViewDialog.java
@@ -142,7 +142,7 @@ public class SaveViewDialog extends ModalDialog
      */
     public void saveView()
     {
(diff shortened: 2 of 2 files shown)


src/org/labkey/test/components/ui/grids/QueryGrid.java
```diff
@@ -435,7 +435,7 @@ public void saveView(String viewName, boolean makeAvailable)
                 dialog.setMakeAvailable(false);
         }
 
-        dialog.saveView();
+        doAndWaitForUpdate(()-> dialog.saveView());
```
labkey-danield: I don't think this is the right fix.
I think the check for a stale element, or preventing one, should be done in the getViewName method.

labkey-danield (changes requested):
I don't think this is the right place to check/prevent this stale element error.

src/org/labkey/test/components/ui/grids/QueryGrid.java
labkey-tchad: I disagree. The method that causes the element to go stale should be responsible for wait for the operation to complete.

labkey-danield: The problem is that there is another save view methods that only returns the dialog. This fix would not address any of those tests that use that method.
Why not change the find in getEditAlertText to be a refindWhenNeeded?

labkey-tchad: Looks like `SaveViewDialog` has access to the grid. Could we move the call to `doAndWaitForUpdate` into the `saveView` method?

labkey-chrisj: It's worth a try!

labkey-danield (approved):
Excellent! Much better and complete solution that should cover all tests. Well done!
As long as the LKSM and LKB runs on TeamCity have no problems with this fix I'm happy with it.

labkey-tchad (approved):

PR #1637, opened 2023-09-14
Refactor `DeferredErrorCollector.awaiting` to actually work

labkey-tchad:
#### Rationale
`awaiting` didn't work as intended. It only works with `wrapAssertion`, not any of the other `verify*` methods.

#### Related Pull Requests
* #1622 

#### Changes
* Make `awaiting` function similarly to `wrapAssertion`

### DIFF
--- a/src/org/labkey/test/util/DeferredErrorCollector.java
+++ b/src/org/labkey/test/util/DeferredErrorCollector.java
@@ -113,17 +113,6 @@ public class DeferredErrorCollector
     }
 
     /**
-     * Create temporary error collector that waits for assertions to succeed. Assertions will be invoked repeatedly
-     * until they succeed within the specified timeout.
-     *
-     * @return Wrapped error collector. Intended to be used only once.
-     */
-    public DeferredErrorCollector awaiting(Duration timeout)
-    {
-        return new DeferredErrorCollectorWrapper(this).awaiting(timeout);
-    }
-
-    /**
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/util/DeferredErrorCollector.java
```diff
@@ -221,6 +210,20 @@ protected boolean isErrorDeferrable(Throwable err)
...
     }
 
+    /**
+     * Record an error if the provided assertion doesn't succeed within the specified timeout.
+     *
+     * @param timeout maximum duration to wait for assertion
+     * @param wrappedAssertion {@link Runnable} that might throw an error that shouldn't fail a test immediately
+     * @return <code>true</code> if the assertion succeeded within the specified timeout
+     */
+    public boolean awaiting(Duration timeout, Runnable wrappedAssertion)
```
labkey-chrisj, line 220: This is a sensible pattern, thanks for doing this.
Existing usages of this method will need refactoring
LMK if you want me to take a crack at those, mostly they should be fresh in my mind

labkey-chrisj (approved):

src/org/labkey/test/util/DeferredErrorCollector.java
labkey-tchad: Other uses are already refactored. PRs are assigned to you.

PR #1646, opened 2023-09-22
Refactor waitForLoaded into waitForReady on BaseReactSelect

labkey-chrisj:
#### Rationale
This change seeks to implement `waitForReady `on `BaseReactSelect`, by refactoring `waitForLoaded `to that method. Most methods on reactSelect defensively call waitForLoaded, but not all do.

Out of caution, I replaced old calls to `waitForLoaded `with new calls to `waitForReady` (which does the same thing, but is also run at instantiation time).  These old calls may not be necessary but I don't think it'll hurt to keep them there.

**Context:**
CrossFolderRegistryTest.testRegistryFilters has been [failing ](https://teamcity.labkey.org/viewLog.html?buildId=2567485&buildTypeId=LabKey_Trunk_Premium_ProductSuites_Biologics_LimsStarter_LimsStarterPostgres&fromSakuraUI=true#testNameId8049891800230736841)intermittently for some time because the test attempts to get filtered options from a reactSelect and it's been failing because it didn't open within 4 seconds.

(the test calls into setFilter, which does not wait for ready before calling `open`, so it's likely that open() doesn't happen because the control is not ready?) This change, which implements `waitForReady `on BaseReactSelect, ensures that all reactSelects will wait for ready when they are instantiated.

I think it's likely that overriding `waitForReady `on `BaseReactSelect `will address the above failure, but against the possibility that it doesn't I've taken the liberty to add fields for `READY_TIMEOUT `and `OPEN_TIMEOUT`, with matching setters for those fields. I've also updated `waitForReady `to respect `READY_TIMEOUT`, and open to respect `OPEN_TIMEOUT`

The idea here might be for pages backed by long queries to then be able to specify appropriate timeouts for their selects, like:

` FilteringReactSelect.finder(getDriver()).waitFor().setReadyTimeout(custom_timeout_msec)`

#### Related Pull Requests
n/a

#### Changes

- [x] Make `READY_TIMEOUT` settable, cause `waitForReady()` to respect it
- [x] Make `OPEN_TIMEOUT` settable, cause `open()` to respect it
- [x] refactor `waitForLoaded `into `waitForReady`, replace usages

### DIFF
--- a/src/org/labkey/test/components/react/BaseReactSelect.java
+++ b/src/org/labkey/test/components/react/BaseReactSelect.java
@@ -36,6 +36,10 @@ public abstract class BaseReactSelect<T extends BaseReactSelect<T>> extends WebD
     private static final String LOADING_TEXT = "loading...";
     protected static final String SELECTOR_CLASS = "select-input-container";
 
+    protected int READY_TIMEOUT = WAIT_FOR_JAVASCRIPT;
+    protected int OPEN_TIMEOUT = 4_000;
+
+
     public BaseReactSelect(WebElement selectOrParent, WebDriver driver)
     {
         // Component needs to be a RefindingWebElement because Select may go stale after loading initial selections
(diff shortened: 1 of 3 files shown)


src/org/labkey/test/components/react/BaseReactSelect.java
```diff
@@ -36,6 +36,10 @@ public abstract class BaseReactSelect<T extends BaseReactSelect<T>> extends WebD
     private static final String LOADING_TEXT = "loading...";
     protected static final String SELECTOR_CLASS = "select-input-container";
 
+    protected int READY_TIMEOUT = WAIT_FOR_JAVASCRIPT;
+    protected int OPEN_TIMEOUT = 4_000;
```
labkey-tchad: [Java convention] All-caps generally indicates that an identifier is `final`
```suggestion
    protected int readyTimeout = WAIT_FOR_JAVASCRIPT;
    protected int openTimeout = 4_000;
```

labkey-tchad (approved):
One nit

PR #1652, opened 2023-09-27
Form Buttons

labkey-alan:
#### Rationale
We are updating various form buttons to be rendered as a sticky element pinned to the bottom of the window.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1294
- https://github.com/LabKey/labkey-ui-premium/pull/195
- https://github.com/LabKey/biologics/pull/2381
- https://github.com/LabKey/inventory/pull/1033
- https://github.com/LabKey/sampleManagement/pull/2110
- https://github.com/LabKey/platform/pull/4777

#### Changes
* Fix selectors, scroll elements into view before interacting

### DIFF
--- a/src/org/labkey/test/Locators.java
+++ b/src/org/labkey/test/Locators.java
@@ -44,6 +44,11 @@ public abstract class Locators
         return Locator.tag("div").withClass("app-navigation");
     }
 
+    public static Locator.XPathLocator domainDesignerFloatingHeader()
+    {
+        return Locator.byClass("domain-floating-hdr");
+    }
+
     public static Locator.XPathLocator bodyPanel()
     {
         return Locator.tagWithClass("div", "lk-body-ct");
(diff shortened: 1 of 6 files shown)


labkey-tchad (approved):
Consider adding a method similar to `WebDriverUtils.ScrollUtil#scrollUnderFloatingHeader` (`scrollOverFloatingFooter`).
It could be used by [`ReclickingWebElement.revealElement`](https://github.com/LabKey/testAutomation/blob/95a161d102f64799418cfbff64d561e2405019a0/src/org/labkey/test/selenium/ReclickingWebElement.java#L207) to scroll things into view automatically.
Not sure how well it will work for these selects but it might help with other blocked clicks.

labkey-alan:
> Consider adding a method similar to `WebDriverUtils.ScrollUtil#scrollUnderFloatingHeader` (`scrollOverFloatingFooter`). It could be used by [`ReclickingWebElement.revealElement`](https://github.com/LabKey/testAutomation/blob/95a161d102f64799418cfbff64d561e2405019a0/src/org/labkey/test/selenium/ReclickingWebElement.java#L207) to scroll things into view automatically. Not sure how well it will work for these selects but it might help with other blocked clicks.

@labkey-tchad something I just experimented with, and appears to work, is to add a call to `scrollIntoView` to `scrollUnderFloatingHeader`, as the first thing that method does. Would that be an acceptable solution? It seems like the appropriate place, but I'm not quite sure I fully understand these methods. Alternatively I could put it in revealElement right before we call `scrollUnderFloadingHeader`.

PR #1653, opened 2023-10-03
Automation test for Issue 48185

labkey-sweta:
#### Rationale
Regression coverage for Update of the User table clears any non-specified custom fields 

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48185

Test steps
- Create two custom fields on the users table.
- Complete some information on at least one user for both fields
- Use the API to updateRows, but only provide the userid and customfield1 fields for update
Result: customfield1 is updated, but customfield2 is nulled out
Expected: when customfield2 is not provided, ignore and don't update

#### Related Pull Requests
* 

#### Changes
New test class UserTableCustomFieldUpdateTest

### DIFF
--- a/src/org/labkey/test/tests/UserTableCustomFieldUpdateTest.java
+++ b/src/org/labkey/test/tests/UserTableCustomFieldUpdateTest.java
@@ -33,7 +33,7 @@ public class UserTableCustomFieldUpdateTest extends BaseWebDriverTest
     @Override
     protected @Nullable String getProjectName()
     {
-        return getClass().getSimpleName() + " Project";
+        return null;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


labkey-klum (approved):
Coverage looks good for the original problem. As a side note, even though this issue was specific to user properties the bug could occur more generally in any domain that was not provisioned (leveraged ontology for custom fields).

src/org/labkey/test/tests/UserTableCustomFieldUpdateTest.java
```diff
@@ -0,0 +1,112 @@
...
+{
+    private static final String CUSTOM_FIELD1 = "CustomField1";
+    private static final String CUSTOM_FIELD2 = "CustomField2";
+    private static final String TEST_USER = "testcustomfieldupdate@test.com";
+
+    @Override
+    protected @Nullable String getProjectName()
+    {
+        return getClass().getSimpleName() + " Project";
+    }
```
labkey-tchad: This can/should return `null` if the test doesn't actually create a project.
```suggestion
    protected @Nullable String getProjectName()
    {
        return null;
    }
```

labkey-tchad (approved):

PR #1658, opened 2023-10-04
Fix inappropriate inheritance in `QueryMetadataEditorPage`

labkey-tchad:
#### Rationale
`QueryMetadataEditorPage` isn't actually a domain designer and doesn't get much utility out of extending `DomainDesigner`. The recent change to the domain designer button bar broke tests that use the QueryMetadataEditorPage.

#### Related Pull Requests
* #1652

#### Changes
* Fix inappropriate inheritance by `QueryMetadataEditorPage`


labkey-chrisj (approved):
Thanks for taking this on.
Just need to resolve conflicts, then good to go

PR #1659, opened 2023-10-04
Add synchronization to more FileBrowserHelper methods

labkey-tchad:
#### Rationale
S3 tests hit intermittent failures because of the slightly longer delay for many file browser operations. `FileBrowserHelper` assumes that many of these operations will finish immediately. `doAndWaitForFileListRefresh` should make sure the file list has updated before allowing tests to continue.

#### Related Pull Requests
* N/A

#### Changes
* Remove redundant `dragAndDropFileInDropZone` method
* Clean up extra test projects created by `FolderExportTest`
* Wrap more file browser operations in `doAndWaitForFileListRefresh`


labkey-tchad:
Fixing in 23.10: #1661

PR #1673, opened 2023-10-12
Test helper to get text from text nodes

labkey-tchad:
#### Rationale
We occasionally see product DOM that contains text in a way that is difficult for tests to extract just the text it wants (e.g. `<div><span>unwanted</span>wanted</div>`). A `WebElement` cannot represent such a text node but JavaScript can. We can use JavaScript to isolate the text children of a WebElement and get their text.
Given a WebElement representing the following div:
```
<div>
    <span>A</span>
    B
    <button>C</button>
    D
    <span>D</span>
</div>
```
`WebDriverUtils.getTextNodesWithin` will return a list containing `("B", "D")`
`WebDriverUtils.getTextNodeWithin` will return just `"B"`

#### Related Pull Requests
* N/A

#### Changes
* Create test helper to get text from text nodes
* Backport timing fix for `SaveViewDialog` (#1636)

### DIFF
--- a/src/org/labkey/test/util/selenium/WebDriverUtils.java
+++ b/src/org/labkey/test/util/selenium/WebDriverUtils.java
@@ -22,7 +22,9 @@ import org.labkey.test.util.DataRegionTable;
 import org.labkey.test.util.TestLogger;
 import org.openqa.selenium.JavascriptExecutor;
 import org.openqa.selenium.Keys;
+import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.WrapsDriver;
 import org.openqa.selenium.WrapsElement;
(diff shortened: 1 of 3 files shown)


src/org/labkey/test/util/selenium/WebDriverUtils.java
```diff
@@ -137,4 +139,67 @@ public static WebDriver extractWrappedDriver(Object peeling)
...
+                let thisNode = iterator.iterateNext();
+
+                while (thisNode) {
+                    texts.push(thisNode.textContent);
+                    thisNode = iterator.iterateNext();
+                }
+                return texts;
+                """;
+
+        List<Object> nodeTexts;
```
labkey-danield, line 178: Clever!

src/org/labkey/test/util/selenium/WebDriverUtils.java
```diff
@@ -137,4 +139,67 @@ public static WebDriver extractWrappedDriver(Object peeling)
...
+        try
+        {
+            nodeTexts = (List<Object>) executor.executeScript(script, element);
+        }
+        catch (WebDriverException retry)
+        {
+            nodeTexts = (List<Object>) executor.executeScript(script, element);
+        }
```
labkey-danield: Maybe add a comment here why this is needed.

labkey-danield (approved):

PR #1685, opened 2023-10-20
Handle `filename*` in Content-Disposition when downloading from URL

labkey-tchad:
#### Rationale
Some trial tests started failing because `SimpleHttpRequest.getResponseAsFile` (used by `ExportDiagnosticsPseudoTest`) couldn't parse the `Content-Disposition` header from a request to export diagnostics.
[[MDN]](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition#filename_2)

#### Related Pull Requests
* N/A

#### Changes
* Handle `filename*` in Content-Disposition when downloading from URL


labkey-danield (approved):

labkey-jeckels (commented):
Thanks for tracking this down. I looked at the failure briefly a couple of days ago. I thought it might have been related to recent Content-Disposition changes but didn't drill into the header interpretation here.

PR #1687, opened 2023-10-23
Wait for ready for EntityInsertPanel

labkey-chrisj:
#### Rationale
Recently [BiologicsSampleCreateTest.testDashboardCreateSamples](https://teamcity.labkey.org/viewLog.html?buildId=2704106&buildTypeId=LabKey_Trunk_Premium_ProductSuites_Biologics_BiologicsPostgres&fromSakuraUI=true#testNameId-1930444423740823537) failed because the `EntityInsertPanel.getEditableGrid` method assumes that if the grid isn't already shown, there must be a means to show it present (a la the 'show grid' or 'upload from file' selector).  The screenshot shows the grid is there, and this is an instance where there isn't a mode-selection; what's likely happened here is that the test asked for the grid before `EntityInsertPanel `was ready.

This change implements `waitForReady `on `EntityInsertPanel `by awaiting the file-upload or the grid or mode-selection tabs to be present.

#### Related Pull Requests
n/a

#### Changes

- [ ] await ready,

### DIFF
--- a/src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
+++ b/src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
@@ -31,6 +31,7 @@ public class EntityInsertPanel extends WebDriverComponent<EntityInsertPanel.Elem
 {
     private final WebDriver _driver;
     private final WebElement _editingDiv;
+    private int _readyTimeout = 10000;
 
     public EntityInsertPanel(WebElement element, WebDriver driver)
     {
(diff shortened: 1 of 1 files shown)


labkey-chrisj:
Started test runs for this in Biologics and SM
https://teamcity.labkey.org/project.html?projectId=LabKey_Trunk_Premium_ProductSuites&branch_LabKey_Trunk_Premium_ProductSuites_Biologics=entityInsertWaitForReady

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
```diff
@@ -50,6 +50,21 @@ public WebDriver getDriver()
...
     }
 
+    @Override
+    protected void waitForReady()
+    {
+        WebDriverWrapper.waitFor(()-> {
+            try{
+                return isFileUploadVisible() ||
+                        isGridVisible() ||
+                        hasTabs();
```
labkey-tchad: I don't think we want the `hasTabs` check here. That could make this check succeed before the grid or file upload are actually ready.

labkey-tchad (changes requested):

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
labkey-chrisj: OK.
Also it occurs to me that as we add waitForReady to query-backed components, it makes sense to make the ready timeout settable

labkey-tchad (approved):

labkey-tchad (changes requested):
I integrated your changes into my PR. Can close this

PR #1688, opened 2023-10-23
Remove react-bootstrap-toggle dependency and replace ReactBootstrapToggle usages

cnathe:
#### Rationale
Selenium test update for ToggleButton conversion from ReactBootstrapToggle to ToggleButtons/ToggleIcon

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/1326

#### Changes
- update ToggleButton to account for button or icon version of ToggleButtons.tsx component

### DIFF
--- a/src/org/labkey/test/components/domain/DomainFormPanel.java
+++ b/src/org/labkey/test/components/domain/DomainFormPanel.java
@@ -406,7 +406,7 @@ public class DomainFormPanel extends DomainPanel<DomainFormPanel.ElementCache, D
     {
         if (!getMode().equalsIgnoreCase(name))
         {
-            boolean isSummary = elementCache().customFieldsViewToggle.isEnabled();
+            boolean isSummary = elementCache().customFieldsViewToggle.isOn();
             elementCache().customFieldsViewToggle.set(!isSummary);
         }
         WebDriverWrapper.waitFor(()-> getMode().equalsIgnoreCase(name),
--- a/src/org/labkey/test/components/domain/DomainFormPanel.java
+++ b/src/org/labkey/test/components/domain/DomainFormPanel.java
@@ -394,11 +394,11 @@ public class DomainFormPanel extends DomainPanel<DomainFormPanel.ElementCache, D
      */
     public String getMode()
     {
-        return elementCache().customFieldsViewToggle.getSelectedStatus(); // will be either "Summary mode" or "Detail mode"
+        return elementCache().customFieldsViewToggle.getSelectedStatus(); // will be either "Summary" or "Detail"
     }
 
     /**
-     * Selects the desired mode.  Possible values are "Detail mode" and "Summary mode"
+     * Selects the desired mode.  Possible values are "Summary" and "Detail"
      * @param name The name of the desired mode, or the text to appear in the toggle in its desired state
      * @return an instance of the current DomainFormPanel
      */
@@ -406,9 +406,8 @@ public class DomainFormPanel extends DomainPanel<DomainFormPanel.ElementCache, D
     {
         if (!getMode().equalsIgnoreCase(name))
         {
-            // get the current mode (on/off), assume the opposite state is desired
-            Boolean modeState = elementCache().customFieldsViewToggle.get();
-            elementCache().customFieldsViewToggle.set(!modeState);
+            boolean isSummary = elementCache().customFieldsViewToggle.isEnabled();
+            elementCache().customFieldsViewToggle.set(!isSummary);
         }
         WebDriverWrapper.waitFor(()-> getMode().equalsIgnoreCase(name),
                 "the mode select toggle did not become [" +name+ "] as expected", 2000);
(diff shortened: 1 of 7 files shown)


src/org/labkey/test/components/domain/DomainFormPanel.java
```diff
@@ -394,21 +394,20 @@ public boolean isDefaultSystemFieldsExpanded()
      */
     public String getMode()
     {
-        return elementCache().customFieldsViewToggle.getSelectedStatus(); // will be either "Summary mode" or "Detail mode"
+        return elementCache().customFieldsViewToggle.getSelectedStatus(); // will be either "Summary" or "Detail"
```
labkey-tchad, line 397: I know this isn't open for review yet but I have some feedback since you're refactoring these anyway. Tests shouldn't have to know the specific toggle text for getting or setting the mode.
Outside of this class, the mode could be represented by an `enum` (e.g. `public enum DomainFormMode { Summary, Detail }`) which the public setter and getter would use.
Another option would be to get rid of (or make private) the existing `getMode`/`switchMode` methods and restrict callers to using `isDetailMode`/`isSummaryMode`/`selectDetailMode`/`selectSummaryMode`.

cnathe: Good suggestion. I'll look into making those changes and ping you for review once ready.

labkey-nicka: Leaving this PR review for @labkey-tchad.

cnathe: Done

src/org/labkey/test/components/react/ToggleButton.java
```diff
@@ -32,31 +32,57 @@ public WebDriver getDriver()
...
+        if (enabled && isDisabled())
+        {
+            if (hasButtons()) selectFirst();
+            else getComponentElement().click();
+        }
+        else if (!enabled && isEnabled())
+        {
+            if (hasButtons()) selectSecond();
+            else getComponentElement().click();
+        }
```
labkey-tchad: This should have some failure state if `isEnabled` and `isDisabled` are both false. That would indicate that the test component is out of sync with the product, right?
```suggestion
        else if (!enabled && isEnabled())
        {
            if (hasButtons()) selectSecond();
            else getComponentElement().click();
        }
        else
        {
            throw new IllegalStateException("Unable to determine toggle state: " + getComponentElement());
        }
```

src/org/labkey/test/components/react/ToggleButton.java
```diff
@@ -32,31 +32,57 @@ public WebDriver getDriver()
...
+    {
+        Locator.tag("button").index(0).findElement(this).click();
+        return this;
+    }
+
+    public ToggleButton selectSecond()
+    {
+        Locator.tag("button").index(1).findElement(this).click();
+        return this;
+    }
```
labkey-tchad: Do `selectFirst`/`selectSecond` need to be public? They seem to complicate the component.

src/org/labkey/test/components/react/ToggleButton.java
```diff
@@ -32,31 +32,57 @@ public WebDriver getDriver()
...
+
+    private boolean hasButtons()
+    {
+        return Locator.tag("button").findElementOrNull(this) != null;
```
labkey-tchad: I would stash the `hasButtons` value once to avoid searching for the buttons with every operation.
```suggestion
    private Boolean hasButtons = null;
    private boolean hasButtons()
    {
        if (hasButtons == null)
            hasButtons = Locator.tag("button").existsIn(this);
        return hasButtons;
```

src/org/labkey/test/components/react/ToggleButton.java
```diff
@@ -32,31 +32,57 @@ public WebDriver getDriver()
...
-    public boolean get()
+    public boolean isEnabled()
     {
-        return !getComponentElement().getAttribute("class").contains("off");
+        return getComponentElement().getAttribute("class").contains("toggle-on");
+    }
+
+    public boolean isDisabled()
+    {
+        return getComponentElement().getAttribute("class").contains("toggle-off");
```
labkey-tchad: Why was `get` split into two methods? I would expect these to be symmetrical and thus redundant to have them both.
Also, `isEnabled` is misleading. In other contexts, that would indicate whether or not one would be allowed to toggle the control (i.e. the `disabled` attribute), not what the on/off state is.

labkey-tchad (changes requested):

src/org/labkey/test/components/react/ToggleButton.java
cnathe: Yes. Added.

src/org/labkey/test/components/react/ToggleButton.java
cnathe: nope, good catch. updated.

src/org/labkey/test/components/react/ToggleButton.java
cnathe: good idea. done

src/org/labkey/test/components/react/ToggleButton.java
cnathe: Good points. I removed `isDisabled` and renamed `isEnabled` to `isOn`

labkey-tchad (approved):

PR #1692, opened 2023-10-25
test coverage for configurable ELN name settings

labkey-chrisj:
#### Rationale
This adds page class support for a new page class in the admin console, used to select notebookID name expressions.

The associated product change ([spec](https://docs.google.com/document/d/1WR4Ti8s-c_mJABCDFf5dJde-DQS9Pkc3p3IcoxO_Z1c/edit#heading=h.7u5qro28q0oc)) moves selection of notebookID name expressions from being folder-specific to being a global setting

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/2474

#### Changes

- [x] page class to wrap notebook settings in admin console
- [x] link to it from ShowAdminPage

### DIFF
--- a/src/org/labkey/test/pages/admin/ElnSettingsPage.java
+++ b/src/org/labkey/test/pages/admin/ElnSettingsPage.java
@@ -1,0 +1,91 @@
+package org.labkey.test.pages.admin;
+
+import org.labkey.test.Locator;
+import org.labkey.test.WebDriverWrapper;
+import org.labkey.test.WebTestHelper;
+import org.labkey.test.components.html.RadioButton;
+import org.labkey.test.pages.LabKeyPage;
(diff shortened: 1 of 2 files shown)


labkey-tchad (changes requested):
I don't think this needs to be in `testAutomation`. `ShowAdminPage` doesn't need to know about all of the things it might link to.

labkey-chrisj:
OK, I'll move it to labbook and close this PR.

I put it in TA because I figured having it linked from ShowAdminPage would be worth something to future-us (or anyone new that doesn't just happen to know that), squirreling these kinds of admin-console pages off into modules just fragments the map and makes discovery of these kinds of page classes harder.  (As I recall, one of my biggest challenges when I was new to Labkey's codebase was that you just had to know things like this)

PR #1696, opened 2023-10-26
ToggleButton test component fix for allowing set() to be called on current state without error

cnathe:
#### Rationale
Fixing some test failures that call the ToggleButton.set() on an element that is already in the desired state.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/1688

#### Changes
* change error state check in set() to throw if we can't determine the current state of the ToggleButton

### DIFF
--- a/src/org/labkey/test/components/react/ToggleButton.java
+++ b/src/org/labkey/test/components/react/ToggleButton.java
@@ -31,6 +31,10 @@ public class ToggleButton extends WebDriverComponent<WebDriverComponent<?>.Eleme
 
     public ToggleButton set(boolean enabled)
     {
+        // if we can't determine the current state of the toggle, then throw without proceeding
+        if (!isOn() && !isOff())
+            throw new IllegalStateException("Unable to determine the current state of the toggle.");
+
         String desiredState = enabled ? "enabled" : "disabled";
         if (enabled && !isOn())
         {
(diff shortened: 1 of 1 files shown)


cnathe:
TC passed the following tests:
StudyScheduleTest.testSteps,StudyExportTest.testSteps
LdapTest.testLdapSearchMiss,LdapTest.testLdapSearch

src/org/labkey/test/components/react/ToggleButton.java
```diff
@@ -31,6 +31,10 @@ public WebDriver getDriver()
 
     public ToggleButton set(boolean enabled)
     {
+        // if we can't determine the current state of the toggle, then throw without proceeding
+        if (!isOn() && !isOff())
+            throw new IllegalStateException("Unable to determine the current state of the toggle.");
```
labkey-tchad: I would put this check in `isOn`. That way, the `waitFor` after toggling will ensure its in a valid state.
```
    public boolean isOn()
    {
        String buttonCls = getComponentElement().getAttribute("class");
        if (buttonCls.contains("toggle-on"))
            return true;
        else if (buttonCls.contains("toggle-off"))
            return false;
        else
            throw new IllegalStateException("Unable to determine the current state of the toggle.");
    }
```

cnathe: I like it! Updated.

labkey-tchad (approved):

PR #1718, opened 2023-11-08
Await insert panel readiness before attempting to select grid

labkey-chrisj:
#### Rationale
Recently, [BiologicsSampleCreateTest.testDashboardCreateSamples](https://teamcity.labkey.org/viewLog.html?buildId=2737125&tab=buildResultsDiv&buildTypeId=LabKey_Trunk_Premium_ProductSuites_Biologics_LimsStarter_LimsStarterB&branch_LabKey_Trunk_Premium_ProductSuites_Biologics_LimsStarter=%3Cdefault%3E#testNameId-1930444423740823537) failed because `EntityInsertPanel.showGrid` makes the wrong assumption that if the grid is not already shown, there will already be a list-item there to allow the user to show the grid.  This used to be reliably true, but perhaps running on a faster browser allows the test to get to the race condition.  

Sometimes, the `EntityInsertPanel `doesn't support file-upload, and grid-upload is the only option (and thus, no mode-selection is offered). In this situation, when showGrid is called and the grid isn't already present, showGrid will try to select grid-mode and fail because no mode-selector will be there.

This change causes` EntityInsertPanel.showGrid `to await the presence of either the editable grid or the mode-selector before figuring out whether the grid is already shown.

#### Related Pull Requests
n/a

#### Changes

- [x] await the presence of either the grid or the mode selector before acting

### DIFF
--- a/src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
+++ b/src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
@@ -189,7 +189,7 @@ public class EntityInsertPanel extends WebDriverComponent<EntityInsertPanel.Elem
 
     private Optional<EditableGrid> optionalGrid()
     {
-        return new EditableGrid.EditableGridFinder(_driver).findOptional(this);
+        return elementCache().optionalGrid();
     }
 
     public EntityInsertPanel setMergeData(boolean allowMerge)
@@ -213,7 +213,7 @@ public class EntityInsertPanel extends WebDriverComponent<EntityInsertPanel.Elem
 
     protected FileUploadPanel fileUploadPanel()
     {
-        return new FileUploadPanel.FileUploadPanelFinder(_driver).timeout(WAIT_FOR_JAVASCRIPT).waitFor(this);
+        return elementCache().fileUploadPanel();
     }
 
     private Optional<FileUploadPanel> optionalFileUploadPanel()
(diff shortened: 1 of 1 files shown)


labkey-tchad (changes requested):
This component already has a `waitForReady` method. The thing is, we only call that method when initializing the element cache. The problem here is that too many methods in this component search for things without triggering `waitForReady`.
The solution is to move more of these methods into the element cache (many of them are private anyway), or to use the element cache as the search context. e.g.:
```
new EditableGrid.EditableGridFinder(_driver).findOptional(elementCache());
```

labkey-chrisj:
OK. I'll pull some of those functions into the elementCache, even if they remain runtime methods and don't cache- I think I started pulling those methods out of the cache to avoid staleness issues (this being a multi-modal component) without considering that this would sidestep waitForReady

labkey-chrisj:
Test results for this branch will be here https://teamcity.labkey.org/project/LabKey_Trunk_Premium_ProductSuites?branch=awaitGridSelectionReady&buildTypeTab=overview&mode=builds

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
```diff
@@ -189,7 +189,7 @@ public FileUploadPanel getFileUploadPanel()
 
     private Optional<EditableGrid> optionalGrid()
     {
-        return new EditableGrid.EditableGridFinder(_driver).findOptional(this);
+        return elementCache().optionalGrid();
     }
```
labkey-tchad: This method is private. Might as well just inline and remove it.

labkey-tchad (approved):

src/org/labkey/test/components/ui/entities/EntityInsertPanel.java
labkey-chrisj: done

PR #1729, opened 2023-11-16
Update location of `shouldDoBuild`

labkey-susanh:
#### Rationale
`shouldDoBuild` has moved from `FileModule` to `BuildUtils` for better accessibility for all

#### Related Pull Requests
* https://github.com/LabKey/gradlePlugin/pull/186

#### Changes
* Update location of `shouldDoBuild`


labkey-susanh:
Too much churn. We'll update this bit later.

PR #1731, opened 2023-11-17
Fix Wiki tests after TinyMCE upgrade

Sigmonia:
#### Rationale
Fix wiki tests related to the upgrade

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4969

#### Changes
* Adjust Wiki link header test
* Added a stopImpersonatingWithUnloadAlert helper

### DIFF
--- a/src/org/labkey/test/util/WikiHelper.java
+++ b/src/org/labkey/test/util/WikiHelper.java
@@ -281,6 +281,7 @@ public class WikiHelper
             if (_test.isElementPresent(Locator.css("#wiki-tab-visual.labkey-tab-inactive")))
             {
                 _test.click(Locator.css("#wiki-tab-visual > a"));
+                _test.waitAndClick(Locator.tagWithText("span","Yes"));
                 _test.waitForElement(Locator.css("#wiki-tab-visual.labkey-tab-active"));
             }
         }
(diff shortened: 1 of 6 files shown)


src/org/labkey/test/util/WikiHelper.java
```diff
@@ -280,7 +282,13 @@ public void switchWikiToVisualView()
...
             if (_test.isElementPresent(Locator.css("#wiki-tab-visual.labkey-tab-inactive")))
             {
-                _test.click(Locator.css("#wiki-tab-visual > a"));
+                Locator tab = Locator.css("#wiki-tab-visual > a");
+                _test.waitForElementToBeVisible(tab);
+                _test.click(tab);
+
+                Locator yesButton = Locator.tagWithText("span","Yes");
+                _test.waitForElementToBeVisible(yesButton);
+                _test.waitAndClick(yesButton);
```
labkey-danield, line 291: I don't think you need to do a waitForElementToBeVisible if you use waitAndClick. waitAndClick should wait for the element to be clickable.

labkey-danield (approved):

src/org/labkey/test/util/WikiHelper.java
Sigmonia: it doesn't. at least not in this instance. ~~It fails to find the element pretty consistently w/o it.~~ the elements are present but not visible, so it finds them but can't click them then throws an error.

PR #1732, opened 2023-11-21
await menu close if not navigating

labkey-chrisj:
#### Rationale
While writing automation to interact with the remirror editor in Notebook tests, I noticed what might have been some timing issues while selecting styles from the styles menu- there may have been some failures arising from the test trying to interact with the remirror editor before the menu was completely out of the way?
Adding a wait for the menu to close or be stale seems to help

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/2526
https://github.com/LabKey/labbook/pull/581

#### Changes
await the close (or staleness) of the menu if it's not navigating

### DIFF
--- a/src/org/labkey/test/components/html/BootstrapMenu.java
+++ b/src/org/labkey/test/components/html/BootstrapMenu.java
@@ -20,6 +20,7 @@ import org.labkey.test.WebDriverWrapper;
 import org.labkey.test.components.react.BaseBootstrapMenu;
 import org.labkey.test.util.LogMethod;
 import org.labkey.test.util.LoggedParam;
+import org.openqa.selenium.StaleElementReferenceException;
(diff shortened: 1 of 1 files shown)


labkey-chrisj:
Hmm, it looks like this change doesn't work with [GpatAssayTest.testRenameSharedAssayDesign](https://teamcity.labkey.org/viewLog.html?buildId=2757911&buildTypeId=LabkeyTrunk_Bvt&fromSakuraUI=true#testNameId-7756507080650135775), apparently we can't wait for the menu to close and handle an expected alert at the same time

labkey-danield (changes requested):
This is failing in the BVTs on TeamCity.
The issue is that clicking a menu item can show a system dialog, especially when working across container. The dropdown remains until the dialog is removed.

PR #1734, opened 2023-11-21
 Assay Transform Script UX update to allow file drop and file path entry

cnathe:
#### Rationale
See rationale from related PRs. This PR updates ReactAssayDesignerPage to fix the existing helper for adding transform script via path and adds an option to also upload the transform script file to the new @scripts dir. A few test cases have been updated to use this new option.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/1351
* https://github.com/LabKey/platform/pull/4973

#### Changes
* ReactAssayDesignerPage fixes for addTransformScript() and setTransformScript()
* Update a few existing assay transform script test cases to use new file upload option


labkey-danield (approved):
I thought I read in the spec that multiple files can be added. Not sure if this is something that needs to be added, or maybe added at a later date, or I just misread the spec.

cnathe:
> I thought I read in the spec that multiple files can be added. Not sure if this is something that needs to be added, or maybe added at a later date, or I just misread the spec.

Yes, multiple transform scripts can be added to the assay design. But it is still just one file upload at a time.

PR #1740, opened 2023-12-01
refactor 'select all/clear all selections' methods out of GridBar, into Grid

labkey-chrisj:
#### Rationale
While investigating failures in biologics test automation, I noticed that `GridBar.clearAllSelections` has an issue that causes it to no-op silently if there isn't a button containing 'Clear' already present.  Thinking it would be a simple matter of waiting for it to appear, I added that wait- and discovered that the check for its presence was looking within the `GridBar`, when the 'Clear' button is now located outside of that, in the 'selection-status' element, which contains the select-all and clear-all buttons.

Investigating further, I noted that `QueryGrid.clearAllSelections `checks to see if the current grid is a `GridPanel` (versus being a `QueryGridPanel`)- and only calls into `GridBar.clearAllSelections` if it is the latter.  (IIRC, `QueryGridPanel `was to be deprecated).

If QGP is in fact deprecated (@labkey-alan, is this the case?) we should probably remove `GridBar's ``clearAllSelections `method now and refactor its usages to `QueryGrid`'s `clearAllSelections`.
If not, the change in this PR updates `GridBar`'s `clearAllSelections` to function regardless of whether the current grid is a QueryGrid, by finding the 'Clear' button in the QueryGrid's componentElement

#### Related Pull Requests
related changes in biologics
https://github.com/LabKey/biologics/pull/2554

#### Changes
find the button in the QueryGrid component instead of the GridBar

### DIFF
--- a/src/org/labkey/test/components/ui/grids/GridBar.java
+++ b/src/org/labkey/test/components/ui/grids/GridBar.java
@@ -196,12 +196,13 @@ public class GridBar extends WebDriverComponent<GridBar.ElementCache>
         // Clear button can have text values of 'Clear', 'Clear both' or 'Clear all ' so just look for clear.
         Locator clearBtn = Locator.xpath("//button[contains(text(), 'Clear')]");
 
-        if(clearBtn.findOptionalElement(this).isPresent())
+        if(WebDriverWrapper.waitFor(()->
+                clearBtn.findOptionalElement(_queryGrid.getComponentElement()).isPresent(), 5_000)) // give it time to appear, else no-op
         {
-            WebElement btn = clearBtn.waitForElement(this, 5_000);
+            WebElement btn = clearBtn.findElement(_queryGrid.getComponentElement());
             btn.click();
 
-            WebDriverWrapper.waitFor(() -> clearBtn.findOptionalElement(this).isEmpty(),
+            WebDriverWrapper.waitFor(() -> clearBtn.findOptionalElement(_queryGrid.getComponentElement()).isEmpty(),
                     WAIT_FOR_JAVASCRIPT);
         }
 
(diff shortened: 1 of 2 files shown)


labkey-danield (approved):

labkey-tchad (changes requested):
There has to be a better solution to this. We expect places to call `clearAllSelections` just in case there's a selection. 90% of the time these should no-op but this change will add a 5 second delay to each of these calls.
Most callers shouldn't need to dig into the grid bar.
e.g. `grid.clearAllSelections();` instead of `grid.getGridBar().clearAllSelections();`
Not least of all because the `QueryGrid` version wraps it in a `doAndWaitForUpdate`.

labkey-alan:
QueryGridPanel no longer exists, so yes, you should be able to get rid of the old code.

labkey-chrisj:
OK, I've taken the liberty to:

- [x] take out the checks for QueryGridPanels, 
- [x] remove the GridBar select-all/deselect-all methods
- [x] refactor usages of the GridBar methods to instead use their corresponding methods on the Grid

This addresses the concern that we'll wait too long for situations that don't call for it- also removes some code that's been effectively dead for some time

labkey-chrisj:
Biologics results [here](https://teamcity.labkey.org/project/LabKey_Trunk_Premium_ProductSuites_Biologics?branch=clearSelectionsNoOp&projectTab=overview&mode=builds)
SM results [here](https://teamcity.labkey.org/project/LabKey_Trunk_Premium_SampleManager?branch=clearSelectionsNoOp&buildTypeTab=overview&mode=builds)

labkey-tchad (approved):

PR #1745, opened 2023-12-06
Add test flag to check for CSP warnings

labkey-tchad:
#### Rationale
A Tomcat Content Security Policy can be configured to report back to a server. LabKey now supports this reporting and logs warnings to `csp-report.log`. Tests should be able to check this file for warnings, similar to how we check the server's error log.
The new `webtest.cspCheck` flag enables this check.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4836

#### Changes
* Add test flag to check for CSP warnings

### DIFF
--- a/src/org/labkey/test/BaseWebDriverTest.java
+++ b/src/org/labkey/test/BaseWebDriverTest.java
@@ -601,6 +601,7 @@ public abstract class BaseWebDriverTest extends LabKeySiteWrapper implements Cle
             checker().withScreenshot("startupErrors").wrapAssertion(this::checkErrors);
             checker().withScreenshot("startupLeaks").wrapAssertion(this::checkLeaks);
             checker().wrapAssertion(() -> CspLogUtil.checkNewCspWarnings(getArtifactCollector()));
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/BaseWebDriverTest.java
```diff
@@ -1141,6 +1143,8 @@ private void doPostamble()
         if (isTestRunningOnTeamCity())
             checkActionCoverage();
 
+        CspLogUtil.checkNewCspWarnings(getArtifactCollector());
```
labkey-danield, line 1146: Since `CspLogUtil.checkNewCspWarnings` can throw an assert or an exception shouldn't this line be in a try/catch? Otherwise this is saying that if there was some issue with the csp we cannot/will not check the links or any of the other actions that follow.

labkey-danield (approved):
Did you want to add @labkey-sweta and @labkey-chrisj to review as well?

src/org/labkey/test/BaseWebDriverTest.java
labkey-tchad: This matches the behavior for all of the existing checks

PR #1748, opened 2023-12-07
Fixing some annoying bugs.

labkey-danield:
#### Rationale
Fixed two issues in common test code.

#### Related Pull Requests
* None

#### Changes
* Fixed issues in createUser that fails creation because of case differences in email address.
* Added a scrollIntoView to fix issue when grid rows are scrolled under the grid header and cannot be clicked.


labkey-tchad (changes requested):
I would leave this as is. The server converts emails to lower-case and, generally, accepts either case but our test helpers do not. This check ensures that the email used to create the user matches what will actually show up in the UI and in `email` fields of various APIs.

labkey-danield:
I'm assuming that because this comment was made globally and not on the individual file the ask was not to change either file.

labkey-tchad (approved):
Looks like somebody else already added that scroll and it got merged in to this branch. This is just an import cleanup now.

PR #1757, opened 2023-12-15
Remove dependency on repacked jetty

labkey-tchad:
#### Rationale
Selenium created a repacked version of jetty for reasons that are unclear to me. It hasn't been updated since 2019 and Selenium no longer uses or updates it and it has a transitive dependency that we are tying to eliminate (`javax.servlet-api`).
Tests only use this for URL encoding, so switching to the non-repacked version should work for our needs.

#### Related Pull Requests
* N/A

#### Changes
* Update test dependency on `org.seleniumhq.selenium:jetty-repacked` to `org.eclipse.jetty:jetty-util`


labkey-jeckels (approved):
Do we need this dependency at all, or can we use URLEncoder/URLDecoder?

https://www.baeldung.com/java-url-encoding-decoding

labkey-danield (approved):

PR #1758, opened 2023-12-15
give caller ability to set parentDetailPanel ready timeout

labkey-chrisj:
#### Rationale
Recently, [BiologicsSampleCreateTest.testCrossfolderSampleDerivation](https://teamcity.labkey.org/viewLog.html?buildId=2798792&buildTypeId=LabKey_Trunk_Premium_ProductSuites_Biologics_BiologicsPostgres&fromSakuraUI=true#testNameId4744865710225756707) failed because a query-backed component (in this case, `ParentDetailPanel`) took longer than the component's hard-coded 4-second timeout period.

The screenshot shows a ready component, which suggests that it took very little time beyond that 4 second window- this change extends the default time the component's `waitForReady `method will wait to 10 seconds, and it gives callers and containing page classes the ability to set that timeout.

Failures like this one are essentially noise, the intention of this change is to give callers and page classes the ability to allow for query-backed components like this one to await the backing query.

#### Related Pull Requests
n/a

#### Changes

- [x] make `_waitForReadyTimeout `a settable class variable, default at `WAIT_FOR_JAVASCRIPT`

### DIFF
--- a/src/org/labkey/test/components/ChartTypeDialog.java
+++ b/src/org/labkey/test/components/ChartTypeDialog.java
@@ -16,6 +16,7 @@
 package org.labkey.test.components;
 
 import org.labkey.test.Locator;
+import org.labkey.test.WebDriverWrapper;
 import org.labkey.test.pages.TimeChartWizard;
 import org.labkey.test.selenium.LazyWebElement;
 import org.labkey.test.util.Ext4Helper;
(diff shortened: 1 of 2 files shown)


labkey-chrisj:
Also took the liberty of piggybacking an attempt at fixing this intermittent failure: [TimeChartDateBasedTest.testSteps](https://teamcity.labkey.org/viewLog.html?buildId=2798846&buildTypeId=LabkeyTrunk_DailyPostgres_2&fromSakuraUI=true#testNameId3569904870967522839)

Note, the theory of this fix is that it just takes slightly > 2 seconds for the y-axis arrow component to become visible, but it's not implausible that the associated typeError in the screenshots is involved in that

labkey-tchad (approved):
The `ChartTypeDialog` failure is product-side

PR #1760, opened 2023-12-18
Issue 48715: Limit Data Region "Show all" to a maximum number of rows

labkey-nicka:
#### Rationale
Regression coverage for [Issue 48715](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48715).

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5058

#### Changes
- Add regression coverage scenario to `DataRegionTest` for Issue 48715.


labkey-matthewb:
I don't love this solution as is. I'm not opposed to removing the UI option for "show all", but then we should remove or rename the menu item if it doesn't do "show all".  We could add a :paging 1000" option, or a rename to "show 1000" or another option.

labkey-nicka:
Screwed up merge. Closing in favor of #1762.

PR #1766, opened 2023-12-21
Log knitr report text

labkey-adam:
#### Rationale
Let's figure out what's causing the last Daily suite CSP violation


labkey-adam:
No longer needed

PR #1767, opened 2023-12-21
even stricter form validation

labkey-matthewb:
#### Rationale
Just because we can

#### Related Pull Requests
* 

#### Changes
*


labkey-tchad (changes requested):
I would prefer to be able to tweak these injection strings on the test side. If you really want to lock this down tighter; what if we made this action reject anybody with more than reader permissions?

labkey-matthewb:
> I would prefer to be able to tweak these injection strings on the test side. If you really want to lock this down tighter; what if we made this action reject anybody with more than reader permissions?

That's an interesting idea.  Sure.

PR #1769, opened 2023-12-26
await query-backed detail table link

labkey-chrisj:
#### Rationale
Periodically, [BiologicsBulkRegistryTest.testAddNucleotidesFromFile](https://teamcity.labkey.org/viewLog.html?buildId=2814046&buildTypeId=LabKey_Trunk_Premium_ProductSuites_Biologics_BiologicsPostgres&fromSakuraUI=true#testNameId-2381955727543163758) and some others fail because they call `DetailTable's ` `clickField` method assuming that the link will appear as selected.

Because the link (and much of the detailTable's content) is query backed, the link can take considerable time to appear depending upon the context- this change extends the default timeout (was 1.5 seconds) and makes that timeout configurable

#### Related Pull Requests
n/a

#### Changes

- [x] wait longer for the link to appear
- [x] make configuring the wait timeout possible

### DIFF
--- a/src/org/labkey/test/components/ui/grids/DetailTable.java
+++ b/src/org/labkey/test/components/ui/grids/DetailTable.java
@@ -14,6 +14,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import static org.labkey.test.WebDriverWrapper.WAIT_FOR_JAVASCRIPT;
+
 /**
  * This is a 'special' table that has only two columns, and no header. An example of this table can be seen in the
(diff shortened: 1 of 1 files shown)


labkey-tchad (approved):
Defaulting to 10 seconds seems a bit long for a default but no big problem.

One thing to bring to a dev's attention: looking at the `html` screenshot, I'm surprised there isn't a "Loading..." indicator if this can take a while to appear.

labkey-chrisj:
Tracking that as https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49340

PR #1787, opened 2024-01-10
wait for the select to have a selection before calling clearSelection

labkey-chrisj:
#### Rationale
[BiologicsAutoPopulateAssayJobSamplesTest.testAssociateTaskToExistingAssayRun](https://teamcity.labkey.org/viewLog.html?tab=queuedBuildOverviewTab&buildId=2842046&buildTypeId=LabKey_2311Release_Premium_ProductSuites_Biologics_LimsStarter_LimsStarterB&fromSakuraUI=true#testNameId-3261285293786488283) has been intermittently failing (23.11, starter) because the opens a detailTable for edit, calls `clearSelectValue ` to empty the Workflow Task select, and expects to be able to click the 'Save Run Details' button.

It happens that BaseReactSelect.clearSelection() does not check to ensure there is a selection (if there isn't, it No-ops). This means it is probably likely that the test called it before the select was populated and then failed to save changes because there were no changes and the save button never became enabled.

This change tells `DetailTableEdit.clearSelectValue()` to wait for the select to have any selection before calling `clearSelection()`

While doing this refactor, I discovered that DetailTableEdit has two clearSelectValue methods, this eliminates one and the related PR consolidates usages in biologics

Targeting develop for this fix, will consider backporting after

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/2634

#### Changes

- [x] eliminate redundant method
- [x] add readyTimeout member, make it settable
- [x] use readyTimeout to await value before clearing
- [x] wait for select to have a selection before calling `clearSelection`

### DIFF
--- a/src/org/labkey/test/components/ui/grids/DetailTableEdit.java
+++ b/src/org/labkey/test/components/ui/grids/DetailTableEdit.java
@@ -329,7 +329,10 @@ public class DetailTableEdit extends WebDriverComponent<DetailTableEdit.ElementC
      **/
     public DetailTableEdit clearSelectValue(String fieldCaption)
     {
-        elementCache().findSelect(fieldCaption).clearSelection();
+        var select = elementCache().findSelect(fieldCaption);
+        WebDriverWrapper.waitFor(()-> select.hasSelection(),
+            String.format("The %s select did not have any selection in time"), WebDriverWrapper.WAIT_FOR_JAVASCRIPT);
+        select.clearSelection();
         return this;
     }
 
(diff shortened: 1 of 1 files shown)


labkey-tchad (approved):
You sure we never call this just in case there's a selection to clear but might not?

labkey-chrisj:
I'm not 100% sure there won't be a use case for the just-in-case clear of the select that might-or-might-not have a selection, but it would be really simple to telescope the method to default to wait, but provide the option to skip waiting.

Or, if we want more-complete work, we could push that logic upstream into BaseReactSelect.clearSelection (that is, give the caller the option to await a selection, vs not)

PR #1789, opened 2024-01-11
provide optional wait for selection

labkey-chrisj:
#### Rationale
Today [BiologicsTest.editNucleotideDetails](https://teamcity.labkey.org/viewLog.html?buildId=2844344&buildTypeId=LabKey_Trunk_Premium_ProductSuites_Biologics_BiologicsPostgres&fromSakuraUI=true#testNameId2291990734087517910) began failing because yesterday https://github.com/LabKey/testAutomation/pull/1787 introduced a change causing tests that call DetailTableEdit.clearSelectionValue when no selection is going to appear

For background, some tests have been failing because they call BaseReactSelect.clearSelection (which no-ops if no selection is present) before the selection appears, and then proceed to fail when trying to save the edit that didn't happen because the DetailTableEdit's save button didn't become enabled.

Apparently some tests need to clear selections that might not exist (just to set state that might not be known beforehand), so making it separately optional to wait and to assert if no selection appears

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/2637

#### Changes

- [x] make wait optional
- [x] make assert optional

### DIFF
--- a/src/org/labkey/test/components/ui/grids/DetailTableEdit.java
+++ b/src/org/labkey/test/components/ui/grids/DetailTableEdit.java
@@ -324,23 +324,28 @@ public class DetailTableEdit extends WebDriverComponent<DetailTableEdit.ElementC
      **/
     public DetailTableEdit clearSelectValue(String fieldCaption)
     {
-        return clearSelectValue(fieldCaption, true);
+        return clearSelectValue(fieldCaption, true, true);
     }
 
     /**
      * Clear a given select field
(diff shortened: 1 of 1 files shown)


labkey-tchad (changes requested):
A 10 second wait is way too long for the non-asserting scenario. I'd say 1 second should be sufficient if this was functional without a wait for most usages.

labkey-tchad (approved):

PR #1831, opened 2024-02-19
Add nonces to script tags

labkey-adam:
#### Rationale
CSPs like nonces. Also, path-first URL are preferred.


src/org/labkey/test/tests/FileAttachmentColumnTest.java
```diff
@@ -100,7 +100,7 @@ private void doSetup()
 
     private void createList()
     {
-        beginAt("/project/" + getProjectName() +"/"+ FOLDER_NAME + "/begin.view?");
+        beginAt(getProjectName() + "/" + FOLDER_NAME + "/begin.view");
```
labkey-tchad: Missed the controller.
```suggestion
        beginAt(getProjectName() + "/" + FOLDER_NAME + "/project-begin.view");
```

src/org/labkey/test/tests/AdminConsoleTest.java
```diff
@@ -68,7 +68,7 @@ public void testRibbonBar()
         waitForElement(Locator.xpath("//div[contains(text(), 'Cannot enable the ribbon message without providing a message to show')]"));
 
         String linkText = "and also click this...";
-        String html = "READ ME!!!  <a href='<%=contextPath%>/project/home/begin.view'>" + linkText + "</a>";
+        String html = "READ ME!!!  <a href='<%=contextPath%>" + "/home/project-begin.view'>" + linkText + "</a>";
```
labkey-tchad, line 71: Looks like some structural search weirdness
```suggestion
        String html = "READ ME!!!  <a href='<%=contextPath%>/home/project-begin.view'>" + linkText + "</a>";
```

labkey-tchad (changes requested):
One bad URL change.
Everything in `data` is used to populate wikis; I thought wikis didn't need to have the nonces explicitly defined?

src/org/labkey/test/tests/AdminConsoleTest.java
labkey-adam: Looks a little funny in this case, but it is intentional. I use `<%=contextPath%>/` to search for controller-first URLs. Path-first URLs usually have `<%=contextPath%><%=containerPath%>`, but obviously not the case of a hard-coded container path like this. Separating like this means the URL won't show up in searches.

labkey-adam:
> One bad URL change. Everything in `data` is used to populate wikis; I thought wikis didn't need to have the nonces explicitly defined?

You're correct... I'm not sure how I misread their usages. I think they're all corrected now.

labkey-tchad (approved):

PR #1834, opened 2024-02-22
Await grid paste success

labkey-chrisj:
#### Rationale
Occasionally, tests like [BiologicsSampleUpdateTest.testBulkEditThenEditInGrid](https://teamcity.labkey.org/viewLog.html?buildId=2915721&buildTypeId=LabKey_Trunk_Premium_ProductSuites_Biologics_BiologicsPostgres&fromSakuraUI=true#testNameId-795420342457657317) fail because they assert grid content they just pasted and then navigated away from the grid page.

This adds a check to pasteFromCell that splits apart the paste content and awaits the presence of those parts in the cell selection range (which is also the paste target).
If this is too aggressive, it would be simple to make this optional/opt-in behavior instead of default behavior

#### Related Pull Requests
(branches for SM, Biologics don't have changes, just there to get test runs against this change)

#### Changes

- [x] verify paste content parts to be present in the dom


labkey-chrisj:
TC test results here:
Biologics https://teamcity.labkey.org/project/LabKey_Trunk_Premium_ProductSuites_Biologics?branch=awaitGridPasteSuccess&projectTab=overview&mode=builds
SM https://teamcity.labkey.org/project/LabKey_Trunk_Premium_SampleManager?branch=awaitGridPasteSuccess&buildTypeTab=overview&mode=builds

src/org/labkey/test/components/ui/grids/EditableGrid.java
```diff
@@ -595,9 +603,28 @@ public EditableGrid pasteFromCell(int row, String columnName, String pasteText)
...
         return this;
     }
 
+    public void waitForPasteContent(String pasteContent)
+    {
+        var contentParts = pasteContent.replace("\n", "\t").split("\t");
+        await().atMost(Duration.ofSeconds(2))
+                .untilAsserted(()-> Assertions.assertThat(getSelectionCellTexts())
+                        .contains(contentParts));
+    }
```
labkey-danield: Why is this public?

src/org/labkey/test/components/ui/grids/EditableGrid.java
```diff
@@ -574,15 +575,22 @@ public List<String> getFilteredDropdownListForCell(int row, String columnName, @
         return lookupSelect.getOptions();
     }
 
+
+    public EditableGrid pasteFromCell(int row, String columnName, String pasteText)
+    {
+        return pasteFromCell(row, columnName, pasteText, true);
+    }
+
```
labkey-danield: Because this is a widely used method I would suggest that the validate parameter default to false.
Why?
This is the current signature, which means that current test using this method are not expecting a validation check. This changes the default behavior.
Defaulting to true could cause problems for Date, Time or DateTime fields. The string passed in may not match the format of the fields but it would still be a valid string to paste.

If we are going to make a global change like this can we _**please finally**_ change the name of this method? The action is not pasting from a cell it is pasting into a cell.

labkey-danield (commented):
I cannot approve this until there is a TeamCity pass for Biologics, SampleManager and Inventory.

src/org/labkey/test/components/ui/grids/EditableGrid.java
labkey-chrisj: I'm fine with this as opt-in behavior. I pushed the change this way to get results with this validation, since we already know what results will look like without validating.
I'm cool with renaming the method (LOL this will make the change very large) but if we want that, cool.

labkey-chrisj:
Initial test pass results show failures  because values like "" and " " weren't in the actual grid cell texts; I pushed a change to filter those values out of the expected set so we aren't trying to validate non-values like that. Hopefully this resolves the failures we saw in the first runs.  If we get acceptable results here we can consider whether it would be best for validating like this to be opt-in (or default) behavior; if we don't it probably makes sense for it to be opt-in only or possibly not worth doing at all.

labkey-chrisj:
Update: test results with the latest change show only error-case pastes (where an invalid value was pasted in) failing; in biologics there were just 3 tests failing as a result.  Given this result I think it would probably worthwhile to opt those 3 cases out and have this sort of validation be default behavior.
If we decide to make validation opt-in-only, I think I'm going to want to update 99% of calls to this method to opt into that.

labkey-tchad:
> [...] If we decide to make validation opt-in-only, I think I'm going to want to update 99% of calls to this method to opt into that.

Assuming a successful operation is pretty typical for our helpers. `doSomethingExpectingError` is a common pattern for skipping validation or checking for errors after an operation. Maybe the validation could be less strict; e.g. just check that some text appeared in a cell but don't check for all text across all cells.

labkey-chrisj:
> Assuming a successful operation is pretty typical for our helpers.

Seems fair enough. I'll flip it to be opt-in, and update calling code where I suspect test failures to involve the test racing ahead of the page consuming pasted content.
I've also loosened the check to be that any of the pasted elements appears in the selection-cells, and made an alternative (which awaits all of them instead of any) publicly available

labkey-chrisj:
One thing I could do here is get rid of the overload that causes `pasteFromCell` to conditionally validate and instead make the `waitForPasteContent `and `waitForAnyPasteContent `methods public as chainable method calls after pasting.

Another thing I could do is just make `getSelectionCellTexts `public and have tests roll their own asserts when they want or need to verify pasted content

I could also wrap the asserts in those methods with a `DeferredErrorCollector `so they're non-fatal

The ability to do this sort of validation/synchronization occurred to me as needed while testing early iterations of plate configuration- the grid was (at that time) not form-submitted, it was dynamic and when there was no waiting for the form-submit button to become clickable in order to leave the page, tests would routinely race ahead of the product and it was hard to know if pasted content had made it into the app or not- (it sure seemed like it wasn't, and that uncertainty seemed at odds with our objectives of creating low-noise, stable tests).
Also, because the plateSet create page doesn't wait for input to be valid before showing the form buttons (it errors if not all plate types have values), I had to wait for valid state myself to prevent tests from running ahead of paste and failing

Also, the results from running tests against this branch (with full validation) seems to show that it was the error cases (in which paste-data into a lookup cell didn't resolve anything but still visually has pasted text in it) that came up with failures.  Perhaps those cells with errors don't appear in `getSelectionCellTexts`?  This could be a useful way of diffing pasted inputs from what's actually there in non-error cells

src/org/labkey/test/components/ui/grids/EditableGrid.java
```diff
@@ -595,9 +611,51 @@ public EditableGrid pasteFromCell(int row, String columnName, String pasteText)
...
+    /**
+     * Awaits any elements (except for empty, or space-only) of the pasted content to be present in the grid
+     * @param pasteContent  tab-separated text
+     */
+    protected void waitForAnyPasteContent(String pasteContent)
+    {
+        // split pasteContent into its parts
+        var contentParts = pasteContent.replace("\n", "\t").split("\t");
+        // filter out empty and space-only values
+        var filteredParts = Arrays.stream(contentParts).filter(a-> !a.isEmpty() && !a.equals(" ")).collect(Collectors.toList());
```
labkey-danield, line 629: This list is not going to be modified so you can just call .toList here. 
`var filteredParts = Arrays.stream(contentParts).filter(a-> !a.isEmpty() && !a.equals(" ")).toList());`

src/org/labkey/test/components/ui/grids/EditableGrid.java
```diff
@@ -595,9 +611,51 @@ public EditableGrid pasteFromCell(int row, String columnName, String pasteText)
...
+    public void waitForPasteContent(String pasteContent)
+    {
+        // split pasteContent into its parts
+        var contentParts = pasteContent.replace("\n", "\t").split("\t");
+        // filter out empty and space-only values
+        var filteredParts = Arrays.stream(contentParts).filter(a-> !a.isEmpty() && !a.equals(" ")).collect(Collectors.toList());
+        await().atMost(Duration.ofSeconds(2))
+                .untilAsserted(()-> Assertions.assertThat(getSelectionCellTexts())
+                        .containsAll(filteredParts));
+    }
```
labkey-danield, line 649: It doesn't look like this method is used.
This should not be public.

labkey-danield (approved):
You have an unused method that should not be public.

src/org/labkey/test/components/ui/grids/EditableGrid.java
labkey-chrisj: It is offered as an option to changing the behavior of `pasteFromCell`; if it's public callers can chain this method after pasting content if they want to assert that content.
Please, if you're going to press for no change and then press for no new public methods, consider that both of these options at once doesn't make sense.  If the method is private and not used internally, it might as well not exist and the purpose for this whole PR is... useless

If you want to tell me what to do with the code, please let's talk about it instead of doing this in text

PR #1836, opened 2024-02-23
Test Automation for Date-only and Time-only Fields

labkey-danield:
#### Test automation for the date-only and time-only fields.

I _honestly tried_ to put the new date-only and time-only tests into ListTest. However the ListTest is an older tests and several "one-off" checks have been added to it that made it difficult to insert new tests. Several of the older tests use a globally defined list that made it very difficult to add new columns to without causing other tests to fail. Several tests used column positions as part of validation which also made adding new fields difficult. I cleaned up the List test a bit but it was still easier to add the date and time only tests in a separate file.

This change also include a consolidation of the site "look and feel" page with the folder "configuration" page. Some of the tests for this story needed to validate the reconciliation of formate setting in site vs. folder. Consolidating the two pages made the tests cleaner.

@XingY The new test cases for date & time only fields are:

- [AssayRenameExportImportTest](https://github.com/LabKey/testAutomation/pull/1836/files#diff-3b0d97147b341ea1af1fa7212600758b8bac45d44ebbd804817a0ad8b78e2995)
- [ListDateAndTimeTest](https://github.com/LabKey/testAutomation/pull/1836/files#diff-b4c0a11348f090c4dfd5e1216bb1f4610e41c472bd4c9b1142abdc088bcfcabc)
- [DataClassTest](https://github.com/LabKey/testAutomation/pull/1836/files#diff-ccb3d72b2e1709d52a21a8ff2d3f3b1466ca61d5d7f8eb0bd948de3393a1eca2)
- [GpatAssayTest](https://github.com/LabKey/testAutomation/pull/1836/files#diff-b07ade1f8cdbda03f71e121dee3d1bbd5e722a1504e4099bd675fcec80f9d585)
- [ParsingPatternForDateTest](https://github.com/LabKey/testAutomation/pull/1836/files#diff-faa2d77ed0d3cda4fceeff97deb983783c8917324851638b445676f17aeeb2e6)
- [ProjectSettingsTest]([src/org/labkey/test/tests/ProjectSettingsTest.java](https://github.com/LabKey/testAutomation/pull/1836/files#diff-55382fad5c4779c0850ff5ec587208c37f39377e869fc6658ea883f8fe492898))

There are also several data files that have been updated I'll leave up to you if want to review.

The [ListTest](https://github.com/LabKey/testAutomation/pull/1836/files#diff-c8d123a9764ac13f825fa342636169063276c7375bfff5603f0853522c4abfbe) was also change but was primarily for clean up, using new helpers, etc... There are no specific date and time only tests in this file.

I have [TeamCity run](https://teamcity.labkey.org/buildConfiguration/bt20?branch=dateAndTimeAutomation&buildTypeTab=overview&mode=builds) that exercises all of theses tests.

@labkey-tchad I believe you have already reviewed the setting page changes. I haven't changed them since the initial draft review. For completeness the changes mostly focused on the form change are:

- [BaseSettingsPage](https://github.com/LabKey/testAutomation/pull/1836/files#diff-8f151f7aeac6a8fb91110493492ada2fb08dc0e686741a820d663511587ef12b)
- [ProjectSettingsPage](https://github.com/LabKey/testAutomation/pull/1836/files#diff-a57dd15e963e6aed4581bd9cdfbd7de3aa82578ec5386a25176f15d3e35054ac)
- [LookAndFeelSettingsPage](https://github.com/LabKey/testAutomation/pull/1836/files#diff-60de459b797da026a785319c181cb91fe6c9de2565c26416c69076ece383c8f1)

[DiscussionLinkTest.java](https://github.com/LabKey/testAutomation/pull/1836/files#diff-7639ec95791436766b816510880a699ca65895fbfafa68df5a4c8d1240e2a7ae) was updated to use the new settings page.

A new method, setTypeWithDialog, was added to [DomainFieldRow.java](https://github.com/LabKey/testAutomation/pull/1836/files#diff-93d81af484cd9ec10be38a17ab418ef2915aff275c3db8398b5f73a5bce942eb). This was added to accommodate the test that validate changing a dateTime field to a date or time only field.

#### Related Pull Requests
* https://github.com/LabKey/limsModules/pull/13
* https://github.com/LabKey/premiumModules/pull/2

#### Changes
* Added DomainFieldRow.setTypeWithDialog. This returns the dialog so the warning message when converting a field can be validated.
* Consolidated the settings page for Site and Folder.
    * Created BaseSettingsPage which has the shared functionality between the site and folder settings page.
    * Changed LookAndFeelSettingsPage to derive from BaseSettingsPage
    * Changed ProjectSettingsPage to derive from BaseSettingsPage
    * Had to update a few tests because of this change.
* Added tests to ProjectSettingsTest to validate site and folder settings are scoped correctly.
* Changed VarListDefinition.addField to validate that the assumed key field is of a valid key type.
* Added date-only and time-only tests to ParsingPatternForDateTest
* Added ListDateAndTimeTest.
* "Cleaned up" ListTest
    * This test needs to be refactored. 
* Added AssayTest
    * Validate date and time only fields in results and run fields.
* Added DataClasses
    * Possibly redundant with the list tests and may not add much value.


src/org/labkey/test/tests/ParsingPatternForDateTest.java
```diff
@@ -123,16 +203,36 @@ public void testAdditionalParsingPatternForPipelineJobs()
...
+
+        checker().verifyEquals("Incorrect date parsed while importing", Arrays.asList("2020-11-29", "2020-11-28", "2024-02-05")
                 , table.getColumnDataAsText("dateCol"));
+
+        checker().verifyEquals("Incorrect time parsed while importing", Arrays.asList("00:23:00", "00:23:00", "16:36:00")
+                , table.getColumnDataAsText("timeCol"));
     }
 
-    private void setUpDataParsing(String pattern)
+    private void setAdditionalParsingPatterns(String dateTimePattern, String datePattern, String timePattern)
```
XingY: Do we have any coverage for site/project level setting for those fields?

src/org/labkey/test/tests/ParsingPatternForDateTest.java
```diff
@@ -65,54 +78,121 @@ private void doSetup() throws IOException, CommandException
...
     }
 
     @Test
     public void testAdditionalParsingPatternDateAndTime()
     {
-        log("Setting the parsing pattern");
-        String pattern = "ddMMMyyyy:HH:mm:ss";
-        setUpDataParsing(pattern);
+        log("Setting the additional parsing patterns.");
+        String dateTimePattern = "ddMMMyyyy:HH:mm:ss";
```
XingY: Do we have test coverage for US vs non-US parsing mode?

src/org/labkey/test/tests/list/ListDateAndTimeTest.java
```diff
@@ -0,0 +1,1288 @@
...
+        log(String.format("Set the format for the other DateTime field to '%s'. This field will be converted to a tine-only field.", dtFormatTime));
+
+        _listHelper.createList(PROJECT_NAME, listName, ListHelper.ListColumnType.AutoInteger, "Key",
+                new FieldDefinition(dateTimeToDateCol, FieldDefinition.ColumnType.DateAndTime).setFormat(dtFormatDate).setLabel(dateTimeToDateCol),
+                new FieldDefinition(dateTimeToTimeCol, FieldDefinition.ColumnType.DateAndTime).setFormat(dtFormatTime).setLabel(dateTimeToTimeCol)
+        );
+
+        List<Date> dates = new ArrayList<>();
+
+        dates.add(new Calendar.Builder()
```
XingY, line 1270: For regression testing for issue 47341, it would be useful to have a datetime value with actual time portion that's before 1970. The test can test 2 things: before converting to time field, export will export datetime as expected. After converting to time, the column in UI and exported are both as expected. This can be a separate test case.

src/org/labkey/test/tests/ParsingPatternForDateTest.java
```diff
@@ -65,54 +78,121 @@ private void doSetup() throws IOException, CommandException
...
+        }
+        clickButton("Submit");
+
+        log("Bulk import with values in additional parsing patterns.");
         listTable.clearAllFilters();
         listTable.clickImportBulkData();
-        click(Locator.tagWithText("h3", "Upload file (.xlsx, .xls, .csv, .txt)"));
+        click(Locator.tagContainingText("h3", "Upload file"));
+
+        // Note: the date and time columns in this xlsx are formatted as date and time types. Excel does not have a date-time format, that is a string.
```
XingY: Hmm.. Date/Time/Datetime in excel are all "Numbers", with different format. The field should be a "Number" type, not a string type.

XingY: Though it would be good to have separate coverage that test import from excel with "String" type date/time/datetime too.

XingY (commented):
Will there be a separate story for checking date/time values in Assay Run domains? Samples/dataclass can be checked additionally in apps, dataset can probably be skipped. But it would be good to have some test coverage for Assay Run fields since its backend implementation is different

src/org/labkey/test/tests/ParsingPatternForDateTest.java
labkey-danield: I have added tests that validate parsing mode set at site and project levels.

src/org/labkey/test/tests/ParsingPatternForDateTest.java
labkey-danield: I have added a xlsx file for import that has Date, Time and DateTime fields in text format.

src/org/labkey/test/tests/ParsingPatternForDateTest.java
labkey-danield: I've added a test for non-US parsing.

src/org/labkey/test/params/list/VarListDefinition.java
```diff
@@ -31,11 +33,19 @@ public ListDefinition setFields(List<? extends PropertyDescriptor> fields)
...
         }
         return super.setFields(fields);
     }
 
+    private void assertKeyIsValidType(PropertyDescriptor field)
+    {
+        String fieldRangeURI = field.getRangeURI();
+        Assert.assertTrue(String.format("Only 'integer' or 'text' fields can be made the primary key, found '%s'.", fieldRangeURI),
+                fieldRangeURI.equalsIgnoreCase("int") || fieldRangeURI.equalsIgnoreCase("string"));
+    }
```
labkey-tchad: Why add this extra validation here? Does the server not throw an error if you specify an invalid key type?

src/org/labkey/test/pages/core/admin/ProjectSettingsPage.java
```diff
@@ -41,14 +39,27 @@ public static ProjectSettingsPage beginAt(WebDriverWrapper driver, String contai
...
     {
-        return elementCache().discussionEnabled;
+        if(value)
+        {
+            elementCache().shouldInherit.check();
+        }
+        else
+        {
+            elementCache().shouldInherit.uncheck();
+        }
```
labkey-tchad: I think this component has a `set` method.
```suggestion
        elementCache().shouldInherit.set(value);
```

src/org/labkey/test/tests/ProjectSettingsTest.java
```diff
@@ -20,61 +20,175 @@
...
+        if(supportLink.isEmpty())
+        {
+            checker().verifyTrue("Support link should be empty.",
+                    settingsPage.getSupportLink().isEmpty());
+        }
+        else
+        {
+            checker().verifyEquals("Support link value not as expected.",
+                    supportLink, settingsPage.getSupportLink());
+        }
```
labkey-tchad: Conditionalizing all of these assertions makes this look much more complicated that they actually are. These each can be done in a single line.
```suggestion
        checker().verifyEquals("Help menu should be " + (helpMenu ? "checked" : "unchecked"), helpMenu, settingsPage.getHelpMenu());
        checker().verifyEquals("Support link value not as expected.", supportLink, settingsPage.getSupportLink());
        //etc.
```

src/org/labkey/test/tests/list/ListDateAndTimeTest.java
labkey-danield: There are test dates that happen before 1970 (Date-only, Time-only and DateTime values). I think the test date is 1952, or sometime in the 50's, so we should be good there.

I've added test that validate the round trip import / export issue with excel called out in issue 47341. The roundtrip does include dates before 1970.

src/org/labkey/test/params/list/VarListDefinition.java
labkey-danield: I thought I had a scenario where it did not, or it didn't give an error until the list was populated.
In either case I could not repro it and have removed the check.

src/org/labkey/test/tests/ProjectSettingsTest.java
labkey-danield: Changed.

src/org/labkey/test/pages/core/admin/ProjectSettingsPage.java
labkey-danield: Changed.

labkey-danield: Changed.

src/org/labkey/test/tests/GpatAssayTest.java
```diff
@@ -128,16 +127,49 @@ public void testSteps()
 
         log("Import XLSX GPAT assay");
         assayDesignerPage = startCreateGpatAssay(GPAT_ASSAY_XLSX, ASSAY_NAME_XLSX);
-        setAssayResultsProperties(assayDesignerPage, 9);
+        setAssayResultsProperties(assayDesignerPage, 11);
+
+        DomainFormPanel domainFormPanel = assayDesignerPage.expandFieldsPanel("Run Fields");
+        domainFormPanel.addField(new FieldDefinition("Date", FieldDefinition.ColumnType.Date));
```
XingY, line 140: Is there coverage for converting assay run fields from datetime -> date/time?

XingY (approved):

src/org/labkey/test/tests/list/ListDateAndTimeTest.java
```diff
@@ -0,0 +1,1434 @@
...
+    {
+        ListDateAndTimeTest init = (ListDateAndTimeTest) getCurrentTest();
+        init.resetSiteSettings();
+    }
+
+    public void resetSiteSettings()
+    {
+        log("Reset site settings.");
+        LookAndFeelSettingsPage.beginAt(this).reset();
+    }
```
labkey-tchad: I don't think this will actually reset anything. The "Reset" button pops up an alert when I do it locally.
On the bright side, doing cleanup in `@After/AfterClass` is a bad idea anyway (especially if it does any navigation); furthermore, the reset button just POSTs to `admin-resetProperties.view`, which the test can easily do without any extra navigation.
The typical place to do this sort of cleanup would be to add another step to `TestScrubber.cleanSiteSettings`.

src/org/labkey/test/tests/list/ListDateAndTimeTest.java
```diff
@@ -0,0 +1,1434 @@
...
+        cmd.setColumns(Arrays.asList("Name", "Container"));
+        cmd.addFilter("Container/DisplayName", projectName, Filter.Operator.getOperator("EQUAL"));
+        cmd.addFilter("Name", listName, Filter.Operator.getOperator("EQUAL"));
+        SelectRowsResponse response = cmd.execute(cn, projectName);
+
+        if(response.getRowCount().intValue() != 0)
+        {
+            _listHelper.beginAtList(projectName, listName);
+            _listHelper.deleteList();
+        }
```
labkey-tchad: There's a helper for this.
```suggestion
        DomainUtils.ensureDeleted(projectName, "lists", listName);
```

src/org/labkey/test/tests/list/ListDateAndTimeTest.java
```diff
@@ -0,0 +1,1434 @@
...
+        sleep(500);
+
+        // Use java's Date object and SimpleDateFormatter (not strings) to enter the data.
+        _defaultDateFormat = new SimpleDateFormat(settingsPage.getDefaultDateDisplay());
+        _defaultTimeFormat = new SimpleDateFormat(settingsPage.getDefaultTimeDisplay());
+        _defaultDateTimeFormat = new SimpleDateFormat(settingsPage.getDefaultDateTimeDisplay());
+
+        deleteListIfPresent(getProjectName(), listName);
+
+    }
```
labkey-tchad: This does a lot of unnecessary UI groveling. Only getting the default formats requires going through the UI and that only needs to happen once (during `doSetup` would be good). The rest of this (resetting the look and feel and deleting the list) can be done via API.

labkey-tchad (changes requested):
There are a lot of places that could be sped up by using APIs

src/org/labkey/test/tests/GpatAssayTest.java
labkey-danield: There is now.

src/org/labkey/test/tests/list/ListDateAndTimeTest.java
labkey-danield: Changed.

src/org/labkey/test/tests/list/ListDateAndTimeTest.java
labkey-danield: Changed.

src/org/labkey/test/tests/list/ListDateAndTimeTest.java
labkey-danield: I'll follow up with you on this feedback offline.

src/org/labkey/test/TestScrubber.java
```diff
@@ -140,6 +140,16 @@ public void cleanSiteSettings()
             TestLogger.error("Failed to reset active user limit after test", e);
         }
 
+        try
+        {
+            new SimplePostCommand("admin", "resetProperties")
+                    .execute(createDefaultConnection(), "/");
```
labkey-tchad: Move this to a static method in `BaseSettingsPage` to be shared with test.

src/org/labkey/test/tests/ParsingPatternForDateTest.java
```diff
@@ -44,16 +65,52 @@ private void doSetup() throws IOException, CommandException
...
+    {
+        log("Reset site settings.");
+        try
+        {
+            resetSettings("/");
+        }
+        catch (IOException | CommandException e)
+        {
+            TestLogger.error("Failed to reset site look and settings.", e);
+        }
```
labkey-tchad: I don't think this error should be ignored. Swallowing the exception could lead to hard-to-diagnose failures.

src/org/labkey/test/tests/ParsingPatternForDateTest.java
```diff
@@ -44,16 +65,52 @@ private void doSetup() throws IOException, CommandException
...
+    {
+        log("Reset project settings.");
+        try
+        {
+            resetSettings(getProjectName());
+        }
+        catch (IOException | CommandException e)
+        {
+            TestLogger.error("Failed to reset project settings.", e);
+        }
```
labkey-tchad: Rethrow this instead of ignoring it.
```suggestion
        catch (IOException | CommandException e)
        {
            throw new RuntimeException("Failed to reset project settings.", e);
        }
```

src/org/labkey/test/tests/ParsingPatternForDateTest.java
```diff
@@ -65,74 +122,352 @@ private void doSetup() throws IOException, CommandException
...
+
+    private void createList(String listName) throws IOException, CommandException
+    {
+        // Delete the list if it already exists.
+        goToProjectHome();
+        if(goToManageLists().getGrid().getListNames().contains(listName))
+        {
+            _listHelper.goToList(listName);
+            _listHelper.deleteList();
+        }
```
labkey-tchad: Did `DomainUtils.ensureDeleted(getProjectName(), "lists", listName);` not work for this?

src/org/labkey/test/tests/ProjectSettingsTest.java
```diff
@@ -86,53 +160,167 @@ public static void setupProject()
...
+            bulkImportData.append(String.format("%s\t%s\t%s\n",
+                    listRow.get(DT_LIST_DATE_COL),
+                    listRow.get(DT_LIST_TIME_COL),
+                    listRow.get(DT_LIST_DATETIME_COL))
+            );
+        }
+
+        _listHelper.bulkImportData(bulkImportData.toString());
+
     }
```
labkey-tchad: This isn't a list or domain designer focused test, it shouldn't spend a bunch of time in UI that is tested elsewhere. Use API helpers for list creation (`IntListDefinition`) and data import (`TestDataGenerator`).

src/org/labkey/test/tests/ProjectSettingsTest.java
```diff
@@ -20,61 +20,135 @@
...
+    {
+        LookAndFeelSettingsPage settingsPage = LookAndFeelSettingsPage.beginAt(this);
+        settingsPage.reset();
+    }
+
+    private void resetProjectSettings()
+    {
+        ProjectSettingsPage settingsPage = ProjectSettingsPage.beginAt(this, PROJ_CHANGE);
+        settingsPage.reset();
     }
```
labkey-tchad: Use API to reset site/project settings

src/org/labkey/test/tests/list/ListDateAndTimeTest.java
```diff
@@ -0,0 +1,1368 @@
...
+import java.util.Map;
+
+@Category({Daily.class, Data.class, Hosting.class})
+public class ListDateAndTimeTest extends BaseWebDriverTest
+{
+
+    private static final String PROJECT_NAME = "List Date And Time Test";
+    private SimpleDateFormat _defaultDateFormat = new SimpleDateFormat("yyyy-MM-dd");
+    private SimpleDateFormat _defaultTimeFormat = new SimpleDateFormat("HH:mm:ss");
+    private SimpleDateFormat _defaultDateTimeFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm");
```
labkey-tchad: It is misleading to give these initial values since they get defined in `doSetup`. They also need to be static for tests to see the value set in `doSetup` (each `@Test` is run with a different instance of the test class).
```suggestion
    // Fetched from look and feel page in 'doSetup'
    private static SimpleDateFormat _defaultDateFormat = null;
    private static SimpleDateFormat _defaultTimeFormat = null;
    private static SimpleDateFormat _defaultDateTimeFormat = null;
```

src/org/labkey/test/tests/list/ListDateAndTimeTest.java
```diff
@@ -0,0 +1,1368 @@
...
+
+    private void validateListDataInUI(DataRegionTable table, List<Map<String, String>> expectedData)
+    {
+        checker().verifyEquals("Number of rows in the UI list not equal to number of expected rows.",
+                expectedData.size(), table.getDataRowCount());
+
+        int rowIndex = 0;
+        for(Map<String, String> expectedRowMap : expectedData)
+        {
+            // Protect against expecting more rows than are present in the grid.
```
labkey-tchad: Why protect against this? You've already verified that the table has the correct number of rows.

labkey-tchad (approved):
Still a lot of places that could be sped up by using APIs. `ListDateAndTimeTest` creates a lot of lists via UI.

src/org/labkey/test/pages/core/admin/BaseSettingsPage.java
```diff
@@ -0,0 +1,303 @@
+package org.labkey.test.pages.core.admin;
+
+import org.labkey.test.Locator;
+import org.labkey.test.components.html.RadioButton;
+import org.labkey.test.pages.LabKeyPage;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+public class BaseSettingsPage extends LabKeyPage<LabKeyPage<?>.ElementCache>
```
labkey-tchad: If you change how this is declared, you won't have to override `elementCache` below.
```suggestion
public class BaseSettingsPage extends LabKeyPage<BaseSettingsPage.ElementCache>
```

src/org/labkey/test/TestScrubber.java
labkey-danield: Done.

src/org/labkey/test/tests/ParsingPatternForDateTest.java
labkey-danield: Changed  to throw a new runtime exception.

src/org/labkey/test/tests/ParsingPatternForDateTest.java
labkey-danield: Changed.

src/org/labkey/test/tests/ParsingPatternForDateTest.java
labkey-danield: It did, but I must have missed this reference or undid the change as I was doing other updates.
Added it back.

src/org/labkey/test/tests/ProjectSettingsTest.java
labkey-danield: Changed.

src/org/labkey/test/tests/ProjectSettingsTest.java
labkey-danield: Changed.

src/org/labkey/test/tests/list/ListDateAndTimeTest.java
labkey-danield: Changed.

src/org/labkey/test/tests/list/ListDateAndTimeTest.java
labkey-danield: So it does not get an index out of bounds error and stops the entire tests.

src/org/labkey/test/pages/core/admin/BaseSettingsPage.java
labkey-danield: Changed.

PR #1840, opened 2024-02-27
Issue 45315: Remove DomainEditorServiceBase and GWT-based assay infer-from-file UI

cnathe:
#### Rationale
See related PR for rationale.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5277

#### Changes
- Selenium test updates for file import data to use field editor instead of GWT assay import action

### DIFF
--- a/src/org/labkey/test/tests/GpatAssayTest.java
+++ b/src/org/labkey/test/tests/GpatAssayTest.java
@@ -30,7 +30,6 @@ import org.labkey.test.WebTestHelper;
 import org.labkey.test.categories.Assays;
 import org.labkey.test.categories.BVT;
 import org.labkey.test.components.domain.DomainFormPanel;
-import org.labkey.test.components.html.SelectWrapper;
 import org.labkey.test.pages.ReactAssayDesignerPage;
 import org.labkey.test.pages.assay.AssayBeginPage;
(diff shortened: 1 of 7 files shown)


src/org/labkey/test/tests/GpatAssayTest.java
```diff
@@ -361,82 +310,6 @@ public void testRenameAssayDesign()
 
     }
 
-    @Test
-    public void testRenameSharedAssayDesign()
```
labkey-jeckels, line 365: I would have expected this to be refactored to match the React UI instead of deleting the test case completely. Is it redundant or irrelevant now?

cnathe: The other test cases from this file that "take a file from the file system, click Import Data > Create Standard Assay Design, and import the file" were refactored here to use the React UI. This test case does that but was selecting for the assay design to be created in the /Shared location. That part is no longer supported in the new UI/UX (as noted in the PR comment) so I just removed this test case.

labkey-jeckels (approved):

src/org/labkey/test/tests/GpatAssayTest.java
labkey-jeckels: Got it - I missed that in the other PR's description.

PR #1864, opened 2024-03-20
use java command wrappers and params to create plates

labkey-chrisj:
#### Rationale
This adds command/response wrappers to create plate sets used in biologics for now, (I'm not sure if this will be limited to lims modules or not, so creating here)

The reason to create via API at this point is that the UI doesn't yet support creating child or parent-type plate sets, this allows us to get automated regression coverage in place until it does

#### Related Pull Requests
https://github.com/LabKey/limsModules/pull/65

#### Changes
create command/response/params classes for use in testing plate automation

### DIFF
--- a/src/org/labkey/remoteapi/plate/PlateSetParams.java
+++ b/src/org/labkey/remoteapi/plate/PlateSetParams.java
@@ -7,7 +7,7 @@ public class PlateSetParams
 {
     private String _name = null;
     private String _description = null;
-    private String _type = null;
+    private PlateSetType _type = null;
     private Integer _id = null;
     private Integer _parentPlateSetId = null;
 
@@ -19,7 +19,7 @@ public class PlateSetParams
     {
         _name = json.getString("name");
         _description = json.optString("description", null);
-        _type = json.optString("type", null);
+        _type = PlateSetType.fromName(json.optString("type", null));
         _id = json.getInt("rowId");
         _parentPlateSetId = json.optIntegerObject("parentPlateSetId", null);
     }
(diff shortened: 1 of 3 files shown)


src/org/labkey/remoteapi/plate/PlateSetParams.java
```diff
@@ -0,0 +1,92 @@
...
+            json.put("name", _name);
+        if (_description != null)
+            json.put("description", _description);
+        if (_type != null)
+            json.put("type", _type);
+        if (_id != null)
+            json.put("rowId", _id);
+        if (_parentPlateSetId != null)
+            json.put("parentPlateSetId", _parentPlateSetId);
+        return json;
```
labkey-tchad: These `null` checks actually aren't necessary. Much to our chagrin, `JSONObject` doesn't allow `null` values. A `put` with a `null` value actually removes any existing entry and does nothing if there is no existing entry.
```suggestion
        JSONObject json = new JSONObject();
        json.put("name", _name);
        json.put("description", _description);
        json.put("type", _type);
        json.put("rowId", _id);
        json.put("parentPlateSetId", _parentPlateSetId);
        return json;
```

labkey-tchad (approved):
Just a small nit about some odd `JSONObject` behavior.

src/org/labkey/remoteapi/plate/PlateSetParams.java
labkey-chrisj: Thank you! (I didn't know that, it feels good to learn something new)
This probably means we have a lot of other writing-of-json to update accordingly; in writing this I followed the pattern from `NotebookParams`

labkey-nicka (approved):

PR #1878, opened 2024-04-02
test components for charting in apps

labkey-chrisj:
#### Rationale
We have added the ability (in biologics, with an experimental feature) to compose, edit, and delete charts on queryGrids. This change adds and updates classes necessary to interact with new UI in the apps.

The new Dialog `QueryChartDialog` supports create and edit of charts, and could (if we think that's a good idea) be refactored into 2 dialog classes that would only have methods related to buttons of the dialog in that mode.  Here it is in create mode: 
image
and in edit mode: 
image

Because we've introduced the ability to add, edit, save, and delete charts in the apps, I've taken the liberty of creating a new component, QueryChartPanel (which contains the chart in the QueryGrid when charts or reports are shown) to surface the ability to find the chart's svg, to click to edit the chart, and to close the chart view- it's highlit in red, here:
image

#### Related Pull Requests
https://github.com/LabKey/limsModules/pull/105 <- contains test coverage using new/changed components in this PR

#### Changes

- [x] new create/edit chart dialog, `QueryChartDialog`
- [ ] new chart/report panel in `QueryGrid`, `QueryChartPanel `(which provides the means to get the edit dialog, to dismiss the chart when it is shown, to find the chart/report)

### DIFF
--- a/src/org/labkey/test/components/react/QueryChartDialog.java
+++ b/src/org/labkey/test/components/react/QueryChartDialog.java
@@ -1,0 +1,323 @@
+package org.labkey.test.components.react;
+
+import org.labkey.test.Locator;
+import org.labkey.test.WebDriverWrapper;
+import org.labkey.test.components.bootstrap.ModalDialog;
+import org.labkey.test.components.html.Checkbox;
+import org.labkey.test.components.html.Input;
+import org.labkey.test.components.ui.grids.QueryGrid;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.ui.ExpectedConditions;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.labkey.test.components.html.Input.Input;
+
+public class QueryChartDialog extends ModalDialog
+{
+    private final QueryGrid _queryGrid;
+
+    /**
+     * Contains
+     * @param title "Create Chart" or "Edit Chart"
+     * @param driver the webDriver
+     * @param grid the associated QueryGrid
+     */
+    public QueryChartDialog(String title, WebDriver driver, QueryGrid grid)
+    {
+        super(new ModalDialogFinder(driver).withTitle(title));
+        _queryGrid = grid;
+    }
+
+    public QueryChartDialog setName(String value)
+    {
+        elementCache().nameInput.set(value);
+        return this;
+    }
+
+    public String getName()
+    {
+        return elementCache().nameInput.get();
+    }
+
+    public QueryChartDialog setShared(boolean checked)
+    {
+        elementCache().sharedCheckbox.set(checked);
+        return this;
+    }
+
+    public boolean getShared()
+    {
+        return elementCache().sharedCheckbox.get();
+    }
+
+    // field selects
+
+    /*
+        X axis is an option for bar, box, line, scatter charts
+     */
+    public QueryChartDialog selectXAxis(String field)
+    {
+        elementCache().reactSelectByLabel("X Axis *").select(field);
+        return this;
+    }
+
+    public String getXAxisSelection()
+    {
+        return elementCache().reactSelectByLabel("X Axis").getValue();
+    }
+
+    /*
+        Y Axis is an option for bar, box, line, scatter charts
+     */
+    public QueryChartDialog selectYAxis(String field)
+    {
+        elementCache().reactSelectByLabel("Y Axis").select(field);
+        return this;
+    }
+
+    public String getYAxisSelection()
+    {
+        return elementCache().reactSelectByLabel("Y Axis").getValue();
+    }
+
+    /*
+        groupBy is an option for bar charts only
+     */
+    public QueryChartDialog selectGroupBy(String field)
+    {
+        elementCache().reactSelectByLabel("Group By").select(field);
+        return this;
+    }
+
+    /*
+     * color is an option for Box, Scatter charts
+     */
+    public QueryChartDialog selectColor(String field)
+    {
+        elementCache().reactSelectByLabel("Color").select(field);
+        return this;
+    }
+
+    /*
+        shape is an option for Scatter, Box charts
+     */
+    public QueryChartDialog selectShape(String field)
+    {
+        elementCache().reactSelectByLabel("Shape").select(field);
+        return this;
+    }
+
+    /*
+        series is an option for line
+     */
+    public QueryChartDialog selectSeries(String field)
+    {
+        elementCache().reactSelectByLabel("Series").select(field);
+        return this;
+    }
+
+    /*
+        categories is an option for pie charts
+     */
+    public QueryChartDialog selectCategories(String field)
+    {
+        elementCache().reactSelectByLabel("Categories").select(field);
+        return this;
+    }
+
+    // chart type selection
+    /*
+        Note: Chart Type is not settable when opened for edit/ it is only settable
+        when creating the chart
+     */
+    public QueryChartDialog setChartType(CHART_TYPE chartType)
+    {
+        elementCache().selectChartType(chartType);
+        return this;
+    }
+
+    public CHART_TYPE getSelectedChartType()
+    {
+        return elementCache().getSelectedChartType();
+    }
+
+    public WebElement waitForPreview()
+    {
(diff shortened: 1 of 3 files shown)


src/org/labkey/test/components/react/QueryChartDialog.java
```diff
@@ -0,0 +1,330 @@
...
+    {
+        elementCache().reactSelectByLabel("Y Axis").select(field);
+        return this;
+    }
+
+    public String getYAxisSelection()
+    {
+        return elementCache().reactSelectByLabel("Y Axis").getValue();
+    }
+
```
labkey-danield, line 91: Wouldn't it be useful to have a getSelectionOptions for XAxis, YAxis & GroupBy? It seems like there should be a test that validates the correct fields/columns are returned for the options.

src/org/labkey/test/components/react/QueryChartDialog.java
```diff
@@ -0,0 +1,330 @@
...
+        WebDriverWrapper.waitFor(()-> elementCache().isPreviewPresent(),
+                "the preview was not present in time", 2000);
+        return elementCache().svg();
+    }
+
+    public WebElement svg()
+    {
+        waitForPreview();
+        return elementCache().svg();
+    }
```
labkey-danield: This method is redundant with the waitForPreview() method.

src/org/labkey/test/components/react/QueryChartDialog.java
```diff
@@ -0,0 +1,330 @@
...
+            if (getSelectedChartType().equals(chartType))
+                return;
+
+            var el = chartBuilderType.withAttribute("data-name", chartType.getChartType())
+                    .waitForElement(this, 1500);
+            getWrapper().shortWait().until(ExpectedConditions.elementToBeClickable(el));
+            el.click();
+            WebDriverWrapper.waitFor(()-> getSelectedChartType().equals(chartType),
+                    "The requested chart type did not become selected", 2000);
+        }
```
labkey-danield: These aren't really elements, these are methods. They shouldn't be in the element cache.

src/org/labkey/test/components/react/QueryChartDialog.java
```diff
@@ -0,0 +1,330 @@
...
+        {
+            return Locator.tagWithClass("div", "gray-text").waitForElement(previewContainer(), 1500).getText();
+        }
+
+        private Locator previewBodyLoc = Locator.tagWithClass("div", "chart-builder-preview-body");
+        private Locator svgLoc = Locator.tagWithClass("div", "svg-chart__chart");
+        public boolean isPreviewPresent()
+        {
+            return previewBodyLoc.existsIn(previewContainer()) && svgLoc.existsIn(previewContainer());
+        }
```
labkey-danield: Not really an element.

src/org/labkey/test/components/react/QueryChartDialog.java
```diff
@@ -0,0 +1,330 @@
...
+    public QueryChartDialog setChartType(CHART_TYPE chartType)
+    {
+        elementCache().selectChartType(chartType);
+        return this;
+    }
+
+    public CHART_TYPE getSelectedChartType()
+    {
+        return elementCache().getSelectedChartType();
+    }
```
labkey-danield:
```suggestion
    public QueryChartDialog setChartType(CHART_TYPE chartType)
    {

        if (getSelectedChartType().equals(chartType))
            return this;

        // note: selecting a chart type will cause most select fields to go stale
        var el = elementCache().chartBuilderType.withAttribute("data-name", chartType.getChartType())
                .waitForElement(this, 1500);
        getWrapper().shortWait().until(ExpectedConditions.elementToBeClickable(el));
        el.click();
        WebDriverWrapper.waitFor(()-> getSelectedChartType().equals(chartType),
                "The requested chart type did not become selected", 2000);

        return this;
    }

    public CHART_TYPE getSelectedChartType()
    {
        var selectedEl = elementCache().chartBuilderType.withAttributeContaining("class", "selected")
                .waitForElement(this, 1500);
        String dataName = selectedEl.getAttribute("data-name");
        return CHART_TYPE.fromChartType(dataName);
    }
```

labkey-danield: This would clean up the elementCache and remove the methods that aren't really elements on the dialog.

src/org/labkey/test/components/react/QueryChartDialog.java
```diff
@@ -0,0 +1,330 @@
...
+        }
+
+    }
+
+    public enum CHART_TYPE{
+        Bar("bar_chart"),
+        Box("box_plot"),
+        Line("line_plot"),
+        Pie("pie_chart"),
+        Scatter("scatter_plot");
```
labkey-danield, line 339: (Something of a nit-pick comment)
Enum type should be upper case in Java: [Oracle Java Doc](https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html)

src/org/labkey/test/components/react/QueryChartPanel.java
```diff
@@ -0,0 +1,119 @@
...
+        editButton.click();
+        return new QueryChartDialog("Edit Chart", getDriver(), _queryGrid);
+    }
+
+    public String getTitle()
+    {
+        return elementCache().getTitle();
+    }
+
+    public WebElement svgChart()
```
labkey-danield: Nit: Really should name this `getSvgChart()`.

src/org/labkey/test/components/react/QueryChartPanel.java
```diff
@@ -0,0 +1,119 @@
...
+        var editButton = elementCache().editButton;
+        getWrapper().shortWait().until(ExpectedConditions.elementToBeClickable(editButton));
+        editButton.click();
+        return new QueryChartDialog("Edit Chart", getDriver(), _queryGrid);
+    }
+
+    public String getTitle()
+    {
+        return elementCache().getTitle();
+    }
```
labkey-danield:
```suggestion
    public String getTitle()
    {
        return elementCache().title.getText();
    }
```

src/org/labkey/test/components/react/QueryChartPanel.java
```diff
@@ -0,0 +1,119 @@
...
+        public WebElement headingEl = Locator.tagWithClass("div", "chart-panel__heading")
+                .findWhenNeeded(this).withTimeout(2000);
+        public WebElement editButton = Locator.tagWithAttribute("button", "title", "Edit chart")
+                .findWhenNeeded(headingEl);
+        public WebElement closeButton = Locator.tagWithAttribute("button", "title", "Hide chart")
+                .findWhenNeeded(headingEl);
+        public String getTitle()
+        {
+            return Locator.tagWithClass("div", "chart-panel__heading-title").findElement(headingEl).getText();
+        }
```
labkey-danield:
```suggestion
        public final WebElement title = Locator.tagWithClass("div", "chart-panel__heading-title").findElement(headingEl);
```

src/org/labkey/test/components/react/QueryChartPanel.java
```diff
@@ -0,0 +1,119 @@
...
+    protected ElementCache newElementCache()
+    {
+        return new ElementCache();
+    }
+
+    @Override
+    protected ElementCache elementCache()
+    {
+        return (ElementCache) super.elementCache();
+    }
```
labkey-danield, line 78: You don't need this method.

src/org/labkey/test/components/react/QueryChartPanel.java
```diff
@@ -0,0 +1,119 @@
...
+
+    protected class ElementCache extends Component<?>.ElementCache
+    {
+        public WebElement headingEl = Locator.tagWithClass("div", "chart-panel__heading")
```
labkey-danield: Nit: These should be `final`.

labkey-danield (approved):
There is some clean up you should do.
Use the IntelliJ linter it will help identify some nit-pick things to clean up.

src/org/labkey/test/components/react/QueryChartDialog.java
labkey-chrisj: Yes, those are probably going to be useful to have.  I'll implement them here while we're thinking about it

PR #1894, opened 2024-04-12
LKSM Test Fixes

labkey-danield:
#### Rationale
Adding some more logging in SampleTypeAPIHelper.getRowIdsForSamples. There have been some random failures where getRows call returned nothing when there should be results.

#### Related Pull Requests
* None

#### Changes
* Added some extra logging.


src/org/labkey/test/util/exp/SampleTypeAPIHelper.java
```diff
@@ -84,6 +85,30 @@ public static Map<String, Integer> getRowIdsForSamples(String containerPath, Str
...
+            cmd = new SelectRowsCommand("samples", sampleTypeName);
+            cmd.setColumns(Arrays.asList("Name"));
+
+            SelectRowsResponse responseCount = cmd.execute(connection, containerPath);
+
+            errorMsg = errorMsg + "\n" + String.format("No rows were returned with filter. The sample type '%s' has %d rows.",
+                    sampleTypeName, responseCount.getRowCount().intValue());
+
+            List<String> names = new ArrayList<>();
+            for(Map<String, Object> row : response.getRows())
```
labkey-susanh: Seems like you want to use `responseCount` here not `response` (since it has 0 rows).

labkey-susanh (approved):

src/org/labkey/test/util/exp/SampleTypeAPIHelper.java
labkey-danield: D"oh! Changed.

PR #1895, opened 2024-04-12
Update Shared Components for Date-Only and Time-Only Tests

labkey-danield:
#### Rationale
Updating shared components used in LKSM Date-only, Time-only and DateTime tests.

The WebDriverWrapper.getCurrentDateFormatString needed to be renamed top getCurrentDateTimeFormatString. Methods getCurrentDateFormatString and getCurrentTimeFormatString were added.

Previously DateTime helpers were named using only the word _date_. With the introduction of date-only field types this would causes confusion. This PR renames the ReactDatePicker to ReactDateTimePicker and is now used as the picker for DateTime and Date-only or Time-only fields. There wasn't much change needed to the component other than to (basically) expose methods that were previously private.

These three pickers are all the ReactDateTimePicker:
Screenshot 2024-04-16 at 9 00 59 AM
Screenshot 2024-04-16 at 9 01 09 AM
Screenshot 2024-04-16 at 9 01 19 AM

DetailTableEdit.setDateTimeField method has been added. This method can be used to set DateTime, Date-only or Time-only fields. The second parameter is an object type and is used to identify the field type. If a string is passed in it is used as a literal value and is simply typed into the text box. The javadoc comments above the method explain this. 

This approach of a single method to be used to set the fields is repeated in various other forms, pages, etc...

EditableGrid.setCellValue now has an overload that takes that indicates if/how the cell update should be validated. This is needed because date and time values have formatting that is applied and can be different from the text sent in. 

#### Related Pull Requests
* https://github.com/LabKey/limsModules/pull/147

#### Changes
* Disambiguate DateTime helpers from Date-only and Time-only helpers.
* Add getters for date and time format settings.
* Exposed selectDate and selectTime methods in ReactDateTimePicker
* Updated ReactDateTimePicker.clear to use the button to clear.
* Updated selectTime method to be smarter about identifying the format of the time in the dropdown list.
* Update EditableGrid to have the option to not wait for an exact match to the value sent in (date and time formatting).


src/org/labkey/test/components/ui/grids/DetailTableEdit.java
```diff
@@ -351,6 +355,64 @@ public DetailTableEdit clearSelectValue(String fieldCaption, boolean waitForSele
...
+        }
+        else if(dateTime instanceof LocalTime localTime)
+        {
+            dateTimePicker.selectTime(localTime);
+        }
+        else if(dateTime instanceof String setValue)
+        {
+            dateTimePicker.set(setValue, true);
+            // When setting DateTime, Date-only or Time-only fields by text there is a slight delay as any formatting is applied.
+            getWrapper().sleep(500);
```
labkey-tchad: Should this sleep (or some sort of wait) be baked in to `ReactDateTimePicker.set`?

labkey-tchad (approved):

src/org/labkey/test/components/ui/grids/DetailTableEdit.java
labkey-danield: Removed from here.

PR #1910, opened 2024-05-01
Corrected definition of Time column type.

labkey-danield:
#### Rationale
Use correct definition for Time column type.
[Issue 50238](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50238): Using CreateDomianCommand to create a sample type with a Time-only field results in a bad picker for the field.

@XingY: This is just an FYI for you.

#### Related Pull Requests
* None

#### Changes
* Changed parameter passed into ColumnTypeImpl for the Time column type.

### DIFF
--- a/src/org/labkey/test/params/FieldDefinition.java
+++ b/src/org/labkey/test/params/FieldDefinition.java
@@ -476,7 +476,7 @@ public class FieldDefinition extends PropertyDescriptor
         ColumnType Subject = new ColumnTypeImpl("Subject/Participant", "string", "http://cpas.labkey.com/Study#ParticipantId", null);
         ColumnType DateAndTime = new ColumnTypeImpl("Date Time", "dateTime");
         ColumnType Date = new ColumnTypeImpl("Date", "date");
-        ColumnType Time = new ColumnTypeImpl("Time", "time");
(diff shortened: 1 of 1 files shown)


XingY (approved):

labkey-tchad (commented):
Have you verified that this change fixes the problem? When I was investigating this, domains created via API and via UI both had the full 'rangeURI' in the `getDomain` response.

labkey-danield:
@labkey-tchad yes I validated using the small sample test code in the issue. It creates a sample type by:
```
        FMSampleTypeDefinition stDefinition = new FMSampleTypeDefinition(stName);
        stDefinition.setInventoryMetricUnit(InventoryMetricUnit.ML);
        stDefinition.setFields(List.of(
                new FieldDefinition(COL_DATETIME_CAPTION, FieldDefinition.ColumnType.DateAndTime),
                new FieldDefinition(COL_DATE_CAPTION, FieldDefinition.ColumnType.Date),
                new FieldDefinition(COL_TIME_CAPTION, FieldDefinition.ColumnType.Time)
        ));

        SampleTypeAPIHelper.createEmptySampleType(getProjectName(), stDefinition);

```
It then uses the time picker to select the time in a bulk edit/update dialog.
I also validated manually in the editable grid with the same sample type that the time picker is correct.

I also validated that running the test with out this change fails as per the issue.

PR #1915, opened 2024-05-03
genid is supposed to be hidden

labkey-matthewb:
#### Rationale
so call showHiddenItems()

#### Related Pull Requests
* 

#### Changes
*


labkey-susanh (approved):
I wonder why this worked before...

labkey-matthewb:
> I wonder why this worked before...

There are places we don't copy hidden by default.

PR #1917, opened 2024-05-03
Don't attempt invalid filter on wrapped column

labkey-tchad:
#### Rationale
Filter on `Pid2Consent/Value` doesn't resolve unless lookup is explicitly defined in the metadata.

#### Related Pull Requests
* #1905 

#### Changes
* Remove check when filter doesn't match lookup column
* Use API to create "Consent" lookup list

### DIFF
--- a/src/org/labkey/test/tests/LinkedSchemaTest.java
+++ b/src/org/labkey/test/tests/LinkedSchemaTest.java
@@ -28,13 +28,14 @@ import org.labkey.test.categories.Daily;
 import org.labkey.test.categories.Data;
 import org.labkey.test.components.CustomizeView;
 import org.labkey.test.pages.list.EditListDefinitionPage;
+import org.labkey.test.pages.list.GridPage;
 import org.labkey.test.pages.query.QueryMetadataEditorPage;
 import org.labkey.test.params.FieldDefinition;
 import org.labkey.test.params.list.IntListDefinition;
 import org.labkey.test.params.list.ListDefinition;
+import org.labkey.test.params.list.VarListDefinition;
 import org.labkey.test.util.ApiPermissionsHelper;
 import org.labkey.test.util.DataRegionTable;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/LinkedSchemaTest.java
```diff
@@ -490,55 +491,61 @@ public void customFilters()
...
 
-        updatedMetaData = STUDY_FILTER_METADATA
-                .replace("<!-- ", "").replace(" -->", "")
-                .replace("ParticipantId/Study", "Pid2Consent/Study");
-        checkLinkedSchema(updatedMetaData, null, 0);
-
         log("Add FK to the wrapped field.");
-        updatedMetaData = updatedMetaData.replace("ParticipantId", "Pid2Consent");
-        checkLinkedSchema(updatedMetaData, null, 2);
+        checkLinkedSchema(buildStudyFilterMetadata("StudyA", "Pid2Consent"), null, 2);
```
labkey-matthewb: We need to change the whole lookup column here.  "Demo Study" and "Pid2Consent/Study/Label" here.

src/org/labkey/test/tests/LinkedSchemaTest.java
```diff
@@ -490,55 +491,61 @@ public void customFilters()
...
 
         log("Looks good, going home.");
         goToProjectHome();
     }
 
+    private String buildStudyFilterMetadata(String study, String lookupColumn)
+    {
+        return buildStudyFilterMetadata(study).replace("ParticipantId", lookupColumn);
```
labkey-matthewb: and change "ParticipantId" here to "Participant/Study" here so we replace the whole filter attribute column="Participant/Study"

labkey-matthewb (changes requested):

labkey-matthewb (approved):

PR #1919, opened 2024-05-07
Fix for WNPRC test failure.

labkey-sweta:
#### Rationale
Fix for wnprc_ehr.WNPRC_EHRTest](https://teamcity.labkey.org/viewLog.html?buildId=2999488&buildTypeId=LabkeyTrunk_EhrPostgres&fromSakuraUI=true#testNameId-4031106202433799349) 
Unable to find element: css=select[id="assay-type-select-container"]

selectStandardAssay() clicks submit as well, which made test fail on develop.
#### Related Pull Requests
* 

#### Changes
*


labkey-tchad (approved):
Oops, looks like I caused this when resolving merge conflicts.

PR #1927, opened 2024-05-14
Disambiguate hasSelectionColumn from hasSelectAllCheckbox

labkey-chrisj:
#### Rationale
While writing test coverage for Issue 50062 (related PR) I ran into the apparently rare case where the EditableGrid has a selection column but does not have a select-all checkbox: 
image
(most editable grids that have a select-column have a select-all-checkbox, like this: 
image

Currently, the method to select a row (`selectRow`) checks to see if there's a selection column present on the grid. The method it checks with (`hasSelectColumn`) simply checks to see if the select-all checkbox is present. Because no select-all-checkbox is present, `hasSelectColumn` wrongly returns false when `selectRow` asks if there is a selection column.

This change adds a separate method (`hasSelectAllCheckbox`) to the class, and refactors callers to use that where it's the presence of the select-all checkbox they're interested in vs. the availability of row-by-row select checkboxes in the table.

#### Related Pull Requests
https://github.com/LabKey/limsModules/pull/256 (pr for issue 50062)

#### Changes

- [x] new method to check for the presence of a select-all checkbox
- [x] refactor usages of hasSelectColumn to use hasSelectAllCheckbox where appropriate


src/org/labkey/test/components/ui/grids/EditableGrid.java
```diff
@@ -998,7 +1004,9 @@ protected class ElementCache extends Component<?>.ElementCache
 
         final WebElement table = Locator.byClass("table-cellular").findWhenNeeded(this);
 
-        private final WebElement selectColumn = Locator.xpath("//th/input[@type='checkbox']").findWhenNeeded(table);
+        private final Locator selectAllHeaderCellLoc = Locator.tagWithClass("th", "grid-header-cell")
+            .withAttribute("id", "__selection__");
```
labkey-tchad, line 1017: A future editor of this file might appreciate a comment indicating why the select-all cell and select-all checkbox are defined separately.
```suggestion
        // Some grids have an placeholder cell without a select-all checkbox
        private final Locator selectAllHeaderCellLoc = Locator.tagWithClass("th", "grid-header-cell")
            .withAttribute("id", "__selection__");
```

labkey-tchad (approved):

labkey-danield (changes requested):
Can you validate that this is not a product bug and add a comment?

labkey-danield:
@labkey-chrisj did you validate that this isn't a product bug?

labkey-chrisj:
@labkey-nicka @labkey-alan @cnathe should it be considered a product bug if we see a grid with a selection row but without a select-all checkbox in it?

labkey-chrisj:
> @labkey-chrisj did you validate that this isn't a product bug?

Nick said in standup that there's no reason to have it that way, he intends to add in the missing select-all checkbox in a PR already in flight, not sure which PR

labkey-nicka:
> > @labkey-chrisj did you validate that this isn't a product bug?
> 
> Nick said in standup that there's no reason to have it that way, he intends to add in the missing select-all checkbox in a PR already in flight, not sure which PR

I've addressed this with https://github.com/LabKey/labkey-ui-components/pull/1495/commits/19535282e.

labkey-chrisj:
oh well

PR #1930, opened 2024-05-14
Check for CSP violations before each page load

labkey-tchad:
#### Rationale
Waiting until the end of a test to check for CSP violations loses some context. If we check before each navigation, we have a greater chance of getting useful information about the violation.

#### Related Pull Requests
* N/A

#### Changes
* Register a `PageLoadListener` to check for CSP violations before each navigation

### DIFF
--- a/src/org/labkey/test/util/CspLogUtil.java
+++ b/src/org/labkey/test/util/CspLogUtil.java
@@ -38,7 +38,7 @@ public class CspLogUtil
 
     public static void checkNewCspWarnings(ArtifactCollector artifactCollector)
     {
-        if (TestProperties.isCspCheckSkipped() || TestProperties.isServerRemote() || missingLog)
+        if (TestProperties.isServerRemote() || missingLog)
             return;
 
         if (!logFile.isFile())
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/util/CspLogUtil.java
```diff
@@ -38,7 +38,7 @@ private CspLogUtil() { }
 
     public static void checkNewCspWarnings(ArtifactCollector artifactCollector)
     {
-        if (TestProperties.isCspCheckSkipped() || TestProperties.isServerRemote() || missingLog)
+        if (TestProperties.isServerRemote() || missingLog)
```
labkey-danield, line 41: I trust this is intentional, just curious why the `isCspCheckSkipped` check is no longer needed.

labkey-danield (approved):

labkey-tchad:
Preparing review...

src/org/labkey/test/util/CspLogUtil.java
labkey-tchad: This way it will still log the CSP violations in the test log, even if the check is disabled

PR #1940, opened 2024-05-16
Fix issue with CustomizeGridViewDialog.removeColumn accidentally clic…

labkey-alan:
#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1494
- https://github.com/LabKey/labkey-ui-premium/pull/412
- https://github.com/LabKey/limsModules/pull/260
- https://github.com/LabKey/platform/pull/5507
- https://github.com/LabKey/testAutomation/pull/1940

#### Changes
* Fix issue with CustomizeGridViewDialog.removeColumn accidentally clicking the edit icon
    * If the column name is long enough the edit icon appears exactly in the middle of the column element
    * Clicking on the listItem has no other effect when removing the column, so I just removed the code

### DIFF
--- a/src/org/labkey/test/components/ui/grids/CustomizeGridViewDialog.java
+++ b/src/org/labkey/test/components/ui/grids/CustomizeGridViewDialog.java
@@ -324,11 +324,6 @@ public class CustomizeGridViewDialog extends ModalDialog
     public CustomizeGridViewDialog removeColumn(String column, int index)
     {
         WebElement listItem = getShownInGridListItems(column).get(index);
-
-        // Make sure the mouse is on the identified row. Tests that remove columns one after another can show the mouse
-        // hovering for a moment over the remove icon of the previous row.
-        listItem.click();
-
         WebElement removeIcon = Locator.tagWithClass("span", "view-field__action").findElement(listItem);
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/components/ui/grids/CustomizeGridViewDialog.java
```diff
@@ -324,11 +324,6 @@ public CustomizeGridViewDialog removeColumn(String column)
     public CustomizeGridViewDialog removeColumn(String column, int index)
     {
         WebElement listItem = getShownInGridListItems(column).get(index);
-
```
labkey-chrisj, line 327: This looks like it should work.
Consider leaving a comment behind to warn future-us about the way repeat-removing columns can appear to leave the cursor over the previous row

labkey-chrisj (approved):

labkey-tchad (approved):

src/org/labkey/test/components/ui/grids/CustomizeGridViewDialog.java
labkey-alan: I think that comment was outdated given that we now move the mouse cursor over the modal title every time.

PR #1947, opened 2024-05-22
Remove deprecated ListColumn and ListColumnType

### LINKED ISSUE #1949: Clean up deprecated list test classes
We've been gradually removing redundant functionality from `ListHelper`. `ListColumn` and and `ListColumnType` were deprecated in `20.4` in favor of the more widely used and generic `FieldDefinition`.

- [x] Update non-list tests to create lists via API
- [x] Fix API helpers to create integer key lists correctly
- [x] Remove deprecated classes from `ListHelper`
- [x] Code review - @labkey-danield 
- [x] TeamCity verification and merge

labkey-tchad:
#### Rationale
We've been gradually removing redundant functionality from `ListHelper`. `ListColumn` and and `ListColumnType` were deprecated in `20.4` in favor of the more widely used and generic `FieldDefinition`. Time to prune the last of the usages.
During this refactor, discovered that `IntListDefinition` and `VarListDefinition` make some incorrect assumptions.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/247

#### Changes
* Refactor uses of `ListColumn` to use `FieldDefinition`
* Refactor uses of `ListColumnType` to use `FieldDefinition.ColumnType`
* Make `IntListDefinition` work for regular (non-auto-incrementing) Integer keys

### DIFF
--- a/src/org/labkey/test/params/FieldDefinition.java
+++ b/src/org/labkey/test/params/FieldDefinition.java
@@ -25,6 +25,7 @@ import org.labkey.remoteapi.domain.PropertyDescriptor;
 import org.labkey.remoteapi.query.Filter;
 import org.labkey.test.components.html.OptionSelect;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -459,15 +460,6 @@ public class FieldDefinition extends PropertyDescriptor
 
     }
 
-    // Temporary, for 'ColumnType.values()'
-    private static final List<ColumnType> COLUMN_TYPES = List.of(
-            ColumnType.MultiLine, ColumnType.Integer, ColumnType.String, ColumnType.Subject,
-            ColumnType.DateAndTime, ColumnType.Date, ColumnType.Time,
-            ColumnType.Boolean, ColumnType.Double, ColumnType.Decimal, ColumnType.File, ColumnType.Flag,
-            ColumnType.Attachment, ColumnType.User, ColumnType.Lookup, ColumnType.OntologyLookup, ColumnType.VisitId,
-            ColumnType.VisitDate, ColumnType.Sample, ColumnType.Barcode, ColumnType.TextChoice, ColumnType.SMILES
-    );
-    
     public interface ColumnType
     {
         ColumnType MultiLine = new ColumnTypeImpl("Multi-Line Text", "multiLine");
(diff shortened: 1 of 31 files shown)


src/org/labkey/test/params/FieldDefinition.java
```diff
@@ -459,30 +466,42 @@ public LookupInfo getLookupInfo()
...
+            {
+                return true;
+            }
+        };
         ColumnType String = new ColumnTypeImpl("Text", "string");
         ColumnType Subject = new ColumnTypeImpl("Subject/Participant", "string", "http://cpas.labkey.com/Study#ParticipantId", null);
         ColumnType DateAndTime = new ColumnTypeImpl("Date Time", "dateTime");
         ColumnType Date = new ColumnTypeImpl("Date", "date");
-        ColumnType Time = new ColumnTypeImpl("Time", "http://www.w3.org/2001/XMLSchema#time");
+        ColumnType Time = new ColumnTypeImpl("Time", "time");
```
labkey-danield, line 484: At first I wasn't sure about this change, but I see that the change in ColumnTypeImpl should address this.

src/org/labkey/test/params/FieldDefinition.java
```diff
@@ -531,13 +550,14 @@ default FieldDefinition.LookupInfo getLookupInfo()
             return null;
         }
 
-        /**
-         * @deprecated Bridge for converting away from enum
-         */
-        @Deprecated (since = "22.10")
```
labkey-danield, line 537: So the values() method was deprecated but that is no longer the case?

src/org/labkey/test/params/list/IntListDefinition.java
```diff
@@ -9,27 +9,45 @@
...
     protected String getKind()
     {
-        return AUTO_INCREMENT_DOMAIN_KIND;
+        return DOMAIN_KIND;
+    }
+
+    @Override
+    protected String getKeyType()
+    {
+        return isAutoIncrementKey ? "AutoIncrementInteger" : "Integer";
```
labkey-danield, line 50: Doesn't this assume that the key will always be an integer or AutoIncrementInteger?

labkey-danield (approved):

src/org/labkey/test/params/list/IntListDefinition.java
labkey-tchad: Those are the only valid options for `IntList`: https://github.com/LabKey/platform/blob/develop/list/src/org/labkey/list/model/IntegerListDomainKind.java#L35

src/org/labkey/test/params/FieldDefinition.java
labkey-tchad: I'd previously thought it would go away eventually but `DomainFieldRow.getType` uses it. We could refactor `DomainFieldRow.getType` to return a `String` instead but that doesn't seem like a great option.

PR #1948, opened 2024-05-24
PlateSet remoteAPI updates, DetailPopover

labkey-chrisj:
#### Rationale
This change separates out parameter classes for `CreatePlateSetCommand`/`PlateSetResponse`.
It introduces the ability for tests to configure plates in plate sets as they are created.

It will hopefully support the ability to correctly establish plate set lineage (parentPlateSet, sourcePlateSet) but at this point that does not appear to be working adequately

#### Related Pull Requests
https://github.com/LabKey/limsModules/pull/303

#### Changes
New classes for CreatePlateSetParam, PlateSetParam, 

- [x] New classes for `CreatePlateSetParam`, `PlateSetParam`, `CreatePlateSetPlate`
- [x] new component wrapper for `DetailPopover`

### DIFF
--- a/src/org/labkey/remoteapi/plate/PlateSetParams.java
+++ b/src/org/labkey/remoteapi/plate/PlateSetParams.java
@@ -1,58 +1,91 @@
 package org.labkey.remoteapi.plate;
 
-
 import org.json.JSONObject;
 
 public class PlateSetParams
 {
-    private String _name;
+    private boolean _archived;
+    private String _containerId;
+    private String _containerName;
+    private String _containerPath;
+    private String _created;
+    private Integer _createdBy;
     private String _description;
-    private PlateSetType _type;
+    private String _modified;
+    private Integer _modifiedBy;
+    private String _name;
+    private Integer _plateCount;
     private String _plateSetId;
-    private Integer _rowId;
+    private Integer _primaryPlateSetId;
     private Integer _rootPlateSetId;
-
-    public PlateSetParams()
-    {
-    }
+    private Integer _rowId;
+    private boolean _template;
+    private CreatePlateSetParams.PlateSetType _plateSetType;
 
     public PlateSetParams(JSONObject json)
     {
-        _name = json.getString("name");
-        _description = json.optString("description", null);
-        _type = PlateSetType.fromName(json.optString("type", null));
-        _rowId = json.getInt("rowId");
-        _plateSetId = json.getString("plateSetId");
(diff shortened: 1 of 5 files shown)


labkey-chrisj:
@labkey-nicka - I attempted to use the remoteAPI in these tests to set up the plate set tree, but I'm missing something- apparently, setting parentPlateSetId isn't all that's required- the errors I'm seeing tell me I need to set sourcePlateSet?  (that's not a field on CreatePlateSetForm, I imagine it's something extra that needs to be wired in somewhere else?)

labkey-chrisj:
> @labkey-nicka - I attempted to use the remoteAPI in these tests to set up the plate set tree, but I'm missing something- apparently, setting parentPlateSetId isn't all that's required- the errors I'm seeing tell me I need to set sourcePlateSet? (that's not a field on CreatePlateSetForm, I imagine it's something extra that needs to be wired in somewhere else?)

Ah, I figured out what I was doing wrong-- nvm

src/org/labkey/remoteapi/plate/PlateSetParams.java
```diff
@@ -1,122 +1,135 @@
...
+    private boolean _template;
+    private CreatePlateSetParams.PlateSetType _plateSetType;
 
-    public PlateSetParams()
+    public PlateSetParams(JSONObject json)
     {
+        if (json.has("archived"))
+            _archived = json.getBoolean("archived");
+        if (json.has("containerId"))
+            _containerId = json.getString("containerId");
```
labkey-tchad, line 30: FYI, you can make this a bit more concise by using `json.opt*` methods to get optional values.
```suggestion
        _archived = json.optBoolean("archived");
        _containerId = json.optString("containerId", null);
        ...etc.
```
One warning though; none of the default values are `null`, so you'll have to specify those defaults explicitly (e.g. `json.optInt("createdBy", null)`)

labkey-tchad (approved):

src/org/labkey/remoteapi/plate/PlateSetParams.java
labkey-adam: Note that `json.optInt("createdBy", null)` is not an option since this method returns a primitive `int`. `json.optString("containerID", null)` is totally valid and useful.

There are also `optIntegerObject(String, Integer)`, `optBooleanObject(String, Boolean)`, etc. if you're stashing `Integer` or `Boolean` values and want to default to `null`.

PR #1956, opened 2024-06-05
Test components for FileField assay tests

labkey-chrisj:
#### Rationale
This adds component support for testing of file upload in assays with FileField results fields.

It introduces two components classes for some otherwise-familiar UI, 
FileAttachmentContainer:
image
and FileAttachmentEntry: 
image

Our existing component, `FileUploadPanel`, wraps [FileAttachmentForm](https://github.com/LabKey/labkey-ui-components/blob/fb_assayResultsFiles/packages/components/src/public/files/FileAttachmentForm.tsx) but FileUploadPanel assumes (based on long evidence) that there will only be one [FileAttachmentContainer](https://github.com/LabKey/labkey-ui-components/blob/fb_assayResultsFiles/packages/components/src/internal/components/files/FileAttachmentContainer.tsx) in it. But, the related feature work puts two of them there: 
image
This change adds a method to `FileUploadPanel `to allow tests to get a `FileAttachmentContainer `based on its input name so it can interact with the second one while making no changes to any existing methods in `FileUploadPanel`. 

It also makes a change to AttachmentCardFinder to distinguish finding it by its title attribute from finding it by the file name. (in most situations, the title is the file name but in others, the title can include its location: 
image)
Previously the withFile method used the title attribute to find the card, on the assumption file and title would always be the same. This change renames `withFile `to `withTitle `(usages are refactored) and adds a new method `withFileName` that finds the component based on the contents of its attachment-card__name element.

Prior to this, there wasn't support for updating file fields in `EntityBulkUpdateDialog`. This adds support for interacting with file fields in it: 
image

#### Related Pull Requests
https://github.com/LabKey/limsModules/pull/302

#### Changes

- [x] New component wrappers, `FileAttachmentContainer`, `FileAttachmentEntry`
- [x] Update `FileUploadPanel `to give tests the ability to reference another `FileAttachmentContainer` but no functional changes to any other part of `FileUploadPanel`
- [x] Add file-field support to `EntityBulkUpdateDialog`
- [x] Rename finder method on AttachmentCard, refactor usages, add finder method to find by file name

### DIFF
--- a/src/org/labkey/test/components/ui/entities/EntityBulkUpdateDialog.java
+++ b/src/org/labkey/test/components/ui/entities/EntityBulkUpdateDialog.java
@@ -11,6 +11,8 @@ import org.labkey.test.components.html.Input;
 import org.labkey.test.components.react.FilteringReactSelect;
 import org.labkey.test.components.react.ReactDateTimePicker;
 import org.labkey.test.components.react.ToggleButton;
+import org.labkey.test.components.ui.files.FileAttachmentContainer;
+import org.labkey.test.components.ui.files.FileUploadField;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.ui.ExpectedConditions;
@@ -123,6 +125,17 @@ public class EntityBulkUpdateDialog extends ModalDialog
     public String getDateField(String fieldKey)
     {
         return elementCache().dateInput(fieldKey).get();
+    }
+
+    public FileAttachmentContainer getFileField(String fieldKey)
+    {
+        return elementCache().fileUploadField(fieldKey);
+    }
+
+    public EntityBulkUpdateDialog removeFile(String fieldKey)
+    {
+        getFileField(fieldKey).removeFile();
+        return this;
     }
 
     public EntityBulkUpdateDialog setBooleanField(String fieldKey, boolean checked)
@@ -275,6 +288,11 @@ public class EntityBulkUpdateDialog extends ModalDialog
         {
             return new ReactDateTimePicker.ReactDateTimeInputFinder(getDriver())
                     .withInputId(fieldKey).waitFor(formRow(fieldKey));
+        }
+
+        public FileAttachmentContainer fileUploadField(String fieldKey)
+        {
+            return new FileAttachmentContainer(formRow(fieldKey), getDriver());
         }
 
         final Locator textInputLoc = Locator.tagWithAttribute("input", "type", "text");
--- a/src/org/labkey/test/components/ui/files/AttachmentCard.java
+++ b/src/org/labkey/test/components/ui/files/AttachmentCard.java
@@ -77,6 +77,7 @@ public class AttachmentCard extends WebDriverComponent<AttachmentCard.ElementCac
             return false;
         elementCache().menu.get().openMenuTo();
         List<String> menuOptions = getWrapper().getTexts(elementCache().menu.get().findVisibleMenuItems());
+        elementCache().menu.get().collapse();
         return menuOptions.contains(DOWNLOAD_ATTACHMENT);
     }
 
--- a/src/org/labkey/test/components/ui/files/AttachmentCard.java
+++ b/src/org/labkey/test/components/ui/files/AttachmentCard.java
@@ -144,15 +144,29 @@ public class AttachmentCard extends WebDriverComponent<AttachmentCard.ElementCac
     {
         private final Locator.XPathLocator _baseLocator = Locator.tagWithClass("div", "attachment-card");
         private String _fileName = null;
+        private String _title = null;
 
         public FileAttachmentCardFinder(WebDriver driver)
         {
             super(driver);
         }
 
-        public FileAttachmentCardFinder withFile(String fileName)
+        /*
+            Finds the card by the componentElement's title attribute
+            This is often the file name but can include more information, such as its location
+         */
+        public FileAttachmentCardFinder withTitle(String title)
         {
-            _fileName = fileName;
+            _title = title;
+            return this;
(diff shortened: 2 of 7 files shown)


src/org/labkey/test/components/ui/files/AttachmentCard.java
```diff
@@ -144,15 +145,29 @@ public static class FileAttachmentCardFinder extends WebDriverComponentFinder<At
...
         private final Locator.XPathLocator _baseLocator = Locator.tagWithClass("div", "attachment-card");
         private String _fileName = null;
+        private String _title = null;
 
         public FileAttachmentCardFinder(WebDriver driver)
         {
             super(driver);
         }
 
-        public FileAttachmentCardFinder withFile(String fileName)
```
labkey-chrisj, line 153: Initially, the `withFile `method caused finder to find the `AttachmentCard `by its title attribute (based on the assumption that it would always be the file).  I discovered during this story that the title could include more info, such as the hierarchy of location in the dav folders, so I changed the existing function (and refactored its usages) to be `withTitle`.

The reason for this is I needed to find `AttachmentCards `by just their file name, and it's right there in the attachment-card__name element, so I implemented another method to do that and named it `withFileName`

src/org/labkey/test/components/ui/files/FileUploadPanel.java
```diff
@@ -78,6 +78,16 @@ public File downloadTemplate()
         return getWrapper().doAndWaitForDownload(()->elementCache().downloadTemplate.click());
     }
 
+    /*
```
labkey-chrisj, line 81: adding this method was the least-intrusive way I could make the framework able to work with the existence of a second attachment container

src/org/labkey/test/components/ui/files/FileAttachmentContainer.java
```diff
@@ -0,0 +1,249 @@
+package org.labkey.test.components.ui.files;
```
labkey-chrisj, line 1: Given the choice to make between implementing the functionality of a component in the Biologics AssayUpload page and writing a wrapper for a component we don't really have a (potentially reusable) wrapper for, I chose to make this after discussing with Cory which components are in play on the dev side

src/org/labkey/test/components/ui/files/FileAttachmentContainer.java
```diff
@@ -0,0 +1,249 @@
...
+/*
+    Wraps https://github.com/LabKey/labkey-ui-components/blob/fb_assayResultsFiles/packages/components/src/internal/components/files/FileAttachmentContainer.tsx
+
+ */
+public class FileAttachmentContainer extends WebDriverComponent<FileAttachmentContainer.ElementCache>
+{
+    private final WebElement _el;
+    private final WebDriver _driver;
+
+    public FileAttachmentContainer(WebElement element, WebDriver driver)
```
labkey-chrisj, line 24: Leaving the c'tor public allows calling code to specify the componentElement itself- this follows the pattern set in `FileUploadField`, and it happens that this component works in cases FileUploadField does (they can be interchangeably used in the compact-mode usages like in `EntityBulkUpdateDialog` )

src/org/labkey/test/components/ui/files/FileAttachmentEntry.java
```diff
@@ -0,0 +1,116 @@
+package org.labkey.test.components.ui.files;
+
```
labkey-chrisj, line 2: A lot of pages have their own implementations of interactions with this, it's possibly a thing we could use to consolidate those implementations into less code  (or, if we don't want to, not)

src/org/labkey/test/components/ui/entities/EntityBulkUpdateDialog.java
```diff
@@ -125,6 +127,17 @@ public String getDateField(String fieldKey)
         return elementCache().dateInput(fieldKey).get();
     }
 
+    public FileAttachmentContainer getFileField(String fieldKey)
```
labkey-chrisj, line 130: Note: for this use case, `FileUploadField `can be used interchangeably
(or possibly, we could use FileAttachmentContainer everywhere `FileUploadField `is and get rid of it if we want)

src/org/labkey/test/components/ui/files/FileUploadPanel.java
labkey-chrisj: If this is too much, I could remove this method here and add it to the page class in Bio

src/org/labkey/test/components/ui/files/FileAttachmentContainer.java
```diff
@@ -0,0 +1,249 @@
...
+import org.labkey.test.components.html.Input;
+import org.openqa.selenium.ElementNotInteractableException;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+import java.io.File;
+import java.util.List;
+
+/*
+    Wraps https://github.com/LabKey/labkey-ui-components/blob/fb_assayResultsFiles/packages/components/src/internal/components/files/FileAttachmentContainer.tsx
```
labkey-tchad: Don't use the feature branch in this url
```suggestion
    Wraps https://github.com/LabKey/labkey-ui-components/blob/develop/packages/components/src/internal/components/files/FileAttachmentContainer.tsx
```

labkey-tchad (approved):
Looks great. Adds the new functionality without being too complicated.

src/org/labkey/test/components/ui/files/FileAttachmentContainer.java
labkey-chrisj: omg thanks nice catch

labkey-danield (approved):

PR #1957, opened 2024-06-06
Allow tests to ignore DatabaseNotSupportedException

labkey-tchad:
#### Rationale
Tests running in production mode have been failing because we can't undeploy SQL Server only modules on embedded tomcat. We can configure the test harness to ignore these errors.

#### Related Pull Requests
* N/A

#### Changes
* Add flag to allow tests to ignore DatabaseNotSupportedException
* Make sure test postamble runs even if test is skipped


labkey-tchad:
Verifying on TeamCity: https://teamcity.labkey.org/buildConfiguration/LabKey_246Release_Premium_ModulesSuites_ModuleEditorSqlserver/3060326?hideTestsFromDependencies=false&hideProblemsFromDependencies=false

labkey-danield (approved):

PR #1961, opened 2024-06-12
InlineImagesAssayTest.testAssayInlineImages fix to check for assay result file path update

cnathe:
#### Rationale
Fixing test failures that came up after the merge of the changes for the assay results files updates.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5533

#### Changes
- InlineImagesAssayTest.testAssayInlineImages fix to check for assay result file path update

### DIFF
--- a/src/org/labkey/test/tests/InlineImagesAssayTest.java
+++ b/src/org/labkey/test/tests/InlineImagesAssayTest.java
@@ -27,6 +27,7 @@ import org.labkey.test.Locator;
 import org.labkey.test.TestFileUtils;
 import org.labkey.test.WebTestHelper;
 import org.labkey.test.categories.Daily;
+import org.labkey.test.components.CustomizeView;
 import org.labkey.test.pages.ReactAssayDesignerPage;
 import org.labkey.test.params.FieldDefinition;
 import org.labkey.test.util.DataRegionExportHelper;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/InlineImagesAssayTest.java
```diff
@@ -172,14 +173,22 @@ public final void testAssayInlineImages() throws Exception
...
         waitForElement(DataRegionTable.updateLinkLocator()); // Wait to make sure the grid has been rendered.
 
+        // add columns to the grid to get assay and run id for file verification path
+        DataRegionTable list = new DataRegionTable("Data", getDriver());
+        CustomizeView customizeView = list.openCustomizeGrid();
+        customizeView.showHiddenItems();
+        customizeView.addColumn(new String[]{"Run", "RowId"});
+        customizeView.addColumn(new String[]{"Run", "Protocol", "RowId"});
+        customizeView.applyCustomView();
+        String helpJpgFilePath = "AssayId_" + list.getDataAsText(0, "Run/Protocol/RowId") + "/RunId_" + list.getDataAsText(0, "Run/RowId") + "/" + HELP_JPG_FILE.getName();
```
labkey-danield: Is the file path OS dependent? I'm not sure that it is but want to check.

In some LKSM test failures where the OS path was an issue I addressed like this:
```
right = SystemUtils.IS_OS_WINDOWS ?
                            "sampletype\\" + right :
                            "sampletype/" + right; 
```

labkey-danield (approved):

src/org/labkey/test/tests/InlineImagesAssayTest.java
cnathe: This code looks to have passed on both PG and SQLServer on TC so I think we are good.

cnathe: Good call, Dan. Chris tested this locally on his Windows machine and it failed. He pushed a change to us the OS file separator instead and TC looks happy with that.

PR #1970, opened 2024-07-03
stream large data to file - Issue 50740

labkey-chrisj:
#### Rationale
This updates `TestDataGenerator `to be able to stream large data directly to file, instead of holding all of it in memory (but also allows it to function as it has).
It also updates `UploadLargeExcelAssayTest `to export imported data in .xlsx format, where before the test tried converting the file in-memory.  This works as a behind-the-back way of converting the large .tsv file `TestDataGenerator `wrote into .xlsx, so the test still imports and exports large .xlsx files

I tried getting `TestFileUtils.convertTabularToXlsx` to stream file-to-file, unfortunately the way I did it wrote bad excel
I'm sure there's a better way using `TsvLoader `and `ExcelWriter`, but I time-boxed this effort to avoid burning too much time

#### Changes

- [x] minimal changes to allow TestDataGenerator to stream large data to file
- [x] update UploadLargeExcelAssayTest to work around the memory-intensive way we convert tabular to .xlsx


labkey-tchad (changes requested):
This should go to 24.7 since it's an ESR release.
Please run the test with this change on TeamCity to verify that it works.
Regarding streaming the excel file; did you see how we do it on the server side?
[`org.labkey.api.data.ExcelWriter.ExcelDocumentType.xlsx`](https://github.com/LabKey/platform/blob/develop/api/src/org/labkey/api/data/ExcelWriter.java#L123)

labkey-chrisj:
> This should go to 24.7 since it's an ESR release. Please run the test with this change on TeamCity to verify that it works. Regarding streaming the excel file; did you see how we do it on the server side? [`org.labkey.api.data.ExcelWriter.ExcelDocumentType.xlsx`](https://github.com/LabKey/platform/blob/develop/api/src/org/labkey/api/data/ExcelWriter.java#L123)

That makes sense, I'll retarget.
re: how it's done on the server side- thanks, pointing that out was helpful.  I will try.
I'm optimistic this will pass; I've been running my jvm on 1gb locally (this ran OOM before these changes) but now it doesn't

PR #1972, opened 2024-07-04
Automation for Issue 50121 (Can't enter multiple lines of text in multi-line data field)

labkey-danield:
#### Rationale
Automation for:
- [Issue 50121](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50121): LKSM: Can't enter multiple lines of text in multi-line data field
- [Issue 49953](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49953): Shift up-arrow does not undo cell selection in editable grid.

Added EditableGrid.setMultiLineCellValue and EditableGrid.doubleClickMultiLineCell helper methods. These are not intended to test editing function of a multi-line fields in the editable grid. They are meant to set the value of a multi-line field in the editable grid.

Added a multi-line field to some of the existing drag and fill / copy tests as extra validation.

The EditableGridTest.testMultiLineCellSizing fails on TeamCity because of the older version of FireFox. Running with a new version of FireFox on TeamCity does pass.

#### Related Pull Requests
* https://github.com/LabKey/limsModules/pull/437

#### Changes
* Added various tests to EditableGridTest.
* Added helper methods to EditableGrid.


src/org/labkey/test/components/ui/grids/EditableGrid.java
```diff
@@ -505,6 +505,51 @@ else if (value instanceof LocalTime localTime)
...
+    }
+
+    /**
+     * Double-clicking a cell that is for a multi-line field will put the cell into "edit mode". This will return the
+     * value that can be used to set the field.
+     * @param row Row to be edited.
+     * @param columnName Column name of the multi-line field.
+     * @return The TextArea component that can be used to edit the field.
+     */
+    public WebElement doubleClickMultiLineCell(int row, String columnName)
```
labkey-nicka, line 544: A couple points: 

1. The rest of `EdtiableGrid` refers to this as `activateCell()`. In editable grid on the client-side we refer to this as the "focused" state.
2. Double-clicking on any editable cell will focus the cell.

Consider renaming this to some variation using those words. Something like `activateCellUsingDoubleClick` `focusCellUsingDoubleClick`. The reason this is distinguished from `activateCell` is that `activateCell` uses send keys ENTER instead.

src/org/labkey/test/components/ui/grids/EditableGrid.java
```diff
@@ -505,6 +505,51 @@ else if (value instanceof LocalTime localTime)
...
+    {
+        WebElement gridCell = getCell(row, columnName);
```
labkey-nicka, line 546: The cell may already be activated. Consider factoring out the common check for activation.

src/org/labkey/test/tests/component/EditableGridTest.java
```diff
@@ -434,6 +482,717 @@ public void testExpandedPasteIntoSkinnySelection() throws Exception
...
+
+        WebElement gridCell = editableGrid.getCell(0, PASTE_ML);
+        log("Cell initial Width & Height: " + gridCell.getSize());
+        log("Empty width: " + emptyWidth + " empty height: " + emptyHeight);
+
+        WebElement editCell = editableGrid.doubleClickMultiLineCell(0, PASTE_ML);
+
+        Actions actions;
+
+        for(int rows = 0; rows < 50; rows++)
```
labkey-nicka, line 610: Why 50 times?

src/org/labkey/test/tests/component/EditableGridTest.java
```diff
@@ -434,6 +482,717 @@ public void testExpandedPasteIntoSkinnySelection() throws Exception
...
+
+        WebElement updatedGridCell04 = editableGrid.getCell(0, PASTE_ML);
+        checker().verifyTrue("Cell should not be updated after hitting <esc>.",
+                waitFor(()->updatedGridCell04.getText().isEmpty(), 1_000));
+
+        checker().screenShotIfNewError("Exit_With_Escape_Error");
+
+    }
+
+    static final Locator SELECTED_CELL_LOCATOR = Locator.tagWithClass("div", "cell-selection");
```
labkey-nicka, line 792: Extract the common locator in `EditableGrid.getSelectionCellTexts` and expose a new method which returns the `List<WebElement>` for the selected cells.

labkey-nicka (approved):

src/org/labkey/test/components/ui/grids/EditableGrid.java
labkey-danield: Changed to activateCellUsingDoubleClick.

src/org/labkey/test/components/ui/grids/EditableGrid.java
labkey-danield: Fair point. Added a check.

src/org/labkey/test/tests/component/EditableGridTest.java
labkey-danield: No particular reason. 
Reduced it to 10, for a total of 30 lines to be added.

src/org/labkey/test/tests/component/EditableGridTest.java
labkey-danield: Was able to leverage functionality in the editable grid for this.
This has been removed.

src/org/labkey/test/tests/component/EditableGridTest.java
```diff
@@ -434,6 +482,717 @@ public void testExpandedPasteIntoSkinnySelection() throws Exception
...
+     * </p>
+     */
+    @Test
+    public void testPasteIntoMultiLine()
+    {
+
+        EditableGrid editableGrid = goToEditableGrid(PASTING_SAMPLE_TYPE);
+        editableGrid.addRows(1);
+
+        List<String> expectedValues = List.of("Line 1",
```
labkey-chrisj, line 504: for future reference, you can use triple-quotes to formulate multi-line text, like : 
 String myMultiLineText = """
                line 1
                line 2
                etc
                more
                """;
That saves you the need to concatenate newlines

labkey-chrisj (approved):

src/org/labkey/test/tests/component/EditableGridTest.java
labkey-danield: Yes, I know that. Using it as a list made it easier to check the count of rows and then the scenario where each row in the column was one value.

PR #1971, opened 2024-07-03
Stop running heap out of memory

labkey-chrisj:
This updates TestDataGenerator to be able to stream large data directly to file, instead of holding all of it in memory (but also allows it to function as it has).
It also updates UploadLargeExcelAssayTest to export imported data in .xlsx format, where before the test tried converting the file in-memory. This works as a behind-the-back way of converting the large .tsv file TestDataGenerator wrote into .xlsx, so the test still imports and exports large .xlsx files

I was also successful getting TestFileUtils.convertTabularToXlsx to stream file-to-file, it turns out `poi.xssf.streaming.SXSSFWorkbook` and its related sheet/row classes do a lot of heavy lifting here

Related Pull Requests
https://github.com/LabKey/testAutomation/pull/1970

Changes
 minimal changes to allow TestDataGenerator to stream large data to file
 update UploadLargeExcelAssayTest to work around the memory-intensive way we convert tabular to .xlsx

Apparently, renaming the branch in order to target 24.7 closed the prior PR, this is the new one

PR #1976, opened 2024-07-10
Restore custom suppliers in TestDataGenerator

labkey-chrisj:
#### Rationale
Recently on 24.7, [BiologicsReportTest.testBarChartWithAggregateMethod](https://teamcity.labkey.org/viewLog.html?buildId=3098170&buildTypeId=LabKey_247Release_Premium_ProductSuites_Biologics_BiologicsPostgres&fromSakuraUI=true#testNameId-6701890804038466177) began failing because the chart-builder expects there to be values in the field it is using for its y-axis column.  The reason that column was empty, it happens, is that this domain was being generated using a custom dataSupplier (its job was to put the value '1' in generated rows.)

My recent changes to TestDataGenerator seem to have inadvertently regressed its ability to use custom dataSuppliers when generating row data.  This change initializes the default field data-generating suppliers to the list of suppliers if no supplier is in the list, and uses suppliers from the list (vs. calling getDefaultDataSupplier every row to get a supplier) to generate data.

#### Related Pull Requests
https://github.com/LabKey/testAutomation/pull/1971

#### Changes

- [x] Initialize suppliers in the suppliers list for fields to receive generated data, and use suppliers from the list instead of getting default suppliers only


src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -257,12 +257,17 @@ private Map<String, Object> generateRow()
         final Map<String, Object> newRow = new CaseInsensitiveHashMap<>();
         for (String columnName : _columns.keySet())
         {
-            if (_autoGeneratedFields.contains(columnName) || _dataSuppliers.containsKey(columnName))
+            if (!_dataSuppliers.containsKey(columnName))
+            {
+                _dataSuppliers.put(columnName, getDefaultDataSupplier(_columns.get(columnName).getRangeURI()));
+            }
```
labkey-tchad, line 257: Sick Map trick 🛹 
```suggestion
            _dataSuppliers.computeIfAbsent(columnName, k -> getDefaultDataSupplier(_columns.get(columnName).getRangeURI()));
```

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -257,12 +257,17 @@ private Map<String, Object> generateRow()
...
+
+            if (_autoGeneratedFields.contains(columnName))
             {
                 continue;
             }
-            // Generate default values
-            Object columnValue = getDefaultDataSupplier(_columns.get(columnName).getRangeURI()).apply(_rows);
+            // Generate values
+            Object columnValue = _dataSuppliers.get(columnName).apply(_rows);
```
labkey-tchad: Consider making `_dataSuppliers` actually be `Suppliers`. We don't actually use the row collection in any existing data suppliers and I don't know that it adds much utility to provide it.
```
Map<String, Supplier<Object>> _dataSuppliers = new CaseInsensitiveHashMap<>()
```
Then this simplifies to:
```suggestion
            // Generate values
            Object columnValue = _dataSuppliers.get(columnName).get();
```
And we could remove the unused method `addDataSupplier(String columnName, Function<List<Map<String, Object>>, Object> supplier)`

labkey-tchad (approved):

labkey-susanh (approved):

src/org/labkey/test/util/TestDataGenerator.java
labkey-chrisj: I was excited to try this sick map trick, but unfortunately it didn't update the suppliers map; this meant that below the call to 

`_dataSuppliers.get(columnName).get();`

...would fail because _dataSuppliers.get(columnName) would be null ☹️

src/org/labkey/test/util/TestDataGenerator.java
labkey-chrisj: Thanks, I like this suggestion a lot.  (applying it to _rows without really understanding why has bothered me, this gets me out of that)
Also, this makes everything else that interacts with `_dataSuppliers `a lot more readable

PR #1986, opened 2024-07-17
Add page state validation to list helpers

labkey-tchad:
#### Rationale
List import was failing for a particular test but didn't cause any problems until later in the test. Helpers should catch these errors when they happen so that the problem is obvious.

#### Related Pull Requests
* N/A

#### Changes
 - Check for errors after importing a list archive
 - Add method to import a list archive expecting errors
 - Define `list.BeginPage.waitForPage`


labkey-chrisj (approved):
looks good

labkey-sweta (approved):

PR #1987, opened 2024-07-17
Add test infrastructure for running API scenarios in parallel

labkey-tchad:
#### Rationale
In order to stress test our cloud infrastructure, we need to have the capability to run multiple API scenarios concurrently.
The existing `ApiTestHelper` does most of what we want but makes a lot of assumptions about the target server. Instead of trying to make the existing helper serve double duty, we are creating a new set of helpers to run multiple API test simulations concurrently against different servers.

Example of tsv results: https://docs.google.com/spreadsheets/d/1DnS2bMvxJ-uWAphY7AGbipcg6h5cHHSZAa01_KuQvtY

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/5690

#### Changes
* Create test helpers for concurrent API testing


src/org/labkey/remoteapi/miniprofiler/CustomTiming.java
```diff
@@ -0,0 +1,39 @@
...
+    /*
+      "message" : "-- <QueryServiceImpl.getSelectSQL(Policies)>\n\tSELECT *\n\tFROM (\n\tSELECT \n\tPolicies.resourceid AS resourceid,\n\tPolicies.resourceclass AS resourceclass,\n\tPolicies.container AS container,\n\tPolicies.modified AS modified\n\tFROM core.policies Policies ) x\n\tWHERE (container = ?) AND (resourceid = ?)\n-- <\/QueryServiceImpl.getSelectSQL()>\n",
+      "duration" : 3,
+      "stackTrace" : null,
+      "detailsURL" : "\/admin\/queryStackTraces.view?sqlHashCode=-584866968",
+      "startOffset" : 49
+     */
+
+    private final String message;
+    private final Long duration;
```
labkey-jeckels: Would a `long` be sufficient?

src/org/labkey/remoteapi/miniprofiler/CustomTiming.java
```diff
@@ -0,0 +1,39 @@
...
+    private final String stackTrace;
+    private final String detailsURL;
+    private final Long startOffset;
```
labkey-jeckels: Same question about a `long`

src/org/labkey/remoteapi/miniprofiler/RequestInfo.java
```diff
@@ -0,0 +1,87 @@
...
+        "ignored" : false,
+        "name" : "project\/begin"
+     */
+
+    public static final RequestInfo BLANK = new RequestInfo(null, null, null, null, null, null, null);
+
+    private final Long id;
+    private final String url;
+    private final String date;
+    private final Long duration;
```
labkey-jeckels, line 26: `long` vs `Long`

src/org/labkey/remoteapi/miniprofiler/Timing.java
```diff
@@ -0,0 +1,110 @@
...
+      "durationExclusive" : 784,
+      "children" : [...],
+      "objects" : {...},
+      "customTimings" : [...]
+     */
+
+    private final String name;
+    private final String id;
+    private final Long duration;
+    private final Long durationExclusive;
```
labkey-jeckels: `long` vs `Long`

src/org/labkey/test/stress/AbstractScenario.java
```diff
@@ -0,0 +1,253 @@
...
+    {
+        if (state != ScenarioState.READY)
+        {
+            throw new IllegalStateException("Unable to start scenario, already " + state);
+        }
+        state = ScenarioState.STARTING;
+        TestLogger.log("Starting %d background simulations".formatted(_simulationDefinitions.size()));
+        try
+        {
+            _simulationDefinitions.parallelStream().forEach(definition -> {
```
labkey-jeckels: Note that this doesn't guarantee parallelism. I don't know the exact rules for how it splits/batches.

src/org/labkey/test/stress/ApiTestCommand.java
```diff
@@ -0,0 +1,121 @@
...
+                    tokener.back();
+                    if (firstChar == '{')
+                    {
+                        json = new JSONObject(tokener);
+                    }
+                    else if (firstChar == '[')
+                    {
+                        json = new JSONObject();
+                        json.put("jsonArray", new JSONArray(tokener));
+                    }
```
labkey-jeckels, line 80: Should this be a failure condition? Maybe the fallback to plain text is OK

src/org/labkey/test/stress/ApiTestCommand.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+
+    @Override
+    protected HttpUriRequest createRequest(URI uri)
+    {
+        HttpUriRequest method;
+        String requestUrl =  uri.getScheme() + "://" + uri.getAuthority() + '/' + _url;
+
+        if (_type.equals("get"))
```
labkey-jeckels, line 100: The possible types are candidates for an enum or constants.

src/org/labkey/test/stress/RequestInfoTsvWriter.java
```diff
@@ -0,0 +1,98 @@
...
+            {
+                value = resultMetadata.get(field);
+            }
+            else if (REQUEST_INFO_MAPPER.containsKey(field))
+            {
+                value = REQUEST_INFO_MAPPER.get(field).apply(ri);
+            }
+
+            if (value != null)
+                row.append(value);
```
labkey-jeckels: If it's impractical to use library code to write TSVs, consider blowing up with a clear error message if the value includes a tab or newline.

labkey-jeckels (approved):
Some minor comments. Good for Dan to review as well.

src/org/labkey/test/stress/AbstractScenario.java
```diff
@@ -0,0 +1,253 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.stream.Collectors;
+
+/**
+ * Scenario: A set of coordinated simulations across multiple servers, simulating a particular usage event (e.g. one server with a heavy workload of importing large amounts of data while other servers perform simple read-only activities).
+ * This will manage a group of simulations. {@link #startScenario()} will begin the scenario by running all provided simulations in parallel. The method will block briefly to allow simulations to collect some baseline performance data. Then the caller should perform other operations that are expected to impact the performance of the background simulations.<br>
```
labkey-danield, line 24: Nit:
Is it 'baseline performance data' or is it just 'performance data'. (repeated in multiple places)

src/org/labkey/test/stress/Simulation.java
```diff
@@ -0,0 +1,434 @@
...
+     * {@link #startSimulation(Function)} is the primary entry point once all parameters have been set
+     */
+    public static class Definition
+    {
+        private final Supplier<Connection> _connectionFactory;
+
+        private List<Activity> activityDefinitions = Collections.emptyList();
+        private int maxActivityThreads = 6; // This seems to be the number of parallel requests browsers handle
+        private int delayBetweenActivities = 5_000;
+        private boolean runOnce = false;
```
labkey-danield, line 254: Would it be valuable to make any of these configurable through environment variables or TeamCity parameters and default to these values if nothing is set?

src/org/labkey/test/util/APIUserHelper.java
```diff
@@ -238,4 +248,79 @@ protected void _deleteUsers(boolean failIfNotFound, String... userEmails)
                 deleteUser(userId);
         }
     }
+
+    private static final Pattern regEmailVerification = Pattern.compile("verification=([A-Za-z0-9]+)");
+    @Override
+    public String setInitialPassword(String email)
```
labkey-danield, line 254: Kind of interesting/funny that this functionality (for APIHelper) had been missing for all this time and wasn't needed until this story.

labkey-danield (approved):

src/org/labkey/test/stress/Simulation.java
labkey-tchad: I was thinking that these values would be part of the scenario definition. In other words, having a different delay or number of threads would be a different scenario.
Now, a particular scenario might want to add TeamCity knobs to adjust these values for that scenario but there wouldn't be any baked in to the helpers.

src/org/labkey/test/util/APIUserHelper.java
labkey-tchad: I only figured out how to do this via API relatively recently to [create the initial user via API](https://github.com/LabKey/testAutomation/blob/develop/src/org/labkey/test/util/ApiBootstrapHelper.java#L174). It wasn't explicitly necessary for this story but one of the scenarios I was considering involved creating 100 users and setting that many passwords via UI would have taken a long time.

src/org/labkey/test/stress/AbstractScenario.java
labkey-tchad: I meant "baseline" to indicate that the data is there for comparison to the perf data that is collected during the stress events (e.g. large sample imports). It is sort of the "zero" for the performance graph.

src/org/labkey/test/stress/ApiTestCommand.java
labkey-tchad: This actually is a failure condition in the main Java API. I was getting errors when trying to get `product-menuSections.api`. These helpers don't actually care about the response data so I could have just ignored the response.
This was my idea for a fix that might go into the published API at some point but maybe we shouldn't be sending raw JSON arrays as API responses?

labkey-jeckels: Yeah, updating the API to not return the array seems like the best long-term approach. Can you log an issue about this?

PR #1989, opened 2024-07-18
Wait for Search Service to boot up before tests start running

labkey-sweta:
#### Rationale
Recently tests have started failing on SQL Sever on windows machine caused due to slower search service. This fix should wait for search service to start before running any tests.

Link to test failures
https://teamcity.labkey.org/buildConfiguration/LabKey_247Release_Community_DailySqlServer?mode=branches#all-projects

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/src/org/labkey/test/LabKeySiteWrapper.java
+++ b/src/org/labkey/test/LabKeySiteWrapper.java
@@ -33,6 +33,7 @@ import org.apache.hc.core5.http.ProtocolException;
 import org.apache.hc.core5.http.io.entity.EntityUtils;
 import org.apache.hc.core5.http.message.BasicNameValuePair;
 import org.apache.hc.core5.http.protocol.HttpContext;
+import org.awaitility.Awaitility;
 import org.jetbrains.annotations.Nullable;
 import org.json.JSONObject;
 import org.junit.Assert;
@@ -727,19 +728,25 @@ public abstract class LabKeySiteWrapper extends WebDriverWrapper
                 SimpleGetCommand command = new SimpleGetCommand("search", "json");
                 command.setParameters(Map.of("q", "pinging to check server is started", "scope", "All"));
                 CommandResponse searchResponse;
-                try
+                do
                 {
-                    do
+                    try
                     {
+                        sleep(10); // to poll the re-request
                         searchResponse = command.execute(connection, "/");
                     }
-                    while (searchResponse.getStatusCode() != 200);
+                    catch (IOException e)
+                    {
+                        throw new RuntimeException(e);
+                    }
+                    catch (CommandException e)
+                    {
+                        throw new RuntimeException(e);
+                    }
+                }
+                while (searchResponse.getStatusCode() != 200);
 
-                }
-                catch (Exception e)
-                {
-                    throw new RuntimeException(e);
-                }
+
             }
             else // Just upgrading
             {
--- a/src/org/labkey/test/LabKeySiteWrapper.java
+++ b/src/org/labkey/test/LabKeySiteWrapper.java
@@ -723,17 +723,19 @@ public abstract class LabKeySiteWrapper extends WebDriverWrapper
                     customizeSitePage.save();
                 }
 
-                //Waiting for search service to boot up
+                /*
+                    Waiting for search service to boot up
+                    Issue 50601: PDF indexing is slow on first file after server startup on Windows
+                 */
                 Connection connection = createDefaultConnection();
                 SimpleGetCommand command = new SimpleGetCommand("search", "json");
                 command.setParameters(Map.of("q", "pinging to check server is started", "scope", "All"));
                 long searchEndTime = System.currentTimeMillis() + 180000; // Wait for maximum of 3 mins.
-                CommandResponse searchResponse = null;
                 do
                 {
                     try
                     {
-                        searchResponse = command.execute(connection, "/");
+                        command.execute(connection, "/");
                     }
                     catch (IOException e)
                     {
@@ -741,10 +743,10 @@ public abstract class LabKeySiteWrapper extends WebDriverWrapper
                     }
                     catch (CommandException e)
                     {
-                        sleep(500); // Retry and to poll the re-request
+                        sleep(500); //poll the re-request
                     }
                 }
-                while ((searchResponse != null) && (searchResponse.getStatusCode() != 200) && (System.currentTimeMillis() < searchEndTime));
+                while (System.currentTimeMillis() < searchEndTime);
             }
             else // Just upgrading
             {
--- a/src/org/labkey/test/LabKeySiteWrapper.java
+++ b/src/org/labkey/test/LabKeySiteWrapper.java
@@ -720,6 +720,26 @@ public abstract class LabKeySiteWrapper extends WebDriverWrapper
                     // Note: leave the self-report setting unchanged
                     customizeSitePage.save();
                 }
+
+                //Waiting for search service to boot up
+                Connection connection = createDefaultConnection();
+
+                SimpleGetCommand command = new SimpleGetCommand("search", "json");
+                command.setParameters(Map.of("q", "pinging to check server is started", "scope", "All"));
+                CommandResponse searchResponse;
+                try
+                {
+                    do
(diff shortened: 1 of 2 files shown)


labkey-sweta:
Teamcity run results

https://teamcity.labkey.org/buildConfiguration/LabKey_247Release_Community_ModuleSuites_SearchSqlserver/3111708?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true

src/org/labkey/test/LabKeySiteWrapper.java
```diff
@@ -720,6 +720,26 @@ else if (getDriver().getTitle().startsWith("Sign In"))
...
+
+                //Waiting for search service to boot up
+                Connection connection = createDefaultConnection();
+
+                SimpleGetCommand command = new SimpleGetCommand("search", "json");
+                command.setParameters(Map.of("q", "pinging to check server is started", "scope", "All"));
+                CommandResponse searchResponse;
+                try
+                {
+                    do
```
labkey-chrisj: Consider adding a polling interval to avoid spamming the service- 
Also consider building an exit condition to the loop- if the service never responds 200, this code should probably have a built-in exit after some number of iterations or time (or both)

labkey-sweta: Good point

src/org/labkey/test/LabKeySiteWrapper.java
```diff
@@ -720,6 +721,32 @@ else if (getDriver().getTitle().startsWith("Sign In"))
...
+                Connection connection = createDefaultConnection();
+
+                SimpleGetCommand command = new SimpleGetCommand("search", "json");
+                command.setParameters(Map.of("q", "pinging to check server is started", "scope", "All"));
+                CommandResponse searchResponse;
+                do
+                {
+                    try
+                    {
+                        sleep(10); // to poll the re-request
```
labkey-tchad: The sleep should be longer (500ms maybe) and it should be done after we've found that the service is not running.

src/org/labkey/test/LabKeySiteWrapper.java
```diff
@@ -720,6 +721,32 @@ else if (getDriver().getTitle().startsWith("Sign In"))
...
+                        searchResponse = command.execute(connection, "/");
+                    }
+                    catch (IOException e)
+                    {
+                        throw new RuntimeException(e);
+                    }
+                    catch (CommandException e)
+                    {
+                        throw new RuntimeException(e);
+                    }
```
labkey-tchad: Don't rethrow these exceptions. If there is a `CommandException` it likely means that we should continue waiting.

src/org/labkey/test/LabKeySiteWrapper.java
```diff
@@ -720,6 +721,32 @@ else if (getDriver().getTitle().startsWith("Sign In"))
...
+                }
+                while (searchResponse.getStatusCode() != 200);
```
labkey-tchad: `searchResponse.getStatusCode()` will always be 200. Anything else will trigger a `CommandException`.
Also, there should be some maximum timeout (3 minutes should be enough).

labkey-tchad (changes requested):

src/org/labkey/test/LabKeySiteWrapper.java
labkey-sweta: DOne

src/org/labkey/test/LabKeySiteWrapper.java
```diff
@@ -720,6 +722,29 @@ else if (getDriver().getTitle().startsWith("Sign In"))
...
+                    catch (IOException e)
+                    {
+                        throw new RuntimeException(e);
+                    }
+                    catch (CommandException e)
+                    {
+                        sleep(500); // Retry and to poll the re-request
+                    }
+                }
+                while ((searchResponse != null) && (searchResponse.getStatusCode() != 200) && (System.currentTimeMillis() < searchEndTime));
```
labkey-tchad: If the command throws an exception, `searchResponse` will be null and this won't loop.

src/org/labkey/test/LabKeySiteWrapper.java
```diff
@@ -720,6 +722,29 @@ else if (getDriver().getTitle().startsWith("Sign In"))
                     // Note: leave the self-report setting unchanged
                     customizeSitePage.save();
                 }
+
+                //Waiting for search service to boot up
```
labkey-tchad: We should include a note here about why this wait is being done:
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50601

labkey-tchad (changes requested):

src/org/labkey/test/LabKeySiteWrapper.java
```diff
@@ -720,6 +722,31 @@ else if (getDriver().getTitle().startsWith("Sign In"))
...
+                    catch (IOException e)
+                    {
+                        throw new RuntimeException(e);
+                    }
+                    catch (CommandException e)
+                    {
+                        sleep(500); //poll the re-request
+                    }
+                }
+                while (System.currentTimeMillis() < searchEndTime);
```
labkey-tchad: This will always loop for the maximum duration (unless there is an `IOException`)

labkey-tchad (changes requested):

src/org/labkey/test/LabKeySiteWrapper.java
```diff
@@ -720,6 +722,31 @@ else if (getDriver().getTitle().startsWith("Sign In"))
...
+                 */
+                Connection connection = createDefaultConnection();
+                SimpleGetCommand command = new SimpleGetCommand("search", "json");
+                command.setParameters(Map.of("q", "pinging to check server is started", "scope", "All"));
+                long searchEndTime = System.currentTimeMillis() + 180000; // Wait for maximum of 3 mins.
+                do
+                {
+                    try
+                    {
+                        command.execute(connection, "/");
```
labkey-danield: How about adding a `break;` here. If the command runs and there is no exception there is no need to continue.

labkey-sweta:
https://teamcity.labkey.org/buildConfiguration/LabKey_247Release_Community_ModuleSuites_SearchSqlserver/3119777?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true&expandBuildDeploymentsSection=false

All the tests passed.

src/org/labkey/test/LabKeySiteWrapper.java
```diff
@@ -720,6 +722,33 @@ else if (getDriver().getTitle().startsWith("Sign In"))
...
+                SimpleGetCommand command = new SimpleGetCommand("search", "json");
+                command.setParameters(Map.of("q", "pinging to check server is started", "scope", "All"));
+                long searchEndTime = System.currentTimeMillis() + 180000; // Wait for maximum of 3 mins.
+                do
+                {
+                    try
+                    {
+                        CommandResponse response = command.execute(connection, "/");
+                        if (response.getStatusCode() == 200) break; // Server is up, we can exit the loop
+                        else throw new RuntimeException("Search server did not start properly");
```
labkey-tchad:
```suggestion
                        else throw new RuntimeException("Search service did not start properly");
```

labkey-tchad (changes requested):
A subsequent search suite run failed
https://teamcity.labkey.org/buildConfiguration/LabKey_247Release_Community_ModuleSuites_SearchSqlserver/3122742
Not sure what's going wrong. Maybe add some logging to this to indicate how long it waited (maybe 3 minutes wasn't enough).

labkey-tchad (commented):
Oh, I see what happened. This suite doesn't create the initial user via the UI. `ApiBootstrapHelper#waitForBootstrap` needs to have a similar wait.

src/org/labkey/test/LabKeySiteWrapper.java
labkey-sweta: Done

labkey-tchad (approved):

PR #1991, opened 2024-07-22
Automation test for Display outliers in heat maps

labkey-sweta:
#### Rationale
Wrapper for the new web part “Outlier Heatmap Summary View” which is added to the Panorama QC/SystemSuitability folder. The Web part will display the total number of outliers for each peptide for a user defined period of time.

New automation test class to test all the scenarios.

#### Related Pull Requests
https://github.com/LabKey/targetedms/pull/932

#### Changes
*


src/org/labkey/test/components/html/Table.java
```diff
@@ -92,6 +92,28 @@ public List<String> getTableHeaderTexts()
...
+    {
+        List<WebElement> headerEls = Locator.xpath("//thead//th").findElements(this);
+        int counter = 1;
+        for(WebElement headerEl : headerEls)
+        {
+            if(headerEl.getText().equalsIgnoreCase(headerText))
+                return counter;
+            counter++;
+        }
+        return counter;
```
labkey-tchad: Don't return an inaccurate value. This should throw an exception if the specified column doesn't exist

labkey-tchad (changes requested):
Why don't the existing `getColumnIndex` and `getColumnAsText` methods work?

labkey-sweta:
> Why don't the existing `getColumnIndex` and `getColumnAsText` methods work?

There is TODO in the Table.java "// TODO: This finder makes an assumption that the column headers will be in a tr in the tbody, that is not always the case.
    // Maybe a possible solution would be to remove "./tbody" from the locator, but that is a thread I am not willing to pull at this time."
Did not work in my case too, instead of touching the old code wrote new helper.

src/org/labkey/test/components/html/Table.java
labkey-sweta: good point

labkey-tchad (approved):
Ugh, I knew this class would come back to haunt us someday. I'm not eager to add more special purpose methods to work around its jank.
There are methods that let you specify the row index for the row that should be treated as the header row; we could have zero indicate that the header row is in a `<thead>`.
```
    private List<WebElement> getColumnHeaderElements(int headerRow)
    {
        WebElement rowEl = (headerRow == 0 ? Locator.xpath("./thead/tr") : Locator.xpath("./tbody/tr[" + headerRow + "]")).findElement(this);
        return Locator.xpath("./*[(name()='TH' or name()='TD' or name()='th' or name()='td')]").findElements(rowEl);
    }
```
Then the test could use `getColumnAsText(header, 0)` to get the column data.

That being said, this seems like something to tackle on another day. This `Table` class needs a major overhaul.

PR #1997, opened 2024-07-28
Automation test for Link to study using visit labels

### LINKED ISSUE #5699: [LabKey CRI] Link to study support for visit labels
[spec](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50224)

When importing data into visit based studies that have a visit map configured, specifying a visit label in the imported data will correctly map to the associated sequence number. However, this is not supported in the link to study scenarios and we always expect a sequence number for sample data.

This work adds the capability to use visit labels in place of sequence numbers when linking sample data to visit based studies.

**Tasks**
-  [x] New data type for sample result domains.
- [x] Client side changes
- [x] Publish confirm action
- [x] Sample type field key resolution
- [x] Derived sample support
- [x] Auto-link support
- [x] Feature verification @labkey-hannah @jryurkanin 
- [x] Manual test @labkey-sweta 
- [x] Automation test

> - [x] Create new CreateNewVisitPage
> - [x] testVisitLabelLinkToStudy() test case
> - [x] testVisitLabelAutoLinkToStudy() test case

- [x]  Code review for automation
- [x] Code review & merge

labkey-sweta:
#### Rationale
Automation test to verify "Link to study using visit labels"

#### Related Pull Requests
https://github.com/LabKey/platform/pull/5712

#### Changes
*

### DIFF
--- a/src/org/labkey/test/pages/study/CreateNewVisitPage.java
+++ b/src/org/labkey/test/pages/study/CreateNewVisitPage.java
@@ -1,0 +1,98 @@
+package org.labkey.test.pages.study;
+
+import org.labkey.test.Locator;
+import org.labkey.test.WebDriverWrapper;
+import org.labkey.test.WebTestHelper;
+import org.labkey.test.components.html.Checkbox;
+import org.labkey.test.components.html.Input;
+import org.labkey.test.components.html.SelectWrapper;
+import org.labkey.test.pages.LabKeyPage;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.ui.Select;
+
+public class CreateNewVisitPage extends LabKeyPage<CreateNewVisitPage.ElementCache>
+{
+    public CreateNewVisitPage(WebDriver driver)
+    {
+        super(driver);
+    }
+
+    public static CreateNewVisitPage beginAt(WebDriverWrapper webDriverWrapper)
+    {
+        return beginAt(webDriverWrapper, webDriverWrapper.getCurrentContainerPath());
+    }
+
+    public static CreateNewVisitPage beginAt(WebDriverWrapper webDriverWrapper, String containerPath)
+    {
+        webDriverWrapper.beginAt(WebTestHelper.buildURL("study", containerPath, "createVisit"));
+        return new CreateNewVisitPage(webDriverWrapper.getDriver());
+    }
+
+    public CreateNewVisitPage setLabel(String label)
+    {
+        elementCache().label.set(label);
+        return this;
+    }
+
+    public CreateNewVisitPage setMinSequence(String range)
+    {
+        elementCache().sequenceNumMin.set(range);
+        return this;
+    }
+
+    public CreateNewVisitPage setMaxSequence(String range)
+    {
+        elementCache().sequenceNumMax.set(range);
+        return this;
+    }
+
+    public CreateNewVisitPage setDescription(String value)
+    {
+        setFormElement(elementCache().description, value);
+        return this;
+    }
+
+    public CreateNewVisitPage setType(String option)
+    {
+        elementCache().type.selectByVisibleText(option);
+        return this;
+    }
+
+    public CreateNewVisitPage setVisitHandling(String value)
+    {
+        elementCache().sequenceNumHandling.selectByVisibleText(value);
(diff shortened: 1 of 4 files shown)


labkey-klum (approved):
Scenarios look fine, the only suggestion I have is to, in the auto-link case, use both a mixture of visit ID and visit label. If both are present, we should accept the visit ID over the visit label.

src/org/labkey/test/pages/study/CreateNewVisitPage.java
```diff
@@ -0,0 +1,98 @@
...
+    {
+        return beginAt(webDriverWrapper, webDriverWrapper.getCurrentContainerPath());
+    }
+
+    public static CreateNewVisitPage beginAt(WebDriverWrapper webDriverWrapper, String containerPath)
+    {
+        webDriverWrapper.beginAt(WebTestHelper.buildURL("study", containerPath, "createVisit"));
+        return new CreateNewVisitPage(webDriverWrapper.getDriver());
+    }
+
```
labkey-danield, line 31: It doesn't look like this page really needs one, but want to check and make sure you considered if there is a need for a `waitForPage` check to make sure the page is loaded.

src/org/labkey/test/pages/study/CreateNewVisitPage.java
```diff
@@ -0,0 +1,98 @@
...
+        elementCache().showByDefault.set(value);
+        return this;
+    }
+
+    public ManageVisitPage clickSave()
+    {
+        clickAndWait(elementCache().save);
+        return new ManageVisitPage(getDriver());
+    }
+
```
labkey-danield, line 79: Do you have / need a test that clicks the Cancel button?
Is there a possibility that clicking Save could throw an error? Like a required field not set, or a data collision with existing data (e.g. duplicate labels, `sequenceNumMin` is greater than `sequenceNumMax`, overlapping ranges, etc...). Then you would need a `clickSaveExpectingError` method and the appropriate test cases.
You may want to sync with @labkey-klum on some of those.

src/org/labkey/test/pages/study/CreateNewVisitPage.java
```diff
@@ -0,0 +1,98 @@
...
+    {
+        Input label = new Input(Locator.name("label").findWhenNeeded(this), getDriver());
+        Input sequenceNumMin = new Input(Locator.name("sequenceNumMin").findWhenNeeded(this), getDriver());
+        Input sequenceNumMax = new Input(Locator.name("sequenceNumMax").findWhenNeeded(this), getDriver());
+        WebElement description = Locator.textarea("description").findWhenNeeded(this);
+        Select type = SelectWrapper.Select(Locator.name("typeCode")).findWhenNeeded(this);
+        Select sequenceNumHandling = SelectWrapper.Select(Locator.name("sequenceNumHandling")).findWhenNeeded(this);
+        Checkbox showByDefault = new Checkbox(Locator.name("showByDefault").findWhenNeeded(this));
+        WebElement save = Locator.linkWithText("Save").findWhenNeeded(this);
+        WebElement cancel = Locator.linkWithText("Cancel").findWhenNeeded(this);
```
labkey-danield: I believe the general practice / style is to mark these as `final`.

labkey-danield (approved):
There is nothing I see that would prevent merging, but I think there might still be some missing test cases that you may want to discuss with @labkey-klum.

For example:

- Canceling out of the form.
- Clicking Save and expecting some validation error.
- Are there scoping considerations? Like the sample type exists in the Shared folder or a child folder, or some of the samples do.

src/org/labkey/test/pages/study/CreateNewVisitPage.java
labkey-sweta: This is very old page, i created a page bare minimum methods that i needed for testing new feature

src/org/labkey/test/pages/study/CreateNewVisitPage.java
labkey-sweta: changed, might be good option if we can update the Page template.

src/org/labkey/test/pages/study/CreateNewVisitPage.java
labkey-sweta: Never needed it.

PR #2001, opened 2024-07-30
Stop testing deprecated hasPermission() function

labkey-adam:
#### Rationale
`LABKEY.Security.hasPermission()` has been deprecated and is no longer useful, since bitmask-based permissions are no longer returned by the server.

Also, remove warnings, address styling inconsistencies, and fix broken `addGroupMembers()` check.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/5711

### DIFF
--- a/modules/simpletest/resources/scripts/validationTest/securityTest.js
+++ b/modules/simpletest/resources/scripts/validationTest/securityTest.js
@@ -25,7 +25,7 @@ function doTest()
     });
 
     // LABKEY.Security.currentUser
-    if(!currentUser.isSystemAdmin)
+    if (!currentUser.isSystemAdmin)
         errors[errors.length] = Error("Security.currentUser = " + Ext.util.JSON.encode(currentUser));
 
     //LABKEY.Security.createContainer()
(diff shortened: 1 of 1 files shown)


modules/simpletest/resources/scripts/validationTest/securityTest.js
```diff
@@ -25,42 +27,42 @@ function doTest()
...
 
     // LABKEY.Security.addGroupMembers()
     result = LABKEY.Security.addGroupMembers({
         groupId: group.id,
         principalIds: [user.userId],
         containerPath: containerPath
     });
-    if(result.added != user.userId )
-        errors[errors.length] = new Error("Security.addGroupMembers() = "+Ext.util.JSON.encode(result));
+    if (result.added[0] !== user.userId)
```
labkey-tchad: Should this guard against the (remote) possibility of `result.added` not having a single item?
```suggestion
    if (result.added.length !== 1 || result.added[0] !== user.userId)
```

labkey-tchad (approved):

modules/simpletest/resources/scripts/validationTest/securityTest.js
labkey-adam: Yes!

PR #2005, opened 2024-08-01
New wait for search indexer fails if the search module isn't present

labkey-sweta:
Rationale
New wait for search indexer fails if the search module isn't present

https://teamcity.labkey.org/buildConfiguration/LabKey_247Release_Premium_GitExtraPostgres/3127873?buildTab=tests&status=failed&expandedTest=build%3A%28id%3A3127860%29%2Cid%3A2000000017

Related Pull Requests
https://github.com/LabKey/testAutomation/pull/1989

Changes

### DIFF
--- a/src/org/labkey/test/LabKeySiteWrapper.java
+++ b/src/org/labkey/test/LabKeySiteWrapper.java
@@ -33,10 +33,8 @@ import org.apache.hc.core5.http.ProtocolException;
 import org.apache.hc.core5.http.io.entity.EntityUtils;
 import org.apache.hc.core5.http.message.BasicNameValuePair;
 import org.apache.hc.core5.http.protocol.HttpContext;
-import org.awaitility.Awaitility;
 import org.jetbrains.annotations.Nullable;
 import org.json.JSONObject;
-import org.junit.Assert;
 import org.junit.Assume;
 import org.labkey.api.data.dialect.DatabaseNotSupportedException;
 import org.labkey.remoteapi.CommandException;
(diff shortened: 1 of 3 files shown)


labkey-tchad (changes requested):
- `lastException.getCause()` will always be `null` here, making this change an effective no-op
- Even if this worked the way you intend, it would always wait the maximum duration if the search service isn't available. A 404 from the search command should be treated as a success, not force the loop to continue until the timer runs out
- The status code (`CommandException.getStatusCode == 404`) would be a more precise comparison than the exception message ("Not Found")
- `LabKeySiteWrapper.checkForUpgrade` also needs to be able to handle this case

Since there are two places that need this wait. I'd move it into a single utility method that can be shared (maybe something like `SearchAdminAPIHelper.waitForSearchServiceBootstrap(connection)`)

labkey-sweta:
Teamcity run looks clean for Fda My Studies and Search SQL server.
[FDA My Studies](https://teamcity.labkey.org/buildConfiguration/LabKey_247Release_FdaMyStudies_MyStudiesRegistrationPostgres/3134387?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true&expandBuildDeploymentsSection=false)

[SQL Search Module](https://teamcity.labkey.org/buildConfiguration/LabKey_247Release_Community_ModuleSuites_SearchSqlserver/3134385?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildDeploymentsSection=false&expandBuildChangesSection=true)

labkey-tchad (approved):
I don't think that BVT failure has anything to do with this change. It only failed in 1 of the 4 runs.

PR #2007, opened 2024-08-02
Issue 50924: LKSM: Importing samples using naming expression referencing parent inputs with #

XingY:
#### Rationale
Add selenium tests for importing samples with parents whose name starts with #.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1548
- https://github.com/LabKey/platform/pull/5732
- https://github.com/LabKey/testAutomation/pull/2007
- https://github.com/LabKey/limsModules/pull/537

#### Changes
*


src/org/labkey/test/tests/SampleTypeNameExpressionTest.java
```diff
@@ -180,6 +192,67 @@ public void testSimpleNameExpression()
...
+        final String sampleTypeName = "ParentNamesExprTest";
+
+        log("Verify import tsv would ignore lines starting with #");
+        String nameExpression = "${MaterialInputs/" + PARENT_SAMPLE_TYPE + "}-child";
+        String data = "MaterialInputs/" + PARENT_SAMPLE_TYPE + "\n";
+        data += PARENT_SAMPLE_01 + "\n";
+        data += PARENT_SAMPLE_01 + "," + PARENT_SAMPLE_02 + "," + PARENT_SAMPLE_03 + "\n";
+        // tsv lines starting with # should be ignored
+        data += PARENT_SAMPLE_03 + "\n";
+        data += PARENT_SAMPLE_03 + "," + PARENT_SAMPLE_02 + "\n";
```
labkey-susanh, line 217: What if the sampleID is not the first field in the file?  That should succeed in creating the sample, right?

XingY: Yes, there is a test case for that immediately below.

labkey-susanh: I see the test for the parent input being not the first field, but I was wondering about when creating the sample itself.

XingY: I see, I will add a test case for that. As long as the line doesn't start with #, it should work

XingY: test added

src/org/labkey/test/tests/SampleTypeNameExpressionTest.java
```diff
@@ -180,6 +192,71 @@ public void testSimpleNameExpression()
...
+        names = materialTable.getColumnDataAsText("Name");
+        log("generated sample names:");
+        names.forEach(this::log);
+
+        assertEquals(7, names.size());
+        assertEquals("[" + PARENT_SAMPLE_01 + ", " + PARENT_SAMPLE_04 + "]-child", names.get(0));
+        assertEquals("[" + PARENT_SAMPLE_04 + ", " + PARENT_SAMPLE_03 + "]-child", names.get(1));
+        assertEquals(PARENT_SAMPLE_04 + "-child", names.get(2));
+
+        log("Verify importing tsv to create sample should ignore lines starting with #");
```
labkey-susanh, line 277: And TSV should work when #-named sample is not the first field? Or is this covered above and I'm not seeing it.

XingY: pushed..

labkey-susanh (approved):

labkey-danield (approved):

PR #2011, opened 2024-08-07
Issue 50826: Add Sample Creation Order option for fill order when adding or moving samples in storage

labkey-susanh:
#### Rationale
Issue [50826](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50826): When samples are created and then users want to add them to storage, they often want them added to storage in the same order in which they were created. Though sometimes this can equate to doing an ascending sort on the sample names, this is not always the case.  Here we add an option to sort the samples by RowId, which is effectively the order in which they were created.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-premium/pull/487
* https://github.com/LabKey/limsModules/pull/545

#### Changes
* Add new utility method `getSortedRowIdToSamplesMap`


labkey-danield (approved):
At some point it might make sense to break out the functionality of `getSortedRowIdToSamplesMap` and `getRowIdsForSamples` into a private function that takes SelectRowsCommand parameter. If that does happen I think the retry code would have to be pushed back into `getSortedRowIdToSamplesMap` and `getRowIdsForSamples`.

PR #2016, opened 2024-08-14
EditableGrid: remove dependence on QueryModel

labkey-alan:
#### Rationale
See rationale in ui-components

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1534
- https://github.com/LabKey/labkey-ui-premium/pull/470
- https://github.com/LabKey/platform/pull/5687
- https://github.com/LabKey/limsModules/pull/469
- https://github.com/LabKey/testAutomation/pull/2016

#### Changes
* EditableGrid: wait for input cells

### DIFF
--- a/src/org/labkey/test/components/ui/grids/EditableGrid.java
+++ b/src/org/labkey/test/components/ui/grids/EditableGrid.java
@@ -1096,6 +1096,9 @@ public class EditableGrid extends WebDriverComponent<EditableGrid.ElementCache>
 
         public WebElement inputCell()
         {
+            // Wait for the input to appear because there are some cases where we type into a cell to trigger the input
+            // to appear, and then immediately try to type into the input and it may not be there.
+            Locators.inputCell.waitForElement(table, WAIT_FOR_JAVASCRIPT);
             return Locators.inputCell.findElement(table);
         }
 


src/org/labkey/test/components/ui/grids/EditableGrid.java
```diff
@@ -1096,6 +1096,9 @@ public List<String> getColumnNames()
 
         public WebElement inputCell()
         {
+            // Wait for the input to appear because there are some cases where we type into a cell to trigger the input
+            // to appear, and then immediately try to type into the input and it may not be there.
+            Locators.inputCell.waitForElement(table, WAIT_FOR_JAVASCRIPT);
```
labkey-chrisj, line 1101: you can return the value of `Locators.inputCell.waitForElement(table, WAIT_FOR_JAVASCRIPT)` here
waiting for it is more robust than using .findElement in the sense that it will tolerate it if the element isn't already there by the time the test thread arrives.

labkey-chrisj (approved):

labkey-alan:
I am going to close this PR as the problem was actually a bug on the client side, and my little test fix was only working due to other coincidences.

PR #2017, opened 2024-08-14
Automation test for Issue 49483: Test permissions on admin console links

labkey-sweta:
#### Rationale
Test coverage added for 
1. As a site admin, collect all the admin console links and ensure they can be accessed
2. Impersonate the Troubleshooter role, collect all the admin console links now shown, and ensure they can be accessed. Maybe fail if "Save", "Submit", or "Update" buttons are seen? Also, determine the "site admin" links that went away after impersonating and ensure that none of those can be accessed.
3. Impersonate a non-admin, non-troubleshooter user and ensure that none of the "site admin" links can be accessed, except known exceptions (e.g., Credits).

Issue link [https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49483](url)

#### Related Pull Requests
* 

#### Changes
*


labkey-tchad (changes requested):
The new checks should avoid hitting the "Dump Heap" page. We don't want to actually do a heap dump.

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
```diff
@@ -71,7 +102,77 @@ public void testAdminNavTrails()
...
+        Map<String, String> linkHrefs = new LinkedHashMap<>();
+        List<WebElement> troubleshooterLinks = adminConsole.getAllAdminConsoleLinks();
+        for (WebElement link : troubleshooterLinks)
+            linkHrefs.put(link.getText(), link.getAttribute("href"));
+
+        for (Map.Entry<String, String> link : linkHrefs.entrySet())
+        {
+            if (!link.getKey().equalsIgnoreCase("SignUp")           //Signup link shows up to the troubleshooter but throws 403 while accessing it.
+                    && !link.getKey().equalsIgnoreCase("Profiler")) //Profiler can be edited by the troubleshooter
+            {
```
labkey-adam, line 138: Creating a `Set.of()` ignoreLinks is going to be easier to maintain vs. multiple if conditions. Use the same pattern as the navtrail test above.

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
```diff
@@ -71,7 +102,77 @@ public void testAdminNavTrails()
...
+            linkHrefs.put(link.getText(), link.getAttribute("href"));
+
+        log("Verifying links can be access by admin");
+        for (Map.Entry<String, String> link : linkHrefs.entrySet())
+        {
+            log("Verifying link " + link.getKey() + " with URL " + link.getValue());
+            verifyLinks(link.getKey(), link.getValue(), 200);
+        }
+
+        log("Verifying links cannot be access by non admin");
```
labkey-adam:
```suggestion
        log("Verifying links cannot be accessed by non admin");
```

labkey-adam (changes requested):

labkey-danield:
Sorry to pile on @labkey-sweta  :)

This isn't getting past doSetup when [running on TeamCity](https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_GitPostgres?branch=adminConsoleLinkTest&buildTypeTab=overview&mode=builds). Please try to get it running on TeamCity before re-submitting for review.

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
labkey-sweta: Done

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
labkey-sweta: Done

labkey-sweta:
> The new checks should avoid hitting the "Dump Heap" page. We don't want to actually do a heap dump.

along with heap dump, should Reset Site Errors and Running Threads be ignored as well? @labkey-adam  and @labkey-tchad

labkey-tchad:
> > The new checks should avoid hitting the "Dump Heap" page. We don't want to actually do a heap dump.
> 
> along with heap dump, should Reset Site Errors and Running Threads be ignored as well? @labkey-adam and @labkey-tchad

Those links go to a confirmation page so won't actually do anything. Arguably, the "Dump Heap" link should also go to a confirmation page.

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
```diff
@@ -71,7 +108,134 @@ public void testAdminNavTrails()
...
+        impersonate(TROUBLESHOOTER);
+        Map<String, String> linkHrefs = new LinkedHashMap<>();
+        List<WebElement> troubleshooterLinks = adminConsole.getAllAdminConsoleLinks();
+        for (WebElement link : troubleshooterLinks)
+            linkHrefs.put(link.getText(), link.getAttribute("href"));
+
+        for (Map.Entry<String, String> link : linkHrefs.entrySet())
+        {
+            if (ignoredLinks.contains(link.getKey()))
+                TestLogger.log("Skipping admin link: " + link.getKey());
```
labkey-danield, line 136: Nit:
You shouldn't need explicit reference to `TestLogger`

labkey-danield (approved):

labkey-adam (approved):

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
```diff
@@ -71,7 +108,134 @@ public void testAdminNavTrails()
...
+    }
+
+    @Test
+    public void testTroubleshooterLinkAccess()
+    {
+        Set<String> ignoredLinks = Collections.newSetFromMap(new CaseInsensitiveHashMap<>());
+        ignoredLinks.addAll(List.of(
+                "SignUp",                   //link shows up to the troubleshooter but throws 403 while accessing it.
+                "Notebook Settings",        //link shows up to the troubleshooter but throws 403 while accessing it.
+                "Puppeteer Service",        //link shows up to the troubleshooter but throws 403 while accessing it.
```
labkey-tchad: I think these are errors. If a troubleshooter can't visit these pages, the links shouldn't show up on the admin console

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
```diff
@@ -71,7 +108,134 @@ public void testAdminNavTrails()
...
+
+    @Test
+    public void testAdminConsoleLinksForAdminAndNonAdmin()
+    {
+        Set<String> ignoredLinksNonAdmin = Collections.newSetFromMap(new CaseInsensitiveHashMap<>());
+        ignoredLinksNonAdmin.addAll(List.of(
+                "Merge sync admin",     //Can be accessed by non admin
+                "Dump Heap",            // Undesired consequences
+                "Reset Site Errors",    // Undesired consequences
+                "Running Threads",      // Undesired consequences
```
labkey-tchad: A non-admin shouldn't be able to visit these pages at all, let alone trigger any undesired side-effects

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
```diff
@@ -71,7 +108,134 @@ public void testAdminNavTrails()
...
+    public void testTroubleshooterLinkAccess()
+    {
+        Set<String> ignoredLinks = Collections.newSetFromMap(new CaseInsensitiveHashMap<>());
+        ignoredLinks.addAll(List.of(
+                "SignUp",                   //link shows up to the troubleshooter but throws 403 while accessing it.
+                "Notebook Settings",        //link shows up to the troubleshooter but throws 403 while accessing it.
+                "Puppeteer Service",        //link shows up to the troubleshooter but throws 403 while accessing it.
+                "Dump Heap",                // Undesired consequences
+                "Reset Site Errors",        // Undesired consequences
+                "Running Threads",          // Undesired consequences
```
labkey-tchad: These don't actually have any undesired consequences; behavior may have changed since `testAdminNavTrails` was written.

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
```diff
@@ -71,7 +108,134 @@ public void testAdminNavTrails()
...
+            throw new RuntimeException(e);
+        }
+        catch (CommandException e)
+        {
+            Assert.assertEquals("Post command should not go through", 404, e.getStatusCode());
+        }
+        catch (JavascriptException e)
+        {
+            return; //Do nothing for admin-showAllErrors.view for error LABKEY is not defined
+        }
```
labkey-tchad: There is no way our remote Java API would throw any kind of selenium exception.
The exception you are seeing is probably from the call to `getCurrentContainerPath` above. All of the admin console links are at the root container (I think), so there's no reason to call that method

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
```diff
@@ -71,7 +108,134 @@ public void testAdminNavTrails()
...
+        beginAt(url);
+        Assert.assertEquals("URL " + url + " is broken for " + text, 200, getResponseCode());
+        Locator.XPathLocator updateButtonLoc = Locator.XPathLocator.union(
+                Locator.linkWithText("Submit"), Locator.linkWithText("Save"), Locator.linkWithText("Update")).notHidden();
+        assertFalse("Either Save/Submit/Update button exists in the page " + text, isElementPresent(updateButtonLoc));
+    }
+
+    private void verifyPostCommandFails(String url)
+    {
+        String[] actionController = url.substring(url.lastIndexOf('/') + 1).split("-");
```
labkey-tchad: This isn't a valid way to extract the action and controller from a URL; some controllers include a hyphen (e.g. `mini-profiler-manage.view`.
`new Crawler.ControllerActionId(url)` will correctly parse all of the url variations we support. It will also give you the container path, which could be used below (though I would expect it to always be root).

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
```diff
@@ -71,7 +108,134 @@ public void testAdminNavTrails()
...
+            command.execute(connection, getCurrentContainerPath());
+        }
+        catch (IOException e)
+        {
+            throw new RuntimeException(e);
+        }
+        catch (CommandException e)
+        {
+            Assert.assertEquals("Post command should not go through", 404, e.getStatusCode());
+        }
```
labkey-tchad: These don't return a 403? 404 is surprising.

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
```diff
@@ -71,7 +108,134 @@ public void testAdminNavTrails()
...
+            {
+                log("Verifying link " + link.getKey() + " with URL " + link.getValue());
+                verifyLinks(link.getKey(), link.getValue(), 403);
+            }
+        }
+        goToHome();
+        stopImpersonating();
+    }
+
+    private void verifyReadOnlyLinks(String text, String url)
```
labkey-tchad: This method only verifies a single link
```suggestion
    private void verifyReadOnlyLink(String text, String url)
```

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
```diff
@@ -71,7 +108,134 @@ public void testAdminNavTrails()
...
+        {
+            Assert.assertEquals("Post command should not go through", 404, e.getStatusCode());
+        }
+        catch (JavascriptException e)
+        {
+            return; //Do nothing for admin-showAllErrors.view for error LABKEY is not defined
+        }
+    }
+
+    private void verifyLinks(String text, String url, int expectedResponseCode)
```
labkey-tchad: Only verifies a single link
```suggestion
    private void verifyLink(String text, String url, int expectedResponseCode)
```

labkey-tchad (changes requested):

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
labkey-adam: `Running Threads` spits a thread dump into the log. I think that's the "undesired consequence", but if that noise isn't a problem for the tests or test investigation, then I agree.

Likewise, if the tests & test framework aren't sensitive to `Reset Site Errors` then it's fine to click it as well.

Any changes made here should be made to the navtrail test as well.

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
labkey-sweta: Nope, i was surprised too

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
labkey-sweta: This ignore list is mainly for admins not non admins

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
labkey-sweta: Done

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
labkey-sweta: Done

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
labkey-sweta: Done

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
labkey-sweta: Done

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
labkey-sweta: Done

src/org/labkey/test/tests/AdminConsoleNavigationTest.java
labkey-sweta: For Notebook settings and Puppeteer service its discussed in this issue [Troubleshooters see error page for Notebook Settings and Puppeteer Service from admin console](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49129) @labkey-adam Should the troubleshooter not see the links at all?

Comments from adam "Signup is the MacCoss-developed thing? We should probably report this to Vagisha and call it a day."

labkey-adam: In all three cases, Troubleshooters should either see the links and be able to access the page (in a read-only mode) OR they shouldn't see the link. I don't know what's most appropriate for these actions or if there are read-only versions of these pages. Of course the easiest thing to do is restrict the link permissions. I would open issues to Vagisha (SignUp) and Nick (the other two), annotate comments with each issue number, leave them as ignored for now, and update the test once the issues are resolved.

labkey-sweta: @labkey-adam Will do that 
Thank you adam

labkey-tchad (approved):
Just needs links to the issues that Adam suggested opening.

PR #2018, opened 2024-08-14
Update MultiMenu for issue 50901

### LINKED ISSUE #2003: Fix for issue 50901
[Issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50901)

MultiMenu's doMenuAction (String... path) has an implicit promise to click the next item under the previous one, but in practice it will click the first item by that name that it finds in the menu list.  
Until recently, this didn't matter because items in menus like this one have always been effectively unique.  Recent changes in the way we export plate data and worklists and instrument instructions in Biologics Plates mean that there will be as many as 3 menu-items with the same text (Excel, CSV, TSV formats) under different headings or expansion-menus (e.g. export plate data, worklists, instrument instructions).

Further discussion with Alan and Rosaline on usages of the underlying shared product component reveal that there are no usages (and will probably never be usages) with the menu-path being more than two deep.  
There currently exists a way to identify which items are under a heading, and the same logic can be used to identify which items are under a sub-menu.  Implementing a method to do this, and making the means to expand/collapse submenus public, should go a long ways towards helping tests and pages to be able to click the intended menu item (and not another one with the same text).

Tasks 

- [x] design discussion .75
- [x] Don't extend BootstrapMenu
     - [x]  implement expand, isExpanded
     - [x] implement collapse
- [x] Method to expand a named submenu .25
- [x] Method to collapse a named submenu .25
- [x] reimplement doMenuAction as 1-argument, 2-argument methods
- [x] Method to click an item by name under a given heading .25
- [x] Method to click an item by name under a named subMenu .5
- [x] method to get items under submenu
- [x] refactor usages where BootstrapMenus use MultiMenu locators and vice versa 2.5
- [x] refactor params, list usages where necessary - 1
- [x] investigate/resolve failing tests -1
- [x] tc and results
- [x] code review
- [x] merge

labkey-chrisj:
#### Rationale
This change updates `MultiMenu `to be able to click list-items under a given heading, or under a given menu-item path.

[related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50901)
[github issue](https://github.com/LabKey/testAutomation/issues/2003)

#### Related Pull Requests
https://github.com/LabKey/limsModules/pull/569

#### Changes

- [x] Add method to get item under menu-item, like `getItemsUnderHeading`
- [x] Refactor `doMenuActions `into 2 methods that take 1 or 2 strings, instead of params string
- [x] Refactor usage of params/list doMenuActions to call the new doMenuActions signatures
- [ ] consider implementing waitForReady method for codepaths that don't include expandAll/waitForData

### DIFF
--- a/src/org/labkey/test/components/react/MultiMenu.java
+++ b/src/org/labkey/test/components/react/MultiMenu.java
@@ -49,6 +49,8 @@ public class MultiMenu extends BootstrapMenu
      */
     public WebElement getMenuItem(String menuItem)
     {
+        expand();
+        waitForData();
         return Locators.menuItem().withText(menuItem).waitForElement(getMenuList(), _menuWaitTmeout);
     }
 
@@ -91,7 +93,6 @@ public class MultiMenu extends BootstrapMenu
     @LogMethod(quiet = true)
     public void doMenuAction(@LoggedParam String menuAction)
     {
-        expand();
         var item = getMenuItem(menuAction);
         getWrapper().shortWait().until(ExpectedConditions.elementToBeClickable(item));
         item.click();
@@ -104,7 +105,6 @@ public class MultiMenu extends BootstrapMenu
      */
     public void doMenuAction(@LoggedParam String toggleText, @LoggedParam String menuAction)
     {
-        expand();
         clickMenuItemUnderToggle(toggleText, menuAction);
     }
 
@@ -213,6 +213,7 @@ public class MultiMenu extends BootstrapMenu
     public WebElement expandToggle(String toggle)
     {
         expand();
+        waitForData();
         int listItemCount = getListItems().size();
 
         // find the toggle-item; it may be expanded already
@@ -326,8 +327,6 @@ public class MultiMenu extends BootstrapMenu
      */
     protected List<WebElement> getMenuItemsUnderToggle(String toggle)
     {
-        expand();
-        waitForData();
         expandToggle(toggle);
 
         List<WebElement> itemsUnderToggle = new ArrayList<>();
--- a/src/org/labkey/test/components/react/MultiMenu.java
+++ b/src/org/labkey/test/components/react/MultiMenu.java
@@ -40,44 +40,43 @@ public class MultiMenu extends BootstrapMenu
         return Locator.tagWithClass("ul", "dropdown-menu").findElement(this);
     }
 
-    /*
-        finds the first menu item with the specified text
-     */
-    public WebElement getMenuItem(String text)
-    {
-        return Locators.menuItem().withText(text).waitForElement(getMenuList(), _menuWaitTmeout);
-    }
-
     /**
-     * Find the first item with specified text under the path of the
-     * @param menuText Text of the menu to expand
-     * @param itemText  Text of the item to find under its path
+     * finds the first menu item with the specified text.
+     * If there are duplicate items in the menu by text and the one you want isn't the first in the menu,
+     * consider using getMenuItemUnderToggle or getMenuItemUnderHeading
+     * @param menuItem Text of the intended menuItem
      * @return
      */
-    public WebElement getMenuItem(String menuText, String itemText)
+    public WebElement getMenuItem(String menuItem)
     {
-        return getItemUnderToggle(menuText, itemText);
+        return Locators.menuItem().withText(menuItem).waitForElement(getMenuList(), _menuWaitTmeout);
     }
 
     /**
      * Checks whether the specified menu item is disabled
-     * @param text Text of the menu item
-     * @return  true if class contains 'disabled'
+     * @param menuItem Text of the menu item
+     * @return  true if the item is not enabled
      */
-    public boolean isMenuItemDisabled(String text)
+    public boolean isMenuItemDisabled(String menuItem)
     {
-        return getMenuItem(text).getAttribute("class").toLowerCase().contains("disabled");
+        return !getMenuItem(menuItem).isEnabled();
     }
 
-    public boolean isMenuItemDisabled(String menuText, String itemText)
+    /**
+     * Checks whether the specified menu item is disabled
+     * @param toggle Text of the toggle under which to find the menu item
+     * @param menuItem  Text of the menu item
+     * @return  true if the item is not enabled
+     */
+    public boolean isMenuItemUnderToggleDisabled(String toggle, String menuItem)
     {
-        return getItemUnderToggle(menuText, itemText).getAttribute("class").toLowerCase().contains("disabled");
+        return !getMenuItemUnderToggle(toggle, menuItem).isEnabled();
     }
 
     /**
      * gets all list-items currently appearing in the menu.  This includes header and separator items,
-     * which are special and don't have links
-     * @return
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/components/react/MultiMenu.java
```diff
@@ -40,48 +40,73 @@ private WebElement getMenuList()
...
+    /**
+     * Find the first item with specified text under the path of the
+     * @param menuText Text of the menu to expand
+     * @param itemText  Text of the item to find under its path
+     * @return
+     */
+    public WebElement getMenuItem(String menuText, String itemText)
+    {
+        return getItemUnderToggle(menuText, itemText);
+    }
```
labkey-tchad: Why not just make `getItemUnderToggle` public? I like that name better because it is more explicit; maybe name it `getMenuItemUnderToggle` to match the name of `getMenuItem`.

src/org/labkey/test/components/react/MultiMenu.java
```diff
@@ -40,48 +40,73 @@ private WebElement getMenuList()
...
+     * @return  true if class contains 'disabled'
+     */
     public boolean isMenuItemDisabled(String text)
     {
         return getMenuItem(text).getAttribute("class").toLowerCase().contains("disabled");
     }
 
+    public boolean isMenuItemDisabled(String menuText, String itemText)
+    {
+        return getItemUnderToggle(menuText, itemText).getAttribute("class").toLowerCase().contains("disabled");
```
labkey-tchad: Does `WebElement.isEnabled` work for these?
```suggestion
        return !getItemUnderToggle(menuText, itemText).isEnabled();
```

src/org/labkey/test/components/react/MultiMenu.java
```diff
@@ -186,34 +206,182 @@ public List<String> getMenuText()
...
+    private WebElement getItemUnderHeading(String toggleLabel, String label)
+    {
+        var itemsUnderHeading = getItemsUnderHeading(toggleLabel);
+        for (WebElement item : itemsUnderHeading)
+        {
+            if (item.getText().trim().equals(label))
+                return item;
+        }
+        throw new NoSuchElementException("No item '" + label + "' under toggle '" + toggleLabel + "' found.");
+    }
```
labkey-tchad: `toggle` ≠ `heading`, right?
```suggestion
    private WebElement getItemUnderHeading(String heading, String label)
    {
        var itemsUnderHeading = getItemsUnderHeading(heading);
        for (WebElement item : itemsUnderHeading)
        {
            if (item.getText().trim().equals(label))
                return item;
        }
        throw new NoSuchElementException("No item '" + label + "' under heading '" + heading + "' found.");
    }
```

src/org/labkey/test/components/react/MultiMenu.java
```diff
@@ -186,34 +206,182 @@ public List<String> getMenuText()
...
+        getWrapper().shortWait().until(ExpectedConditions.elementToBeClickable(item));
+        item.click();
+    }
+
+    /**
+     * gets the names of dropdown-section__menu-items between a heading and the next divider
+     *
+     * @param heading The text of the heading
+     * @param collapse Whether to collapse the menu when finished
+     * @return
```
labkey-tchad: `heading` ≠ `toggle`
```suggestion
     * gets the names of dropdown-section__menu-items between a toggle and the next toggle or divider
     *
     * @param heading The text of the toggle
     * @param collapse Whether to collapse the menu when finished
     * @return texts of menu items under toggle
```

src/org/labkey/test/components/react/MultiMenu.java
```diff
@@ -186,34 +206,182 @@ public List<String> getMenuText()
...
+                break;
+        }
+
+        return itemsUnderToggle;
+    }
+
+    private WebElement getItemUnderToggle(String toggleLabel, String label)
+    {
+        waitForData();
+        var itemsUnderToggle = getItemsUnderToggle(toggleLabel);
```
labkey-tchad: `getItemsUnderToggle` already waits for data
```suggestion
        var itemsUnderToggle = getItemsUnderToggle(toggleLabel);
```

src/org/labkey/test/components/ui/grids/GridBar.java
```diff
@@ -205,7 +205,10 @@ public void doMenuAction(String buttonText, List<String> menuActions)
 
         if(found)
         {
-            multiMenu.doMenuAction(menuActions);
+            if (menuActions.size() == 1)
+                multiMenu.doMenuAction(menuActions.get(0));
+            else
+                multiMenu.doMenuAction(menuActions.get(0), menuActions.get(1));
```
labkey-tchad: Make sure we didn't miss some place that does pass in three actions
```suggestion
            if (menuActions.size() == 1)
                multiMenu.doMenuAction(menuActions.get(0));
            else if (menuActions.size() == 2)
                multiMenu.doMenuAction(menuActions.get(0), menuActions.get(1));
            else
                throw new IllegalArgumentException();
```

labkey-tchad (changes requested):
Quite a few places conflating "header"s and "toggle"s. I think I flagged them all but please double-check. This could be a huge point of confusion for folks using this component.

src/org/labkey/test/components/react/MultiMenu.java
labkey-chrisj: Let's find out!

src/org/labkey/test/components/react/MultiMenu.java
labkey-chrisj: Thank you, good catch

src/org/labkey/test/components/ui/grids/GridBar.java
labkey-chrisj: Yes, I kind of went back and forth over calling dropdown-section-toggles menuItems and toggles, wanted to avoid conflating the menu-item-toggles with the menu toggle
Also some of those in there were copy-pasta problems my eyeballs just glazed over.

Let's call dropdown-section-toggles toggles now

src/org/labkey/test/components/react/MultiMenu.java
labkey-chrisj: The only reason for that, really, was to align to the convention where the parameters of like-named methods with 1 and 2 params signify that the second one, if any, is to be found under a menu specified by the first.
But, I'll take this question as a suggestion

src/org/labkey/test/components/react/MultiMenu.java
labkey-chrisj: Sadly, I found out that webElement.isEnabled can't tell the difference between having 'disabled' in the class and not.
restoring use of class to determine disabled for now
I'll file a backlog issue to track our problem of having lots of non-standard enable/disable implementations in our products

src/org/labkey/test/components/react/MultiMenu.java
labkey-chrisj: I'm now considering whether to put waitForData into the path for getting first-level menu items as well

labkey-tchad (approved):

PR #2030, opened 2024-08-27
wait for the save button to be ready before clicking

labkey-chrisj:
#### Rationale
Yesterday (and recently), [CrossSiteScriptingForDeleteTest.verifyReportNameIsEncoded](https://teamcity.labkey.org/viewLog.html?buildId=3110547&tab=buildResultsDiv&buildTypeId=LabKey_247Release_Community_DailySuites_DailyASqlserver#testNameId-1213030100997127121) has failed in a way that seems likely to be due to clicking the 'save' button before it was ready- the most-recent failure waited for navigation for the default page timeout.  This one-line change will hopefully address this.

#### Related Pull Requests
n/a

#### Changes
wait for the 'save' button before clicking

### DIFF
--- a/src/org/labkey/test/tests/CrossSiteScriptingForDeleteTest.java
+++ b/src/org/labkey/test/tests/CrossSiteScriptingForDeleteTest.java
@@ -9,6 +9,7 @@ import org.labkey.test.WebTestHelper;
 import org.labkey.test.categories.Daily;
 import org.labkey.test.components.ext4.Window;
 import org.labkey.test.pages.reports.ManageViewsPage;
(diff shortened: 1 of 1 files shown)


labkey-sweta (commented):
The Save button is already present, i dont think this wait would fix the problem
What do you think?

labkey-chrisj:
> The Save button is already present, i dont think this wait would fix the problem What do you think?

My thought was that the test might be clicking it before it is ready 
I've changed it to explicitly wait for the button to be clickable, which might make no difference, but probably can't hurt

labkey-sweta (approved):

PR #2037, opened 2024-09-03
add regression coverage for Issue 22743

labkey-chrisj:
#### Rationale
This adds test coverage for Issue 22743, which is essentially that there isn't adequate/comprehensive test coverage for file and attachment columns in various domain kinds

#### Related Pull Requests
n/a

#### Changes

- [x] test for list, samples, assay, dataset with files/attachments, export and import
- [x] test for sample file fields, ui
- [x] test for assay run and result domain file fields, ui

### DIFF
--- a/src/org/labkey/test/tests/FileAttachmentColumnTest.java
+++ b/src/org/labkey/test/tests/FileAttachmentColumnTest.java
@@ -276,7 +276,6 @@ public class FileAttachmentColumnTest extends BaseWebDriverTest
         var otherResultFileTexts = resultsPage.getDataTable().getColumnDataAsText(OTHER_RESULT_FILE_COL);
         var runFileTexts = resultsPage.getDataTable().getColumnDataAsText("Run/runFile");
         var expectedRunFileLinkTexts = resultFiles.stream().map(File::getName).toList();
-        var expectedOtherResultFiles = SAMPLE_FILES.stream().map(File::getName).toList();
 
         checker().withScreenshot("unexpected_results_texts")
                 .wrapAssertion(()-> Assertions.assertThat(resultTxts)
@@ -295,7 +294,7 @@ public class FileAttachmentColumnTest extends BaseWebDriverTest
                 .wrapAssertion(()-> Assertions.assertThat(otherResultFileTexts.stream().map(String::trim).toList())
                         .as("expect other results files to have resolved")
                         .containsExactlyInAnyOrder("csv_sample.csv", "pdf_sample.pdf",
-                                "pdf_sample_with+%$@+%%+#-+=.pdf", "tif_sample.tif"));
+                                "pdf_sample_with+%$@+%%+#-+=.pdf", "tif_sample.tif", ""));  // empty value is for jpg, which doesn't get text/is rendered inline
         checker().withScreenshot("unexpected_run_file_links")
                 .wrapAssertion(()-> Assertions.assertThat(runFileTexts.stream().map(String::trim).toList())
                         .as("expect complete run files")
--- a/src/org/labkey/test/tests/FileAttachmentColumnTest.java
+++ b/src/org/labkey/test/tests/FileAttachmentColumnTest.java
@@ -110,6 +110,11 @@ public class FileAttachmentColumnTest extends BaseWebDriverTest
         createListWithData(EXPORT_FOLDER_PATH);
 
         //create sample types with file columns
+        WebDavUploadHelper uploadHelper = new WebDavUploadHelper(EXPORT_FOLDER_PATH);
+        for (File file : DATAFILE_DIRECTORY.listFiles())
+        {
+            uploadHelper.uploadFile(file);
+        }
         createSampleTypeWithData(EXPORT_SAMPLETYPE_NAME, EXPORT_FOLDER_PATH);
 
         // create an assay in the export folder
@@ -378,7 +383,7 @@ public class FileAttachmentColumnTest extends BaseWebDriverTest
         for (File file : DATAFILE_DIRECTORY.listFiles())
         {
             sampleFileData.add(Map.of("Name", file.getName(), "Color", "green",
-                    "File", String.format("\"%s\"",file.getAbsolutePath())));
+                    "File", file.getName()));
         }
 
         SampleTypeHelper sampleHelper = new SampleTypeHelper(this);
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/FileAttachmentColumnTest.java
```diff
@@ -15,153 +15,609 @@
...
+        {
+            var queryUpdatePage = samplesRegion.clickInsertNewRow();
+            queryUpdatePage.setField("Name", file.getName());
+            queryUpdatePage.setField("file", file)
+                    .submit();
+        }
+        validateSampleData(SUBA_SAMPLETYPE_NAME, SUBFOLDER_A_PATH, SAMPLE_FILES);
+
+        var fileContentPage = FileContentPage.beginAt(this, SUBFOLDER_A_PATH);
+        fileContentPage.fileBrowserHelper().selectFileBrowserItem("/sampletype/csv_sample.csv");
```
XingY, line 222: why "csv_sample.csv" needs to selectFileBrowserItem?

src/org/labkey/test/tests/FileAttachmentColumnTest.java
```diff
@@ -15,153 +15,609 @@
...
+    @Test
+    public void testSampleFileFields()
     {
-        _containerHelper.deleteProject(getProjectName(), afterTest);
+        SampleTypeDefinition subASampleType = new SampleTypeDefinition(SUBA_SAMPLETYPE_NAME)
+                .setFields(List.of(new FieldDefinition("color", ColumnType.String),
+                        new FieldDefinition("file", ColumnType.File)));
+        SampleTypeAPIHelper.createEmptySampleType(SUBFOLDER_A_PATH, subASampleType);
+
+        // give the sampleType actual files by adding individual rows; for the nonce not all files are resolving on import
```
XingY: Could you explain more why/what is not resolving on import?

src/org/labkey/test/tests/FileAttachmentColumnTest.java
```diff
@@ -15,153 +15,609 @@
...
-        init.doSetup();
+        String runName = "assay_run.tsv";
+
+        // generate some resultFiles
+        var resultFiles = new ArrayList<File>();
+        WebDavUploadHelper uploadHelper = new WebDavUploadHelper(SUBFOLDER_A_PATH);
+        List<Map<String, Object>> importData = new ArrayList<>();
+        for (int i = 0; i < 5; i++)
+        {
+            String fileName = String.format("field_file_for_results_domain-%d.tsv", i);
```
XingY: Why using 5 different result files? This seems they could be 5 data rows in a single result file. Or, have 1 result file that has N rows, and another result file with 1 row.

src/org/labkey/test/tests/FileAttachmentColumnTest.java
```diff
@@ -15,153 +15,609 @@
...
+            uploadHelper.uploadFile(resultFile);
+            resultFiles.add(resultFile);
+
+            importData.add(Map.of(RESULT_TXT_COL, result, RESULT_FILE_COL, resultFile.getName(),
+                    OTHER_RESULT_FILE_COL, otherResultFile.getName(),
+                    RUN_TXT_COL,"run text", RUN_FILE_COL, "runFile.tsv"));
+        }
+
+        // generate a run file, referencing the result files
+        String importDataFileContents = TestDataUtils.tsvStringFromRowMaps(importData,
```
XingY: Could use any file for run file, the content doesn't seem to matter, but using an existing static file should simplify the test

src/org/labkey/test/tests/FileAttachmentColumnTest.java
```diff
@@ -15,153 +15,609 @@
...
+        beginAt(SUBFOLDER_A_PATH + "/project-begin.view");
+        goToModule("Assay");
+        clickAndWait(Locator.linkContainingText(SUB_A_ASSAY));
+        clickButton("Import Data");
+        clickButton("Next");    // batch properties
+
+        // run properties
+        setFormElement(Locator.input("name"), runName);
+        setFormElement(Locator.input(RUN_TXT_COL), "run text");
+        setFormElement(Locator.input(RUN_FILE_COL), importFile);
```
XingY: Could use any existing sample file for importFile. Also, could you please rename this to runFileFieldFile?

src/org/labkey/test/tests/FileAttachmentColumnTest.java
labkey-chrisj: Ah.  That *was* true when that code was written- your fix in platform resolved the tricky-characters pdf file import that wasn't working in the sampleType in export folder.  I'll update

src/org/labkey/test/tests/FileAttachmentColumnTest.java
labkey-chrisj: selecting it with this path was the only way I could get the file browser to open into the /sampleType container.
It was just a file I knew I could expect to be there

src/org/labkey/test/tests/FileAttachmentColumnTest.java
labkey-chrisj: It was just the way that occurred for me to do it at the time.
If you would like me to change this, I will.

src/org/labkey/test/tests/FileAttachmentColumnTest.java
labkey-chrisj: Come to think of it, now that the samples are importing correctly we may not need this test at all- I wrote it when the test was trying (and failing) to bulk-import.  We get equivalent coverage from the export/import test now

src/org/labkey/test/tests/FileAttachmentColumnTest.java
labkey-chrisj: OK, I will modify the test accordingly

src/org/labkey/test/tests/FileAttachmentColumnTest.java
```diff
@@ -15,153 +15,628 @@
...
+        for (File file : OTHER_SAMPLE_FILES)
+        {
+            uploadHelper.uploadFile(file);
+        }
+
+        // generate some resultFiles
+        var resultFiles = new ArrayList<File>();
+
+
+        String runFileContent = """
```
XingY, line 251: I meant that we could just use a file from the existing testAutomation/data folder so we don't have to generate those temp files

XingY (approved):

PR #2040, opened 2024-09-05
Test Automation for Calculated Columns

labkey-danield:
#### Rationale
Updated some helpers for calculated columns tests in LIMS module.
Added a test in the LKS domain designer.

#### Related Pull Requests
* https://github.com/LabKey/limsModules/pull/661

#### Changes
* Updated DomainFieldRow:
    * Remove redundant code that was setting format.
    * Added method to set and get the calculated column expression.
    * Added a getter for the warning message.
    * Added a getter to see if 'Required Field' checkbox is present.
* Added Calculation column type.
* Added DomainDesignerTest.testCalculatedColumn
    * Validates that setting attributes before changing field to a Calculated type.


src/org/labkey/test/components/domain/DomainFieldRow.java
```diff
@@ -916,6 +975,40 @@ public DomainFieldRow deleteTextChoiceValue(String value)
...
+    }
+
+    public String getValueExpressionStatusMessage()
+    {
+        // Need to remove focus from the expression field to have the updated status message show up.
+        getWrapper().fireEvent(elementCache().expressionInput, WebDriverWrapper.SeleniumEvent.blur);
+        WebDriverWrapper.sleep(500);
+
+        String statusMsg = "";
+        if(Boolean.TRUE.equals(waitFor(elementCache().expressionStatusMsg::isDisplayed, 1_000)))
```
labkey-tchad: No need to do `TRUE.equals` here, `WebElement.isDisplayed()` won't ever return `null`.
```suggestion
        if(waitFor(elementCache().expressionStatusMsg::isDisplayed, 1_000))
```

src/org/labkey/test/params/FieldDefinition.java
```diff
@@ -509,7 +509,7 @@ public boolean isMeasureByDefault()
         ColumnType Barcode = new ColumnTypeImpl("Unique ID", "string", "http://www.labkey.org/types#storageUniqueId", null);
         ColumnType TextChoice = new ColumnTypeImpl("Text Choice", "string", "http://www.labkey.org/types#textChoice", null);
         ColumnType SMILES = new ColumnTypeImpl("SMILES", "string", "http://www.labkey.org/exp/xml#smiles", null);
-
+        ColumnType Calculation = new ColumnTypeImpl("Calculation", "http://www.labkey.org/exp/xml#calculated");
```
labkey-tchad: "`http://www.labkey.org/exp/xml#calculated`" is the `conceptURI` for this type, not the `rangeURI`. I think we want something like this:
```suggestion
        ColumnType Calculation = new ColumnTypeImpl("Calculation", null, "http://www.labkey.org/exp/xml#calculated", null);
```
Though it doesn't really matter unless we add API support for calculated columns

labkey-tchad (approved):
Would it be worth adding API support for calculated columns on the test side? It would allow us to do more thorough testing of various calculations without having to go through the designer.

cnathe (approved):

src/org/labkey/test/components/domain/DomainFieldRow.java
labkey-danield: Changed.

src/org/labkey/test/params/FieldDefinition.java
labkey-danield: Changed.
The plan is to add API support for calculated columns. I have to open a bug and assign it to Cory.

PR #2046, opened 2024-09-11
Update Editable Grid Test Component

labkey-danield:
#### Rationale
Addressing some failures because of changes in the editable grid behavior. Previously the way the text field was activated in the editable grid would clear the existing text (click the field and then send a 'space' key). That doesn't work the same as before. This change uses the input.clear() method to clear the text.

Some of the calculated columns test were failing because the validation msg wasn't getting updated. The message shows up after a expression is entered and the input control loses focus. This change changes the set method to remove focus when done.

#### Related Pull Requests
* https://github.com/LabKey/limsModules/pull/701

#### Changes
* Fixing how the editable grid clears a text field before setting the value.
* Have the grid bar allow for a 'More' menu.
* Update the DomainFieldRow.setValueExpression to lose focus after setting the value.

### DIFF
--- a/src/org/labkey/test/components/domain/DomainFieldRow.java
+++ b/src/org/labkey/test/components/domain/DomainFieldRow.java
@@ -16,6 +16,7 @@ import org.labkey.test.components.ui.ontology.ConceptPickerDialog;
 import org.labkey.test.params.FieldDefinition;
 import org.labkey.test.util.LabKeyExpectedConditions;
 import org.openqa.selenium.ElementNotInteractableException;
+import org.openqa.selenium.Keys;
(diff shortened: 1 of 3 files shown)


cnathe (approved):

labkey-tchad:
This behavior change sounds like a bug. In other spreadsheet apps, typing into a selected cell will overwrite the previous value.

labkey-danield:
> This behavior change sounds like a bug. In other spreadsheet apps, typing into a selected cell will overwrite the previous value.

The behavior works as expected manually, it only fails if done with Selenium.

labkey-tchad (approved):

PR #2065, opened 2024-09-20
Improve synchronization of `ProductMenu`

labkey-tchad:
#### Rationale
`ProductMenu`  doesn't synchronize its operations as much as it could. There are some cases that might trigger a page load and those that don't should wait for the menu to close and clear the element cache.
Removing instances of `refindWhenNeeded` will help enforce synchronization as well.

#### Related Pull Requests
- N/A

#### Changes
- Improve synchronization of `ProductMenu`


src/org/labkey/test/components/ui/navigation/ProductMenu.java
```diff
@@ -54,16 +53,21 @@ protected boolean isExpanded()
...
                 ariaExpanded, menuContentDisplayed));
 
-        return  ariaExpanded && menuContentDisplayed &&
-                ExpectedConditions.invisibilityOfAllElements(BootstrapLocators.loadingSpinner.findElements(this)).apply(getDriver());
+        return ariaExpanded && menuContentDisplayed;
+    }
+
+    protected boolean isLoaded()
+    {
+        return isExpanded() && Locator.tagWithClass("li", "menu-section-header").existsIn(elementCache().sectionContent);
```
labkey-chrisj, line 61: Perhaps `isLoaded `is where that check for a `loadingSpinner `belongs?

labkey-chrisj (approved):

PR #2067, opened 2024-09-23
dist.group no longer needs to be specified

labkey-adam:
#### Related Pull Requests
* https://github.com/LabKey/gradlePlugin/pull/219

### DIFF
--- a/distributions/teamcity/build.gradle
+++ b/distributions/teamcity/build.gradle
@@ -16,7 +16,6 @@ dist.description = "Distribution that includes all modules, for use in the conti
 
 project.tasks.register("distribution", ModuleDistribution) {
     ModuleDistribution dist ->
-        dist.group = GroupNames.DISTRIBUTION
         dist.description = "Make a LabKey modules distribution for 'test'"
 
         dist.subDirName = "test"
--- a/distributions/teamcity/build.gradle
+++ b/distributions/teamcity/build.gradle
@@ -21,6 +21,8 @@ project.tasks.register("distribution", ModuleDistribution) {
         dist.subDirName = "test"
         dist.extraFileIdentifier = '-test'
         dist.versionPrefix = 'Test'
+
+        dist.extraProperties = [supportedDatabases: "pgsql, mssql"]
 }
 
 


labkey-adam:
@labkey-tchad Should we update this distribution like all the others? That would mean renaming the directory to "test" and removing the properties. Or just removing the properties and going with "teamcity" as the canonical name.

Second question: I'm also tagging distributions with a "supportedDatabases" property, just like the module.xml property. Should I specify "pgsql, mssql" for the test distribution, for now?

labkey-tchad:
> @labkey-tchad Should we update this distribution like all the others? That would mean renaming the directory to "test" and removing the properties. Or just removing the properties and going with "teamcity" as the canonical name.
> 
> Second question: I'm also tagging distributions with a "supportedDatabases" property, just like the module.xml property. Should I specify "pgsql, mssql" for the test distribution, for now?

I think changing the project name would cause gradle problems; "test" is a predefined task name. We could change it to `teamcity`, it would just require some configuration updates on TeamCity.

"supportedDatabases" isn't really applicable to test distributions. It would vary depending on the modules present and whether the test distribution is inheriting from another distribution ("inheritedDistPath" property). Test distributions often aren't fully compatible with either pgsql or mssql; they include postgres-only and mssql-only modules.

labkey-tchad (approved):

labkey-adam:
> > @labkey-tchad Should we update this distribution like all the others? That would mean renaming the directory to "test" and removing the properties. Or just removing the properties and going with "teamcity" as the canonical name.
> > Second question: I'm also tagging distributions with a "supportedDatabases" property, just like the module.xml property. Should I specify "pgsql, mssql" for the test distribution, for now?
> 
> I think changing the project name would cause gradle problems; "test" is a predefined task name. We could change it to `teamcity`, it would just require some configuration updates on TeamCity.
> 
> "supportedDatabases" isn't really applicable to test distributions. It would vary depending on the modules present and whether the test distribution is inheriting from another distribution ("inheritedDistPath" property). Test distributions often aren't fully compatible with either pgsql or mssql; they include postgres-only and mssql-only modules.

@labkey-tchad I'm fine leaving the naming alone... up to you if you ever want to change it.

I neglected to explain how "supportedDatabases" would work. When running in production mode, LabKey will check that the primary database is one of the supported databases that the distribution specifies. The default (nothing specified, the case for most distributions) is PostgreSQL only. The 10 or so distributions that need to run against SQL Server in production mode must specify this new property, otherwise LabKey will refuse to start. This is how we limit the distributions that run on SQL Server, prevent new ones from getting added, and start paring down the modules that support SQL Server. So the question really is: do the distributions produced by this build file ever need to run on SQL Server in production mode?

labkey-tchad:
> ... So the question really is: do the distributions produced by this build file ever need to run on SQL Server in production mode?

A subset of moduleEditor functionality requires production mode. There are related tests running with SQL Server on TeamCity.
There is an argument to be made that that is unnecessary test coverage (no moduleEditor schema) but I'd prefer to have the test distributions support "pgsql, mssql" just to avoid any confusion down the road.

labkey-adam:
> > ... So the question really is: do the distributions produced by this build file ever need to run on SQL Server in production mode?
> 
> A subset of moduleEditor functionality requires production mode. There are related tests running with SQL Server on TeamCity. There is an argument to be made that that is unnecessary test coverage (no moduleEditor schema) but I'd prefer to have the test distributions support "pgsql, mssql" just to avoid any confusion down the road.

It is so

PR #2087, opened 2024-10-10
Ignore errors caused by cancelling a pipeline job

labkey-tchad:
#### Rationale
Recent changes made cancelling pipeline jobs more aggressive, which can cause some server-side errors.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5892

#### Changes
- Ignore errors caused by cancelling a pipeline job

### DIFF
--- a/src/org/labkey/test/ClientErrorAssertionError.java
+++ b/src/org/labkey/test/ClientErrorAssertionError.java
@@ -1,8 +1,12 @@
 package org.labkey.test;
 
-public class ClientErrorAssertionError extends AssertionError
+/**
+ * Indicates that a client-side error was found in the server's error log
+ * Requires the 'javascriptErrorServerLogging' optional feature to be enabled on the server
+ */
+public class ClientErrorAssertionError extends ServerErrorAssertionError
 {
-    public ClientErrorAssertionError(Object detailMessage)
+    public ClientErrorAssertionError(String detailMessage)
     {
         super(detailMessage);
     }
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/ClientErrorAssertionError.java
```diff
@@ -0,0 +1,9 @@
+package org.labkey.test;
+
+public class ClientErrorAssertionError extends AssertionError
```
labkey-jeckels: Having `Error` in the class name makes sense but reads strangely. A line of JavaDoc on intended use here and in the server-side equivalent would be nice.

src/org/labkey/test/ClientErrorAssertionError.java
```diff
@@ -0,0 +1,9 @@
...
+{
+    public ClientErrorAssertionError(Object detailMessage)
```
labkey-jeckels: AssertionError is very atypical of Throwable subclasses in taking an Object for its constructor. It really wants a String or a Throwable. I'd suggest using String as the argument type here.

src/org/labkey/test/tests/PipelineCancelTest.java
```diff
@@ -52,10 +53,23 @@ public void testSteps()
...
         goToProjectHome();
         assertTextPresent("This folder does not contain a study."); //part of the import will be done, but it shouldn't have gotten to participants.
+
+        try
+        {
+            checkErrors();
+        }
+        catch (ServerErrorAssertionError error)
+        {
+            log("Ignoring server errors cause by cancelling pipeline job");
```
labkey-jeckels:
```suggestion
            log("Ignoring server errors caused by cancelling pipeline job");
```

labkey-jeckels (approved):

PR #2088, opened 2024-10-10
Update API key tests to use ApiKeyPanel and ApiKeyDialog

labkey-tchad:
#### Rationale
Tests should use shared components when able. `ApiKeyPanel` should live in `testAutomation` since it isn't specific to sample manager.

#### Related Pull Requests
* https://github.com/LabKey/nlp/pull/251
* https://github.com/LabKey/limsModules/pull/816

#### Changes
* Update API key tests to use ApiKeyPanel and ApiKeyDialog
* Relocate ApiKeyPanel and ApiKeyDialog to `org.labkey.test.components.core`

### DIFF
--- a/src/org/labkey/test/components/core/ApiKeyDialog.java
+++ b/src/org/labkey/test/components/core/ApiKeyDialog.java
@@ -1,10 +1,11 @@
-package org.labkey.test.credentials;
+package org.labkey.test.components.core;
 
 import org.labkey.test.Locator;
 import org.labkey.test.components.bootstrap.ModalDialog;
 import org.labkey.test.components.html.Input;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.ui.ExpectedConditions;
 
 import java.awt.*;
 import java.awt.datatransfer.DataFlavor;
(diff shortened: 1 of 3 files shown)


src/org/labkey/test/components/core/ApiKeyDialog.java
```diff
@@ -27,7 +28,10 @@ public ApiKeyDialog(WebDriver driver, String title)
     public ApiKeyDialog generateApiKey()
     {
         elementCache().generateApiKeyButton.click();
-        return new ApiKeyDialog(getDriver(), _title);
+        getWrapper().shortWait().until(ExpectedConditions.invisibilityOf(elementCache().descriptionInput.getComponentElement()));
+        clearElementCache();
+        getWrapper().shortWait().until(ExpectedConditions.visibilityOf(elementCache().inputField.getComponentElement()));
+        return this;
```
labkey-susanh, line 34: I would have thought using `this` might produce a stale element exception. If not, then so much the better.

labkey-susanh (approved):

src/org/labkey/test/components/core/ApiKeyDialog.java
labkey-tchad: The body of the dialog does go stale (hence the `clearElementCache`) but the outer div of the dialog is stable.

PR #2089, opened 2024-10-10
Test coverage for Issue 39496

labkey-chrisj:
#### Rationale
This adds regression coverage for Issue 39496, which verifies that missing values written to an assay via the `SaveBatch `API are handled as expected.

On a related note, this test sets up a simple way to verify whether or not missingValue indicators should be exported or not; [Issue 37610](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=37610) tracks this question and is a very stale issue; current behavior is to export empty values (not showing missing-value indicators as they appear in the dataregion) and in the years this issue has been around, nobody has asked for a fix.  (so, this might be an opportune time to decide whether or not to expect missingValue indicators to be exported)

It also attempts to address the intermittent test failure in [BiologicsSampleTimelineTest.testEditSampleDetailCausesTimelineEvent](https://teamcity.labkey.org/viewLog.html?buildId=3226197&buildTypeId=LabKey_2410Release_Premium_ProductSuites_Biologics_LimsStarter_LimsStarterB&fromSakuraUI=true#testNameId-6192646934553344580), which seems to be caused by a race condition in which the test looks for a select in `DetailTableEdit `before it is rendered, without waiting for it to appear.

#### Related Pull Requests
n/a

#### Changes

- [ ] add a new test case
- [ ] try to fix intermittent test failure

### DIFF
--- a/src/org/labkey/test/tests/assay/AssayMissingValuesTest.java
+++ b/src/org/labkey/test/tests/assay/AssayMissingValuesTest.java
@@ -10,7 +10,6 @@ import org.labkey.test.Locator;
 import org.labkey.test.TestFileUtils;
 import org.labkey.test.categories.Daily;
 import org.labkey.test.pages.admin.ExportFolderPage;
-import org.labkey.test.pages.samplemanagement.assay.AssayResultsPage;
 import org.labkey.test.params.FieldDefinition;
 import org.labkey.test.params.assay.GeneralAssayDesign;
 import org.labkey.test.tests.MissingValueIndicatorsTest;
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/tests/assay/AssayMissingValuesTest.java
```diff
@@ -217,6 +222,54 @@ public void testAssayMV()
...
+        // navigate to the results view
+        goToProjectHome();
+        clickAndWait(Locator.linkWithText(assayName));
+        clickAndWait(Locator.linkWithText("view results"));
+        var dataRegion = DataRegionTable.DataRegion(getDriver()).waitFor();
+
+        // expect 3 rows in this assay, p2 and p3 should get mv indicators in the count column
+        Map<String, List<Integer>> expectedMVIndicators = new HashMap<>();
+        expectedMVIndicators.put("Count", List.of(1, 2));
+        checkMvIndicatorPresent(dataRegion, expectedMVIndicators);
```
cnathe, line 275: this looks good and matches the approach we discussed!
looks like the other test cases in this file also have a call to `checkDataregionData()` that goes with the `checkMvIndicatorPresent()` call.

labkey-chrisj: I've added a check for dataRegion data to the test, thanks for the suggestion

labkey-tchad (approved):

PR #2091, opened 2024-10-11
Add a long report ETL for testing pipeline job cancellation

labkey-tchad:
#### Rationale
In order to verify that cancelling a report job terminates the process, we need a long-running ETL.

#### Related Pull Requests
- https://github.com/LabKey/dataintegration/pull/277

#### Changes
- Add a long report ETL for testing pipeline job cancellation
- Rename and clean up `ETLScheduler`


src/org/labkey/test/pages/dataintegration/ETLSchedulerPage.java
```diff
@@ -127,7 +128,7 @@ public TransformRow setEnabled(boolean enable)
 //
 //        public TransformRow setVerboseLoggingEnabled(boolean enable)
```
labkey-jeckels, line 129: Delete this commented out code?

src/org/labkey/test/pages/dataintegration/ETLSchedulerPage.java
```diff
@@ -22,6 +22,7 @@
 import org.labkey.test.components.Component;
```
labkey-jeckels, line 22: Probably we should wait for when we consolidate into `premiumModules`, but it seems like it would be nice to have this test code and the `ETLtest` module live into the same repo as `dataintegration`

labkey-jeckels (approved):

PR #2094, opened 2024-10-15
Revert changes to use special characters in test

XingY:
#### Rationale
[SampleTypeLimitsTest](https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyPostgres_2/3232211?buildTab=tests&suite=org.labkey.test.Runner%3A+&package=org.labkey.test.tests&class=SampleTypeLimitsTest).[testInsertLargeLineageGraph](https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyPostgres_2/3232211?buildTab=tests)

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/src/org/labkey/test/tests/SampleTypeLimitsTest.java
+++ b/src/org/labkey/test/tests/SampleTypeLimitsTest.java
@@ -32,6 +32,7 @@ import java.util.Map;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.fail;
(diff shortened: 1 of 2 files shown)


labkey-tchad (approved):
The seems like it might be a product bug. Shouldn't we throw some kind of error if these imports fail?

XingY:
> The seems like it might be a product bug. Shouldn't we throw some kind of error if these imports fail?

The product bug is tracked in Issue 51432

PR #2098, opened 2024-10-16
Calculated fields error handling

cnathe:
#### Rationale
See related PR for rationale

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5938

#### Changes
- Selenium test helpers for calculated fields: clickValidateExpression(), hasGridError(), getGridError()
- Selenium test for calc field and parent alias check


labkey-susanh (approved):

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
```diff
@@ -644,6 +644,19 @@ public List<String> getGridMessages()
...
+    {
+        return getGridError() != null;
+    }
+
+    public String getGridError()
+    {
+        Locator.XPathLocator errorLocator = Locator.tagWithClass("div", "alert-danger");
+        if (errorLocator.existsIn(this))
+            return errorLocator.findElement(this).getText();
+        return null;
```
labkey-tchad: Could do something like this to avoid finding the element multiple times.
```suggestion
        Locator.XPathLocator errorLocator = Locator.tagWithClass("div", "alert-danger");
        return errorLocator.findOptionalElement(this).map(WebElement::getText).orElse(null);
```

labkey-tchad (approved):

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
cnathe: done

PR #2109, opened 2024-10-25
Issue 51521: Add switch to disable Server HTTP header

labkey-jeckels:
#### Rationale
Some organizations don't like having the product and version identified in HTTP headers. We can make it optional.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5989

#### Changes
- Test the presence and absence of the header when toggling the setting

### DIFF
--- a/src/org/labkey/test/tests/AdminConsoleTest.java
+++ b/src/org/labkey/test/tests/AdminConsoleTest.java
@@ -18,6 +18,10 @@ package org.labkey.test.tests;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
+import org.labkey.remoteapi.CommandException;
+import org.labkey.remoteapi.CommandResponse;
+import org.labkey.remoteapi.Connection;
+import org.labkey.remoteapi.SimpleGetCommand;
 import org.labkey.test.BaseWebDriverTest;
 import org.labkey.test.Locator;
 import org.labkey.test.TestTimeoutException;
@@ -26,12 +30,16 @@ import org.labkey.test.categories.Daily;
 import org.labkey.test.pages.core.login.LoginConfigRow;
 import org.labkey.test.pages.core.login.LoginConfigurePage;
 import org.labkey.test.util.ApiPermissionsHelper;
+import org.labkey.test.util.LogMethod;
 import org.labkey.test.util.PermissionsHelper;
+import org.labkey.test.util.TestLogger;
 import org.openqa.selenium.WebElement;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/tests/AdminConsoleTest.java
```diff
@@ -51,6 +59,74 @@ public String getProjectName()
...
+            GetServerHeaderCommand command = new GetServerHeaderCommand();
+            command.execute(cn, "/home");
+            return command._server;
+        }
+        catch (IOException | CommandException e)
+        {
+            throw new RuntimeException("Failed to get Server HTTP response header", e);
+        }
+    }
+        @Test
```
labkey-tchad: Spacing looks messed up
```suggestion
    }
    
    @Test
```

src/org/labkey/test/tests/AdminConsoleTest.java
```diff
@@ -51,6 +59,74 @@ public String getProjectName()
     }
 
     @Test
+    public void testServerHttpHeaderSetting()
+    {
+        goToAdminConsole().clickSiteSettings();
+        waitForElement(Locator.name("includeServerHttpHeader"));
+        WebElement checkbox = Locator.checkboxByName("includeServerHttpHeader").findElement(getDriver());
+
+        boolean originalValue = "true".equals(checkbox.getAttribute("checked"));
```
labkey-tchad: "checked" is a valueless attribute. It's mere existence indicates that the checkbox is checked, even with `checked="false"`. This code implies that the `includeServerHttpHeader` checkbox is not written correctly.

That being said, Selenium understands standard HTML checkboxes.
```suggestion
        boolean originalValue = checkbox.isSelected());
```

labkey-tchad (changes requested):

src/org/labkey/test/tests/AdminConsoleTest.java
labkey-jeckels: > "checked" is a valueless attribute. It's mere existence indicates that the checkbox is checked, even with `checked="false"`. This code implies that the `includeServerHttpHeader` checkbox is not written correctly.
> 
> That being said, Selenium understands standard HTML checkboxes.

Thanks Trey. I also fixed up the other code that I patterned my change after.

labkey-tchad (changes requested):
Sorry I didn't catch this the first time.

src/org/labkey/test/tests/AdminConsoleTest.java
```diff
@@ -50,6 +55,75 @@ public String getProjectName()
...
+        goToAdminConsole().clickSiteSettings();
+        waitForElement(Locator.name("includeServerHttpHeader"));
+        WebElement checkbox = Locator.checkboxByName("includeServerHttpHeader").findElement(getDriver());
+
+        boolean originalValue = checkbox.isSelected();
+
+        // Try with the setting on
+        if (!originalValue)
+            click(Locator.checkboxByName("includeServerHttpHeader"));
+        clickButton("Save");
```
labkey-tchad: This checkbox should be manipulated via `CustomizeSitePage`.
```suggestion
        CustomizeSitePage customizeSitePage = goToAdminConsole().clickSiteSettings();
        boolean originalValue = customizeSitePage.isServerHttpHeaderEnabled();

        // Try with the setting on
        if (!originalValue)
            customizeSitePage.enableServerHttpHeader(true);
        customizeSitePage.save();
```

labkey-tchad (changes requested):

src/org/labkey/test/tests/AdminConsoleTest.java
labkey-jeckels: OK. Made another update, including another improvement to the section I used as a reference.

labkey-tchad (approved):

PR #2110, opened 2024-10-25
Roll back test changes to FolderTest.deleteWorkbooksTest

labkey-danield:
#### Rationale
Roll back test changes to FolderTest.deleteWorkbooksTest

#### Related Pull Requests
- 

#### Changes
- Roll back test changes to FolderTest.deleteWorkbooksTest

### DIFF
--- a/src/org/labkey/test/tests/FolderTest.java
+++ b/src/org/labkey/test/tests/FolderTest.java
@@ -372,14 +372,14 @@ public class FolderTest extends BaseWebDriverTest
             SelectRowsCommand sr = new SelectRowsCommand("core", "workbooks");
             sr.setColumns(Arrays.asList("EntityId"));
             SelectRowsResponse srr = sr.execute(cn, getProjectName() + "/WorkbookParent");
-            Object rowId = srr.getRows().get(0).get("RowId");
+            Object entityId = srr.getRows().get(0).get("EntityId");
 
             DeleteRowsCommand drc = new DeleteRowsCommand("core", "workbooks");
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/FolderTest.java
```diff
@@ -372,14 +372,14 @@ public void deleteWorkbooksTest() throws Exception
...
+            Object entityId = srr.getRows().get(0).get("EntityId");
 
             DeleteRowsCommand drc = new DeleteRowsCommand("core", "workbooks");
             Map<String, Object> row1 = new HashMap<>();
-            row1.put("RowId", rowId);
+            row1.put("EntityId", entityId);
             drc.addRow(row1);
 
-            SaveRowsResponse resp = drc.execute(cn, getContainerId());
+            SaveRowsResponse resp = drc.execute(new Connection(WebTestHelper.getBaseURL(), PasswordUtil.getUsername(), PasswordUtil.getPassword()), getContainerId());
```
labkey-adam:
```suggestion
            SaveRowsResponse resp = drc.execute(cn, getContainerId());
```
Not required, but I would keep this change. Might as well re-use the connection from above, regardless of key.

labkey-adam (approved):
Thanks for taking care of this!

src/org/labkey/test/tests/FolderTest.java
labkey-danield: Done (reusing existing connection).

PR #2112, opened 2024-10-26
Adjust test for session invalidation change

labkey-adam:
#### Rationale
Server now invalidates a user's existing sessions on password change, so expect to NOT be impersonating afterwards

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5994


labkey-adam:
Not needed

PR #2114, opened 2024-10-28
await select in edit mode

labkey-chrisj:
#### Rationale
Recently, [BiologicsSampleTimelineTest.testEditSampleDetailCausesTimelineEvent](https://teamcity.labkey.org/viewLog.html?buildId=3247197&buildTypeId=LabKey_247Release_Premium_ProductSuites_Biologics_LimsStarter_LimsStarterB&fromSakuraUI=true#testNameId-6192646934553344580) has failed in 24.7 because the 'status' select was not found (but was present in the after-failure screenshot) when the test attempted to edit a sample's status in order to verify that doing so would cause a timeline event.

This change backports the fix for this intermittent failure from develop; the test will now await the select as it does in  develop

#### Related Pull Requests
n/a

#### Changes
one-line backport from develop


labkey-danield (approved):

labkey-chrisj:
Unfortunately, this PR's remote branch didn't match TeamCity's naming pattern and thus didn't get a build check (which is required). So, deleting this branch, pushing the same change to a branch that will get a build check

PR #2119, opened 2024-11-04
Set window size for new browser

labkey-tchad:
#### Rationale
Leaving the default browser size resulted in too many test failures. We can get most of the perf improvement by only setting the window size when we initially launch the browser.

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/2118

#### Changes
- Set window size when launching new browser instance


labkey-tchad:
Verifying with a couple of the most affected suites:
https://teamcity.labkey.org/buildConfiguration/LabKey_247Release_Premium_ProductSuites_Biologics_BiologicsCPostgres/3254616
https://teamcity.labkey.org/buildConfiguration/LabKey_247Release_External_Panorama_PanoramaASqlserver/3254618

labkey-chrisj (approved):

PR #2161, opened 2024-11-29
Use findElementOrNull for look and feel settings that are now deprecated

cnathe:
#### Rationale
See related PR for rationale. This PR just fixes the misc test failures that are now erroring out because the deprecated look and feel setting is no longer present on the page.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6097

#### Changes
- Use findElementOrNull for look and feel settings that are now deprecated


labkey-adam:
@cnathe thanks... I created https://github.com/LabKey/testAutomation/pull/2160 yesterday for this. Maybe we go with that one since it fixes the test as well?

cnathe:
> @cnathe thanks... I created #2160 yesterday for this. Maybe we go with that one since it fixes the test as well?

Perfect, @labkey-adam. Yes let's go with that one. I'll close this PR.

PR #2165, opened 2024-12-02
Automation test for Editor without Delete users are able to delete files

labkey-sweta:
#### Rationale
Automation test for Editor without Delete users are able to delete files

Test case:
Create a folder.
Create a user.
Give the user Editor without Delete.
As that new user, upload a file.
As that new user, verify you can't see the Delete option in the file browser UI.
As that new user, verify that a DELETE HTTP request fails with a permission error and the file persists.

[Issue link](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=51500)

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/src/org/labkey/test/tests/UserEditorWithoutDeleteTest.java
+++ b/src/org/labkey/test/tests/UserEditorWithoutDeleteTest.java
@@ -1,0 +1,86 @@
+package org.labkey.test.tests;
+
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.labkey.test.BaseWebDriverTest;
+import org.labkey.test.TestFileUtils;
+import org.labkey.test.categories.Daily;
+import org.labkey.test.components.FilesWebPart;
+import org.labkey.test.util.FileBrowserHelper;
+import org.labkey.test.util.PortalHelper;
+import org.labkey.test.util.TestUser;
+import org.labkey.test.util.core.webdav.WebDavUploadHelper;
+import org.labkey.test.util.core.webdav.WebDavUrlFactory;
+import org.labkey.test.util.core.webdav.WebDavUtils;
+
+import java.io.File;
+import java.util.Arrays;
+import java.util.List;
+
+@Category({Daily.class})
+public class UserEditorWithoutDeleteTest extends BaseWebDriverTest
+{
+    private static final TestUser EDITOR_WITHOUT_DELETE = new TestUser("editorwithoutdelete@user.test");
+    private static final File TEST_FILE = TestFileUtils.getSampleData("studies/Read_Me.txt");
+
+    @BeforeClass
+    public static void setupProject()
+    {
+        UserEditorWithoutDeleteTest init = (UserEditorWithoutDeleteTest) getCurrentTest();
+        init.doSetup();
+    }
+
+    @Override
+    protected void doCleanup(boolean afterTest)
+    {
+        _containerHelper.deleteProject(getProjectName(), afterTest);
+        _userHelper.deleteUsers(afterTest, EDITOR_WITHOUT_DELETE);
+    }
+
+    private void doSetup()
+    {
+        _containerHelper.createProject(getProjectName(), null);
+        EDITOR_WITHOUT_DELETE.create(this)
+                .setInitialPassword()
+                .addPermission("EditorWithoutDelete", getProjectName());
+
+        goToProjectHome();
+        new PortalHelper(this).addWebPart("Files");
+    }
+
+    /*
+        Regression coverage for Secure Issue 51500: Editor without Delete users are able to delete files
+     */
+    @Test
+    public void testFileDeleteWithEditorWithoutDeleteUser()
+    {
+        goToProjectHome();
+        impersonate(EDITOR_WITHOUT_DELETE.getEmail());
+        FilesWebPart filesWebPart = FilesWebPart.getWebPart(getDriver());
+        filesWebPart.fileBrowser().uploadFile(TEST_FILE);
+        Assert.assertFalse("Delete action should not be present for " + EDITOR_WITHOUT_DELETE.getEmail(),
+                filesWebPart.fileBrowser().isActionPresent(FileBrowserHelper.BrowserAction.DELETE));
+
+        WebDavUploadHelper uploadHelper = new WebDavUploadHelper(WebDavUrlFactory.webDavUrlFactory(getProjectName()),
+                WebDavUtils.beginSardine(getCurrentUser()));
+        uploadHelper.uploadFile(TEST_FILE);
+        uploadHelper.deleteFile(WebDavUtils.buildBaseWebDavUrl(getProjectName()) + TEST_FILE.getName(), true);
+        Assert.assertTrue("File should not be deleted by " + EDITOR_WITHOUT_DELETE.getEmail(),
+                uploadHelper.fileExists(WebDavUtils.buildBaseWebDavUrl(getProjectName()) + TEST_FILE.getName()));
+        stopImpersonating();
+    }
+
+    @Override
+    protected String getProjectName()
+    {
+        return "UserEditorWithoutDeleteTest Project";
+    }
+
+    @Override
+    public List<String> getAssociatedModules()
+    {
+        return Arrays.asList();
+    }
+}
--- a/src/org/labkey/test/tests/UserEditorWithoutDeleteTest.java
+++ b/src/org/labkey/test/tests/UserEditorWithoutDeleteTest.java
@@ -27,7 +27,7 @@ public class UserEditorWithoutDeleteTest extends BaseWebDriverTest
     @BeforeClass
     public static void setupProject()
     {
-        UserEditorWithoutDeleteTest init = (UserEditorWithoutDeleteTest) getCurrentTest();
+        UserEditorWithoutDeleteTest init = getCurrentTest();
         init.doSetup();
     }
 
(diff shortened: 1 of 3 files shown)


src/org/labkey/test/util/core/webdav/WebDavUploadHelper.java
```diff
@@ -133,6 +133,37 @@ public void uploadFile(@NotNull File file)
         uploadFile(file, "");
     }
 
+    public void deleteFile(@NotNull String file, boolean shouldFail)
+    {
+        try
+        {
+            if (_sardine.exists(file))
```
labkey-jeckels: I probably wouldn't check if it exists here, as the caller may be wanting to test what the server does when you try deleting a file that doesn't exist.

src/org/labkey/test/tests/UserEditorWithoutDeleteTest.java
```diff
@@ -0,0 +1,86 @@
...
+    @Override
+    protected String getProjectName()
+    {
+        return "UserEditorWithoutDeleteTest Project";
+    }
+
+    @Override
+    public List<String> getAssociatedModules()
+    {
+        return Arrays.asList();
```
labkey-jeckels:
```suggestion
        return Arrays.asList("FileContent");
```

src/org/labkey/test/tests/UserEditorWithoutDeleteTest.java
```diff
@@ -0,0 +1,86 @@
...
+        uploadHelper.deleteFile(WebDavUtils.buildBaseWebDavUrl(getProjectName()) + TEST_FILE.getName(), true);
+        Assert.assertTrue("File should not be deleted by " + EDITOR_WITHOUT_DELETE.getEmail(),
+                uploadHelper.fileExists(WebDavUtils.buildBaseWebDavUrl(getProjectName()) + TEST_FILE.getName()));
+        stopImpersonating();
+    }
+
+    @Override
+    protected String getProjectName()
+    {
+        return "UserEditorWithoutDeleteTest Project";
```
labkey-jeckels: Is there a reason to avoid using the tricky characters here or with the file name itself?

labkey-jeckels (commented):
I can confirm that this covers the exact use case for the bug. A few comments or suggestions. I'll defer to Trey for the full approval.

src/org/labkey/test/tests/UserEditorWithoutDeleteTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+    private void doSetup()
+    {
+        _containerHelper.createProject(getProjectName(), null);
+        EDITOR_WITHOUT_DELETE.create(this)
+                .setInitialPassword()
+                .addPermission("EditorWithoutDelete", getProjectName());
+
+        goToProjectHome();
+        new PortalHelper(this).addWebPart("Files");
```
labkey-tchad: No need to add a Files webpart, it just adds extra navigations by entering and exiting page admin mode. The test can just use `goToModule("FileContent")` to see the file browser.

src/org/labkey/test/util/core/webdav/WebDavUploadHelper.java
```diff
@@ -133,6 +133,36 @@ public void uploadFile(@NotNull File file)
         uploadFile(file, "");
     }
 
+    public void deleteFile(@NotNull String file, boolean shouldFail)
```
labkey-tchad: Please avoid this pattern of having a flag to specify whether failure is expected. Something like `deleteExpectingError` is preferred.

src/org/labkey/test/util/core/webdav/WebDavUploadHelper.java
```diff
@@ -133,6 +133,36 @@ public void uploadFile(@NotNull File file)
...
+    {
+        try
+        {
+            _sardine.delete(file);
```
labkey-tchad: This should use the UrlFactory so that callers don't have to.
```suggestion
            _sardine.delete(_urlFactory.getPath(file));
```

src/org/labkey/test/util/core/webdav/WebDavUploadHelper.java
```diff
@@ -133,6 +133,36 @@ public void uploadFile(@NotNull File file)
...
+    }
+
+    public boolean fileExists(String file)
+    {
+        try
+        {
+            if (_sardine.exists(file))
+                return true;
+            else
+                return false;
```
labkey-tchad:
```suggestion
            return _sardine.exists(_urlFactory.getPath(file)));
```

src/org/labkey/test/tests/UserEditorWithoutDeleteTest.java
```diff
@@ -0,0 +1,86 @@
...
+     */
+    @Test
+    public void testFileDeleteWithEditorWithoutDeleteUser()
+    {
+        goToProjectHome();
+        impersonate(EDITOR_WITHOUT_DELETE.getEmail());
+        FilesWebPart filesWebPart = FilesWebPart.getWebPart(getDriver());
+        filesWebPart.fileBrowser().uploadFile(TEST_FILE);
+        Assert.assertFalse("Delete action should not be present for " + EDITOR_WITHOUT_DELETE.getEmail(),
+                filesWebPart.fileBrowser().isActionPresent(FileBrowserHelper.BrowserAction.DELETE));
```
labkey-tchad: This check should also be done as somebody who **can** see the delete button to ensure that the check is valid. At the end, after attempting deletion would be fine.

src/org/labkey/test/util/core/webdav/WebDavUploadHelper.java
```diff
@@ -133,6 +133,36 @@ public void uploadFile(@NotNull File file)
...
+        {
+            _sardine.delete(file);
+        }
+        catch (IOException e)
+        {
+            if (shouldFail)
+              ;
+            else
+                throw new RuntimeException(e);
+        }
```
labkey-tchad: This is a bit too wide of a net for an expected error. An IOException could be all sorts of things. I would at least make sure the correct exception type was thrown; then return it so the test could verify the results.
e.g.
```
public SardineException deleteExpectingError(String file)
{
    try
    {
        _sardine.delete(_urlFactory.getPath(file));
        throw new RuntimeException("Expected error did not occur");
    }
    catch (SardineException expected)
    {
        return expected;
    }
    catch (IOException e)
    {
        throw new RuntimeException(e);
    }
}
```
`SardineException` will include a status code which would be good for verifying in this test.

src/org/labkey/test/tests/UserEditorWithoutDeleteTest.java
```diff
@@ -0,0 +1,86 @@
...
+@Category({Daily.class})
+public class UserEditorWithoutDeleteTest extends BaseWebDriverTest
+{
+    private static final TestUser EDITOR_WITHOUT_DELETE = new TestUser("editorwithoutdelete@user.test");
+    private static final File TEST_FILE = TestFileUtils.getSampleData("studies/Read_Me.txt");
+
+    @BeforeClass
+    public static void setupProject()
+    {
+        UserEditorWithoutDeleteTest init = (UserEditorWithoutDeleteTest) getCurrentTest();
```
labkey-tchad: This cast should no longer be necessary.
```suggestion
        UserEditorWithoutDeleteTest init = getCurrentTest();
```
Do you have warnings disabled in your IntelliJ? It should flag this.

labkey-tchad (changes requested):

src/org/labkey/test/tests/UserEditorWithoutDeleteTest.java
labkey-sweta: I never got any warning, so I think the warning is disabled. 

But strangely enough if i delete it, i get the error.

And, this statement was added as part of template when we add new test class.

labkey-tchad: Your branch is a bit behind develop. It's missing the change that makes the cast unnecessary. It is more recent that I remembered (https://github.com/LabKey/testAutomation/pull/2156)

src/org/labkey/test/tests/UserEditorWithoutDeleteTest.java
```diff
@@ -57,19 +53,27 @@ private void doSetup()
...
         Assert.assertFalse("Delete action should not be present for " + EDITOR_WITHOUT_DELETE.getEmail(),
-                filesWebPart.fileBrowser().isActionPresent(FileBrowserHelper.BrowserAction.DELETE));
+                _fileBrowserHelper.isActionPresent(FileBrowserHelper.BrowserAction.DELETE));
 
         WebDavUploadHelper uploadHelper = new WebDavUploadHelper(WebDavUrlFactory.webDavUrlFactory(getProjectName()),
                 WebDavUtils.beginSardine(getCurrentUser()));
         uploadHelper.uploadFile(TEST_FILE);
-        uploadHelper.deleteFile(WebDavUtils.buildBaseWebDavUrl(getProjectName()) + TEST_FILE.getName(), true);
+        SardineException sardineException = uploadHelper.deleteExpectingError(WebDavUtils.buildBaseWebDavUrl(getProjectName()) + TEST_FILE.getName());
+        Assert.assertEquals("Incorrect response for deletion", 404, sardineException.getStatusCode());
```
labkey-tchad: The correct status code is 403. You're getting a 404 because you're passing a full URL to `deleteExpectingError` instead of just the file name.

labkey-tchad (changes requested):

labkey-tchad (approved):

PR #2169, opened 2024-12-04
Encode URL query parameters when building a URL

labkey-tchad:
#### Rationale
I updated a test to use `WebTestHelper.buildURL` to create a URL and it started failing because the URL query isn't properly encoded. The page just says "Invalid characters in request." and the test fails with:
```
java.lang.AssertionError: Incorrect response code expected:<404> but was:<200>
  at org.junit.Assert.failNotEquals(Assert.java:835)
  at org.junit.Assert.assertEquals(Assert.java:647)
  at org.labkey.test.tests.list.ListTest.testCustomViews(ListTest.java:691)
```

#### Related Pull Requests
- #2154 

#### Changes
- Encode URL query parameters when building a URL


labkey-chrisj (approved):
I hope this works!

PR #2173, opened 2024-12-05
Fix error when splitting DateTime format string

labkey-danield:
#### Rationale
Fixing a bad string split.

#### Related Pull Requests
- none

#### Changes
- Be smarter about splitting a DateTime format string into a date part and a time part.
- Remove unused 'throw' annotation.

### DIFF
--- a/src/org/labkey/test/components/domain/DomainFormPanel.java
+++ b/src/org/labkey/test/components/domain/DomainFormPanel.java
@@ -139,21 +139,28 @@ public class DomainFormPanel extends DomainPanel<DomainFormPanel.ElementCache, D
             else if (fieldDefinition.getType().equals(FieldDefinition.ColumnType.DateAndTime))
             {
 
-                // Identify the part of the format that is the date and the part that is the time. Take into account
-                // a data format may include spaces (time does not).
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/components/domain/DomainFormPanel.java
```diff
@@ -139,21 +139,28 @@ else if (fieldDefinition.getType().equals(FieldDefinition.ColumnType.Time))
             else if (fieldDefinition.getType().equals(FieldDefinition.ColumnType.DateAndTime))
             {
 
-                // Identify the part of the format that is the date and the part that is the time. Take into account
-                // a data format may include spaces (time does not).
-                String format = fieldDefinition.getFormat().trim();
-                int index = format.lastIndexOf(" ");
+                // Identify the part of the format that is the date and the part that is the time.
```
labkey-chrisj, line 142: Ah, that explains it.  
This is probably out of scope for this fix, but this code (parsing date format/time format out of a format string) seems like it ought to be in a utils class like `TestDateUtils.java`

labkey-chrisj (approved):

PR #2180, opened 2024-12-09
Issue 51822: Assay import API to support result field mapping via field Name, Label, and Import Aliases

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51822

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6135
- https://github.com/LabKey/testAutomation/pull/2180

#### Changes
- add API test via Rlabkey to test assay import using result field import alias

### DIFF
--- a/data/api/rlabkey-api-experiment.xml
+++ b/data/api/rlabkey-api-experiment.xml
@@ -185,6 +185,26 @@
             ]]>
         </response>
     </test>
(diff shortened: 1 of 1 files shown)


labkey-klum (approved):
The `guspec` import alias came from SCHARP originally, I wonder if we can retire it at some point 😄

PR #2186, opened 2024-12-10
Change locators for additional format elements

labkey-danield:
#### Rationale
Additional parsing formats are only available is the deprecated / optional feature is enabled. Locators for those elements needs to take that into account.

#### Related Pull Requests
- None

#### Changes
- Change locators for alternate formats to findWhenNeeded.


labkey-tchad (changes requested):
This class should extend `BaseSettingsPage`. That class received a [similar fix](https://github.com/LabKey/testAutomation/pull/2156) last week.
`admin-lookAndFeelSettings.view`, `admin-projectSettings.view`, and `admin-folderSettings.view` all come from `lookAndFeelProperties.jsp`. The first two already have page classes that extend `BaseSettingsPage`. (`SiteValidationPage` also extends it but I think that was done in error)

labkey-tchad (approved):

PR #2192, opened 2024-12-12
FilterSelectionDialog: handle non-removable fields

labkey-nicka:
#### Rationale
Update `FieldSelectionDialog` test component to handle removing all selected fields edge case where some fields are not actually removable.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/971

#### Changes
- Skip all fields removed check if not all fields can be removed.

### DIFF
--- a/src/org/labkey/test/components/ui/grids/FieldSelectionDialog.java
+++ b/src/org/labkey/test/components/ui/grids/FieldSelectionDialog.java
@@ -386,20 +386,25 @@ public class FieldSelectionDialog extends ModalDialog
     public FieldSelectionDialog removeAllSelectedFields()
     {
         List<WebElement> allItems = elementCache().getListItemElements(elementCache().selectedFieldsPanel);
+        boolean removedAll = true;
 
-        for(WebElement listItem : allItems)
+        for (WebElement listItem : allItems)
         {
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/components/ui/grids/FieldSelectionDialog.java
```diff
@@ -386,20 +386,25 @@ public FieldSelectionDialog removeFieldFromSelected(String field, int index)
     public FieldSelectionDialog removeAllSelectedFields()
     {
         List<WebElement> allItems = elementCache().getListItemElements(elementCache().selectedFieldsPanel);
+        boolean removedAll = true;
 
-        for(WebElement listItem : allItems)
+        for (WebElement listItem : allItems)
```
labkey-chrisj, line 391: Occasionally this method will fail because not every item was removed-- one theory I have is that it might be due to how they slide up when the previous one is removed.  It might be more reliable to iterate from the bottom up, if it's possible for the click to 'miss' if it happens when the target is moving

labkey-chrisj (approved):

src/org/labkey/test/components/ui/grids/FieldSelectionDialog.java
labkey-nicka: I'm going to leave as-is since I'm not looking to triage intermittent failures with this work. What you are proposing seems reasonable.

PR #2207, opened 2024-12-20
Fix test issues with file upload

labkey-alan:
#### Rationale
There are some failing tests on team city

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/2207
* https://github.com/LabKey/provenance/pull/194
* https://github.com/LabKey/limsModules/pull/1018

#### Changes
- Fix selectors
- Add backwards compat mode for FileUploadPanel
  - Needed because the provenance module is using a rather old version of ui-components

### DIFF
--- a/src/org/labkey/test/components/ui/files/FileUploadPanel.java
+++ b/src/org/labkey/test/components/ui/files/FileUploadPanel.java
@@ -18,11 +18,13 @@ public class FileUploadPanel extends WebDriverComponent<FileUploadPanel.ElementC
 {
     final WebElement _el;
     final WebDriver _driver;
+    final boolean _backwardsCompat;
 
-    public FileUploadPanel(WebElement element, WebDriver driver)
+    public FileUploadPanel(WebElement element, WebDriver driver, boolean backwardsCompat)
     {
         _el = element;
         _driver = driver;
+        _backwardsCompat = backwardsCompat;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


src/org/labkey/test/components/ui/files/FileUploadPanel.java
```diff
@@ -120,18 +125,28 @@ WebElement removeBtn(String fileName)
...
         {
             super(driver);
+            _backwardsCompat = false;
+        }
+
+        public FileUploadPanelFinder(WebDriver driver, boolean backwardsCompat)
+        {
+            super(driver);
+            _backwardsCompat = backwardsCompat;
+            _baseLocator = Locator.tagWithClass("div", "file-upload--container").parent();
```
labkey-tchad: `backwardsCompat=false` would not work as expected here.
```suggestion
            super(driver);
            _backwardsCompat = backwardsCompat;
            if (backwardsCompat)
                _baseLocator = Locator.tagWithClass("div", "file-upload--container").parent();
```

labkey-tchad (approved):

src/org/labkey/test/components/ui/files/FileUploadPanel.java
```diff
@@ -18,11 +18,13 @@ public class FileUploadPanel extends WebDriverComponent<FileUploadPanel.ElementC
 {
     final WebElement _el;
     final WebDriver _driver;
+    final boolean _backwardsCompat;
 
-    public FileUploadPanel(WebElement element, WebDriver driver)
+    public FileUploadPanel(WebElement element, WebDriver driver, boolean backwardsCompat)
```
labkey-chrisj, line 23: I don't love the pattern of parameterizing backwards compatibility here (it seems to require a bit of extra complexity) but I think it does have the upside of making it simple in the future to know when no more callers needing back-support exist.

The other alternative might be to use an xpath-union locator that supports both cases, like 
`Locator.XPathLocator.union(
    Locator.tagWithClass("div", "attached-file--container")
                .withChild(Locator.tagWithClass("span", "fa-times-circle")), 
    Locator.tagWithClass("div", "attached-file__container"))` 
and skip the conditional locator getter

labkey-chrisj (approved):

PR #2226, opened 2025-01-11
New helpers encodeUriPath() and decodeUriPath()

labkey-adam:
#### Rationale
The Jetty utils dependency is used only for encoding and decoding URL paths. We can likely use other existing libraries for these purposes and eliminate this dependency.

Introduce helper methods and route all callers to them. These methods can be updated to test and migrate to new implementations.

### DIFF
--- a/src/org/labkey/test/WebDriverWrapper.java
+++ b/src/org/labkey/test/WebDriverWrapper.java
@@ -24,7 +24,6 @@ import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 import org.assertj.core.api.Assertions;
 import org.awaitility.Awaitility;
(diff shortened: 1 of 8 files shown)


labkey-tchad (approved):
I'm fine with centralizing this but I have been unable to find any other library that clearly distinguishes between the path and query encoding/decoding.
Other methods tend to implement `www-form-urlencoded` which causes some wonkiness when you have a path containing `'+'` and `' '`

labkey-adam:
> I'm fine with centralizing this but I have been unable to find any other library that clearly distinguishes between the path and query encoding/decoding. Other methods tend to implement `www-form-urlencoded` which causes some wonkiness when you have a path containing `'+'` and `' '`

Agreed... I was planning to do a quick replace with a standard method, but it's trickier than I expected, so I'm just bailing out and merging the centralization. Too many higher priority things going on.

PR #2234, opened 2025-01-15
Automation test for invalidate sessions.

labkey-sweta:
#### Rationale
Automation test for 

[Secure Issue 51523](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=51637): Invalidate sessions on password change

[Secure Issue 51637](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=51523): Invalidate sessions when their API key becomes invalid

[Secure Issue 31493:](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=31493) Test for session and cookie persistence through login and logout 

[Secure Issue 51187](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=51187): Additional automation testing for group audit logs

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/src/org/labkey/test/tests/ApiKeyTest.java
+++ b/src/org/labkey/test/tests/ApiKeyTest.java
@@ -214,8 +214,11 @@ public class ApiKeyTest extends BaseWebDriverTest
         log("Deleting the apikey");
         deleteAPIKeys(_generatedApiKeys);
 
+        /*
+            Regression coverage for Issue 52004: Session associated with APIKey can used even after APIKey is deleted.
+         */
         log("Verifying the session associated with deleted apikey is invalid");
-//        verifyInvalidAPIKey(cn, false);
+        verifyInvalidAPIKey(cn, false);
 
         log("Verifying that new connection cannot be created after apikey is deleted");
         verifyInvalidAPIKey(createApiKeyConnection(apiKey1, false), false);
--- a/src/org/labkey/test/tests/ApiKeyTest.java
+++ b/src/org/labkey/test/tests/ApiKeyTest.java
@@ -75,7 +75,7 @@ public class ApiKeyTest extends BaseWebDriverTest
     @BeforeClass
     public static void setupProject()
     {
-        ApiKeyTest init = (ApiKeyTest) getCurrentTest();
+        ApiKeyTest init = getCurrentTest();
 
         init.doSetup();
     }
@@ -107,74 +107,21 @@ public class ApiKeyTest extends BaseWebDriverTest
 
         String apiKey = generateSessionKey();
 
-        verifyValidAPIKey(apiKey);
+        verifyValidAPIKey(createApiKeyConnection(apiKey, false));
 
         log("Verify session key remains valid if key generation is turned off");
         goToAdminConsole()
                 .clickSiteSettings()
                 .setAllowSessionKeys(false)
                 .save();
-        verifyValidAPIKey(apiKey);
+        verifyValidAPIKey(createApiKeyConnection(apiKey, false));
 
         signOut();
         log("Verify that logging out invalidates session keys");
-        verifyInvalidAPIKey(apiKey, null);
+        verifyInvalidAPIKey(createApiKeyConnection(apiKey, false), false);
         simpleSignIn();
         log("Verify that session keys remain invalid after logging back in");
-        verifyInvalidAPIKey(apiKey, null);
-    }
-
-    private void verifyValidAPIKey(String apiKey) throws IOException
-    {
-        verifyValidAPIKey(apiKey, false, null);
-    }
-
-    private Connection verifyValidAPIKey(String apiKey, boolean basicAuth, @Nullable Connection connection) throws IOException
-    {
-        Connection cn;
-        if (connection == null)
-            cn = new Connection(WebTestHelper.getBaseURL(), basicAuth ? new BasicAuthCredentialsProvider(API_USERNAME, apiKey)
-                    : new ApiKeyCredentialsProvider(apiKey));
-        else
-            cn = connection;
-        try
-        {
-            GetSchemasCommand cmd = new GetSchemasCommand();
-            GetSchemasResponse resp = cmd.execute(cn, getProjectName());
-            List<String> schemaNames = resp.getSchemaNames().stream().map(String::toLowerCase).collect(Collectors.toList());
-            Set<String> missingSchemas = new HashSet<>(Arrays.asList("pipeline", "lists", "core"));
-            missingSchemas.removeAll(schemaNames);
-            assertTrue("Some expected schemas missing. Schemas missing: " + missingSchemas, missingSchemas.isEmpty());
-        }
-        catch (CommandException e)
-        {
-            throw new RuntimeException("Response: " + e.getStatusCode(), e);
-        }
-        return cn;
-    }
-
-    private void verifyInvalidAPIKey(String apiKey, @Nullable Connection connection) throws IOException
-    {
-        boolean isSessionKey = !apiKey.startsWith(API_USERNAME);
-        Connection cn;
-        if (connection == null)
-            cn = new Connection(WebTestHelper.getBaseURL(), new ApiKeyCredentialsProvider(apiKey));
-        else
-            cn = connection;
-        try
-        {
-            GetSchemasCommand cmd = new GetSchemasCommand();
-            cmd.execute(cn, getProjectName());
-            if (isSessionKey)
-                fail("Session key didn't invalidate after logout");
-            else
-                fail("API key should no longer be valid");
-        }
-        catch (CommandException e)
-        {
-            assertEquals("Wrong response for invalid " + (isSessionKey ? "session" : "API") + " key", HttpStatus.SC_UNAUTHORIZED, e.getStatusCode());
-            log("Success: command failed as expected.");
-        }
+        verifyInvalidAPIKey(createApiKeyConnection(apiKey, false), false);
     }
 
     @Test
@@ -192,14 +139,14 @@ public class ApiKeyTest extends BaseWebDriverTest
         signIn(EDITOR_USER.getEmail(), EDITOR_USER.getPassword());
         String keyDescription = "Key for editing";
         String apiKey = generateAPIKey(keyDescription);
-        verifyValidAPIKey(apiKey);
+        verifyValidAPIKey(createApiKeyConnection(apiKey, false));
 
         QueryGrid grid = new QueryGrid.QueryGridFinder(getDriver()).waitFor();
         int beforeDeleteCount = grid.getRecordCount();
         assertFalse("Row with description not found", grid.getRowMap("Description", keyDescription).isEmpty());
         grid = deleteAPIKeyViaUI();
         assertEquals("Number of keys after UI deletion not as expected", beforeDeleteCount - 1, grid.getRecordCount());
-        verifyInvalidAPIKey(apiKey, null);
+        verifyInvalidAPIKey(createApiKeyConnection(apiKey, false), false);
     }
 
     @Test
@@ -215,9 +162,9 @@ public class ApiKeyTest extends BaseWebDriverTest
 
         String apiKey = generateAPIKeyAndRecord(_generatedApiKeys);
         log("Verify active API key via api authentication");
-        verifyValidAPIKey(apiKey);
+        verifyValidAPIKey(createApiKeyConnection(apiKey, false));
         log("Verify active API key via basic authentication");
-        verifyValidAPIKey(apiKey, true, null);
+        verifyValidAPIKey(createApiKeyConnection(apiKey, true));
 
         log("Generate two other keys for use in testing deletion.");
         generateAPIKey(null);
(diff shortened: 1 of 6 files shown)


src/org/labkey/test/tests/ApiKeyTest.java
```diff
@@ -118,20 +118,25 @@ public void testSessionKey() throws IOException
...
     }
 
     private void verifyValidAPIKey(String apiKey) throws IOException
     {
-        verifyValidAPIKey(apiKey, false);
+        verifyValidAPIKey(apiKey, false, null);
     }
 
-    private void verifyValidAPIKey(String apiKey, boolean basicAuth) throws IOException
+    private Connection verifyValidAPIKey(String apiKey, boolean basicAuth, @Nullable Connection connection) throws IOException
```
labkey-tchad: These arguments are misleading because there's no guarantee that the `apiKey` provided is actually used by the `connection`. Rather than having the `connection` parameter be nullable with duplicate connection creation code, I would have `verifyValidAPIKey` and `verifyInvalidApiKey` take only a connection. Have a separate method for connection creation:
```
private Connection createApiKeyConnection(String apiKey, boolean basicAuth)
{
    return new Connection(WebTestHelper.getBaseURL(), basicAuth ? new BasicAuthCredentialsProvider(API_USERNAME, apiKey) : new ApiKeyCredentialsProvider(apiKey));
}
```
_Note: `isSessionKey` could become a parameter or the logging could just not make any distinction between regular API and session keys._

src/org/labkey/test/tests/InvalidateSessionTest.java
```diff
@@ -0,0 +1,164 @@
...
+    @Override
+    protected void doCleanup(boolean afterTest)
+    {
+        _containerHelper.deleteProject(getProjectName(), afterTest);
+        _userHelper.deleteUsers(afterTest, USER);
+    }
+
+    private void doSetup()
+    {
+        _containerHelper.createProject(getProjectName(), null);
```
labkey-tchad: Does this test need a project?

src/org/labkey/test/tests/InvalidateSessionTest.java
```diff
@@ -0,0 +1,164 @@
...
+            fail("Should fail when logged out");
+        }
+        catch (CommandException e)
+        {
+            Assert.assertEquals("Command execution should fail when logged out", HttpStatus.SC_UNAUTHORIZED, e.getStatusCode());
+        }
+
+        log("Capture the cookie after logout");
+        Set<Cookie> afterCookie = getDriver().manage().getCookies();
+        Assert.assertFalse("Before and after log out cookie should be different", beforeCookie.equals(afterCookie));
```
labkey-tchad: This is a bit too broad. There might be other cookies that change. Just verify the session id (`Connection.JSESSIONID`).

src/org/labkey/test/tests/InvalidateSessionTest.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    private SetPasswordForm goToChangePassword()
+    {
+        if (PasswordUtil.getUsername().equals(getCurrentUser()))
+            throw new IllegalArgumentException("Don't change the primary site admin user's password");
+
+        goToMyAccount();
+        clickButton("Change Password");
```
labkey-tchad: Add the "Change Password" button to the `UserDetailsPage` returned by `goToMyAccount`. Also, move the primary user assertion into the method that clicks it.
Also, update `PasswordTest` where this appears to be copied from.

src/org/labkey/test/tests/ApiKeyTest.java
```diff
@@ -226,13 +236,60 @@ public void testStandardApiKey() throws IOException
...
+        goToExternalToolPage();
+        String apikey2 = ApiKeyPanel.panelFinder(getDriver()).find().generateApiKey();
+
+        log("Verify valid apikey cannot be created before expiring");
+        verifyValidAPIKey(apikey2, false, null);
+
+        sleep(10000); // Wait for apikey to expire
+
+        log("Verify apikey cannot be created after the timer is expired");
+        verifyInvalidAPIKey(apikey2, null);
```
labkey-tchad: Inaccurate logging
```suggestion
        log("Verify apikey cannot be used after it has expired");
        verifyInvalidAPIKey(apikey2, null);
```

Also, we should verify that existing sessions are invalidated when a key expires.

src/org/labkey/test/tests/ApiKeyTest.java
```diff
@@ -226,13 +236,60 @@ public void testStandardApiKey() throws IOException
...
+                .setAllowApiKeys(true)
+                .setApiKeyExpiration(CustomizeSitePage.KeyExpirationOptions.TEN_SECONDS)
+                .save();
+
+        log("Verify apikey generation fails");
+        goToExternalToolPage();
+        String apikey2 = ApiKeyPanel.panelFinder(getDriver()).find().generateApiKey();
+
+        log("Verify valid apikey cannot be created before expiring");
+        verifyValidAPIKey(apikey2, false, null);
```
labkey-tchad: Inaccurate logging
```suggestion
        log("Verify apikey can be used before expiring");
        verifyValidAPIKey(apikey2, false, null);
```

src/org/labkey/test/tests/ApiKeyTest.java
```diff
@@ -226,13 +236,60 @@ public void testStandardApiKey() throws IOException
...
+        verifyInvalidAPIKey(apiKey1, null);
+
+        log("Generating the apikey which expires in ten seconds");
+        goToAdminConsole()
+                .clickSiteSettings()
+                .setAllowApiKeys(true)
+                .setApiKeyExpiration(CustomizeSitePage.KeyExpirationOptions.TEN_SECONDS)
+                .save();
+
+        log("Verify apikey generation fails");
```
labkey-tchad: Generation doesn't fail.
```suggestion
        log("Verify apikey expiration");
```

labkey-tchad (changes requested):

src/org/labkey/test/tests/ApiKeyTest.java
```diff
@@ -145,12 +150,17 @@ private void verifyValidAPIKey(String apiKey, boolean basicAuth) throws IOExcept
         {
             throw new RuntimeException("Response: " + e.getStatusCode(), e);
         }
+        return cn;
     }
 
-    private void verifyInvalidAPIKey(String apiKey) throws IOException
+    private void verifyInvalidAPIKey(String apiKey, @Nullable Connection connection) throws IOException
     {
         boolean isSessionKey = !apiKey.startsWith(API_USERNAME);
```
labkey-adam: This is no longer a valid check... API keys and session keys no longer start with these prefixes. You can't distinguish them just by looking at them.

src/org/labkey/test/tests/InvalidateSessionTest.java
labkey-sweta: Not needed

src/org/labkey/test/tests/InvalidateSessionTest.java
labkey-sweta: Done

src/org/labkey/test/tests/InvalidateSessionTest.java
```diff
@@ -0,0 +1,151 @@
...
+        try
+        {
+            response = selectCmd.execute(cn, "Home");
+            Assert.assertEquals("Did not establish the database connection before the password change", 200,
+                    response.getStatusCode());
+        }
+        catch (IOException | CommandException e)
+        {
+            throw new RuntimeException(e);
+        }
```
labkey-tchad: No need for a try/catch here. The method can just declare that it throws these exceptions. It already has `throws IOException`

src/org/labkey/test/tests/InvalidateSessionTest.java
```diff
@@ -0,0 +1,151 @@
...
+        try
+        {
+            response = selectCmd.execute(cn, getProjectName());
+            Assert.assertEquals("Did not establish the database connection before the password change", 200,
+                    response.getStatusCode());
+        }
+        catch (CommandException e)
+        {
+            throw new RuntimeException(e);
+        }
```
labkey-tchad: Unnecessary try/catch. Just update the method signature

src/org/labkey/test/tests/InvalidateSessionTest.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    /*
+        Regression coverage for Secure Issue 31493: Test for session and cookie persistence through login and logout
+     */
+    @Test
+    public void testCookieAndSessionFromLogout() throws IOException
+    {
+        log("Capture the cookie after login");
+        Cookie beforeCookie = getCookies(getCurrentUser()).get(Connection.JSESSIONID);
```
labkey-tchad: `WebTestHelper.getCookies` gets cached cookies. Use `getDriver().manage().getCookies()` to ensure that you're getting the live cookies from the browser.

src/org/labkey/test/tests/InvalidateSessionTest.java
```diff
@@ -0,0 +1,151 @@
...
+            selectCmd.execute(cn, getProjectName());
+            fail("Should fail when logged out");
+        }
+        catch (CommandException e)
+        {
+            Assert.assertEquals("Command execution should fail when logged out", HttpStatus.SC_UNAUTHORIZED, e.getStatusCode());
+        }
+
+        log("Capture the cookie after logout");
+        Cookie afterCookie = getCookies(getCurrentUser()).get(Connection.JSESSIONID);
```
labkey-tchad: Use `getDriver().manage().getCookies()`

src/org/labkey/test/tests/InvalidateSessionTest.java
```diff
@@ -0,0 +1,151 @@
...
+        Assert.assertFalse("Before and after log out cookie should be different", beforeCookie.equals(afterCookie));
```
labkey-tchad: Sorry, I should have been more specific. Just verify the `value` of the cookie instead of the entire cookie object. I wouldn't count on `Cookie.equals` to do what this test wants. For example, the expiration date of the cookie might change without the value changing.

labkey-tchad (changes requested):
The test should not use `WebTestHelper.getCookies`.

src/org/labkey/test/tests/InvalidateSessionTest.java
```diff
@@ -139,8 +125,13 @@ public void testCookieAndSessionFromLogout() throws IOException
...
         log("Capture the cookie after logout");
-        Cookie afterCookie = getCookies(getCurrentUser()).get(Connection.JSESSIONID);
-        Assert.assertFalse("Before and after log out cookie should be different", beforeCookie.equals(afterCookie));
+        Set<Cookie> afterCookie = getDriver().manage().getCookies();
+        Assert.assertFalse("Before and after log out cookie should be different", getJSessionIdValue(beforeCookie).equals(getJSessionIdValue(afterCookie)));
+    }
+
+    private String getJSessionIdValue(Set<Cookie> cookies)
+    {
+        return cookies.stream().toList().get(1).getValue();
```
labkey-tchad: Don't assume that the session cookie is at a particular index. Get the cookie named `Connection.JSESSIONID`.

src/org/labkey/test/tests/UserPermissionsTest.java
```diff
@@ -123,24 +123,44 @@ private void userPermissionRightsTest()
...
          * impersonating on my local labkey, so may require special page?
          */
+    }
+
+    @Test
+    public void testSteps()
+    {
+        enableEmailRecorder();
+        userPermissionRightsTest();
+    }
```
labkey-tchad: Just get rid of `testSteps` and make `userPermissionRightsTest ` the `@Test` method. Call `enableEmailRecorder ` from within it and update to be consistent with our test naming scheme (`testUserPermissionRights`)

src/org/labkey/test/tests/UserPermissionsTest.java
```diff
@@ -236,6 +256,33 @@ private void userPermissionRightsTest()
...
+        permissionsHelper.removeUserFromGroup(GAMMA_SUBMITTER_GROUP_NAME, GAMMA_SUBMITTER_USER);
+        verifyAuditLog("User: " + GAMMA_SUBMITTER_USER + " was deleted from Group: " + GAMMA_SUBMITTER_GROUP_NAME);
+    }
+
+    private void verifyAuditLog(String expectedComment)
+    {
+        ShowAuditLogPage showAuditLogPage = goToAdminConsole().clickAuditLog();
+        showAuditLogPage.selectView("Group and role events");
+        DataRegionTable table = showAuditLogPage.getLogTable();
+        Assert.assertEquals("Incorrect audit log record for user getting added to group", expectedComment, table.getDataAsText(0, "Comment"));
```
labkey-tchad: This message isn't accurate for both test cases.

labkey-tchad (changes requested):

labkey-tchad (approved):

PR #2238, opened 2025-01-21
Handle MultipleFailureException thrown by tests

labkey-tchad:
#### Rationale
Sometimes test use `@After` or `@AfterClass` annotations against our recommendation. If a test fails and the after method also throws an exception, JUnit combines the errors into a `MultipleFailureException`. This can mask the actual error, making `BaseWebDriverTest.handleFailure` unable to perform its duties (e.g. close alerts). We mostly care about the primary test failure so we can just pull that out of the `MultipleFailureException`.

#### Related Pull Requests
- N/A

#### Changes
- Handle `MultipleFailureException` thrown by tests


labkey-tchad:
Problematic tests are now collecting screenshots: https://teamcity.labkey.org/buildConfiguration/bt310/3351706

labkey-danield (approved):

PR #2241, opened 2025-01-23
Simple Test Fixes

labkey-danield:
#### Rationale
Some test fixes.

#### Related Pull Requests
- None

#### Changes
- Empty grid cell for an integer field should only show "required" message.
- Mark upgrade tests as disabled so they don't show up in other git tests.

### DIFF
--- a/src/org/labkey/test/tests/ServerUpgradeTest.java
+++ b/src/org/labkey/test/tests/ServerUpgradeTest.java
@@ -6,13 +6,14 @@ import org.junit.experimental.categories.Category;
 import org.labkey.test.BaseWebDriverTest;
 import org.labkey.test.Locator;
 import org.labkey.test.WebTestHelper;
-import org.labkey.test.categories.Git;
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/tests/ServerUpgradeTest.java
```diff
@@ -6,13 +6,14 @@
...
-import org.labkey.test.categories.Git;
+import org.labkey.test.categories.Disabled;
 import org.openqa.selenium.WebElement;
 
 import java.util.Collections;
 import java.util.List;
 
-@Category({Git.class})
+// Don't run as a git test, explicitly add this test to a suite file.
+@Category({Disabled.class})
```
labkey-tchad: `Disabled` will remove it from all suites, even file-based ones. You can just leave the category empty.
```suggestion
@Category({})
```

labkey-tchad (approved):

PR #2244, opened 2025-01-27
Test updates

labkey-sweta:
#### Rationale
image

Hopefully fixes this error.

#### Related Pull Requests
- 

#### Changes
-


labkey-tchad (changes requested):
This wouldn't fix the error, that element won't ever appear at the point where the test is failing.
I have a fix in CDS: https://github.com/LabKey/cds/pull/682

labkey-sweta:
Fixed in https://github.com/LabKey/cds/pull/682#pullrequestreview-2576337955

PR #2277, opened 2025-02-14
coverage for assay transform UI

labkey-chrisj:
#### Rationale
This change adds coverage for assay transform script UI configuration to support transform on insert, update

github issue https://github.com/LabKey/kanban/issues/481

#### Related Pull Requests
related test coverage in limsModules https://github.com/LabKey/limsModules/pull/1168
https://github.com/LabKey/premiumModules/pull/170

#### Changes
Adds a new method to AssayTransformWarningTest to validate insert/update configuration of assay transform script
Updates to ReactAssayDesignerPage to support configuration of transform scripts for assays


src/org/labkey/test/pages/ReactAssayDesignerPage.java
```diff
@@ -359,6 +379,13 @@ final Optional<OptionSelect<MetadataInputFormat>> metadataInputSelect()
             return OptionSelect.finder(Locator.id("assay-design-selectedMetadataInputFormat"), MetadataInputFormat.class)
                     .findOptional(propertiesPanel);
         }
+        Locator runOnActionCheckboxLoc(String fileName, String actionLabel) // actionLabel is Run on Import, or Run on Edit
```
labkey-tchad: Make an `enum` for the possible actions. Tests shouldn't need to know the exact text for the checkbox labels.

labkey-tchad (changes requested):
This doesn't seem to verify that the checkboxes actually do anything. We should verify that unchecking a checkbox skips the transform script for the corresponding action.

labkey-tchad (approved):
If you import them directly, you can make the references to `Import` and `Edit` less verbose.
```
import static org.labkey.test.pages.ReactAssayDesignerPage.ScriptFileEvent.Edit;
import static org.labkey.test.pages.ReactAssayDesignerPage.ScriptFileEvent.Import;
```
Which would allow
```
assayDesignerPage.getScriptActionCheckbox(insertOrUpdateTransform, Import));
```

labkey-tchad (commented):
Now that I look again. This test case doesn't have anything to do with assay transform warnings.
I would create a new assay transform test for this and possible future assay transform test cases.

labkey-danield (commented):
Has the AssayTransformImportUpdateTest run in TeamCity? It looks very similar to the tests in the linked PR, and would be interesting to see if generates the same error that happened when that similar test was run in Biologics.

PR #2290, opened 2025-02-20
List Tests for Special Characters

labkey-danield:
#### Rationale
Test automation to cover issues 52070 and 52069

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6325

#### Changes
- Updated existing list to use special characters in the key field name. Tested in testCustomViews.
- Added new test testTrickyCharacterFields to validate CrUD operations.

### DIFF
--- a/src/org/labkey/test/tests/list/ListTest.java
+++ b/src/org/labkey/test/tests/list/ListTest.java
@@ -30,6 +30,7 @@ import org.labkey.remoteapi.domain.DomainResponse;
 import org.labkey.remoteapi.domain.PropertyDescriptor;
 import org.labkey.remoteapi.domain.SaveDomainCommand;
 import org.labkey.remoteapi.query.Filter;
+import org.labkey.serverapi.reader.TabLoader;
 import org.labkey.test.BaseWebDriverTest;
 import org.labkey.test.Locator;
 import org.labkey.test.SortDirection;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/list/ListTest.java
```diff
@@ -548,7 +552,7 @@ public void testCustomViews()
 
         log("4725: Check Customize View can't remove all fields");
         _customizeViewsHelper.openCustomizeViewPanel();
-        _customizeViewsHelper.removeColumn(LIST_KEY_NAME2);
+        _customizeViewsHelper.removeColumn(EscapeUtil.fieldKeyEncodePart(LIST_KEY_NAME2));
```
labkey-adam, line 555: Shouldn't helper methods like this always do the field key encoding by default? (If there are tests that test field keys, then maybe they should use a different method.)

src/org/labkey/test/tests/list/ListTest.java
```diff
@@ -644,11 +648,11 @@ public void testCustomViews()
 
         log("Check that reference worked");
         _customizeViewsHelper.openCustomizeViewPanel();
-        _customizeViewsHelper.addColumn(_list2Col1.getName() + "/" + _listColDesc.getName(), _list2Col1.getLabel() + " " + _listColDesc.getLabel());
-        _customizeViewsHelper.addColumn(_list2Col1.getName() + "/" + _listColMonth.getName(), _list2Col1.getLabel() + " " + _listColMonth.getLabel());
-        _customizeViewsHelper.addColumn(_list2Col1.getName() + "/" + _listColGood.getName(), _list2Col1.getLabel() + " " + _listColGood.getLabel());
-        _customizeViewsHelper.addFilter(_list2Col1.getName() + "/" + _listColGood.getName(), _listColGood.getLabel(), "Is Less Than", "10");
-        _customizeViewsHelper.addSort(_list2Col1.getName() + "/" + _listColGood.getName(), _listColGood.getLabel(), SortDirection.ASC);
+        _customizeViewsHelper.addColumn(EscapeUtil.fieldKeyEncodePart(_list2Col1.getName()) + "/" + _listColDesc.getName(), _list2Col1.getLabel() + " " + _listColDesc.getLabel());
```
labkey-adam, line 651: Escape by default here as well?

labkey-adam (approved):

labkey-chrisj (approved):
This looks good.
I notice that Adam suggested in the issue that this issue might not be specific to lists alone- may need to consider other coverage, perhaps another story

src/org/labkey/test/tests/list/ListTest.java
labkey-danield: Fair point. I talked with @labkey-tchad about this and it is a bit more involved to update the _customizeViewsHelper.

I believe we might have other test helpers that would be tripped up by tricky characters. I'm going to add a test backlog story to identify as many test helpers as we can that could be updated to work better with tricky characters.

labkey-danield:
> This looks good. I notice that Adam suggested in the issue that this issue might not be specific to lists alone- may need to consider other coverage, perhaps another story

Good point @labkey-chrisj worth investigating but beyond the scope of this change. I can add a story to the backlog to add additional coverage for additional areas.

PR #2291, opened 2025-02-21
coverage for Issue 51382

labkey-chrisj:
#### Rationale
This change adds regression coverage for [Issue 51382](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=51382), Wikis unable to be accessed/deleted due to bad characters

The issue can't be repro'd via the UI, so this uses post commands to cover the create/update scenarios.

It also cleans up tests failing on TC-- [AssayTransformImportUpdateTest](https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailySqlServer/3390042?buildTab=tests&status=failed&suite=org.labkey.test.Runner%3A+&package=org.labkey.test.tests.assay&class=AssayTransformImportUpdateTest) was failing because the  test disabled transform settings prior to checking their measures; the test now enables those measures before checking them

and, [AssayTransformWarningTest](https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyPostgres_2/3390033?buildTab=tests&status=failed&suite=org.labkey.test.Runner%3A+&package=org.labkey.test.tests.assay&class=AssayTransformWarningTest) was failing because the asserts expected the wrong file to have caused errors 

#### Related Pull Requests
n/a

#### Changes
adds 2 test methods

### DIFF
--- a/src/org/labkey/test/tests/assay/AssayTransformImportUpdateTest.java
+++ b/src/org/labkey/test/tests/assay/AssayTransformImportUpdateTest.java
@@ -188,6 +188,13 @@ public class AssayTransformImportUpdateTest extends BaseWebDriverTest
                 " ", dataMap2.get("M2"));
         checker().screenShotIfNewError("unexpected update transform data");
 
+        // re-enable protocol run-on-import and edit so we can measure their metrics
+        assayDesignerPage = ReactAssayDesignerPage.beginAt(this, getProjectName(), protocolResponse.getProtocolId(),
+                "general", getURL().toString());
+        assayDesignerPage.setScriptActionCheckbox(transformFile.getName(), Edit, true);
+        assayDesignerPage.setScriptActionCheckbox(transformFile.getName(), Import, true);
+        assayDesignerPage.clickSave();
+
         // validate some metrics for this feature
         var metricsPage = UsageStatisticsPage.beginAt(this);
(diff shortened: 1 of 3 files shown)


labkey-chrisj:
@RosalineP  One curious behavior I observed when writing the update case here: if I used a new command instance to do the update, the action never went into the `updateWiki  `method -- (it checks for 'form.isNew'' on line 2204 of WikiController) and falls into the `insertWiki `case. 'isNew' appears to only check to see if entityId is null.  (this is curious because setting the same json in different command objects seems to behave differently somehow)
This might just be a thing actual users will never encounter, but I thought it was worth calling out

labkey-sweta (approved):

RosalineP:
> @RosalineP One curious behavior I observed when writing the update case here: if I used a new command instance to do the update, the action never went into the `updateWiki `method -- (it checks for 'form.isNew'' on line 2204 of WikiController) and falls into the `insertWiki `case. 'isNew' appears to only check to see if entityId is null. (this is curious because setting the same json in different command objects seems to behave differently somehow) This might just be a thing actual users will never encounter, but I thought it was worth calling out

Thanks for the heads up on this Chris! I agree this is weird but after a bit of poking around I'm not too concerned. Thank you for writing up this coverage!

RosalineP (approved):

PR #2295, opened 2025-02-25
Add tests and helpers for using tricky characters for Source type fields

labkey-susanh:
#### Rationale
We are looking to add more tests that use realistic names for fields, and names for fields that include characters known to need special handling for field names. This PR and its relative generalize some existing functionality for creating fields and populating data and add other constants and utility methods to ease the incorporation of such fields into tests.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1196

#### Changes
- Add constant lists to hold realistic field definitions for samples, sources, and assay data domains
- Port `BaseColumnInfo.labelFromName` to static method in `FieldDefinition` to help with mapping the realistic names to their labels in grids


src/org/labkey/test/params/FieldDefinition.java
```diff
@@ -78,6 +78,42 @@ public FieldDefinition(String name)
         this(name, ColumnType.String);
     }
 
+    // See BaseColumnInfo.labelFromName
+    public static String labelFromName(String name)
```
cnathe, line 82: bummer we have to duplicate this labelFromName method. When Rosaline and I were talking about the usages of TestDataGenerator.randomFieldName, we discussed the name -> label part since it was hard to know what the label would be for a generated name. Our solution was to just set the field label to be the name so that we always know what it would be to get data back out of the grid. What do you think about making that the default for the `setFieldCaptionsFromNames` helper you made? Having the label be generated by the server's labelFromName isn't really what we are trying to test in these cases, we just need to be able to get the data back out. So if we have the helpers just setLabel to be the name if the label isn't already set, then it would be predictable (if that is the goal). Thoughts?

labkey-susanh: I had thought of something similar, but also wanted this to be set up as much like reality as possible. For example, if I had done that for the current test I wrote, I don't know that I would have seen the caching issue it has pointed out. Since `BaseColumnInfo` is in api module, I had hoped to just reference the `labelFromName` method there and may still be able to do that by adjusting dependencies (I think maybe the jar file needs to be a run-time dependency not just an implementation dependency, but I'm not sure.)

labkey-danield (approved):

RosalineP (approved):

PR #2296, opened 2025-02-25
create windows postgres suite

labkey-chrisj:
#### Rationale
This change creates the Windows Premium suite, which will run tests we want coverage for on Windows/develop

#### Related Pull Requests
n/a

#### Changes
adds the suite file to source control


labkey-chrisj:
Initial run results are here https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_ProductSuites_WindowsPremium/3395725

There are several flavors of test failure here:

- [x] Some tests fail to create containers in hdrl.HDRLContainerListener, "Data source "labware" has not been configured"
- [x] The Mothership tests that try to reference TEST_localhost fail because instead it's TEST_172.31.1.39
- [x] The suite file contained some wrong test names (fixed)

labkey-tchad (commented):
There are still several failures in this suite.

labkey-chrisj:
Looking at the failing mothershipReportTest results right now- it looks like they're mostly a matter of a mismatch between the local installation ID and the server name-- locally, I'm seeing 2 installations- 1 for TEST_localhost, the other for 127.0.0.1.  

`MothershipHelper.setIgnoreExceptions` gets the installationId by asking for `LABKEY.serverName` from the javascript client, 
`MothershipReportTest.testServerHostName` gets it by parsing the host part of the baseURL in `CustomizeSitePage`

For me, both of the above means of getting the host name return 'localhost'
This has me wondering if this is a matter of mothership showing 127.0.0.1 instead of localhost when running on windows
image

The fact that these tests pass on linux suggests to me that either the product behaves differently on windows, or the way we get the serverName yields different results on windows than it does on linux

labkey-chrisj:
At this point, I'm seeking feedback on whether to exclude mothership tests from the suite, or special-case the test on ways to get the local host name that get the same value on windows and linux, or to see if there's a straightforward way for the test to know the local host's name that works the same on either platform.
@labkey-martyp has some thoughts on how windows gets the remote hostname vs. linux
@josheckels probably knows the most about mothership

We only have one mothership and it runs on Linux (so why test on windows?), but the automated tests try to talk to the mothership module on the test machine as if it's the real one and that in turn exercises the reporting itself

labkey-jeckels:
> At this point, I'm seeking feedback on whether to exclude mothership tests from the suite, or special-case the test on ways to get the local host name that get the same value on windows and linux, or to see if there's a straightforward way for the test to know the local host's name that works the same on either platform. @labkey-martyp has some thoughts on how windows gets the remote hostname vs. linux @josheckels probably knows the most about mothership
> 
> We only have one mothership and it runs on Linux (so why test on windows?), but the automated tests try to talk to the mothership module on the test machine as if it's the real one and that in turn exercises the reporting itself

In terms of production usage, the mothership module only runs on labkey.org, so it's Postgres-only. It doesn't really care about Windows vs Linux. I suggest making the tests that check if exceptions were reported as expected implement `PostgresOnlyTest`.

The `TEST_localhost` hostname is a special case within the mothership code for testing purposes. It's not OS-specific code, but it may only be relevant when running a Postgres suite and therefore might only be showing on Linux on TeamCity. https://github.com/LabKey/platform/blob/42b18ec993892599c29551433a0ee43523ce1fdb/api/src/org/labkey/api/util/MothershipReport.java#L173

labkey-chrisj:
https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_ProductSuites_WindowsPremium/3429719
tests are passing, mothershipreporttest is out

labkey-tchad (approved):

PR #2297, opened 2025-02-25
KnitrReportTest: use checkExpectedErrors()

labkey-nicka:
#### Rationale
Switch `KnitrReportTest` to use `checkExpectedErrors()` to handle expected client-side errors when test is running.

#### Changes
- Utilize `checkExpectedErrors(1)` to handle expected client-side error.

### DIFF
--- a/src/org/labkey/test/tests/KnitrReportTest.java
+++ b/src/org/labkey/test/tests/KnitrReportTest.java
@@ -199,10 +199,11 @@ public class KnitrReportTest extends AbstractKnitrReportTest
 
         createKnitrReport(rmdDependenciesReport, RReportHelper.ReportOption.knitrMarkdown);
 
+        int errorCountBefore = getServerErrorCount();
         _rReportHelper.clickReportTab();
         waitForElement(Locator.id("mtcars_table"));
         assertElementNotPresent(Locator.id("mtcars_table_wrapper")); // Created by jQuery
-        checkExpectedErrors(1); // JavaScript error: "$(...).dataTable is not a function"
+        checkExpectedErrors(errorCountBefore + 1); // JavaScript error: "$(...).dataTable is not a function"
 
         // now set the dependencies
         _rReportHelper.clickSourceTab();
--- a/src/org/labkey/test/tests/KnitrReportTest.java
+++ b/src/org/labkey/test/tests/KnitrReportTest.java
@@ -16,6 +16,7 @@
 package org.labkey.test.tests;
 
 import org.apache.hc.core5.http.HttpStatus;
+import org.assertj.core.api.Assertions;
 import org.jetbrains.annotations.Nullable;
 import org.junit.Assume;
 import org.junit.Test;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/KnitrReportTest.java
```diff
@@ -199,13 +199,10 @@ private void verifyAdhocReportDependencies(String viewName, String dependencies)
...
-        {
-            _rReportHelper.clickReportTab();
-            waitForElement(Locator.id("mtcars_table"));
-            assertElementNotPresent(Locator.id("mtcars_table_wrapper")); // Created by jQuery
-        }
-        resumeJsErrorChecker();
+        _rReportHelper.clickReportTab();
+        waitForElement(Locator.id("mtcars_table"));
+        assertElementNotPresent(Locator.id("mtcars_table_wrapper")); // Created by jQuery
+        checkExpectedErrors(1); // JavaScript error: "$(...).dataTable is not a function"
```
XingY, line 207: Use `int errorCountBefore = getServerErrorCount();`  and `checkExpectedErrors(errorCountBefore + 1);` ?

XingY (approved):

src/org/labkey/test/tests/KnitrReportTest.java
labkey-nicka: Done.

labkey-tchad: That would ignore errors from earlier in the test. Better to just fail if it isn't exactly 1.

src/org/labkey/test/tests/KnitrReportTest.java
```diff
@@ -199,13 +199,11 @@ private void verifyAdhocReportDependencies(String viewName, String dependencies)
...
-            _rReportHelper.clickReportTab();
-            waitForElement(Locator.id("mtcars_table"));
-            assertElementNotPresent(Locator.id("mtcars_table_wrapper")); // Created by jQuery
-        }
-        resumeJsErrorChecker();
+        int errorCountBefore = getServerErrorCount();
+        _rReportHelper.clickReportTab();
+        waitForElement(Locator.id("mtcars_table"));
+        assertElementNotPresent(Locator.id("mtcars_table_wrapper")); // Created by jQuery
+        checkExpectedErrors(errorCountBefore + 1); // JavaScript error: "$(...).dataTable is not a function"
```
labkey-tchad: `checkExpectedErrors(1)` is preferable. We want exactly one error.
We should also make sure that it is the error we expect.
```suggestion
        _rReportHelper.clickReportTab();
        waitForElement(Locator.id("mtcars_table"));
        assertElementNotPresent(Locator.id("mtcars_table_wrapper")); // Created by jQuery
        Assertions.assertThat(getServerErrors()).as("Server errors").contains("$(...).dataTable is not a function");
        checkExpectedErrors(1); // JavaScript error: "$(...).dataTable is not a function"
```

labkey-tchad (changes requested):

src/org/labkey/test/tests/KnitrReportTest.java
```diff
@@ -16,6 +16,7 @@
 package org.labkey.test.tests;
 
 import org.apache.hc.core5.http.HttpStatus;
+import org.assertj.core.api.Assertions;
```
labkey-nicka, line 19: Test run here https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyPostgres_2/3396408

labkey-tchad (approved):

PR #2306, opened 2025-02-28
Remove the ability to select specific objects to import and apply import to multiple folders under the advanced import options

labkey-bpatel:
#### Rationale
Issue [49544](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49544): Deprecate the ability to select specific objects to import and apply import to multiple folders under the advanced import options

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6393
- https://github.com/LabKey/premiumModules/pull/183

#### Changes
- Delete AdvancedImportOptionsTest


labkey-klum (changes requested):
Why delete the test now? I think we usually wait until the feature is deleted before doing this. Can we enable the feature flag for the test and then reset it on clean up/

labkey-bpatel:
Will be removed in 25.7 instead, closing the PR.

PR #2310, opened 2025-03-01
Template download button updates

labkey-nicka:
#### Rationale
Test updates for associated changes to application template download button.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1733
- https://github.com/LabKey/labkey-ui-premium/pull/694
- https://github.com/LabKey/limsModules/pull/1209
- https://github.com/LabKey/testAutomation/pull/2310

#### Changes
- Introduce a component wrapper for the `TemplateDownloadButton` React component.
- Utilize the new component wrapper in `FileUploadPanel`.

### DIFF
--- a/src/org/labkey/test/components/react/TemplateDownloadButton.java
+++ b/src/org/labkey/test/components/react/TemplateDownloadButton.java
@@ -1,0 +1,122 @@
+package org.labkey.test.components.react;
+
+import org.labkey.test.Locator;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+import java.io.File;
+
+import static org.labkey.test.WebDriverWrapper.WAIT_FOR_JAVASCRIPT;
+import static org.labkey.test.WebDriverWrapper.waitFor;
+
+public class TemplateDownloadButton extends MultiMenu
+{
+    private static final String DEFAULT_TEMPLATE_NAME = "Default Template";
+    private static final String BUTTON_CLASS = "template-download-button";
+    public static final Locator LOCATOR = Locator.tagWithClass("button", BUTTON_CLASS);
+
+    public TemplateDownloadButton(WebElement element, WebDriver driver)
+    {
+        super(element, driver);
+    }
+
+    private File download(String templateName)
+    {
+        return getWrapper().doAndWaitForDownload(() -> {
+            clickAndWaitForLoaded();
+
+            boolean hasTemplates = hasCustomTemplates();
+            boolean isDefaultDownload = DEFAULT_TEMPLATE_NAME.equals(templateName);
+
+            if (hasTemplates)
+            {
+                if (isDefaultDownload)
+                    doMenuAction(DEFAULT_TEMPLATE_NAME);
+                else
+                    doMenuAction(templateName);
+            }
+            else if (!isDefaultDownload)
+                throw new Error("Custom Templates are not available.");
+            // else the act of clicking the button will download the file
+        });
+    }
+
+    // This will error if there are no custom templates.
+    public MultiMenu expandDropdown()
+    {
+        if (!isDropdown())
+            throw new Error("Custom Templates are not available.");
+
+        return this;
+    }
+
+    public boolean isDropdown()
+    {
+        if (!isLoaded())
+            clickAndWaitForLoaded();
+        return hasCustomTemplates();
+    }
+
+    public File downloadDefaultTemplate()
+    {
+        return download(DEFAULT_TEMPLATE_NAME);
+    }
+
+    public File downloadCustomTemplate(String templateName)
+    {
+        return download(templateName);
+    }
+
+    // This will induce a file download of the default template if there are no custom templates configured.
+    private void clickAndWaitForLoaded()
+    {
+        boolean wasLoaded = isLoaded();
+        getComponentElement().click();
+        if (!wasLoaded)
+            waitFor(this::isLoaded, "Template download button failed to load in time", WAIT_FOR_JAVASCRIPT);
+    }
+
+    private boolean hasButtonClass(String cls)
+    {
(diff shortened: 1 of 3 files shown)


src/org/labkey/test/components/ui/files/FileUploadPanel.java
```diff
@@ -128,7 +118,11 @@ FileInput fileUploadInput()
                 .withChild(Locator.tagWithClass("span", "fa-times-circle"));
         Locator.XPathLocator attachedFileContainerOld = Locator.tagWithClass("div", "attached-file--container")
                 .withChild(Locator.tagWithClass("span", "fa-times-circle"));
-        WebElement downloadTemplate = Locator.linkWithTitle("Download Template").findElementOrNull(getDriver());
+
+        TemplateDownloadButton templateDownloadButton()
+        {
+            return new TemplateDownloadButton.Finder(getDriver()).findWhenNeeded();
+        }
```
labkey-tchad, line 130: Why is this hidden behind a method? Does it appear and disappear depending on user interactions?
```
final TemplateDownloadButton templateDownloadButton = new TemplateDownloadButton.Finder(getDriver()).findWhenNeeded(this);
```
Also, `findWhenNeeded` isn't generally needed when the component isn't cached:
```
TemplateDownloadButton templateDownloadButton()
{
    return new TemplateDownloadButton.Finder(getDriver()).find(this);
}
```

Regardless, the button should be found within the panel (`find*(this)`).

src/org/labkey/test/components/react/TemplateDownloadButton.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    // This will error if there are no custom templates.
+    public MultiMenu expandDropdown()
+    {
+        if (!isDropdown())
+            throw new Error("Custom Templates are not available.");
+
+        return this;
+    }
```
labkey-tchad: Consider overriding `expand` instead of adding a separate, similar method.
```suggestion
    // This will error if there are no custom templates.
    @Override
    public void expand()
    {
        if (!isDropdown())
            throw new Error("Custom Templates are not available.");

        super.expand();
    }
```

src/org/labkey/test/components/react/TemplateDownloadButton.java
```diff
@@ -0,0 +1,122 @@
...
+        boolean wasLoaded = isLoaded();
+        getComponentElement().click();
+        if (!wasLoaded)
+            waitFor(this::isLoaded, "Template download button failed to load in time", WAIT_FOR_JAVASCRIPT);
+    }
+
+    private boolean hasButtonClass(String cls)
+    {
+        return Locator.tagWithClass("button", cls).existsIn(getComponentElement());
+    }
```
labkey-tchad: The element cache contains this button
```suggestion
    private boolean hasButtonClass(String cls)
    {
        return elementCache().toggleAnchor.getDomAttribut("class").contains(cls);
    }
```

src/org/labkey/test/components/react/TemplateDownloadButton.java
```diff
@@ -0,0 +1,122 @@
...
+    public File downloadCustomTemplate(String templateName)
+    {
+        return download(templateName);
+    }
+
+    // This will induce a file download of the default template if there are no custom templates configured.
+    private void clickAndWaitForLoaded()
+    {
+        boolean wasLoaded = isLoaded();
+        getComponentElement().click();
```
labkey-tchad: This should click the button element:
```suggestion
        elementCache().toggleAnchor.click();
```

src/org/labkey/test/components/react/TemplateDownloadButton.java
```diff
@@ -0,0 +1,122 @@
...
+
+    public TemplateDownloadButton(WebElement element, WebDriver driver)
+    {
+        super(element, driver);
+    }
+
+    private File download(String templateName)
+    {
+        return getWrapper().doAndWaitForDownload(() -> {
+            clickAndWaitForLoaded();
```
labkey-tchad: Make it explicit here that this might be what triggers the download.
```suggestion
            clickAndWaitForLoaded(); // Will trigger download if there are no custom templates
```

src/org/labkey/test/components/react/TemplateDownloadButton.java
```diff
@@ -0,0 +1,122 @@
...
+
+            boolean hasTemplates = hasCustomTemplates();
+            boolean isDefaultDownload = DEFAULT_TEMPLATE_NAME.equals(templateName);
+
+            if (hasTemplates)
+            {
+                if (isDefaultDownload)
+                    doMenuAction(DEFAULT_TEMPLATE_NAME);
+                else
+                    doMenuAction(templateName);
```
labkey-tchad, line 47: This seems like an unnecessary `if`
```suggestion
                doMenuAction(templateName);
```

labkey-tchad (approved):
A couple of suggestions.
Maybe add a little summary to the top describing the unusual lifecycle of this button/menu.

src/org/labkey/test/components/ui/files/FileUploadPanel.java
labkey-nicka: Unfortunately, it's not within the panel as you would think. This `FileUploadPanel` test component wrapper should be rewritten slightly to take this into account. It looks like it most closely is attempting to wrap the `<FileAttachmentForm/>` React component, but that doesn't provide any appropriate marker classes for the test to pick up on. I'm going to leave as-is for the purposes of this PR.

image

src/org/labkey/test/components/react/TemplateDownloadButton.java
labkey-nicka: It is certainly necessary. I've added a comment describing what is happening.

labkey-chrisj (approved):

PR #2311, opened 2025-03-02
Additional Test Coverage for Special Characters in Field Names: Registry Sources

RosalineP:
#### Rationale
See spec [here](https://docs.google.com/document/d/14B65GqAEAOXSpLQPH8SaU9ohYxm1HAu4KML4KhttUj0/edit?tab=t.0).

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/2311
- https://github.com/LabKey/limsModules/pull/1207

#### Changes
- Add clickCancel fn and cancel button element


src/org/labkey/test/components/ui/grids/FieldSelectionDialog.java
```diff
@@ -564,6 +564,14 @@ public void clickUpdateGrid()
         });
     }
 
+    public void clickCancel()
```
RosalineP, line 567: Not sure why this is logging 'Misused element cache in org.labkey.test.components.ui.entities.EntityBulkUpdateDialog'—it seems the same to me as the fn above?

labkey-susanh: @labkey-danield or @labkey-tchad Do you have advice on getting rid of this warning?

labkey-tchad: That error isn't a problem with this component.
`EntityBulkUpdateDialog.elementCache` needs to be updated to not instantiate a `new ElementCache`. It should just cast the result of `super.elementCache()`

RosalineP: This did get rid of the log warning—thank you!

src/org/labkey/test/components/ui/grids/FieldSelectionDialog.java
```diff
@@ -564,6 +564,11 @@ public void clickUpdateGrid()
...
+    {
+        elementCache().cancelButton.click();
```
labkey-tchad, line 569: Should this wait for the dialog to close?

PR #2326, opened 2025-03-05
Issue 49544: Deprecate the ability to select specific objects to import and apply import to multiple folders under the advanced import options

labkey-bpatel:
#### Rationale
Issue 49544: Deprecate the ability to select specific objects to import and apply import to multiple folders under the advanced import options

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6419

#### Changes
- Enable 'Advanced Import Options' flag during setup, and disable during cleanup.


src/org/labkey/test/tests/AdvancedImportOptionsTest.java
```diff
@@ -109,6 +111,9 @@ public void doCleanup(boolean afterTest) throws TestTimeoutException
         if (_studyDesignPreviouslyEnabled != null)
             OptionalFeatureHelper.setOptionalFeature(createDefaultConnection(), "studyDesignFlag", _studyDesignPreviouslyEnabled);
 
+        if (_advancedImportOptionsEnabled != null)
+            OptionalFeatureHelper.setOptionalFeature(createDefaultConnection(), "advancedImportFlag", false);
```
labkey-klum: Don't we want to use the initial state of the flag that was stashed away in `_advancedImportOptionsEnabled`?

labkey-klum (approved):

src/org/labkey/test/tests/AdvancedImportOptionsTest.java
labkey-danield: ProgressReportTest could be used as an example:
```
    @Override
    protected void doCreateSteps()
    {
        _studyDesignPreviouslyEnabled = OptionalFeatureHelper.enableOptionalFeature(createDefaultConnection(), "studyDesignFlag");

```
...
```
    @Override
    protected void doCleanup(boolean afterTest)
    {
        if (_studyDesignPreviouslyEnabled != null)
            OptionalFeatureHelper.setOptionalFeature(createDefaultConnection(), "studyDesignFlag", _studyDesignPreviouslyEnabled);
    }

```

labkey-danield (approved):

src/org/labkey/test/tests/AdvancedImportOptionsTest.java
labkey-bpatel: Thank you, both. Now updated.

PR #2328, opened 2025-03-06
Issue 52478: ELN assay run reference via bulk add not working for assay name with special characters

cnathe:
#### Rationale
See related PR for rationale. This PR updates the randomDomainName() test helper to take a prefix string from the user.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1739
- https://github.com/LabKey/labkey-ui-premium/pull/698
- https://github.com/LabKey/limsModules/pull/1222
- https://github.com/LabKey/testAutomation/pull/2328

#### Changes
- TestDataGenerator.randomDomainName to allow for string prefix set by caller

### DIFF
--- a/src/org/labkey/test/util/TestDataGenerator.java
+++ b/src/org/labkey/test/util/TestDataGenerator.java
@@ -344,12 +344,17 @@ public class TestDataGenerator
 
     public static String randomDomainName(int size)
     {
+        return randomDomainName(size, null);
+    }
+
+    public static String randomDomainName(int size, @Nullable String prefix)
+    {
         String domainName = "";
         do
         {
-            String prefix = randomString(1, null, ALPHANUMERIC_STRING); // domain needs to start with alphanumeric char
-            final String charset = ALPHANUMERIC_STRING + DOMAIN_SPECIAL_STRING;
-            domainName = prefix + randomString(size - 1, null, charset);
+            String _prefix = prefix != null ? prefix : randomString(1, null, ALPHANUMERIC_STRING); // domain needs to start with alphanumeric char
+            final String charset = prefix != null ? DOMAIN_SPECIAL_STRING : ALPHANUMERIC_STRING + DOMAIN_SPECIAL_STRING;
+            domainName = _prefix + randomString(size - 1, null, charset);
             domainName = domainName.trim();
         }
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -343,13 +343,18 @@ public static String randomInvalidDomainName(int size)
     }
 
     public static String randomDomainName(int size)
+    {
+        return randomDomainName(size, null);
+    }
+
+    public static String randomDomainName(int size, @Nullable String prefix)
```
labkey-susanh, line 356: Maybe add a comment about the difference in behavior between a null prefix and a not null prefix.

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -343,13 +343,18 @@ public static String randomInvalidDomainName(int size)
...
     {
         String domainName = "";
         do
         {
-            String prefix = randomString(1, null, ALPHANUMERIC_STRING); // domain needs to start with alphanumeric char
-            final String charset = ALPHANUMERIC_STRING + DOMAIN_SPECIAL_STRING;
-            domainName = prefix + randomString(size - 1, null, charset);
+            String _prefix = prefix != null ? prefix : randomString(1, null, ALPHANUMERIC_STRING); // domain needs to start with alphanumeric char
+            final String charset = prefix != null ? DOMAIN_SPECIAL_STRING : ALPHANUMERIC_STRING + DOMAIN_SPECIAL_STRING;
```
labkey-susanh, line 362: Not sure why we'd want to skip the alphanumeric characters, but I'm sure there's a reason...

labkey-susanh (approved):

src/org/labkey/test/util/TestDataGenerator.java
cnathe: That was just because I assumed that if a caller is providing their own prefix, they are using AlphaNumeric values and would just want to tack on the special chars. Let me know if you think I should put that back to always include `ALPHANUMERIC_STRING`.

labkey-susanh: Seems simpler to not have the conditional there, but I don't feel strongly about it.

PR #2339, opened 2025-03-11
Fix NonStandard Time Format Tests

labkey-danield:
#### Rationale
The tests for non-standard formats needs to work better with daylight saving time. Specifically, time only fields don't care about daylight savings.
Also some fixes for the ExcelHelper and TestDataGenerator.

@labkey-susanh added as FYI only.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1231

#### Changes
- Update test so time only fields work correctly and ignore daylight savings.
- Update ExcelHelper.getCellStringValue to work correctly with date only and time only fields.
- Modify TestDataGenerator.randomFieldName to not include double spaces (it's just easier).


labkey-sweta (approved):

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -382,7 +382,13 @@ public static String randomFieldName(String part, int numStartChars, int numEndC
     {
         // use the characters that we know are encoded in fieldKeys plus characters that we know clients are using
         String chars = ALL_ILLEGAL_QUERY_KEY_CHARACTERS + " %()=+-[]_|*`'\":;<>?!@#^";
-        String randomFieldName = (randomString(numStartChars, exclusion, chars) + part + randomString(numEndChars, null, chars)).trim();
+
+        // Having double space is allowed in a field name but is a problem for automation.
+        // We render double spaces as a single space as column headers in grids and the like, but we maintain the double
+        // space in the name. This trips up helpers for various components. See Issue 52193 for details.
+        String randomFieldName = (randomString(numStartChars, exclusion, chars).replace("  ", " ") +
```
labkey-susanh, line 389: Would probably be best to do a `replaceAll("\\s\\s+"," ")`, just to be safe

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -382,7 +382,13 @@ public static String randomFieldName(String part, int numStartChars, int numEndC
...
+                part +
+                randomString(numEndChars, null, chars).replace("  ", " ")).trim();
```
labkey-tchad, line 391: This is really an issue with QueryGrid and EditableGrid (and others I'm sure). We should open an issue to track and prioritize a refactor of our tests and helpers to key off of field names (or fieldKey) instead of using field labels.

labkey-tchad (approved):

labkey-tchad (changes requested):
These failures are happening in 25.3. We should apply these fixes there.

labkey-danield:
> These failures are happening in 25.3. We should apply these fixes there.

Yes, I know. I've made that comment several times in stand up.
Our merge rate is too slow. I want to get this into develop now, and will cherry pick into 25. 3.

PR #2340, opened 2025-03-11
Assay Designs additional test coverage for special characters in field names

cnathe:
#### Rationale
This set of PRs adds additional test coverage for special characters in field names of assay designs. It covers assay import via grid and import via file along with assay results editing via grid and in bulk. It adds usages of `TestDataGenerator.randomFieldName()` and `verifyEntityData()` to some existing test cases. It also adds a new test case for the various assay import and edit results options using the set of `REALISTIC_ASSAY_FIELDS`.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1232
- https://github.com/LabKey/platform/pull/6431
- https://github.com/LabKey/testAutomation/pull/2340

#### Changes
- createFieldList() to explicitly set field label based on name to prevent variability for test checks
- EntityBulkUpdateDialog fix for setDateField to use fieldKey param
- EditableGrid selectCell to use "scroll to center" to handle edit case where first column is locked on scroll

### DIFF
--- a/src/org/labkey/test/components/ui/entities/EntityBulkUpdateDialog.java
+++ b/src/org/labkey/test/components/ui/entities/EntityBulkUpdateDialog.java
@@ -1,6 +1,5 @@
 package org.labkey.test.components.ui.entities;
 
-import org.labkey.api.query.QueryKey;
 import org.labkey.test.BootstrapLocators;
 import org.labkey.test.Locator;
 import org.labkey.test.WebDriverWrapper;
@@ -13,6 +12,7 @@ import org.labkey.test.components.react.FilteringReactSelect;
 import org.labkey.test.components.react.ReactDateTimePicker;
 import org.labkey.test.components.react.ToggleButton;
 import org.labkey.test.components.ui.files.FileAttachmentContainer;
+import org.labkey.test.util.EscapeUtil;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.ui.ExpectedConditions;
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/components/ui/entities/EntityBulkUpdateDialog.java
```diff
@@ -286,7 +287,7 @@ public Input numericInput(String fieldKey)
         public ReactDateTimePicker dateInput(String fieldKey)
         {
             return new ReactDateTimePicker.ReactDateTimeInputFinder(getDriver())
-                    .withInputId(fieldKey).waitFor(formRow(fieldKey));
+                    .withInputId(QueryKey.encodePart(fieldKey)).waitFor(formRow(fieldKey));
```
labkey-tchad: Avoid using server API. It can cause trouble. We have a test equivalent of this method.
```suggestion
                    .withInputId(EscapeUtil.fieldKeyEncodePart(fieldKey)).waitFor(formRow(fieldKey));
```

labkey-tchad (changes requested):
Don't use the server API in tests

labkey-tchad (approved):
Nevermind. I guess we use `QueryKey` in lots of tests.

labkey-tchad:
> - createFieldList() to explicitly set field label based on name to prevent variability for test checks

What sort of variability? Does `FieldDefinition.labelFromName` not match reality in some cases?

src/org/labkey/test/components/ui/entities/EntityBulkUpdateDialog.java
cnathe: If using `EscapeUtil.fieldKeyEncodePart` is preferred on the test side, I'm happy to change to this. Done.

cnathe:
> > * createFieldList() to explicitly set field label based on name to prevent variability for test checks
> 
> What sort of variability? Does `FieldDefinition.labelFromName` not match reality in some cases?

"variability" is probably the wrong word. It would be things like if the field name was "random_1" in the FieldDefinition object and no label was set, the server would set the label by default to "Random 1". So tests that get values from the data grid would need to use that generated label. But good call on the `FieldDefinition.labelFromName` helper. I'll switch to using that instead of explicitly setting label on many of the FieldDefinition objects.

PR #2342, opened 2025-03-11
Ensure we use legal characters for artifact file names

labkey-tchad:
#### Rationale
I noticed a failure that was unable to take a screenshot with `checker().screenShotIfNewError` on Windows because it was trying to include illegal characters in the file name.
There are several places that include field names or sample type names in screenshot names. As we use more tricky characters more broadly, we should make `ArtifactCollector` more resilient to invalid file names.

#### Related Pull Requests
- N/A

#### Changes
- Ensure we use legal characters for artifact file names
- Use `FileUtils.forceMkdir` to ensure that directories are created


labkey-danield (approved):
Good catch!

PR #2133, opened 2024-11-07
wait for job without navigating to job status page

labkey-chrisj:
#### Rationale
Recently, [UploadLargeExcelAssayTest.testUpload200kRows](https://teamcity.labkey.org/viewLog.html?tab=queuedBuildOverviewTab&buildId=3261497&buildTypeId=LabKey_2411_Premium_CommunitySqlserver_DailyCSqlserver&fromSakuraUI=true#testNameId3537965916196401765) began failing on sql because it takes > 3 minutes to navigate to the job status page.
This change awaits the job completion without attempting to navigate to the job status page.

I've opened https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51630 to track the underlying issue

#### Related Pull Requests
n/a

#### Changes
use waitForPipelineJobsToFinish() instead

PR #2345, opened 2025-03-12
go back to covering pipelineStatusDetailsPage pathway (Issue 51630 coverage)

labkey-chrisj:
#### Rationale
This change restores `UploadLargeExcelAssayTest `to verify we're not hanging or unresponsive in `pipelineStatusDetails `page, in effect un-doing the workaround introduced in https://github.com/LabKey/testAutomation/pull/2149.
Related issue: [51630](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51630)

#### Related Pull Requests
https://github.com/LabKey/testAutomation/pull/2133

#### Changes
Rather than use `waitForPipelineJobsToComplete `(as many tests do), navigate the lesser-travelled clicking of job status to monitor assay import progress

### DIFF
--- a/src/org/labkey/test/tests/assay/UploadLargeExcelAssayTest.java
+++ b/src/org/labkey/test/tests/assay/UploadLargeExcelAssayTest.java
@@ -112,7 +112,9 @@ public class UploadLargeExcelAssayTest extends BaseWebDriverTest
         clickButton("Save and Finish");
 
         // wait for import complete
-        waitForPipelineJobsToComplete(1, "200k", false, 12 * 60000);
+        var assayJobsPage1 = new AssayUploadJobsPage(getDriver());
+        var pipelineDetailsPage1 = assayJobsPage1.clickJobStatus("200k", 3 * WebDriverWrapper.WAIT_FOR_PAGE);
+        pipelineDetailsPage1.waitForComplete(12 * WebDriverWrapper.WAIT_FOR_PAGE);
 
         // export assay1 data to excel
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/assay/UploadLargeExcelAssayTest.java
```diff
@@ -133,7 +135,9 @@ public void testUpload200kRows() throws Exception
         setFormElement(Locator.input("__primaryFile__"), largeExportExcelFile);
         clickButton("Save and Finish");
 
-        waitForPipelineJobsToComplete(2, "200k take 2", false, 12 * 60000);
+        var assayJobsPage2 = new AssayUploadJobsPage(getDriver());
+        var pipelineDetailsPage2 = assayJobsPage2.clickJobStatus("200k take 2", 3 * getDefaultWaitForPage());
```
labkey-jeckels, line 139: This multiplies by `getDefaultWaitForPage()` whereas line 116 multiplies by `WebDriverWrapper.WAIT_FOR_PAGE`. Are they the same, and is one superior to the other?

labkey-jeckels (approved):

src/org/labkey/test/tests/assay/UploadLargeExcelAssayTest.java
labkey-chrisj: Yes, they are the same, and no, I don't think one is any better than the other

PR #2347, opened 2025-03-13
Attribute column-name -> data-column-name

labkey-adam:
#### Rationale
`data-` attributes are best practice


labkey-tchad (approved):
There are a couple of references to this attribute in premiumModules as well

labkey-adam:
> There are a couple of references to this attribute in premiumModules as well

https://github.com/LabKey/premiumModules/pull/193

labkey-matthewb (approved):

PR #2351, opened 2025-03-17
25.3 fb test investigation

labkey-danield:
#### Rationale
The tests for non-standard formats needs to work better with daylight saving time. Specifically, time only fields don't care about daylight savings.
Also some fixes for the ExcelHelper and TestDataGenerator.

#### Related Pull Requests
- None

#### Changes
- Update test so time only fields work correctly and ignore daylight savings.
- Update ExcelHelper.getCellStringValue to work correctly with date only and time only fields.
- Modify TestDataGenerator.randomFieldName to not include double spaces (it's just easier).


labkey-danield:
Not sure what I did wrong here, but closing this PR and creating a new one for 25.3.

PR #2358, opened 2025-03-18
TestDataGenerator.randomFieldName to replace duplicate spaces in name as they don't work well with test helpers

cnathe:
#### Rationale
Note that this is applying the change from Dan's 25.3 PR: https://github.com/LabKey/testAutomation/pull/2352

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1243

#### Changes
- replace duplicate spaces in name as they don't work well with test helpers


src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -392,7 +392,13 @@ public static String randomFieldName(String part, int numStartChars, int numEndC
...
         // use the characters that we know are encoded in fieldKeys plus characters that we know clients are using
         String chars = ALL_ILLEGAL_QUERY_KEY_CHARACTERS + " %()=+-[]_|*`'\":;<>?!@#^";
-        String randomFieldName = (randomString(numStartChars, exclusion, chars) + part + randomString(numEndChars, null, chars)).trim();
+
+        // Having double space is allowed in a field name but is a problem for automation.
+        // We render double spaces as a single space as column headers in grids and the like, but we maintain the double
+        // space in the name. This trips up helpers for various components. See Issue 52193 for details.
+        String randomFieldName = (randomString(numStartChars, exclusion, chars).replaceAll("\\s+"," ") +
+                part +
+                randomString(numEndChars, null, chars).replaceAll("\\s+"," ")).trim();
```
labkey-susanh: Dan merged almost this exact same change in 25.3-SNAPSHOT, so it will get merged forward sometime soon.

We also, separately, just saw a failure because a domain type (sample type, I think) contained multiple spaces, so we will likely need to update some helper method for that as well.

labkey-susanh (approved):

labkey-danield (approved):

src/org/labkey/test/util/TestDataGenerator.java
cnathe: yup, I copied over Dan's change here to get things running in my fb. I'll handle the merge conflict into my branch later today....which I guess likely means that this PR will get closed as a no-op.

cnathe:
No longer needed after merge from 25.3 to develop

PR #2362, opened 2025-03-19
Update LabkeyErrorPageTest to match new error page content

labkey-adam:
#### Rationale
Error page content changed based on https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52495

LabkeyErrorPageTest didn't get the memo

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6437

### DIFF
--- a/src/org/labkey/test/tests/LabkeyErrorPageTest.java
+++ b/src/org/labkey/test/tests/LabkeyErrorPageTest.java
@@ -101,8 +101,8 @@ public class LabkeyErrorPageTest extends BaseWebDriverTest
 
         checker().verifyEquals("Incorrect error heading message", "Oops! An error has occurred.",
                 errorPage.getErrorHeading());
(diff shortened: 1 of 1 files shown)


labkey-tchad (approved):
I noticed that we don't hide the reference code messaging when no such code exists. We only generate the reference code when reporting to mothership. This test server didn't have self-reporting enabled:
image
This is longstanding behavior (just checked 24.11)

labkey-adam:
> I noticed that we don't hide the reference code messaging when no such code exists. We only generate the reference code when reporting to mothership. This test server didn't have self-reporting enabled: image This is longstanding behavior (just checked 24.11)

I noticed the same. I'm not going to bother correcting in this PR (or 25.3), but I've added this to my todo list.

labkey-adam:
Failing test passed here: https://teamcity.labkey.org/buildConfiguration/LabKey_253Release_Community_DailySuites_DailyAPostgres/3428413?buildTab=tests

PR #2363, opened 2025-03-20
Additional Test Coverage: Story #1 Sample Types

labkey-susanh:
#### Rationale
We are updating more of our tests to use random tricky characters so need to continue to improve the helper methods to support this.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1258

#### Changes
- Fix in `FieldDefinition.labelFromName` to account for display collapsing multiple spaces to one space
- Update SampleTypeNameExpressionTest to use random characters for naming pattern
- Fix `TestDataGenerator.randomFieldName` to use exclusion characters in both leading and trailing random strings


labkey-chrisj (approved):
Only note is that as we get clever and helpful in these helper methods, we should probably consider adding javadoc/explanations so that our future selves can have the benefit of what today-us were thinking

labkey-tchad (approved):

PR #2364, opened 2025-03-21
Don't store "realistic" FieldDefinitions as global constants

labkey-tchad:
#### Rationale
`FieldDefinition` is a mutable class; we shouldn't store globally accessible instances of them. There are already a couple of places that are inadvertently modifying the "realistic" field definitions (setting labels and import aliases).

Also, `BaseWebDriverTest` is massive enough as is; moving the realistic field definitions into `TestDataUtils` to avoid bloating this almost 3000 line file any more.

#### Related Pull Requests
- #2295 

#### Changes
- Convert "REALISTIC" field definitions into `Suppliers` to prevent persistent modification
- Moving realistic field definitions into `TestDataUtils`
- Make random number generation consistent in `TestDataGenerator`

### DIFF
--- a/src/org/labkey/test/util/TestDataUtils.java
+++ b/src/org/labkey/test/util/TestDataUtils.java
@@ -15,7 +15,7 @@ import java.util.function.Supplier;
 
 public class TestDataUtils
 {
-    // 'FieldDefinition's are mutable, don't store them as global constants
+    // 'FieldDefinition's are mutable; don't store them as global constants
     public static final List<Supplier<FieldDefinition>> REALISTIC_ASSAY_FIELDS = List.of(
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/util/TestDataUtils.java
```diff
@@ -3,16 +3,109 @@
...
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Supplier;
 
 public class TestDataUtils
 {
+    // 'FieldDefinition's are mutable, don't store them as global constants
```
labkey-susanh:
```suggestion
    // 'FieldDefinition's are mutable; don't store them as global constants
```

labkey-susanh (approved):

labkey-sweta (approved):

PR #2365, opened 2025-03-21
log character set during test setup

labkey-chrisj:
#### Rationale
This introduces a change to log the current (default) character set, which will be useful to know when troubleshooting encoding issues, particularly in tests running on Windows agents.
On Linux, the default seems to be reliably `UTF-8` and on some (?) Windows agents the default is `windows-1252` (I say "some" here because locally I get UTF-8 when running on Windows 🤷 

#### Related Pull Requests
n/a

#### Changes
add log entry to record the character set

### DIFF
--- a/src/org/labkey/test/WebDriverWrapper.java
+++ b/src/org/labkey/test/WebDriverWrapper.java
@@ -116,6 +116,7 @@ import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLDecoder;
+import java.nio.charset.Charset;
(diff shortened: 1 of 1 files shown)


labkey-chrisj:
Apparently we already set the character set to UTF-8 when running UI tests (or at least, we set ` -Dfile.encoding-UTF-8` as a jvm arg, according to https://github.com/LabKey/gradlePlugin/blob/develop/src/main/groovy/org/labkey/gradle/task/RunUiTest.groovy#L60) so we may need further investigation on this issue

labkey-tchad (approved):

PR #2368, opened 2025-03-24
Collapse multiple spaces to one in random domain names

labkey-susanh:
#### Rationale
Multiple spaces in names are collapsed into single spaces in the UI. Change the domain name generator to also do this collapsing so we can find the entities that are generated with usual UI locators.

#### Changes
- Update `TestDataGenrator.randomDomainName` to collapse multiple spaces into one.


labkey-tchad (approved):
Should this just be added to the `Pattern` above that triggers this to retry? How important is it to get a String that is exactly the requested size?

labkey-susanh:
> Should this just be added to the `Pattern` above that triggers this to retry? How important is it to get a String that is exactly the requested size?

The exact size of the field is not really important, and I'd rather have fewer things to check in that while loop than more. (Looping until random data matches a certain pattern makes me nervous.)

labkey-chrisj (approved):

PR #2378, opened 2025-03-28
Fix MalformedUrlException in crawler

labkey-tchad:
#### Rationale
`UrlToCheck._relativeURL` has inconsistent format. It might have a leading '/' or not and it might not even be relative. Attempting to to create a `URL` with a relative URL that doesn't have a leading '/' is causing errors. We should make `UrlToCheck._relativeURL` has a consistent format (actually relative and no leading '/') and account for that structure elswhere.

```
java.lang.RuntimeException: java.net.MalformedURLException: Error at index 4 in: "8111FilesQueryTestProject"
  at org.labkey.test.util.Crawler.crawlLink(Crawler.java:1142)
  at org.labkey.test.util.Crawler.crawl(Crawler.java:842)
  at org.labkey.test.util.Crawler.crawlAllLinks(Crawler.java:805)
```

#### Related Pull Requests
- #2369 

#### Changes
- Fix MalformedUrlException in crawler
- Make `UrlToCheck._relativeURL` has a consistent format
- Remove workaround for fixed issue: [42661](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42661)

### DIFF
--- a/src/org/labkey/test/util/Crawler.java
+++ b/src/org/labkey/test/util/Crawler.java
@@ -43,7 +43,6 @@ import org.labkey.test.WebDriverWrapper;
 import org.labkey.test.WebTestHelper;
 import org.labkey.test.components.core.ProjectMenu;
 import org.labkey.test.util.selenium.WebDriverUtils;
-import org.openqa.selenium.TimeoutException;
 import org.openqa.selenium.UnhandledAlertException;
 import org.openqa.selenium.WebDriverException;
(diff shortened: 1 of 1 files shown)


labkey-tchad:
Verifying on TeamCity: https://teamcity.labkey.org/buildConfiguration/LabKey_253Release_External_Panorama_PanoramaAPostgres/3443474

labkey-sweta (approved):

PR #2380, opened 2025-03-29
March dependency updates

labkey-martyp:
#### Rationale
Update dependencies

#### Changes
- Bump versions

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,9 +1,9 @@
 aspectjVersion=1.9.22.1
-assertjVersion=3.26.3
+assertjVersion=3.27.3
 awaitilityVersion=4.2.2
 
-lookfirstSardineVersion=5.12
-jettyVersion=12.0.15
+lookfirstSardineVersion=5.13
+jettyVersion=12.0.16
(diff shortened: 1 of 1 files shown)


labkey-adam (changes requested):
Looks like there are newer versions of Jetty and AspectJ available. Note that the version numbers listed in the "Latest" column in the upgrade doc are just hints... they aren't necessarily the most recent version of each. Use the link on the far left to check for each dependency's latest version.

labkey-adam:
Probably a good time to update the LabKey Schemas version as well

labkey-martyp:
> Looks like there are newer versions of Jetty and AspectJ available. Note that the version numbers listed in the "Latest" column in the upgrade doc are just hints... they aren't necessarily the most recent version of each. Use the link on the far left to check for each dependency's latest version.

Ok I updated more of these packages. I had looked at a few in the doc in a previous PR and looked like you had the latest so I wasn't checking any further. These should now all be up to the latest.

labkey-adam (approved):

labkey-tchad (approved):

PR #2390, opened 2025-04-02
Fix custom data class fuzz tests

labkey-susanh:
#### Rationale
In the details panel, we need to use labels or fieldKeys when setting values.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1304

#### Changes
- Update `DetailTableEdit.setDetails` method to not attempt any setting if given a null value
- Update keys used in `DetailTableEidt.setDetails` method.

### DIFF
--- a/src/org/labkey/test/components/ui/entities/EntityBulkUpdateDialog.java
+++ b/src/org/labkey/test/components/ui/entities/EntityBulkUpdateDialog.java
@@ -1,6 +1,5 @@
 package org.labkey.test.components.ui.entities;
 
-import org.labkey.api.query.QueryKey;
 import org.labkey.test.BootstrapLocators;
 import org.labkey.test.Locator;
 import org.labkey.test.WebDriverWrapper;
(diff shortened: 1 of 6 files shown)


src/org/labkey/test/components/ui/entities/EntityBulkUpdateDialog.java
```diff
@@ -62,18 +61,18 @@ private WebDriverWait waiter()
...
+            setNumericField(EscapeUtil.fieldKeyEncodePart(field.getName()), String.valueOf(newValue));
         else if (field.getType() == FieldDefinition.ColumnType.Date || field.getType() == FieldDefinition.ColumnType.DateAndTime || field.getType() == FieldDefinition.ColumnType.Time)
-            bulkEditDialog.setCustomDateField(EscapeUtil.fieldKeyEncodePart(field.getName()), (String) newValue);
+            setDateField(field.getLabel() == null ? field.getName() : field.getLabel(), (String) newValue);
         else if (field.getType() == FieldDefinition.ColumnType.Boolean)
-            bulkEditDialog.setBooleanField(EscapeUtil.fieldKeyEncodePart(field.getName()), (Boolean) newValue);
+            setBooleanField(EscapeUtil.fieldKeyEncodePart(field.getName()), (Boolean) newValue);
         else
-            bulkEditDialog.setTextField(EscapeUtil.fieldKeyEncodePart(field.getName()), (String) newValue);
+            setTextField(EscapeUtil.fieldKeyEncodePart(field.getName()), (String) newValue);
```
labkey-chrisj, line 75: Might be worth asserting or throwing if we fall out of this if/else block here; let's have the code tell us we're missing a case if we possibly can 😺

src/org/labkey/test/components/ui/grids/DetailTableEdit.java
```diff
@@ -419,14 +420,17 @@ private ReactDateTimePicker getDateTimePicker(String fieldCaption)
     // For use when the field is of an unknown type, as can occur in fuzz tests
     public void setDetails(FieldDefinition field, Object newValue, DetailTableEdit editTable)
```
labkey-chrisj: Do we need to pass `editTable `in here as a parameter? Seems like we could use the current instance 🤔 

Also, consider an assert or throwing if we fall out of these ifs

labkey-chrisj (approved):
Thank you!

src/org/labkey/test/components/ui/grids/EditableGrid.java
```diff
@@ -529,6 +530,20 @@ else if (value instanceof LocalTime localTime)
         return gridCell;
     }
 
+    public void setEntityData(List<Map<String, Object> >data, List<FieldDefinition> fields)
+    {
+        for (int i = 0; i < data.size(); i++)
+        {
+            Map<String, Object> rowData = data.get(i);
+            for (FieldDefinition field : fields) {
+                String key = field.getLabel() != null ? field.getLabel() : field.getName();
```
labkey-danield, line 539: It's beyond the scope of this PR, but I'm wondering if it would be worthwhile to make a change to FieldDefinition to have a method like `getIdentifier` which would have this conditional.

May not be worth it, but just an idea.

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -72,6 +77,13 @@ public class TestDataGenerator
     public static final String ALPHANUMERIC_STRING = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789abcdefghijklmnopqrstuvxyz";
     public static final String DOMAIN_SPECIAL_STRING =  "+- _.:&()/";
     public static final String ILLEGAL_DOMAIN_NAME_CHARSET = "<>[]{};,`\"~!@#$%^*=|?\\";
+    // Used to set value of date pickers
+    public static final Supplier<SimpleDateFormat> INPUT_DATE_FORMAT = () -> new SimpleDateFormat("MM/dd/yyyy");
+    public static final Supplier<SimpleDateFormat> INPUT_DATETIME_FORMAT = () -> new SimpleDateFormat("MM/dd/yyyy HH:mm:ss");
+    // Default display format
+    public static final Supplier<SimpleDateFormat> UI_DATE_FORMAT = () -> new SimpleDateFormat("yyyy-MM-dd");
+    public static final Supplier<SimpleDateFormat> UI_DATE_TIME_FORMAT = () -> new SimpleDateFormat("yyyy-MM-dd HH:mm");
+    public static final Supplier<SimpleDateFormat> TIME_FORMAT = () -> new SimpleDateFormat("HH:mm:ss");
```
labkey-danield, line 86: These formats are enums in `BaseSettingsPage.DATE_FORMAT`.
I don't know what the best solution is but I would like to see the test automation changed to have values like these formats, or various "tricky characters" string and system defined field names / labels (e.g. Sample ID, Sample ID *, etc...) in a more generalized resource file or class.

Obviously beyond the scope of this PR.

labkey-danield (approved):

src/org/labkey/test/components/ui/grids/EditableGrid.java
labkey-susanh: I had a similar thought. (And, sorry, I saw this after I pushed the Merge button.) Certainly worth revision in a later PR.

src/org/labkey/test/util/TestDataGenerator.java
labkey-susanh: Ah, shoot. Also missed this in my eagerness to merge. If these are available elsewhere, I can update to use the other values in my current branch (assuming I remember).

PR #2393, opened 2025-04-03
Disable crawler for WikiCspTest

labkey-tchad:
#### Rationale
The crawler can trigger a CSP warning if it hits the wiki created by this test. The crawler was intended to be skipped but the wrong method was overridden (disabling the mem check instead of the link check).

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/2156

#### Changes
- Disable crawler for WikiCspTest


labkey-chrisj (approved):
Thank you!

PR #2394, opened 2025-04-03
Additional Test Coverage for Lists

labkey-danield:
#### Rationale
Primarily adding additional test cover for the list object.

Adding coverage for:
- [52071](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=52071) List PK with special characters results in mangled View List Item page (Must Do)
- [51572](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51572) Bug with creating a new list by uploading a csv file in "UTF-8 with BOM" format
- [52480](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52480): Updating a field name that previously had a 'tricky character' in the name results in an error on MSSQL.
- [52694](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52694): Links broken to list, data classes from the (list)-begin.view page if their names end with a /
- Change List Name

Moved other lists test to the list package under testAutomation.

#### Related Pull Requests
- https://github.com/LabKey/premiumModules/pull/208

#### Changes
- Added coverage for various issues.
- Moved ListArchiveExportTest, ListArchiveImportTest & ListExportTest to be under the list package in testAutomation.
- Renamed ListTest.testChangeListName to testChangeListNameOverAPI and removed ignore annotation.
- Added ListTest.testChangeListName


src/org/labkey/test/tests/list/ListTest.java
```diff
@@ -905,6 +952,36 @@ public void testChangeListName() throws Exception
         assertEquals("remoteAPIAfterRename", saveResponse.getDomain().getName());
     }
 
+    @Test
```
labkey-sweta, line 948: Missing the comment with Issue information.

src/org/labkey/test/tests/list/ListTest.java
```diff
@@ -505,6 +507,52 @@ public void testNameTrimming()
...
+                data[i][j] = switch (j)
+                {
+                    case 0 -> Integer.toString(i + 101);
+                    case 1 -> String.format("A%d", i + 2);
+                    case 2 -> String.format("B%d", i + 2);
+                    case 3 -> String.format("C%d", i + 2);
+                    default -> throw new IllegalStateException("Unexpected value: " + j);
+                };
+            }
+        }
```
labkey-sweta: Why cant we have a constant expected value like `String [][] data = {
                {"101","A2","B2","C2"},
                {"102","A3","B3","C3"},
                {"103","A4","B4","C4"},
                {"104","A5","B5","C5"},
                {"105","A6","B6","C6"},
                {"106","A7","B7","C7"},
                {"107","A8","B8","C8"},
                {"108","A9","B9","C9"},
                {"109","A10","B10","C10"},
                {"110","A11","B11","C11"}
        };
    }`
?

labkey-sweta (changes requested):

src/org/labkey/test/tests/list/ListTest.java
labkey-danield: Initially I did, but that tied the test very tightly to the test data. This allows someone to change the number of rows in the test file without having to change the test code.

src/org/labkey/test/tests/list/ListTest.java
labkey-danield: There is no issue for this test. This was just a testing gap that I filled.

src/org/labkey/test/tests/list/ListTest.java
labkey-sweta: In my opinion if someone is changing the file, they can change data here as well. Secondly data can be changed to anything based on future requirement without restricting to pattern here 

But i let you decide.

src/org/labkey/test/tests/list/ListTest.java
labkey-sweta: Issue 52694 i thought this was the issue, since its used for listNameAfter.

labkey-sweta (approved):

src/org/labkey/test/tests/list/ListTest.java
```diff
@@ -1196,23 +1273,27 @@ private void customFormattingTest()
...
     {
+        // Issue 8329
         log("8329: Test that renaming a field then creating a new field with the old name doesn't result in awful things");
         String listName = "new";
-        String origFieldName = "BarBar";
-        String newFieldName = "FooFoo";
+
+        // Issue 52480
+        String origFieldName = ": random 1 /@\".";
+        String newFieldName = "random 1";
```
labkey-susanh: I might not use "random" in the names here lest we mistake these for names that are actually generated.

src/org/labkey/test/tests/list/ListTest.java
```diff
@@ -1690,6 +1792,27 @@ private void testTricky(String listName, boolean autoKey) throws IOException
...
+        // Validates Issue 52071
+        dataRegionTable.clickRowDetails(0);
+
+        List<String> actualFields = Locator.tagWithClass("td", "lk-form-label").findElements(getDriver()).stream().map(WebElement::getText).toList();
+
+        List<String> expectedFields = new ArrayList<>();
+        // Replace the _ with a space and add a : at the end.
+        if (!autoKey)
+        {
+            expectedFields.add(keyField.replace("_", " ") + ":");
```
labkey-susanh: Not needed since we know there's only one, but `replaceAll` seems like it would be better for future-proofing.

labkey-susanh (approved):
Test cases look good. A few minor comments.

PR #2396, opened 2025-04-04
WebDriverWrapper helper for getCurrentContainerName()

cnathe:
#### Rationale
See related PR for rationale

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1311

#### Changes
- helper for getCurrentContainerName()

### DIFF
--- a/src/org/labkey/test/WebDriverWrapper.java
+++ b/src/org/labkey/test/WebDriverWrapper.java
@@ -1321,6 +1321,11 @@ public abstract class WebDriverWrapper implements WrapsDriver
         return (String)executeScript("return LABKEY.container.id;");
     }
 
+    public String getCurrentContainerName()
+    {
+        return (String)executeScript("return LABKEY.container.name;");
+    }
+
     public String getContainerId(String url)
     {
         pushLocation();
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/WebDriverWrapper.java
```diff
@@ -1321,6 +1321,11 @@ public String getContainerId()
         return (String)executeScript("return LABKEY.container.id;");
     }
 
+    public String getCurrentContainerName()
+    {
+        return (String)executeScript("return LABKEY.container.name;");
+    }
+
```
labkey-tchad: I would say that `LabKeySiteWrapper.getCurrentContainer` should be moved here and be updated to return the name instead of the title.
We should deal with names by default and I bet all of the existing usages actually want the name and would break if the container title was different from its name.

labkey-tchad (dismissed):

src/org/labkey/test/WebDriverWrapper.java
labkey-danield: Trey will be on vacation starting tomorrow. @cnathe you can ping me after you make the change @labkey-tchad recommends and I can re-review.

cnathe: Yup, I can make that change.

labkey-danield (approved):

PR #2399, opened 2025-04-04
Use correct noun for field labels

labkey-tchad:
#### Rationale
[[Story](https://github.com/LabKey/kanban/issues/620)]

Numerous tests refer to a field's "name" when they really mean its label or fieldKey. Methods and variable names should accurately reflect what they represent.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1313
- https://github.com/LabKey/premiumModules/pull/232

#### Changes
- Use correct and consistent nouns for labels and fieldKeys


src/org/labkey/test/components/ui/grids/DetailTable.java
```diff
@@ -88,9 +88,9 @@ public Boolean isLoaded()
      */
     private WebElement getField(String identifier)
```
labkey-chrisj, line 89: I wonder if `getField `should become `getFieldByLabel`?  (or perhaps, `identifier `could become '`label`'?)
Just offering suggestions at this point

labkey-chrisj (approved):

src/org/labkey/test/components/domain/DomainFormPanel.java
```diff
@@ -496,7 +496,7 @@ public DomainFormPanel clickSummaryGridNameLink(String linkText)
 
     public List<String> getSummaryModeColumns()
     {
-        var rawColumns = getSummaryModeGrid().getColumnNames();
+        var rawColumns = getSummaryModeGrid().getColumnLabels();
```
labkey-danield, line 499: Would it be more consistent to rename `getColumnNames` to `getFieldLabels`? Many of these other changes replaced Column with Field.

labkey-danield (approved):

PR #2401, opened 2025-04-07
Remove support for extra date/time parsing patterns

labkey-adam:
#### Rationale
We are removing support for extra data/time parsing patterns; see related issue.

The test change deserves some scrutiny. I renamed it and removed a good chunk of the code, since the parsing patterns have been removed. The "study load" method now really only tests that a non-standard date format is applied after import; not sure how valuable this is. Feedback welcome.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6541


labkey-danield (approved):
I think the test look good.
I've run the test locally as well, and I think the coverage is ok.

PR #2405, opened 2025-04-09
25.3 fb fix intermittent failing tc tests

labkey-chrisj:
#### Rationale
This change attempts to stabilize a couple of recently-failing tests on TeamCity

[SimpleModuleTest](https://teamcity.labkey.org/viewLog.html?buildId=3458584&tab=buildResultsDiv&buildTypeId=LabKey_253Release_Community_DailySuites_DailyBPostgres#testNameId5981971736586896608) fails intermittently because at the moment the test needs to interact with a webPart permissions menu, there's a tooltop hovering over it. 

[ClientAPITest](https://teamcity.labkey.org/viewLog.html?buildId=3453890#testNameId5684422414308035516), one of our BVT tests, failed recently attempting to navigate via the project menu in a test that isn't actually about navigating that way. This replaces that call (and several like it) with beginAt actions.

#### Related Pull Requests
n/a

#### Changes
try to make these tests more reliable

### DIFF
--- a/src/org/labkey/test/tests/SimpleModuleTest.java
+++ b/src/org/labkey/test/tests/SimpleModuleTest.java
@@ -467,6 +467,8 @@ public class SimpleModuleTest extends BaseWebDriverTest
         assertTextPresentInThisOrder("A customized web part", "Data Pipeline", "Experiment Runs", "Sample Type", "Assay List");
         assertTextPresent("Run Groups");
         assertElementNotPresent(Locator.linkWithText("Create Run Group")); // Not in small Run Groups web-part.
+        // ensure the mouse isn't inadvertently hovering something else, causing obscuring popup
+        mouseOver(portalHelper.getBodyWebPart("A customized web part").getComponentElement());
         portalHelper.checkWebpartPermission("A customized web part", "Read", null);
         portalHelper.checkWebpartPermission("Data Pipeline", "Read", null);
     }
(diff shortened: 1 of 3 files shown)


src/org/labkey/test/tests/SimpleModuleTest.java
```diff
@@ -467,6 +467,8 @@ private void doTestCustomFolder()
         assertTextPresentInThisOrder("A customized web part", "Data Pipeline", "Experiment Runs", "Sample Type", "Assay List");
         assertTextPresent("Run Groups");
         assertElementNotPresent(Locator.linkWithText("Create Run Group")); // Not in small Run Groups web-part.
+        // ensure the mouse isn't inadvertently hovering something else, causing obscuring popup
+        mouseOver(portalHelper.getBodyWebPart("A customized web part").getComponentElement());
```
labkey-danield: I would move the mouse off the web part, maybe to the icon at the top of the page.
```suggestion
        mouseOver(Locator.tagWithClass("a", "brand-logo"));
```

labkey-danield (changes requested):
If there is a problem with `projectMenu().navigateToFolder` then it should be fixed, not avoided.

labkey-chrisj:
> If there is a problem with `projectMenu().navigateToFolder` then it should be fixed, not avoided.

It's pretty reliable; what I'm trying to avoid is the mouse being left hovering things with tooltips after it's been used.

labkey-sweta (approved):

labkey-danield (approved):

PR #2189, opened 2024-12-11
Add restrictions on names of domains - automated tests

XingY:
#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1662
- https://github.com/LabKey/labkey-ui-premium/pull/616
- https://github.com/LabKey/limsModules/pull/984
- https://github.com/LabKey/platform/pull/6140
- https://github.com/LabKey/testAutomation/pull/2189
- https://github.com/LabKey/commonAssays/pull/826

#### Changes
-

PR #2409, opened 2025-04-10
Make more Locators work with tricky characters

labkey-tchad:
#### Rationale
Wider use of tricky characters and string has revealed some defects in how we construct Locators.

We use XPath's `normalize-space()` method for most text matching but we don't normalize whitespace in the text we are comparing to. We need to perform normalization to both sides of comparisons we perform.

We have some Locators that do case-insensitive text matching. The current implementation only works for `A-Z`; an XPath translate statement that simulates a toLowerCase sort of operation (`translate(normalize-space(), 'ABCDEFGHIJKLMNOPQRSTUVWXYZ', 'abcdefghijklmnopqrstuvwxyz')`). We know the String the Locator will be comparing to, so we can construct a translate statement that does a `toLowerCase` for just the characters we care about matching, including ones with diacritics (e.g. `Ä` -> `ä`).

#### Related Pull Requests
- #2189 

#### Changes
- Consistently normalize whitespace of strings
- Support a wider set of characters for case-insensitive text matching

### DIFF
--- a/src/org/labkey/test/Locator.java
+++ b/src/org/labkey/test/Locator.java
@@ -46,8 +46,10 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Function;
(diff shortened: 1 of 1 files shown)


XingY (approved):

src/org/labkey/test/Locator.java
```diff
@@ -975,6 +977,16 @@ public static String xq(String value)
         return Quotes.escape(value);
     }
 
+    /**
+     * Equivalent to XPath {@code normalize-space()}:<br>
+     * "The normalize-space function strips leading and trailing white-space from a string, replaces sequences of
+     * whitespace characters by a single space, and returns the resulting string."
+     */
+    private static String ns(String value)
```
labkey-danield, line 985: I would suggest a better, more descriptive method name.

labkey-danield (approved):

PR #2411, opened 2025-04-14
Expand audit entry diff count check to sample/source details update and to source bulk update dialog

cnathe:
#### Rationale
In the related PR from last week, I added a check in the sample bulk update test component save button click to verify the last audit entry event newRecordMap diff change count. This set of PRs moves that code to the AuditLogHelper class and extends it to be checked for both sample and source/registry bulk update and single sample/source overview detail update.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1311
- https://github.com/LabKey/limsModules/pull/1332

#### Changes
- move audit entry diff count check helper code from SMTestUtils to AuditLogHelper
- EntityBulkUpdateDialog to check audit entry diff by default for clickUpdate()
- DetailTableEdit to check audit entry diff by default for clickSave


src/org/labkey/test/util/AuditLogHelper.java
```diff
@@ -69,6 +76,116 @@ public DataRegionTable goToAuditEventView(String eventType)
...
+        for (int i = 0; i < expectedDiffCounts.size(); i++)
+        {
+            Integer expectedDiffCount = expectedDiffCounts.get(i);
+            Map<String, Object> event = events.get(i);
+            String dataChangesStr = (String) event.get("NewRecordMap");
+            String[] dataChanges = dataChangesStr != null ? dataChangesStr.split("&") : new String[0];
+            // filter out SampleStateLabel as that is not a change, it is added for display purposes
+            dataChanges = Stream.of(dataChanges).filter(s -> !s.toLowerCase().startsWith("samplestatelabel=")).toArray(String[]::new);
+            BaseWebDriverTest.getCurrentTest().log("Audit record data changes diff count check: " + dataChangesStr);
+            BaseWebDriverTest.getCurrentTest().checker().verifyEquals("Audit record data changes did not include the expected number of diffs, expected " + expectedDiffCount + " but was " + dataChanges.length + ": " + dataChangesStr,
```
cnathe: @labkey-tchad You mentioned via chat that `BaseWebDriverTest.getCurrentTest()` isn't correct here because of the test scoping. Do you have a recommended approach here? Should I just throw or should I pass through the BaseWebDriverTest in some way?

labkey-tchad: These methods shouldn't be `static`. Our standard pattern is that `Helper` classes are concrete classes that are constructed with all of the context they will need. `Util` classes are collections of static methods like these.
An instance of `AuditLogHelper` has everything it needs for these methods to be non-static, you'd just need to change `_wrapper` to be a `BaseWebDriverTest`; that would give easy access to the `checker()`.

src/org/labkey/test/util/AuditLogHelper.java
```diff
@@ -69,6 +76,116 @@ public DataRegionTable goToAuditEventView(String eventType)
...
+     * @return A rowResponse with the query logs.
+     * @throws IOException Can be thrown by the SelectRowsCommand.
+     * @throws CommandException Can be thrown by the SelectRowsCommand.
+     */
+    public static SelectRowsResponse getAuditLogsFromLKS(String projectName, String folderName, String auditEventName, List<String> columnNames,
+                                                         List<Filter> filters, @Nullable Integer maxRows) throws IOException, CommandException
+    {
+        SelectRowsCommand cmd = new SelectRowsCommand("auditLog", auditEventName);
+        cmd.setColumns(columnNames);
+        cmd.addFilter("ProjectId/DisplayName", projectName, Filter.Operator.EQUAL);
```
labkey-tchad: This filter should use the project path, not the display name. We don't, generally, set custom container display names but we might some day.
```suggestion
        cmd.addFilter("ProjectId/Path", projectName, Filter.Operator.EQUAL);
```

src/org/labkey/test/util/AuditLogHelper.java
```diff
@@ -69,6 +76,116 @@ public DataRegionTable goToAuditEventView(String eventType)
...
+     * @param auditEventName Name of the audit event to filter on. Example 'SamplesWorkflowAuditEvent'.
+     * @param columnNames The name of the columns to return.
+     * @param filters The filters to be applied
+     * @return A rowResponse with the query logs.
+     * @throws IOException Can be thrown by the SelectRowsCommand.
+     * @throws CommandException Can be thrown by the SelectRowsCommand.
+     */
+    public static SelectRowsResponse getAuditLogsFromLKS(String projectName, String folderName, String auditEventName, List<String> columnNames,
+                                                         List<Filter> filters, @Nullable Integer maxRows) throws IOException, CommandException
+    {
```
labkey-tchad: This should just take the containerPath as a single parameter. Having separate projectName and folderName is confusing when they are equal; technically, you could have a subfolder with the same name as the project.
```suggestion
    public static SelectRowsResponse getAuditLogsFromLKS(String containerPath, String auditEventName, List<String> columnNames,
                                                         List<Filter> filters, @Nullable Integer maxRows) throws IOException, CommandException
    {
        String projectName = StringUtils.strip(containerPath, " /").split("/")[0];
```

src/org/labkey/test/util/AuditLogHelper.java
```diff
@@ -69,6 +76,116 @@ public DataRegionTable goToAuditEventView(String eventType)
...
+        List<Map<String, Object>> events = getAuditLogsFromLKS(projectName, folderName, auditEventName, List.of("NewRecordMap"), Collections.emptyList(), maxRows).getRows();
+        for (int i = 0; i < expectedDiffCounts.size(); i++)
+        {
+            Integer expectedDiffCount = expectedDiffCounts.get(i);
+            Map<String, Object> event = events.get(i);
+            String dataChangesStr = (String) event.get("NewRecordMap");
+            String[] dataChanges = dataChangesStr != null ? dataChangesStr.split("&") : new String[0];
+            // filter out SampleStateLabel as that is not a change, it is added for display purposes
+            dataChanges = Stream.of(dataChanges).filter(s -> !s.toLowerCase().startsWith("samplestatelabel=")).toArray(String[]::new);
+            BaseWebDriverTest.getCurrentTest().log("Audit record data changes diff count check: " + dataChangesStr);
```
labkey-tchad: You can use `TestLogger.log` for logging from a static context.
```suggestion
            TestLogger.log("Audit record data changes diff count check: " + dataChangesStr);
```

labkey-tchad (changes requested):

src/org/labkey/test/util/AuditLogHelper.java
cnathe: ah, ok perfect. that makes sense. thank you

cnathe: @labkey-tchad Trey, I'm going to need some help with this. I am able to change the _wrapper in AuditLogHelper to type BaseWebDriverTest, that part makes sense. However, I am still unsure how I can then use the AuditLogHelper inside of the component class such as EntityBulkUpdateDialog or DetailTableEdit. I don't see a way that they have access to the BaseWebDriverTest. It seems like what I'd really like is for that AuditLogHelper _wrapper to be of type WebDriverWrapper, but then I wouldn't have access to the checker().

src/org/labkey/test/util/AuditLogHelper.java
cnathe: I agree. I was just moving this method over from SMTestUtils, but I'll make this change.

src/org/labkey/test/util/AuditLogHelper.java
cnathe: done

src/org/labkey/test/util/AuditLogHelper.java
cnathe: done

src/org/labkey/test/util/AuditLogHelper.java
labkey-tchad: In that case, I'd say to just to a normal assertion. Making the `checker` available in a place like this might require some pretty significant refactoring and redesigning of our component patterns. Probably not worth the trouble at this time but something to consider in the future.

cnathe: Gotcha, that makes sense to me. I'll make it an assertEquals(). Thank you.

src/org/labkey/test/util/AuditLogHelper.java
```diff
@@ -69,6 +80,117 @@ public DataRegionTable goToAuditEventView(String eventType)
...
+                                                         List<Filter> filters, @Nullable Integer maxRows) throws IOException, CommandException
+    {
+        SelectRowsCommand cmd = new SelectRowsCommand("auditLog", auditEventName);
+        cmd.setColumns(columnNames);
+        cmd.addFilter("ProjectId/Name", _wrapper.getCurrentProject(), Filter.Operator.EQUAL);
+        filters.forEach(cmd::addFilter);
+        if (maxRows != null)
+            cmd.setMaxRows(maxRows);
+        if (!containerPath.startsWith("/"))
+            containerPath = "/" + containerPath;
```
labkey-tchad: FYI, the Java client api cares whether the containerPath has a leading slash. It handles either case.

src/org/labkey/test/util/AuditLogHelper.java
```diff
@@ -69,6 +80,117 @@ public DataRegionTable goToAuditEventView(String eventType)
...
+            return "SampleTimelineEvent";
+        else if (isDataClassRoute())
+            return "SourcesAuditEvent";
+        return null;
+    }
+
+    public static boolean isSamplesRoute()
+    {
+        URL url = BaseWebDriverTest.getCurrentTest().getURL();
+        if (url != null)
```
labkey-tchad: Doe these route checking methods need to be static? Want to avoid using `BaseWebDriverTest.getCurrentTest()`.
```suggestion
    public boolean isSamplesRoute()
    {
        URL url = _wrapper.getURL();
        if (url != null)
```

src/org/labkey/test/util/AuditLogHelper.java
```diff
@@ -69,6 +80,117 @@ public DataRegionTable goToAuditEventView(String eventType)
...
+    {
+        SelectRowsCommand cmd = new SelectRowsCommand("auditLog", auditEventName);
+        cmd.setColumns(columnNames);
+        cmd.addFilter("ProjectId/Name", _wrapper.getCurrentProject(), Filter.Operator.EQUAL);
+        filters.forEach(cmd::addFilter);
+        if (maxRows != null)
+            cmd.setMaxRows(maxRows);
+        if (!containerPath.startsWith("/"))
+            containerPath = "/" + containerPath;
+        return cmd.execute(WebTestHelper.getRemoteApiConnection(), containerPath);
```
labkey-tchad: For consistency, use the connection supplier in this class rather than the `WebTestHelper` one.
```suggestion
        return cmd.execute(_connectionSupplier.get(), containerPath);
```

labkey-tchad (approved):

src/org/labkey/test/util/AuditLogHelper.java
cnathe: Thanks, removed.

src/org/labkey/test/util/AuditLogHelper.java
cnathe: No need to be static anymore. Changed.

src/org/labkey/test/util/AuditLogHelper.java
cnathe: Done

src/org/labkey/test/util/AuditLogHelper.java
```diff
@@ -69,6 +77,112 @@ public DataRegionTable goToAuditEventView(String eventType)
...
+     * transactionId. If no transactionId is provided, it will check just the latest row.
+     */
+    public void checkTimelineAuditEventDiffCount(String containerPath, List<Integer> expectedDiffCounts) throws IOException, CommandException
+    {
+        checkTimelineAuditEventDiffCount(containerPath, getAuditEventNameFromURL(), expectedDiffCounts);
+    }
+    public void checkTimelineAuditEventDiffCount(String containerPath, String auditEventName, List<Integer> expectedDiffCounts) throws IOException, CommandException
+    {
+        Integer maxRows = expectedDiffCounts.size();
+        List<Map<String, Object>> events = getAuditLogsFromLKS(containerPath, auditEventName, List.of("NewRecordMap"), Collections.emptyList(), maxRows).getRows();
```
labkey-danield, line 115: Should you assert the the events count matches the expectedDiffsCounts?

labkey-danield (approved):

src/org/labkey/test/util/AuditLogHelper.java
cnathe: yup, added.

PR #2412, opened 2025-04-14
add escape util methods for sql and markup

labkey-chrisj:
#### Rationale
This change adds very basic escape/enquote helpers to EscapeUtil to safely use randomly-generated identifiers (such as field names and domain names) in sql queries and in xml markup like ETL definitions or query metadata.

It also adds a quick-and-dirty method to ETLSchedulerPage to navigate to the transform view if it's not there already and to return a transform/etl schedule by its name.

#### Related Pull Requests
https://github.com/LabKey/premiumModules/pull/211

#### Changes
Helper methods to escape text inputs for sql and markup languages

### DIFF
--- a/src/org/labkey/test/pages/dataintegration/ETLSchedulerPage.java
+++ b/src/org/labkey/test/pages/dataintegration/ETLSchedulerPage.java
@@ -50,6 +50,27 @@ public class ETLSchedulerPage extends LabKeyPage<ETLSchedulerPage.Elements>
         return new ETLSchedulerPage(test);
     }
 
+    /**
+     * gets the ETL given its name.
+     * @param transformName
+     * @return
+     */
+    public TransformRow getTransform(String transformName)
+    {
+        var queryRowLoc = Locator.tag("tr").withDescendant(Locator.tagWithText("td", transformName));
+
+        if (!queryRowLoc.isDisplayed(getDriver()) && Locator.lkButton("View ETL Scheduler").isDisplayed(getDriver()))
+        {
+            clickButton("View ETL Scheduler");
+        }
+        else if (!queryRowLoc.isDisplayed(getDriver()) && Locator.lkButton("View Custom ETL Definitions").isDisplayed(getDriver()))
+        {
+            clickButton("View Custom ETL Definitions");
+            clickButton("View ETL Scheduler");
+        }
+        return new TransformRow(queryRowLoc.findElement(getDriver()));
+    }
+
     public TransformRow transform(String transformId)
     {
         return elementCache().findTransformRow(transformId);
--- a/src/org/labkey/test/pages/dataintegration/ETLSchedulerPage.java
+++ b/src/org/labkey/test/pages/dataintegration/ETLSchedulerPage.java
@@ -57,17 +57,9 @@ public class ETLSchedulerPage extends LabKeyPage<ETLSchedulerPage.Elements>
      */
     public TransformRow getTransform(String transformName)
     {
-        var queryRowLoc = Locator.tag("tr").withDescendant(Locator.tagWithText("td", transformName));
+        var queryRowLoc = Locator.tag("tr").withAttribute("transformid")
+                .withChild(Locator.tagWithText("td", transformName));
 
-        if (!queryRowLoc.isDisplayed(getDriver()) && Locator.lkButton("View ETL Scheduler").isDisplayed(getDriver()))
-        {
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/util/EscapeUtil.java
```diff
@@ -149,4 +149,32 @@ public static String getTextChoiceValidatorExpression(List<String> options)
                 .map(value -> value.replaceAll("([\\\\|])", "\\\\$1"))
                 .collect(Collectors.joining("|"));
     }
+
+    /**
+     * Takes an identifier (a field or domain value) and enquotes it, while also escaping any quotes
+     * it might have.  This his helpful for using randomly-generated identifiers in sql queries.
+     * @param value
+     * @return
```
labkey-jeckels: These aren't adding any value. I suggest deleting them, and for the other new method.

labkey-jeckels (approved):

src/org/labkey/test/util/EscapeUtil.java
```diff
@@ -149,4 +149,19 @@ public static String getTextChoiceValidatorExpression(List<String> options)
...
+
+    public static String getMarkupEscapedValue(String value)
+    {
+        return value
+                .replace("&", "&amp;")
+                .replace("<", "&lt;")
+                .replace(">", "&gt;")
+                .replace("\"", "&quot;")
+                .replace("'", "&apos;");
+    }
```
labkey-tchad: The Apache commons text library provides this functionality. No need to reinvent. If you must define a method, just farm out the work to the established library.
```suggestion
    public static String getMarkupEscapedValue(String value)
    {
        return StringEscapeUtils.escapeXml11(value);
    }
```

src/org/labkey/test/pages/dataintegration/ETLSchedulerPage.java
```diff
@@ -50,6 +50,27 @@ public static ETLSchedulerPage beginAt(BaseWebDriverTest test, String containerP
...
+
+        if (!queryRowLoc.isDisplayed(getDriver()) && Locator.lkButton("View ETL Scheduler").isDisplayed(getDriver()))
+        {
+            clickButton("View ETL Scheduler");
+        }
+        else if (!queryRowLoc.isDisplayed(getDriver()) && Locator.lkButton("View Custom ETL Definitions").isDisplayed(getDriver()))
+        {
+            clickButton("View Custom ETL Definitions");
+            clickButton("View ETL Scheduler");
+        }
```
labkey-tchad: Please try this without all the extra navigation and checking. This is the `ETLSchedulerPage`; It should be able to just get the transform row. `DataIntegrationController.BeginAction` is pretty straightforward; I'm not sure what you were seeing that prompted this.

src/org/labkey/test/pages/dataintegration/ETLSchedulerPage.java
```diff
@@ -50,6 +50,27 @@ public static ETLSchedulerPage beginAt(BaseWebDriverTest test, String containerP
...
     }
 
+    /**
+     * gets the ETL given its name.
+     * @param transformName
+     * @return
+     */
+    public TransformRow getTransform(String transformName)
+    {
+        var queryRowLoc = Locator.tag("tr").withDescendant(Locator.tagWithText("td", transformName));
```
labkey-tchad: This Locator is not very precise. Make sure the row has the `transformId` attribute and that the transformName is in the first `td`.

src/org/labkey/test/util/EscapeUtil.java
labkey-chrisj: Thanks!

src/org/labkey/test/pages/dataintegration/ETLSchedulerPage.java
labkey-chrisj: While I was writing this test, I found I'd sometimes have to click that 'view ETL Scheduler' button in order to get to the ETLScheduler page (I forget whether at that point I was trying to get there via folderManagement -> ETLs tab) so I wrote a way to handle that.
Now that I'm using the ETLSchedulerPage class instead of solving it in-test, it seems I don't need that checking; it's gone and if this test fails now as a result my head will probably essplode 😆

src/org/labkey/test/pages/dataintegration/ETLSchedulerPage.java
labkey-chrisj: Done

PR #2420, opened 2025-04-21
Add a `labelFromName` version that doesn't collapse spaces

labkey-susanh:
#### Rationale
Sometimes, such as when getting data from the UI using a `data-` property, we need the `labelFromName` that doesn't collapse multiple spaces.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1342

#### Changes
- Add new variant of `FeidlDefinition.labelFromName`


labkey-nicka (approved):

labkey-tchad:
Don't use this too much. I'm working on a change that will get field labels without losing the extra spaces: https://github.com/LabKey/testAutomation/compare/develop...fb_doubleSpaces

PR #2424, opened 2025-04-22
surface a method to empty the clipboard contents

labkey-chrisj:
#### Rationale
This change adds a public method to allow tests to clear the clipboard without resorting to copying empty grid cells

#### Related Pull Requests
https://github.com/LabKey/limsModules/pull/1355

#### Changes
- pull the implementation of setting the clipboard out into its own protected method
- surface a public method to allow clearing the clipboard

### DIFF
--- a/src/org/labkey/test/WebDriverWrapper.java
+++ b/src/org/labkey/test/WebDriverWrapper.java
@@ -3553,7 +3553,7 @@ public abstract class WebDriverWrapper implements WrapsDriver
 
     public void clearClipboardContent()
     {
-        setClipboardContent("");
+        setClipboardContent(" ");
     }
 
     protected void setClipboardContent(String text)
--- a/src/org/labkey/test/WebDriverWrapper.java
+++ b/src/org/labkey/test/WebDriverWrapper.java
@@ -3530,9 +3530,7 @@ public abstract class WebDriverWrapper implements WrapsDriver
     {
         Keys cmdKey = WebDriverUtils.MODIFIER_KEY;
 
-        Clipboard c = Toolkit.getDefaultToolkit().getSystemClipboard();
-        StringSelection sel = new StringSelection(text);
-        c.setContents(sel, sel);
+        setClipboardContent(text);
 
         if (input == null)
         {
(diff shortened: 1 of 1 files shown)


labkey-tchad (commented):
Does this actually work? This isn't really how clipboards work; copying with no selection doesn't generally clear the clipboard (maybe OS dependent?). You might need to set it to a single space (" ") to actually clear it. That's what my password manager does to avoid leaving password in the clipboard for too long.

labkey-chrisj:
> Does this actually work? This isn't really how clipboards work; copying with no selection doesn't generally clear the clipboard (maybe OS dependent?). You might need to set it to a single space (" ") to actually clear it. That's what my password manager does to avoid leaving password in the clipboard for too long.

Does it work?  Good question. At least on Windows, I can confirm that it does work.  With some text in my clipboard, running that method means whatever was in my clipboard isn't any more.

For better context, [PlateEditorTest.testCopyPasteColumnContent](https://teamcity.labkey.org/viewLog.html?buildId=3477385&buildTypeId=LabKey_Trunk_Premium_ProductSuites_Biologics_BiologicsWindows&fromSakuraUI=true#testNameId472877620687488123) is the failure I'm trying to address.
If I understand the failure correctly, it's failing because selecting empty grid cells to copy that empty value results in throwing an UnsupportedFlavorException.  The test here just wants to empty the clipboard before copying from the grid, so we know the copy worked.

labkey-tchad (approved):
Does this work on Windows on TeamCity? If not, consider adding a workaround similar to the one in `getClipboardContent`

labkey-chrisj:
> Does this work on Windows on TeamCity? If not, consider adding a workaround similar to the one in `getClipboardContent`

run queued to see: https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_ProductSuites_Biologics_BiologicsWindows?branch=chris_misc_tc_failures_fixes_4_22&buildTypeTab=overview

Result: [PlateEditorTest.testCopyPasteColumnContent](https://teamcity.labkey.org/viewLog.html?tab=queuedBuildOverviewTab&buildId=3482591&buildTypeId=LabKey_Trunk_Premium_ProductSuites_Biologics_BiologicsWindows&fromSakuraUI=true#testNameId472877620687488123) still fails, but because the copied content from the grid didn't include line breaks 🤔 

I don't see the logging info from `getClipboardContent` about available/best flavors in the logs, but the value is non-null. Not sure what's happening here.  Also kind of wondering if I should give up trying to copy from the grid on windows

PR #2425, opened 2025-04-23
EditableGridTest: Disable the test cases that drag fill multiline column values

labkey-alan:
#### Rationale
This PR disables the test cases in EditableGridTest that drag fill a column with a multiline text value. Our drag fill code now uses our paste handling code (in order to reduce complexity, and in order to address other issues with drag fill), but unfortunately we cannot handle pasting multiline text fields. See Issues [52226](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52226), [51927](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51927), which are slated to be fixed in 25.7.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1751

#### Changes
- EditableGridTest: Disable the test cases that drag fill multiline column values

### DIFF
--- a/src/org/labkey/test/tests/component/EditableGridTest.java
+++ b/src/org/labkey/test/tests/component/EditableGridTest.java
@@ -288,7 +288,8 @@ public class EditableGridTest extends BaseWebDriverTest
                 testGrid.getColumnData(DESC_STRING));
     }
 
-    @Test
+    // Test disabled until Issue 52226, Issue 51927 are resolved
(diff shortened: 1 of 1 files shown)


labkey-danield (approved):
There is also a @Ignore annotation as well. The @Ignore annotation will list the test if you run it locally so you see that it was ignored.
But this fix will achieve the desired result on TeamCity.

PR #2426, opened 2025-04-23
Get raw field labels from app grids and tables

labkey-tchad:
#### Rationale
[[Story](https://github.com/LabKey/kanban/issues/620)] (Phase 2)
The browser modifies whitespace when rendering text (removing duplicate spaces and trimming leading/trailing whitespace). As such, we can't trust it to accurately represent field labels. Currently, we use `WebElement.getText` to extract field labels from the page which might not match a field's actual label. We should be able to extract accurate labels from the page via the `testContext` property or by inspecting HTML text nodes. This will allow us to remove the multi-space restriction from `TestDataGenerator.randomName` and `FieldDefinition.labelFromName`.

#### Related Pull Requests
- 

#### Changes
- Get raw field labels from app grids and tables

### DIFF
--- a/src/org/labkey/test/components/ui/grids/EditableGrid.java
+++ b/src/org/labkey/test/components/ui/grids/EditableGrid.java
@@ -49,6 +49,7 @@ import static org.labkey.test.WebDriverWrapper.waitFor;
 import static org.labkey.test.util.TestLogger.log;
 import static org.labkey.test.util.selenium.ScrollUtils.Alignment.center;
 import static org.labkey.test.util.selenium.WebDriverUtils.MODIFIER_KEY;
+import static org.labkey.test.util.selenium.WebDriverUtils.getTextContent;
 
 public class EditableGrid extends WebDriverComponent<EditableGrid.ElementCache>
 {
@@ -1136,7 +1137,7 @@ public class EditableGrid extends WebDriverComponent<EditableGrid.ElementCache>
             {
                 for (WebElement el : headerCells)
                 {
-                    fieldLabels.add(el.getText().trim());
+                    fieldLabels.add(getTextContent(el));
                 }
 
                 int rowNumberColumn = 0;
--- a/src/org/labkey/test/components/ui/grids/EditableGrid.java
+++ b/src/org/labkey/test/components/ui/grids/EditableGrid.java
@@ -15,7 +15,7 @@ import org.labkey.test.components.ui.entities.EntityBulkInsertDialog;
 import org.labkey.test.components.ui.entities.EntityBulkUpdateDialog;
 import org.labkey.test.params.FieldDefinition;
 import org.labkey.test.util.selenium.ScrollUtils;
-import org.labkey.test.util.selenium.WebDriverUtils;
+import org.labkey.test.util.selenium.WebElementUtils;
 import org.openqa.selenium.By;
 import org.openqa.selenium.Keys;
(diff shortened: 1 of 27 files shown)


labkey-susanh (approved):

src/org/labkey/test/components/ui/grids/EditableGrid.java
```diff
@@ -1155,6 +1156,35 @@ public List<String> getColumnLabels()
...
+        private String getLabelFromHeaderCell(WebElement el)
+        {
+            // Use text nodes to ignore browser whitespace formatting
+            List<String> textNodes = WebElementUtils.getTextNodesWithin(el);
+            if (textNodes.isEmpty())
+            {
+                List<WebElement> children = Locator.xpath("./*").findElements(el);
+                if (children.isEmpty())
+                {
+                    return ""; // probably the selection checkbox column
```
labkey-danield, line 1172: I think the number of places where the editable grid has a checkbox column is few and far between.

labkey-danield (approved):

labkey-danield (changes requested):
I marked this as approved before realizing that there are several failures in LKSM and Biologics for this branch.
Can you get a cleaner run first?

labkey-tchad:
> I marked this as approved before realizing that there are several failures in LKSM and Biologics for this branch. Can you get a cleaner run first?

Yeah, I flipped the lims PR to draft when I noticed the failures. I made a temporary change that ensures there is a double-space in every random field name and it uncovered some places that still need updates.

src/org/labkey/test/util/DeferredErrorCollector.java
```diff
@@ -1,6 +1,8 @@
 package org.labkey.test.util;
 
 import org.apache.commons.lang3.StringUtils;
+import org.assertj.core.api.Assertions;
+import org.assertj.core.api.ObjectAssert;
```
labkey-danield: Did you mean to add these? It doesn't look like there are any references to these imports.

src/org/labkey/test/util/DeferredErrorCollector.java
```diff
@@ -15,6 +17,7 @@
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.function.Consumer;
```
labkey-danield: And this one as well.

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -587,7 +581,7 @@ public static String randomFieldName(String part)
 
     public static String randomFieldName(String part, @Nullable String exclusion)
     {
-        return randomFieldName(part, randomInt(0, 5), randomInt(0, 5), exclusion);
+        return randomFieldName(part, randomInt(0, 5), randomInt(1, 5), exclusion);
```
labkey-danield: Why not make the min value in both 1 or both 0?

labkey-danield (approved):

src/org/labkey/test/util/DeferredErrorCollector.java
labkey-tchad: Oops, these unused imports are remnants of some experimentation I was doing.

src/org/labkey/test/util/TestDataGenerator.java
labkey-tchad: A remnant of some additional validation I was doing. Reverting

PR #2452, opened 2025-05-16
clean up the right schema this time

labkey-chrisj:
#### Rationale
Recently, [LinkedSchemaTest.testCoreLinkedSchemaFilters](https://teamcity.labkey.org/buildConfiguration/bt20/3512466?buildTab=tests&status=failed&name=LinkedSchemaTest.testCoreLinkedSchemaFilters) has been failing because at the end of the test where it should be cleaning up the schema it created in the current test, it's cleaning up another one 🤦 

#### Related Pull Requests
n/a

#### Changes
clean up the schema created in this test case, not the other

### DIFF
--- a/src/org/labkey/test/tests/LinkedSchemaTest.java
+++ b/src/org/labkey/test/tests/LinkedSchemaTest.java
@@ -895,7 +895,7 @@ public class LinkedSchemaTest extends BaseWebDriverTest
                 .isGreaterThan(-1));
 
         // clean up after ourselves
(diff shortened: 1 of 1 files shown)


labkey-danield (approved):
Since this was failing on TeamCity it would be useful to add a pointer to a TC run showing this does fix the issue.

labkey-chrisj:
> Since this was failing on TeamCity it would be useful to add a pointer to a TC run showing this does fix the issue.

OK, here you go https://teamcity.labkey.org/buildConfiguration/bt20/3515579?buildTab=tests&name=org.labkey.test.tests.LinkedSchemaTest.testCoreLinkedSchemaFilters

PR #2453, opened 2025-05-19
Chris misc issues

labkey-chrisj:
#### Rationale
Covers test automation for Misc. Issues Needs Automation github issue https://github.com/LabKey/kanban/issues/640

#### Related Pull Requests
n/a

#### Changes
new tests


src/org/labkey/test/BaseWebDriverTest.java
```diff
@@ -1905,7 +1905,7 @@ public void setPipelineRoot(String rootPath, boolean inherit)
         {
             _setPipelineRoot(rootPath, inherit);
 
-            waitForElement(Locators.labkeyMessage.withText("The pipeline root was set to '" + Paths.get(rootPath).normalize() + "'"));
+            waitForElement(Locators.labkeyMessage.withTextMatching("The pipeline root was set to '" + Paths.get(rootPath).normalize() + "'"));
```
labkey-chrisj: This makes the matching for the pipeline root path case-insensitive; this is helpful for reproducing locally where for some reason the test expects some camel-cased path components but they're all lower-case

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
```diff
@@ -193,6 +194,30 @@ public T filterBooleanColumn(String fieldLabel, boolean value)
...
 
+    /**
+     *
+     * @param index
+     * @param fieldLabel
+     * @param operator
+     * @param dateString1
+     * @param dateString2
+     * @return
+     */
```
labkey-chrisj: The repro for  Issue 52661 called for configuring a date filter with an invalid date, which called for the ability to pass in date strings to the filter dialog- this change makes it possible for tests to do that (to filter on date, the test must pass in a date object but I couldn't figure a way to configure an invalid date object)

src/org/labkey/test/components/ui/search/FilterExpressionPanel.java
```diff
@@ -117,7 +117,7 @@ else if (value2 != null)
         }
     }
```
labkey-chrisj: I'm not usually a fan of making private methods public, I did this to allow` ResponsiveGrid.filterDateColumn` to set the operator of its expression

src/org/labkey/test/tests/SampleTypeTest.java
```diff
@@ -137,6 +138,75 @@ protected void doCleanup(boolean afterTest)
...
+                .isEqualTo(dateTime));
+        checker().screenShotIfNewError("unexpected data update");
+
+        // fill in the required text field and submit
+        updatePage.setField(txtField.getName(), "sample01");
+        updatePage.submit();
+
+        var afterSubmit = DataRegionTable.DataRegion(getDriver()).withName("Material").waitFor();
+        var rowData = afterSubmit.getRowDataAsText(0);
+        checker().withScreenshot("unexpected data persisted")
```
labkey-chrisj, line 198: At the moment, milliseconds are being truncated from time again- I've reactivated the issue and set its needs automation status to 'done'

src/org/labkey/test/tests/assay/AssayTransformImportUpdateTest.java
```diff
@@ -219,6 +220,74 @@ public void testEnableTransformForUpdate() throws Exception
...
+        String transformContent = """
+                library(Rlabkey);
+                
+                run.props = labkey.transform.readRunPropertiesFile("${runInfo}");
+                
+                run.data.file = labkey.transform.getRunPropertyValue(run.props, "runDataFile");
+                run.output.file = run.props$val3[run.props$name == "runDataFile"];
+                error.file = labkey.transform.getRunPropertyValue(run.props, "errorsFile");
+                
+                # sleep a bit before writing the table, give the test time to cancel the job before it is complete
```
labkey-chrisj, line 238: @josheckels This part of the transform script attempts to write what it's doing to the console, but the pipeline log does not appear to capture any of it.  I suspect there's a way to do what I'd like to do here but this doesn't seem to be it.  I'm not clear at this point that this coverage actually validates the issue.
The existing log does reflect a 4-second time window between cancelling the Job and the Error reflecting its cancellation so I think it's pretty likely we're actually cancelling the job

src/org/labkey/test/tests/component/GridPanelViewTest.java
```diff
@@ -1260,6 +1263,48 @@ public void testRemoveAllFields()
...
+        grid.filterDateColumn(0, "Created", Filter.Operator.GT, "May 27, 2024", null);
+        var filterStatusValueTexts = grid.getFilterStatusValuesText();
+        checker().withScreenshot("Filter_Texts_Error")
+                .wrapAssertion(()-> Assertions.assertThat(filterStatusValueTexts)
+                        .hasSize(1)
+                        .containsExactly("Created > 2024-05-27"));
+        grid.clearFilters();
+
+        log("try to filter on an invalid date");
+        grid.filterDateColumn(0, "Created", Filter.Operator.GT, "05/37/2024", null);
```
labkey-chrisj: I wanted to verify some expected parsed date here, but the behavior is wildly wrong (for example, instead of 5/3/2024 it will parse this date to a different year) and I think that's bug-worthy but the upside is that it won't let itself be configured with an invalid date

src/org/labkey/test/tests/query/QueryLookupTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+        // create a list
+        var dgen = new VarListDefinition(LIST_NAME)
+                .setFields(List.of(NAME_COLUMN, TSHIRT_COLUMN))
+                .create(createDefaultConnection(), PROJECT_NAME);
+        dgen.withGeneratedRows(10)
+                .insertRows();
+    }
+
+    // Issue 49511 Setting lookup to custom query shows "Error: Lookup target table does not exist."
```
labkey-chrisj, line 63: It's possible this coverage belongs somewhere else- DomainDesignerTest.testInvalidLookupDomainField contains related testing- if you think I should put it in DomainDesignerTest the refactor should be simple

src/org/labkey/test/tests/component/GridPanelViewTest.java
```diff
@@ -1260,6 +1263,48 @@ public void testRemoveAllFields()
 
     }
 
+    @Test
```
labkey-chrisj, line 1270: This coverage is on QueryGrid views, if you think it would be helpful to do the same for DataRegion views I'm happy to add

labkey-chrisj (commented):
Some comments to make this PR more readable

src/org/labkey/test/BaseWebDriverTest.java
labkey-tchad: Did you try this?
`Locator.withTextMatching` doesn't actually work; I've been meaning to remove it. Selenium uses XPath 1.0 but the `matches` function was added in v2. Selenium throws an `InvalidSelectorException` when I try to use it.
Even if this locator did work, it should have any impact on case-sensitivity. That's not how regex works.

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
```diff
@@ -193,6 +194,30 @@ public T filterBooleanColumn(String fieldLabel, boolean value)
...
+        doAndWaitForUpdate(() -> {
+            clickColumnMenuItem(fieldLabel, "Filter...", false);
+            GridFilterModal filterModal = new GridFilterModal(getDriver(), this);
+            filterModal.selectExpressionTab().setFilterType(index, operator);
+            Input.Input(Locator.input("field-value-date-0"), getDriver()).waitFor().set(dateString1);
+            if (dateString2 != null)
+                Input.Input(Locator.input("field-value-date-1"), getDriver()).waitFor().set(dateString2);
+            filterModal.confirm();
+        });
+        return _this;
```
labkey-tchad: Why do these filter methods stash `_this` and return it later? Seems unnecessary.
```suggestion
        return getThis();
```

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
labkey-tchad: I have an idea for an alternative to pulling this functionality out of the filter dialog.
Since `FilterExpressionPanel.setFilter` already takes `Object`s, we can create a wrapper class that it knows to treat like a date. It could even extend `Date` so that the existing `instanceof` checks would go to the date handling branch of `setFilter`; it would format actual Date objects as it currently does but would just pull the provided date string from the new class.
It would be an internal class to `FilterExpressionPanel` or `GridFilterModal`; something super simple, like this:
```
public static class DateString extends Date
{
    private final String _dateStr;
    public DateString(String dateStr)
    {
        _dateStr = dateStr;
    }
}
```

Then you can just use the normal filter methods:
```
grid.filterColumn("Created", Filter.Operator.GT, new DateString("05/27/2024"));
```

src/org/labkey/test/tests/AdminConsoleTest.java
```diff
@@ -151,6 +154,73 @@ public void testRibbonBar()
         assertElementNotPresent(ribbonLink);
     }
 
+    // Issue 51843  allow site banner configuration via api
+    @Test
+    public void testSiteBannerAPIConfiguration() throws Exception
+    {
+        goToAdminConsole();
+
+        String bannerMessage = "test banner message" + TRICKY_CHARACTERS;
```
labkey-tchad, line 164: Consider using the injection test string here as well.
```suggestion
        String bannerMessage = "test banner message" + TRICKY_CHARACTERS + INJECT_CHARS_1;
```

src/org/labkey/test/tests/AdminConsoleTest.java
```diff
@@ -151,6 +154,73 @@ public void testRibbonBar()
...
+        refresh();
+        // verify it remains shown
+        checker().withScreenshot("banner is hidden when not expected")
+                .verifyTrue("expect banner to be shown", bannerLoc.isDisplayed(getDriver()));
+
+        stopImpersonating();
+
+        // clean up after ourselves as site admin
+        var clearAndHideBannerCmd = new SimplePostCommand("admin", "setRibbonMessage.api");
+        clearAndHideBannerCmd.setParameters(Map.of("show", false, "message", ""));
```
labkey-tchad, line 222: What happens if you set show=true with a blank message?
Mostly, I'm just curious, but it might be worth testing.

src/org/labkey/test/tests/MenuBarTest.java
```diff
@@ -211,6 +225,49 @@ public void testSteps()
...
+                        .contains("labkey.org/Documentation"));
+        switchToMainWindow();
+        closeExtraWindows();
+
+        goToProjectHome();
+        clickUserMenuItem(true,"Support");
+        checker().withScreenshot("unexpected_destination")
+                .wrapAssertion(()-> Assertions.assertThat(getURL().toString())
+                        .as("expect navigation to localhost/home/support")
+                        .contains("localhost", "/home/support"));
```
labkey-tchad: Dev instances aren't necessarily `localhost`.
```suggestion
        checker().withScreenshot("unexpected_destination")
                .verifyEquals("Support menu destination",
                        WebTestHelper.buildUrl("project", "home/support", "begin"),
                        getDriver().getCurrentUrl());
```

src/org/labkey/test/tests/SampleTypeTest.java
```diff
@@ -137,6 +138,75 @@ protected void doCleanup(boolean afterTest)
         _userHelper.deleteUsers(false, USER_FOR_FILTERTEST.getEmail());
     }
 
+    // Issue 52390: milliseconds are truncated from time fields on update or reshow
+    @Test
+    public void testDateAndTimeValueUpdates() throws Exception
+    {
```
labkey-tchad, line 144: This test doesn't really have anything to do with samples in particular. Why add it to `SampleTypeTest`?
Maybe `NonStandardDateAndTimeFormatTest`?

src/org/labkey/test/tests/SampleTypeTest.java
```diff
@@ -137,6 +138,75 @@ protected void doCleanup(boolean afterTest)
...
+
+        SampleTypeDefinition sampleTypeDefinition = new SampleTypeDefinition(sampleTypeName).setFields(fields);
+        sampleTypeDefinition.create(createDefaultConnection(), getProjectName());
+
+        goToProjectHome();
+        waitAndClickAndWait(Locator.linkWithText(sampleTypeName));
+        var dataRegion = DataRegionTable.DataRegion(getDriver()).withName("Material").waitFor();
+        var updatePage = dataRegion.clickInsertNewRow();
+        String date = "January 01 2025";
+        String time = "23:56:54:123";
```
labkey-tchad: Are milliseconds missing below because this is the wrong format?
```suggestion
        String time = "23:56:54.123";
```

src/org/labkey/test/tests/component/GridPanelViewTest.java
```diff
@@ -1,5 +1,7 @@
 package org.labkey.test.tests.component;
 
+import org.aspectj.org.eclipse.jdt.core.dom.InfixExpression;
```
labkey-tchad: Accidental import?
```suggestion
```

labkey-tchad (changes requested):

src/org/labkey/test/BaseWebDriverTest.java
labkey-chrisj: Yes, I did try it- and somehow it worked for me locally.
If you'd prefer, I can revert this and maybe one day we sort out a better way to match text case-insensitively in xpath

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
labkey-chrisj: ooo, that's filthy.  I like it and will try it

src/org/labkey/test/tests/AdminConsoleTest.java
labkey-chrisj: If I recall correctly, it shows an empty banner.  Not guaranteeing that's the case tho.  I'll look

src/org/labkey/test/tests/component/GridPanelViewTest.java
labkey-chrisj: Yes, thanks

src/org/labkey/test/tests/SampleTypeTest.java
labkey-chrisj: I looked there, just wasn't sure how changing settings like project settings for date and time would interact with other tests in the class.  I'll try
Yep, adding this test to that class violates most of the assumptions the other tests make about the date or time formats that should be there ☹️

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
labkey-jeckels: This JavaDoc doesn't add any value. Please either fill in some content or delete it.

src/org/labkey/test/tests/assay/AssayTransformImportUpdateTest.java
labkey-jeckels: It's a 4 second gap in the log and a 4 second sleep in the script? I'd have more confidence if the script slept for, say, 30 seconds.

Also, have you confirmed that this script runs successfully without the cancel attempt?

src/org/labkey/test/tests/component/GridPanelViewTest.java
labkey-jeckels: A different year? In quick testing on the issues list, `05/37/2024` resolves to `2024-06-06`. It must be parsing in a lenient way that "wraps" the 37th day of May across the month boundary into June. I suspect we've been running this way for a long time but haven't dug into it. Feel free to log an issue.

src/org/labkey/test/tests/SampleTypeTest.java
labkey-chrisj: Yes, it is. Thank you!
(also, I'm a little mad that I missed that)

src/org/labkey/test/tests/SampleTypeTest.java
labkey-chrisj: Nope, that was a test error. 😮‍💨 
I've re-closed the issue, it's still fixed

src/org/labkey/test/tests/assay/AssayTransformImportUpdateTest.java
labkey-chrisj: I've confirmed that it writes values into the TransformType and M2 columns for every record being inserted, and when I bump the sleep to 30 seconds and don't cancel the job it finishes successfully in... about 31 seconds.
image

When I set the sleep to 30 seconds and do cancel the job, its status shows as cancelling for ~30 seconds but the log says the task was cancelled and the run doesn't import.  This tells me cancelling doesn't actually stop the process, it just doesn't import the run at the end of it.
image

It also tells me that` Sys.sleep(seconds)` works but `print() `and `labkey.setDebugMode `don't do what I was hoping they would. In RStudio, they happily wrote to the console for me but I'm not sure if there isn't some secret configuration sauce in Labkey to make that happen 🤔 
image

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
labkey-chrisj: Ah, I tried doing that and although it worked well enough to parse the value out of the `DateString `object, unfortunately it uses a static `DATE_FORMAT `constant and it's got a fixed pattern of `"yyyy-MM-dd H:m"`.

This defeats the attempt to pass in arbitrary date strings for test purposes 😞

labkey-tchad: You don't need to parse or format anything. Just take the string value as it is and skip the `DATE_FORMAT`.

src/org/labkey/test/tests/SampleTypeTest.java
labkey-tchad: Bummer. I figured it would assume that the test cases were messing with the format settings and get the project back to some known state between them.
Could you move this to that class but have it work in a subfolder?

labkey-chrisj: I tried that as well, but couldn't figure a way to set the date/time formats just in the folder- it looks like we do that at the project level only. I could probably add another project to that test if you think that's worthwhile

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
labkey-chrisj: done

src/org/labkey/test/components/ui/search/FilterExpressionPanel.java
labkey-chrisj: reverted, no longer necessary

src/org/labkey/test/components/ui/search/FilterExpressionPanel.java
```diff
@@ -92,11 +92,19 @@ public void setFilter(int index, Object operator, Object value1, Object value2)
             }
             else if (value1 instanceof Date dateVal1)
             {
-                final String dateStr1 = DATE_FORMAT.format(dateVal1);
+                String dateStr1;
+                if (value1 instanceof DateString)
+                    dateStr1 = ((DateString)value1).getDateStr();   // use the test-supplied date string
```
labkey-tchad: New Java syntax for `instanceof` to avoid casting
```suggestion
                if (value1 instanceof DateString ds1)
                    dateStr1 = ds1.getDateStr();   // use the test-supplied date string
```

src/org/labkey/test/components/ui/search/FilterExpressionPanel.java
```diff
@@ -92,11 +92,19 @@ public void setFilter(int index, Object operator, Object value1, Object value2)
...
+                else
+                    dateStr1 = DATE_FORMAT.format(dateVal1);        // format the supplied date value to string
                 elementCache().dateValues.get(index).set(dateStr1);
                 if (value2 instanceof Date dateVal2)
                 {
-                    final String dateStr2 = DATE_FORMAT.format(dateVal2);
+                    String dateStr2;
+                    if (value2 instanceof DateString)
+                        dateStr2 = ((DateString)value2).getDateStr();
```
labkey-tchad:
```suggestion
                    if (value2 instanceof DateString ds2)
                        dateStr2 = ds2.getDateStr();
```

src/org/labkey/test/tests/query/QueryLookupTest.java
```diff
@@ -0,0 +1,148 @@
...
+@Category({Daily.class})
+public class QueryLookupTest extends BaseWebDriverTest
+{
+    private static final String PROJECT_NAME = "QueryLookupTest" + TRICKY_CHARACTERS_FOR_PROJECT_NAMES;
+    private static final String LIST_NAME = "l&ist q";
+
+    private static final FieldDefinition NAME_COLUMN =
+            new FieldDefinition("Name", FieldDefinition.ColumnType.String);
+    private static final FieldDefinition TSHIRT_COLUMN =
+            new FieldDefinition("TShirt", FieldDefinition.ColumnType.String);
```
labkey-tchad: `FieldDefinition` is not immutable, so isn't really appropriate to define as a constant. Consider using the shiny new `FieldInfo`.
```suggestion
    private static final FieldInfo NAME_COLUMN =
            new FieldInfo("Name", FieldDefinition.ColumnType.String);
    private static final FieldInfo TSHIRT_COLUMN =
            new FieldInfo("TShirt", FieldDefinition.ColumnType.String);
```
Then you can use `getFieldDefinition()` to create the list (e.g. `NAME_COLUMN.getFieldDefinition()`).

labkey-tchad (approved):

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
```diff
@@ -12,6 +12,7 @@
 import org.labkey.test.components.Component;
 import org.labkey.test.components.UpdatingComponent;
 import org.labkey.test.components.WebDriverComponent;
+import org.labkey.test.components.html.Input;
```
labkey-danield: Unused import should be removed.

labkey-danield (changes requested):
Can you point to a TC run?
The banner test is failing for me locally.

src/org/labkey/test/tests/AdminConsoleTest.java
```diff
@@ -151,6 +154,73 @@ public void testRibbonBar()
...
+        //As site admin
+        // set the message and show it
+        var showBannerCmd = new SimplePostCommand("admin", "setRibbonMessage.api");
+        showBannerCmd.setParameters(Map.of("message", bannerMessage,
+                "show", true));
+        showBannerCmd.execute(createDefaultConnection(), "/");
+        refresh();
+        // verify it is shown
+        checker().withScreenshot("banner not shown or not as expected")
+                .verifyTrue("expect banner to be shown", bannerLoc.isDisplayed(getDriver()));
```
labkey-danield: Are we sure the banner will show up immediately after a refresh? Would changing this to use a waitFor the banner to be shown be more reliable?

src/org/labkey/test/tests/AdminConsoleTest.java
```diff
@@ -151,6 +154,73 @@ public void testRibbonBar()
...
+
+        // hide the banner
+        var hideBannerCmd = new SimplePostCommand("admin", "setRibbonMessage.api");
+        hideBannerCmd.setParameters(Map.of("show", false));
+        hideBannerCmd.execute(createDefaultConnection(), "/");
+        refresh();
+        // verify it is hidden
+        checker().withScreenshot("banner is shown when not expected")
+                .verifyFalse("expect banner not to be shown", bannerLoc.isDisplayed(getDriver()));
```
labkey-danield: This test is conditional on the previous call to show the banner, if that fails then there is no value with running this check. Maybe consider putting the previous call the checker as a conditional in a `if` check

src/org/labkey/test/tests/SampleTypeTest.java
```diff
@@ -137,6 +138,75 @@ protected void doCleanup(boolean afterTest)
...
+        updatePage.submitExpectingError();
+
+        checker().wrapAssertion(()-> Assertions.assertThat(updatePage.getTextInputValue(dateField.getName()))
+                .isEqualTo("2025-01-01")); // expect reformat of date
+        checker().wrapAssertion(()-> Assertions.assertThat(updatePage.getTextInputValue(timeField.getName()))
+                .as("expect time to retain milliseconds")
+                .isEqualTo(time));
+        checker().wrapAssertion(()-> Assertions.assertThat(updatePage.getTextInputValue(dateTimeField.getName()))
+                .as("expect dateTime to post back as entered")
+                .isEqualTo("1986-05-06 23:58:34.123"));
```
labkey-danield, line 189: I'm unclear why you are using `wrapAssertion`. `checker().verifyEquals` should validate the strings are equal.

src/org/labkey/test/tests/SampleTypeTest.java
labkey-danield: What about putting it in the DomainDesignerTest?

src/org/labkey/test/tests/component/GridPanelViewTest.java
labkey-danield: Adam added this comment to the issue: "All QueryViews should now have a non-null Errors", I would interpret that as saying testing only QueryGrids is ok.

src/org/labkey/test/tests/component/GridPanelViewTest.java
```diff
@@ -1261,6 +1263,50 @@ public void testRemoveAllFields()
...
+                        .filteredOn(a-> a.getText().equals("Created > 2024-05-27"))
+                                .isNotEmpty());
+        grid.clearFilters();
+
+        log("try to filter on an invalid date");
+        grid.filterColumn("Created", Filter.Operator.GT, new FilterExpressionPanel.DateString("05/37/2024"));
+        // don't expect the parser to get the invalid date right; current behavior won't do that
+
+        log("ensure the view can be edited after");
+        grid.filterColumn("Created", Filter.Operator.LTE, new Date());  // filter on right now
```
labkey-danield: I don't think this is the repro steps in the issue. The issue says:

- Create a new grid view with a valid date filter (e.g., Created > 3/23/25 or whatever)
- Save it as "Broken View"
- Customize "Broken View"
- Change filter to use an invalid date "3/32/25"
- **Re-save**

You need to re-save the grid view after creating the invalid filter.

It may also be worthwhile trying an invalid filter other than a date. This sample type has an Int field, you could try saving a filter of EQ "ABC".

src/org/labkey/test/tests/query/QueryLookupTest.java
```diff
@@ -0,0 +1,148 @@
...
+        goToSchemaBrowser();
+        createQuery(getProjectName(), queryName, "lists", querySql, queryXml, false);
+
+        // now create another list, with a lookup to the custom query
+        new IntListDefinition(secondList, "Key")
+                .setFields(List.of(NAME_COLUMN,
+                        new FieldDefinition("lookup", new FieldDefinition.StringLookup(getProjectName(), "lists", queryName))))
+                .create(createDefaultConnection(), PROJECT_NAME);
+
+        // insert data into the list
```
labkey-danield, line 100: I don't know that you need to add data to the list to validate the issue, but it doesn't hurt anything having it.

labkey-danield (changes requested):
The `AdminConsoleTest.testSiteBannerAPIConfiguration` is failing because the banner is not showing the `INJECT_CHARS_1`, which is probably by design. If so the test needs to check for the expected message.

The `GridPanelViewTest.testWarningOnInvalidDateFilter` is missing a step necessary for the repro.

labkey-jeckels:
> The `AdminConsoleTest.testSiteBannerAPIConfiguration` is failing because the banner is not showing the `INJECT_CHARS_1`, which is probably by design. If so the test needs to check for the expected message.
> 
> The `GridPanelViewTest.testWarningOnInvalidDateFilter` is missing a step necessary for the repro.

The site banner accepts HTML (hence the "Message HTML" label for the field Site Settings) so it's not a good candidate for `INJECT_CHARS_1`.

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
labkey-chrisj: it's gone

src/org/labkey/test/BaseWebDriverTest.java
labkey-chrisj: Looks like my change caused a TC failure, I'm reverting it

labkey-chrisj:
Results with success here:
MenuBarTest https://teamcity.labkey.org/viewLog.html?buildId=3530913&buildTypeId=LabkeyTrunk_DailyPostgres&tab=testsInfo&branch_LabkeyTrunk_DailySuites=chris_misc_issues
SampleTypeTest https://teamcity.labkey.org/viewLog.html?buildId=3530913&buildTypeId=LabkeyTrunk_DailyPostgres&tab=testsInfo&branch_LabkeyTrunk_DailySuites=chris_misc_issues
GridPanelViewTest https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyCPostgres/3532846?buildTab=tests&name=GridPanelViewTest
QueryLookupTest https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailySuites_DailyDPostgres/3530917?buildTab=tests&name=QueryLookupTest
WikiTest https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailySuites_DailyEPostgres/3530919?buildTab=tests&name=wikitest
AdminConsoleTest https://teamcity.labkey.org/buildConfiguration/bt20/3534423?buildTab=tests&name=adminCon
AssayTransformImportUpdateTest https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyCPostgres/3532846?buildTab=tests&name=AssayTransformI

labkey-danield (approved):

PR #2455, opened 2025-05-19
Issue 52098, 49422: when looking up by alternate keys do that first so number names will resolve well

labkey-susanh:
#### Rationale
Issue [52098](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52098): Sample statuses with number labels may not resolve to the correct status
Issue [49422](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49422): List lookups with number-like values may not resolve to the correct status

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6642
- https://github.com/LabKey/limsModules/pull/1416

#### Changes
- Add test case for Issue 49422
- Convert a test field to a TextChoice field


labkey-danield (approved):

src/org/labkey/test/tests/list/ListTest.java
```diff
@@ -1812,6 +1812,210 @@ private void testTricky(String listName, boolean autoKey) throws IOException
...
+        checker().withScreenshot().verifyEquals("Error message for invalid number-like alternate key not as expected", "Value '1234' was not present in lookup target 'lists." + lookupListName + "' for field '" + baseLookupFieldName + "'", error);
+
+    }
+
+    private void resetList()
+    {
+        DataRegionTable dataRegionTable = new DataRegionTable("query", getDriver());
+        dataRegionTable.checkAllOnPage();
+        dataRegionTable.deleteSelectedRows();
+    }
```
labkey-tchad: Consider resetting list via api
```suggestion
    private void truncateList(String listName) throws IOException, CommandException
    {
        new QueryApiHelper(createDefaultConnection(), getProjectName(), "lists", listName).truncateTable();
    }
```

src/org/labkey/test/tests/list/ListTest.java
```diff
@@ -1812,6 +1812,210 @@ private void testTricky(String listName, boolean autoKey) throws IOException
...
+        String baseListName = TestDataGenerator.randomDomainName("base");
+        String baseKeyFieldName = TestDataGenerator.randomFieldName("base Key Field");
+        String baseLookupFieldName = TestDataGenerator.randomFieldName("base Lookup Field");
+        String baseLookupFieldKey = EscapeUtil.fieldKeyEncodePart(baseLookupFieldName);
+        _listHelper.createList(PROJECT_VERIFY, baseListName, baseKeyFieldName);
+        EditListDefinitionPage listDefinitionPage = _listHelper.goToEditDesign(baseListName);
+        listDefinitionPage.getFieldsPanel()
+                .addField(baseLookupFieldName)
+                .setType(ColumnType.Lookup)
+                .setLookup(new FieldDefinition.IntLookup("lists", lookupListName));
```
labkey-tchad: Regarding your question about `ColumnType.Lookup` being deprecated. `setType(Lookup)` isn't actually necessary; `setLookup` will set the type correctly.
```suggestion
                .addField(baseLookupFieldName)
                .setLookup(new FieldDefinition.IntLookup("lists", lookupListName));
```
Alternatively, you can pass a `FieldDefinition` to `addField`.

labkey-tchad (approved):
I would consider making a new test class for this and breaking it into multiple test cases. It doesn't seem to take advantage of the users or lists created by `ListTest`.

src/org/labkey/test/tests/list/ListTest.java
labkey-susanh: Thanks. Done.

src/org/labkey/test/tests/list/ListTest.java
labkey-susanh: Done.

labkey-susanh (commented):
Good idea to move to its own class. Done.

PR #2478, opened 2025-06-06
TargetedMSSampleManagerIntegrationTest update to check for LKSM and LIMS product configuration cases

cnathe:
#### Rationale
We recently made a change to how we determine the primary application Id for a container/project. This means that if a distribution has the LKSM and LIMS products registered and the container has the sampleManager module enabled but does not have the folder type set as LIMS or LKSM, the container will default to the "highest" product based on the site product configuration.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6726
- https://github.com/LabKey/targetedms/pull/1079
- https://github.com/LabKey/limsModules/pull/1454
- https://github.com/LabKey/testAutomation/pull/2478

#### Changes
- Move ProductKey and setProductConfiguration helpers to BaseWebDriverTest from SMBaseTest


labkey-tchad (approved):
These should be added to `LabKeySiteWrapper` (or a separate utility/helper class) instead. `BaseWebDriverTest` should* just contain stuff relevant to controlling test flow and failure handling.
(* we still have work to do in this regard)

src/org/labkey/test/WebDriverWrapper.java
```diff
@@ -769,20 +769,14 @@ public List<String> getFormAddresses()
...
-    public String getCurrentRelativeURL()
-    {
-        return getCurrentRelativeURL(true);
-    }
-
     /**
      * Get the relative URL for the current page. The host, port, and context path are removed.
      * Trailing '?' will be removed (optionally) to make comparisons more consistent when enabling the 'noQuestionMarkUrl' experimental feature.
      * Leaving the trailing question mark can be useful when comparing to a URL generated by 'URLBuilder' because it
      * knows about the experimental feature.
```
labkey-tchad: Remove this comment about the non-existent `noQuestionMarkUrl` feature.
```suggestion
```

labkey-tchad (approved):

src/org/labkey/test/WebDriverWrapper.java
cnathe: good catch, I missed that. done

PR #2479, opened 2025-06-08
Minor auto-refactor code cleanup on a massive scale

labkey-jeckels:
#### Rationale
We can cleanup tens of thousands of warnings across our codebase with IntelliJ's autorefactors. In some cases, this adopts newer, leaner syntax. In others, it simply removes code that's not serving any purpose.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6742

#### Changes
- Auto-refactors
  - Eliminate redundant toString()
  - Eliminate redundant cast
  - Empty JavaDoc annotations
  - Member variable can be final
  - length() and size() -> isEmpty()
  - Add missing @Override
  - Class can be static
  - Remove unused imports
  - toArray() passed array length
  - Remove unnecessary semicolon
  - Explicit type syntax can be replaced by <>
  - Redundant access modifiers
  - Remove redundant initializer
  - "".equals() -> isEmpty()
  - StringBuilder can be replaced by String
  - Fix misordered arguments to assertEquals()
  - StringUtils.defaultString() - Objects.toString()
  - Cast can be replaced with pattern variable
  - Collapse identical catch blocks
  - Type may be primitive
- Manual fixups
  - Delete unused GWT code
  - Improve generics
  - new UnexpectedException() -> UnexpectedException.wrap()


labkey-tchad (approved):
Looks good. I fixed some incorrectly ordered arguments in lookup constructors and removed some unnecessary overrides to `elementCache`.

src/org/labkey/test/tests/SampleTypeFolderExportImportTest.java
```diff
@@ -150,13 +150,12 @@ protected boolean areDataListEqual(List<Map<String, String>> list01, List<Map<St
...
-                    errorMsg.append("\nlist01(" + i + "): " + list01.get(i));
-                    errorMsg.append("\nlist02(" + i + "): " + list02.get(i));
-                    errorMsg.append("\n*************** ERROR ***************");
-                    log(errorMsg.toString());
+                    String errorMsg = "\n*************** ERROR ***************" +
+                            "\nFound a mismatch in the lists." +
+                            "\nlist01(" + i + "): " + list01.get(i) +
+                            "\nlist02(" + i + "): " + list02.get(i) +
+                            "\n*************** ERROR ***************";
+                    log(errorMsg);
```
labkey-danield, line 158: Not a suggestion for a change, just a note.
This "style" of error recording was done before the `checker()` was implemented. This, and other code like this, should be replaced with calls to `checker()`.

labkey-danield (approved):

PR #2481, opened 2025-06-10
Test component audit event row count and diff count checks for Sample and Source delete and move

cnathe:
#### Rationale
Expanding on recent AuditLogHelper checks for audit event row/diff counts on actions like sample/source bulk update and details update, this PR adds similar checks within the test components for deleting and moving samples and sources.

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/2481
- https://github.com/LabKey/platform/pull/6745
- https://github.com/LabKey/limsModules/pull/1461

#### Changes
* `ModalDialog`: Added a new method `getCountFromTitle(String prefix)` to extract a numeric count from the dialog's title based on a given prefix. This method handles potential parsing errors gracefully by returning `null` if the number cannot be parsed.
* `DeleteConfirmationDialog`: Modified the `confirmDelete()` method to include an overloaded version that accepts a `skipAuditEventCheck` parameter, allowing optional verification of audit events during delete confirmation.
* `DeleteConfirmationDialog`: Added a static helper method `verifyAuditEvents()` to validate audit event counts for the last transaction, ensuring consistency between the expected and actual audit log entries. This method handles exceptions and throws a runtime error if validation fails.


src/org/labkey/test/components/bootstrap/ModalDialog.java
```diff
@@ -84,6 +84,26 @@ public String getBodyText()
...
+                title = title.replaceFirst(prefix, "").trim();
+                String countString = title.contains(" ") ? title.substring(0, title.indexOf(' ')) : title;
+                return Integer.parseInt(countString);
+            }
+            catch (NumberFormatException e)
+            {
+                // If we can't parse the number, return null
+                return null;
+            }
+        }
```
labkey-tchad: Rather than specifying a prefix, could you just use regex to grab whatever number it can find in the dialog title? I can see how that might not be precise enough but it would simplify usage a bit.
```
Pattern pattern = Pattern.compile("(\\d+)");
Matcher matcher = pattern.matcher(title);
if (matcher.find())
{
    return matcher.group(1).trim();
}
```

labkey-danield (approved):

src/org/labkey/test/components/ui/DeleteConfirmationDialog.java
```diff
@@ -75,12 +79,38 @@ public void cancelDelete()
...
+    }
+
+    public ConfirmPage confirmDelete(boolean skipAuditEventCheck, Integer waitSeconds)
+    {
+        Integer count = getCountFromTitle("Permanently Delete");
+        String auditEventName = new AuditLogHelper(getWrapper()).getAuditEventNameFromURL();
+
+        var confirmPage = _confirmationSynchronizationFunction.apply(() -> this.dismiss("Yes, Delete", waitSeconds));
+
+        if (!skipAuditEventCheck && count != null && auditEventName != null)
```
labkey-tchad: I'm a little worried about unknowingly losing this coverage if our count or eventName calculations start failing at some point.
We should be strict when we're know we can be. Maybe make an `EntityDeleteConfirmationDialog` subclass that throws an error if `getCountFromTitle` or `getAuditEventNameFromURL` are `null`. Then switch as many places as possible to use that instead.

src/org/labkey/test/components/bootstrap/ModalDialog.java
cnathe: I like that better. Thanks. Done

src/org/labkey/test/components/ui/DeleteConfirmationDialog.java
cnathe: I agreed. I had similar thoughts. I was hoping to put the verifyAuditEvents checks into the SampleDeleteDialog and SourceDeleteDialog for this reason but there are so many usages of DeleteConfirmationDialog directly that don't use those subclasses. Maybe we can sync to discuss this part.

cnathe: As a starting point, I added a null check to the SampleDeleteDialog and SourceDeleteDialog with this commit:
https://github.com/LabKey/limsModules/pull/1461/commits/9a0712d69e990ecb8e688e86b573d6f5d022e581

PR #2482, opened 2025-06-10
Add check to ensure the R process is ended successfully

labkey-chrisj:
#### Rationale
This adds verification to `AssayTransformImportUpdateTest.testCancelAsyncAssayTransformJob` to ensure that canceling the job does in fact end it; during earlier testing we found that in some cases child processes of the initial R process could persist- Josh made changes in https://github.com/LabKey/platform/pull/6740 that terminate these child processes appropriately.

Because this test case also falls afoul of [Issue 53240](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53240) when run on SQL server, I've added an `Assume `call to prevent it from running on SQL until this issue can be sorted out appropriately.  I've also marked up the issue to reflect this, to give future-us a cue to update this test when resolving it.

I've attempted to cancel the job from the pipeline list in the past, but I didn't see 'cancel' as an option (either it wasn't there, or I didn't see it).  I'll try to do that again as a way to work around Issue 53240 that doesn't require not doing the test on SQL

#### Related Pull Requests
https://github.com/LabKey/platform/pull/6740

#### Changes
Verify the duration between launch and done is less than how long it would take to complete unterminated


labkey-tchad (approved):

labkey-chrisj:
For context, this is what I'm seeing locally when I run this test on Windows/SQL up to the point that I click 'save and finish' for the assay import: no cancel option in the data pipeline list: 
image

I did the same check on PG, with exactly the same result: no option to cancel.
This could be info relevant to Issue 53240? I'll add that to the issue.

src/org/labkey/test/tests/assay/AssayTransformImportUpdateTest.java
```diff
@@ -224,6 +229,9 @@ public void testEnableTransformForUpdate() throws Exception
     @Test
     public void testCancelAsyncAssayTransformJob() throws Exception
     {
+        Assume.assumeTrue("Issue 53240: User cannot cancel pipeline job on SQL",
+                WebTestHelper.getDatabaseType() != WebTestHelper.DatabaseType.MicrosoftSQLServer);
+
```
labkey-danield, line 234: I don't understand the value of this code. If this test fails in MSSQL you are just trading one exception for another. This adds better logging but not much else. It will still add to the failure noise in a MSSQL run.

I would suggest, either:
1. Remove this and let the test fail in MSSQL and add the comment in TeamCity and mute the failure.
2. Use this as a condition to skip part, or all(?), of the test if run on MSSQL.
3. Put it in its own test class that implements PostgressOnly.
4. Something else?

Options 2 & 3 aren't great options. They would require updates to the test when the issue is fixed. Not terrible, just something that needs to be tracked in the issue.

Option 1, muting in TeamCity, has the advantage of no update needed to the test code and it would start passing once the issue is fixed. On the down side if you don't scope the mute correctly you could hide a failure inPostgres.

labkey-chrisj:
Also for consideration: with Josh's changes in the related PR, the console spew includes this kind of information when cancelling a job: 
image
I'm curious to know if our recent logging refactors might make it possible for tests to peek at the server log at times like this

labkey-jeckels:
> For context, this is what I'm seeing locally when I run this test on Windows/SQL up to the point that I click 'save and finish' for the assay import: no cancel option in the data pipeline list: image
> 
> I did the same check on PG, with exactly the same result: no option to cancel. This could be info relevant to Issue 53240? I'll add that to the issue.

This is a difference in the job grid in the assay UI vs the pipeline UI. We could add the buttons here, or the test go to Gear->Go To Module->Pipeline

labkey-jeckels:
> Also for consideration: with Josh's changes in the related PR, the console spew includes this kind of information when cancelling a job: image I'm curious to know if our recent logging refactors might make it possible for tests to peek at the server log at times like this

The recent logging changes don't really change this. The tests always had the option of looking at this through the Admin Console. Here's an example I just added: https://github.com/LabKey/testAutomation/blob/f01937d12d848da9b57fdde7d45dad7003ffa6d7/src/org/labkey/test/tests/BasicTest.java#L68

labkey-jeckels (approved):

labkey-danield (approved):

PR #2483, opened 2025-06-10
Introduce a subclass of DomainProps for Datasets

labkey-klum:
#### Rationale
Add to the existing list of domain helpers to manage `Dataset` domains. Refactor existing tests to use the new class.

#### Related Pull Requests
https://github.com/LabKey/platform/pull/6732


src/org/labkey/test/params/study/DatasetDefinition.java
```diff
@@ -0,0 +1,131 @@
...
+    private String _keyPropertyName;
+    private Boolean _timeKeyField;
+
+    public static final String VISIT_BASED_STUDY = "StudyDatasetVisit";
+    public static final String DATE_BASED_STUDY = "StudyDatasetDate";
+
+    public static DatasetDefinition create(String name)
+    {
+        return new DatasetDefinition(name);
+    }
```
labkey-tchad, line 30: This static `create` method is a little confusing because `DomainProps` has a non-static `create` method that actually creates the domain.

labkey-tchad (approved):

src/org/labkey/test/params/study/DatasetDefinition.java
labkey-klum: that's true, I was thinking of maybe `init`, not sure if there is a better verb.

PR #2487, opened 2025-06-11
Add explicit test coverage for trigger script type conversion

labkey-jeckels:
#### Rationale
We want trigger scripts to have the ability to do custom parsing on incoming values. A recent change broke that, and we were somewhat lucky to catch it in another automated test. We'll benefit from more intentional coverage.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6748

#### Changes
- Send values that can't be parsed normally and make sure the trigger script can intercept and transform them

### DIFF
--- a/src/org/labkey/test/tests/TriggerScriptTest.java
+++ b/src/org/labkey/test/tests/TriggerScriptTest.java
@@ -86,6 +86,7 @@ public class TriggerScriptTest extends BaseWebDriverTest
 
     private static final String COMMENTS_FIELD = "Comments";
     private static final String COUNTRY_FIELD = "Country";
+    public static final String PEOPLE_LIST_NAME = "People";
 
     protected final PortalHelper _portalHelper = new PortalHelper(this);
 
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/tests/TriggerScriptTest.java
```diff
@@ -171,37 +172,45 @@ public static void projectSetup()
...
+//                new FieldDefinition("Crazy", ColumnType.Boolean).setDescription("Crazy?"));
+//
+//        importFolderFromZip(TestFileUtils.getSampleData("studies/LabkeyDemoStudy.zip"));
+//
+//        //Add webparts for dataset, data class, sample type setup
+//        goToProjectHome();
+//
+//        _portalHelper.addWebPart("Datasets");
+//        _portalHelper.addWebPart("Data Classes");
+//        _portalHelper.addWebPart("Sample Types");
```
labkey-susanh: Is this meant to be commented out?

labkey-susanh (changes requested):

src/org/labkey/test/tests/TriggerScriptTest.java
labkey-jeckels: On my dev machine, yes. In the repo, no. Now fixed.

labkey-susanh (approved):

PR #2491, opened 2025-06-13
Add test automation for miscellaneous sample and sample type issues.

labkey-danield:
#### Rationale
Updating shared test components to aid with the test automation identified in the related PR.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1466

#### Changes
- Update EntityTypeDesigner.setParentAlias to take an optional parameter that will set a parent as required.
- Update SampleTypeNameExpressionTest. testWithTrickyFieldNames to use all the tricky characters for a name expression.
- Updated TestDataGenerator to have REALISTIC_DOMAIN_NAMES
- Update TestDataGenerator.getRealisticDomainName to take an optional exclusion list (just to avoid duplicate name collisions).
- Removed deprecated method SampleTypeAPIHelper.getSampleIdFromName (and the 2 or 3 calls to it).


src/org/labkey/test/tests/SampleTypeNameExpressionTest.java
```diff
@@ -418,6 +440,68 @@ public void testWithTrickyCharacters()
...
+        goToProjectHome();
+
+        SampleTypeHelper sampleHelper = new SampleTypeHelper(this);
+
+        final String sampleType = "Issue 52180 Sample Type";
+
+        StringBuilder sbNameExpression = new StringBuilder();
+
+        // Covers Issue 52180
+        String parentsampleTypeEncoded = PARENT_SAMPLE_TYPE.replace("/", "\\/");
```
labkey-tchad: Isn't this what `PARENT_SAMPLE_TYPE_INPUT` is for?

src/org/labkey/test/tests/SampleTypeNameExpressionTest.java
```diff
@@ -71,15 +72,13 @@
 public class SampleTypeNameExpressionTest extends BaseWebDriverTest
 {
     private static final String PROJECT_NAME = "SampleType_Name_Expression_Test";
-    private static final String DEFAULT_SAMPLE_PARENT_VALUE = "SS" + TestDataGenerator.randomString(3).replaceAll("[_)]", "."); // '_' is used as delimiter to get batchRandomId and ) is used to close the defaultValue()
+
+    private static final String DEFAULT_SAMPLE_PARENT_VALUE = "SS" +
+            EscapeUtil.escapeForNameExpression(TestDataGenerator.randomString(3).replaceAll("[_)]", ".")); // '_' is used as delimiter to get batchRandomId and ) is used to close the defaultValue()
```
labkey-tchad: When I was messing with this test, escaping didn't seem to work for default values. The backslashes just ended up being part of the default value. If escaping worked, you should be able to not exclude `_` and `)` from the value.

labkey-tchad (approved):

src/org/labkey/test/tests/SampleTypeNameExpressionTest.java
```diff
@@ -637,14 +722,17 @@ public void testDeriveSampleFromSampleDetailsPage() throws Exception
...
         String flagString = "Hello, I'm a derived sample.";
         String intVal = "987";
-        String derivedSampleName = deriveSample(PARENT_SAMPLE_01, PARENT_SAMPLE_TYPE, sampleType, flagString, intVal);
+        String derivedSampleName = deriveSample(PARENT_SAMPLE_01, PARENT_SAMPLE_TYPE, sampleType,
+                Map.of(COL_DSTR.getName(), flagString,
+                        COL_DINT.getName(), intVal));
 
-        checker().verifyTrue("Name of derived sample doesn't look correct. Should contain 'Parent Sample'.",
-                derivedSampleName.contains("Parent Sample"));
+        checker().verifyTrue("Name of derived sample doesn't look correct. Should contain 'tricky06'.",
```
labkey-susanh: The use of 'tricky06' is a little magical here. Perhaps capture this in a constant somewhere like `PARENT_FIELD_VALUE_PREFIX = "tricky06"`?

labkey-susanh (approved):

PR #2504, opened 2025-06-19
Don't check the new "CSP Report" link for a navtrail

labkey-adam:
#### Rationale
Fix `AdminConsoleNavigationTest.testAdminNavTrails()` -- new action doesn't have a navtrail, so ignore that check

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6769


labkey-chrisj (approved):
Makes sense

labkey-tchad (approved):

PR #2508, opened 2025-06-23
expose grid headers to tests

labkey-chrisj:
#### Rationale
This change exposes the recently-added `getHeaders()` method on `ResponsiveGrid` to tests, such that they can resolve fields for which the test might only have fieldKeys or names.  This will be helpful for tests using randomly-generated/tricky field names, as some tricky chars that are legal for field names but aren't in field labels can mean the field I created (either via UI or API) won't match any in the grid if what the grid shows in its headers is labels.

#### Related Pull Requests
example usages are in https://github.com/LabKey/limsModules/pull/1481

#### Changes
Make `ResponsiveGrid.getHeaders()` public


labkey-tchad (approved):
Consider adding `getColumnNames` and `getColumnFieldKeys` methods instead (similar to the existing `getColumnLabels` method). Tests are pretty much going to always map `getHeaders` into one of those. Might as well skip the step for them.

PR #2510, opened 2025-06-24
Improve failure behavior in AssayTransformImportUpdateTest

labkey-jeckels:
#### Rationale
Improve failure condition when AssayTransformImportUpdateTest fails to kill the job

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6752

#### Changes
- Shorten the wait for killing the job, enabling the test to fail when it doesn't cancel and capture a thread dump from the server

### DIFF
--- a/src/org/labkey/test/tests/assay/AssayTransformImportUpdateTest.java
+++ b/src/org/labkey/test/tests/assay/AssayTransformImportUpdateTest.java
@@ -286,7 +286,7 @@ public class AssayTransformImportUpdateTest extends BaseWebDriverTest
         Instant before = Instant.now();
         importPage.clickSaveAndFinish();
 
-        waitAndClickAndWait(Locator.linkWithText("Assay upload RUNNING"));
+        waitAndClick(WAIT_FOR_JAVASCRIPT, Locator.linkWithText("Assay upload RUNNING"), WAIT_FOR_JAVASCRIPT);
         PipelineStatusDetailsPage pipelineStatusDetailsPage = new PipelineStatusDetailsPage(getDriver());
         pipelineStatusDetailsPage.clickCancel();
 


src/org/labkey/test/tests/assay/AssayTransformImportUpdateTest.java
```diff
@@ -286,7 +286,7 @@ public void testCancelAsyncAssayTransformJob() throws Exception
         Instant before = Instant.now();
         importPage.clickSaveAndFinish();
 
-        waitAndClickAndWait(Locator.linkWithText("Assay upload RUNNING"));
+        waitAndClick(WAIT_FOR_JAVASCRIPT, Locator.linkWithText("Assay upload RUNNING"), WAIT_FOR_JAVASCRIPT);
```
labkey-chrisj, line 289: This seems an acceptable change, but I'm curious as to how it serves to avoid the assertion?
waitAndClickAndWait wraps waitAndClick, the only difference is it passes in a longer wait for the page load

labkey-chrisj (approved):
I trust that this will work, but I'm curious to understand how 😅

src/org/labkey/test/tests/assay/AssayTransformImportUpdateTest.java
labkey-jeckels: It doesn't change the server-side behavior. That's addressed in the platform PR that's linked here.

This change simply makes the test fail after 10 seconds instead of 60 seconds if it doesn't see the link it expects. When it does, the failure triggers a thread dump on the server, which helps understand why it's taking so long to render. That thread dump helped me track down a deadlock-type problem.

labkey-chrisj: Thanks!

PR #2512, opened 2025-06-25
LKSM Folder Archives Test Automation

labkey-danield:
#### Rationale
Updating various shared test code for folder archive testing in LKSM.

The FolderArchiveDialog extends the DeleteConfirmationDialog. It's not identical but similar in behavior. I thought is would easier and quicker to derive from the delete dialog than to create a "new" dialog. Especially since I doubt the archive dialog will be used outside of these tests.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1490

#### Changes
- Added BaseWebDriverTest.TRICKY_DOMAIN_NAME_CHARACTERS (Maybe not needed anymore?)
- Add overloaded DeleteConfirmationDialog.confirmDelete to allow for different delete button text.
- Add FolderArchiveDialog
- Made public ResponsiveGrid.clickColumnMenuItem
- Correct locator in NavBar.


src/org/labkey/test/components/ui/search/SampleFinder.java
```diff
@@ -70,11 +70,13 @@ public void waitForReady()
...
      * Open the entity filter dialog for the specified filter type.
      *
-     * @param filterKind "Source" or "Parent" or "Assay" in SM. "Registry Parent" or "Sample Parent" or "Assay" in Biologics
+     * @param filterKind "Sample", "Source" or "Parent" or "Assay" in SM. "Registry Parent" or "Sample Parent" or "Assay" in Biologics
      * @return component wrapper for the EntityFieldFilterModal
      */
     public EntityFieldFilterModal clickAddParent(String filterKind)
     {
+        // Why is this method called clickAddParent?
+
```
labkey-tchad: This name is a relic of when the functionality was limited to searching by "Registry Parent" or "Sample Parent". Perhaps a questionable method name at the time but has gotten progressively worse with time.

src/org/labkey/test/components/ui/FolderArchiveDialog.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public ConfirmPage clickYesArchive()
+    {
+        return clickYesArchive(10);
+    }
+
+    public ConfirmPage clickYesArchive(Integer waitSeconds)
+    {
+        return super.clickConfirmButton(waitSeconds, "Yes, Archive Folder");
+    }
```
labkey-tchad: Extending `DeleteConfirmationDialog` is going to make this class have half a dozen "delete" related methods.
Possibly out of scope for this but we should pull a bunch of functionality from `DeleteConfirmationDialog` up into a more generic `ConfirmationDialog` so that all these confirmation dialogs aren't stepping on each other.

labkey-tchad (approved):

labkey-susanh (commented):
Seems good to me. I agree with Trey's comment about refactoring to have a more generic `ConfirmationDialog`. I remember seeing the `DeleteConfirmationDialog` as a base for something else giving me pause during some other PR.

src/org/labkey/test/components/ui/FolderArchiveDialog.java
labkey-danield: Yeah, I was being a little lazy with this. I've changed the `FolderArchiveDialog` to only extend the `ModalDialog`, and have reverted the changes to the `DeleteConfirmationDialog`.
If refactoring the `DeleteConfirmationDialog` is an important task we should discuss in stand up and see where it fits into the priority of other tasks.

src/org/labkey/test/components/ui/search/SampleFinder.java
labkey-danield: I've changed this to `clickAddSearchCard`. I think that name aligns it with more with the `removeSearchCard` and `removeAllSearchCards` methods. Let me know if yo disagree.

labkey-danield:
> Seems good to me. I agree with Trey's comment about refactoring to have a more generic `ConfirmationDialog`. I remember seeing the `DeleteConfirmationDialog` as a base for something else giving me pause during some other PR.

I've changed the FolderArchiveDialog to extend the ModalDialog and not the DeleteConfirmationDialog. I'll bring up in stand up the possibility of refactoring parts of the DeleteConfirmationDialog.

PR #2514, opened 2025-06-25
Automation for 51695

labkey-klum:
#### Rationale
Adds automation to verify we don't allow users to get into a bad state with a multi value FK in the metadata source editor.

#### Related Pull Requests
https://github.com/LabKey/platform/pull/6793


labkey-chrisj (approved):
No notes, looks good to me

labkey-danield (approved):

PR #2517, opened 2025-06-26
Support for clearing more form input elements

labkey-jeckels:
#### Rationale
It's handy to clear values from more than just file `<input>` elements

#### Changes
- Move `clear()` method up the hierarchy


labkey-danield (approved):
I suspect the clear was done by calling `set("")`

labkey-jeckels:
> I suspect the clear was done by calling `set("")`

That's what I tried initially (and then with `null`) but both failed trying to parse it as a date. I didn't save the stack trace.

PR #2522, opened 2025-06-27
Issue 52339: Change list domain URI construction to not use encoded user-provided names

labkey-susanh:
#### Rationale
Issue [52339](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52339): When names are too long, we exceed the limit of the domain URI field. We don't really need to use names in these URIs.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6805

#### Changes
- Add a test for a list with a long name that includes lots of UTF-8 characters and sets default values

### DIFF
--- a/src/org/labkey/test/tests/list/ListTest.java
+++ b/src/org/labkey/test/tests/list/ListTest.java
@@ -48,6 +48,7 @@ import org.labkey.test.components.list.AdvancedListSettingsDialog;
 import org.labkey.test.pages.ImportDataPage;
 import org.labkey.test.pages.list.EditListDefinitionPage;
 import org.labkey.test.pages.list.GridPage;
+import org.labkey.test.pages.query.UpdateQueryRowPage;
 import org.labkey.test.params.FieldDefinition;
 import org.labkey.test.params.FieldDefinition.StringLookup;
 import org.labkey.test.tests.AuditLogTest;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/list/ListTest.java
```diff
@@ -508,6 +509,32 @@ public void testNameTrimming()
         checker().withScreenshot().verifyEquals("Name not trimmed as expected", trimmedName, editList.getName());
     }
 
+    @Test // Issue 52339
+    public void testLongName()
+    {
+        String listName = "A_+-:''.¡™£¢∞§¶•ªº–≠œ∑´®†¥¨ˆøπ“‘«æ…¬˚∆˙©√ƒ∂ßΩ≈ç√∫µ≤≥÷‹›ﬁﬂ‡°·‚—±⁄€‹›‡‰Æ«»¢∫√∑∏∂";
```
XingY, line 515: Consider append a random string?

XingY (approved):

src/org/labkey/test/tests/list/ListTest.java
labkey-susanh: Seems superfluous when the name is already ridiculously complex.

XingY: Perhaps. It does lack certain pattern though, like whitespaces.

labkey-danield: I don't think any of these need to be escaped to unicode `\u` values. Was that your impression as well?

labkey-danield (approved):

src/org/labkey/test/tests/list/ListTest.java
labkey-susanh: The name of the list looks like it's as provided here. And, in any case, the particular characters are not all that important, just the fact that this leads to a really long URI encoded name.

PR #2513, opened 2025-06-25
Refactor DataRegionTable to handle tricky characters better

labkey-tchad:
#### Rationale
DataRegionTable doesn't actually handle field names with tricky characters. What it calls field names are actually fieldKeys. This update adopts a similar pattern to the recent app grid test update. I had to replicate some existing non-standard behavior to avoid breaking existing tests (removing spaces from fieldKeys and ignoring spaces when looking up field labels).

#### Related Pull Requests
- #2450 
- https://github.com/LabKey/premiumModules/pull/288

#### Changes
- Update `DataRegionTable` to use `FieldReferenceManager.java` to organize and identify columns
- Clean up numerous IntelliJ warnings
- Remove usages of some deprecated `DataRegionTable` methods

PR #2526, opened 2025-06-30
Fix data region column name references

labkey-tchad:
#### Rationale
My previous DataRegionTable refactor attempted to recreate the previous behavior that would remove spaces when matching field labels. I covered the case where the test passed in an unnecessary space but not when they omitted a space. Rather that duplicating that particular behavior, this updates the handful of tests that are passing incorrect field labels.

I've also updated the `CustomizeView` helper to accept `FieldKey` objects instead of String arrays.

#### Related Pull Requests
- #2513 

#### Changes
- Remove unnecessary logging parameters from the CustomizeView helper
- Update CustomizeViewHelper to use FieldKey
- Fix problematic data region column name references

### DIFF
--- a/src/org/labkey/test/components/CustomizeView.java
+++ b/src/org/labkey/test/components/CustomizeView.java
@@ -29,7 +29,6 @@ import org.labkey.test.components.ext4.ComboBox;
 import org.labkey.test.components.ext4.RadioButton;
 import org.labkey.test.components.ext4.Window;
 import org.labkey.test.params.FieldKey;
-import org.labkey.test.selenium.LazyWebElement;
 import org.labkey.test.selenium.RefindingWebElement;
 import org.labkey.test.util.DataRegionTable;
 import org.labkey.test.util.Ext4Helper;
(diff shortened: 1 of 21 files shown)


src/org/labkey/test/components/CustomizeView.java
```diff
@@ -347,84 +342,66 @@ public enum ViewItemType
      * expand customize view menu to all but the last of fieldKeyParts
      * @return The data-recordid property of the &lt;tr&gt; element for the specified field in the "Available Fields" column tree.
      */
-    private WebElement expandPivots(String[] fieldKeyParts)
+    private WebElement expandPivots(CharSequence fieldIdentifier)
```
labkey-danield: nit:
Maybe update the javadoc comment for the method to reference fieldIdentifier and not fieldKeyParts.

labkey-danield (approved):
Several of the changes I'm taking at face value. If the test passes on TeamCity then I'm assuming it is good.

labkey-chrisj (approved):

PR #2530, opened 2025-07-02
A couple more CSP enhancements

labkey-tchad:
#### Rationale
Add test coverage for CSP enhancements 

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6809

#### Changes
- Configure object-src for Knitr tests and remove exclusion from `CspLogUtil`
- Add regression test for script nonce in report webpart (`AbstractKnitrReportTest`)
- Remove redundant methods from `PortalHelper`
- Move some methods from `WikiHelper` to `wiki.EditPage`


src/org/labkey/test/tests/AbstractKnitrReportTest.java
```diff
@@ -224,4 +237,66 @@ protected void moduleReportDependencies()
...
     }
+
+    /**
+     * Issue 53211: CSP reports when an R/Plotly graph is displayed in Reports web part, same thing wrapped in a wiki works fine with strict csp
+     */
+    @Test
+    public void testEmbeddedReportNonce()
+    {
+        CspConfigHelper.debugCspWarnings();
+        new CspConfigHelper(this).setEnforceCsp(false);
```
labkey-adam, line 248: So we're still disabling the enforce CSP and checking for logged warnings?

labkey-adam (approved):

src/org/labkey/test/tests/AbstractKnitrReportTest.java
labkey-tchad: I suppose we don't have to now that you mention it.
I'll remove these in a separate PR after this is merged. I don't want to touch this branch since it is merging to 25.7 but TeamCity is running it against develop.

labkey-chrisj (approved):

PR #2463, opened 2025-05-29
Selenium test for file issues

XingY:
#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1801
- https://github.com/LabKey/platform/pull/6699
- https://github.com/LabKey/limsModules/pull/1431
- https://github.com/LabKey/commonAssays/pull/890
- https://github.com/LabKey/testAutomation/pull/2463

#### Changes
-

PR #2532, opened 2025-07-03
Issue 53394: FileInput revert removal of the "-fileUpload" suffix from inputId

cnathe:
#### Rationale
Issue [53394](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53394): LKSM: Selection of file doesn't work on develop

Revert changes from https://github.com/LabKey/labkey-ui-components/pull/1800 

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1825
- https://github.com/LabKey/testAutomation/pull/2532
- https://github.com/LabKey/limsModules/pull/1506

#### Changes
- Test identifier updates for -fileUpload suffix revert

### DIFF
--- a/src/org/labkey/test/components/ui/entities/EntityBulkUpdateDialog.java
+++ b/src/org/labkey/test/components/ui/entities/EntityBulkUpdateDialog.java
@@ -253,9 +253,9 @@ public class EntityBulkUpdateDialog extends ModalDialog
      * @param fieldIdentifier Identifier for the field; name ({@link String}) or fieldKey ({@link FieldKey})
      * @return file attachment component
      */
-    public FileAttachmentContainer getFileField(CharSequence fieldIdentifier)
+    private FileAttachmentContainer getFileField(CharSequence fieldIdentifier)
     {
-        String identifier = FileAttachmentContainer.fileUploadIdentifier(fieldIdentifier.toString());
+        FieldKey identifier = FileAttachmentContainer.fileUploadFieldKey(fieldIdentifier);
         return enableAndWait(identifier, elementCache().fileUploadField(identifier));
     }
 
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/components/ui/files/FileAttachmentContainer.java
```diff
@@ -195,6 +195,10 @@ protected class ElementCache extends Component<?>.ElementCache
         public Locator fileUploadScrollFooterLoc = Locator.tagWithClass("div", "file-upload__scroll-footer");
     }
 
+    public static String fileUploadIdentifier(String fieldIdentifier)
+    {
+        return fieldIdentifier + "-fileUpload"; // Issue 53394
+    }
```
labkey-tchad: We should [revert](https://github.com/LabKey/testAutomation/pull/2473) to how this worked before the removal. This way it will work for anything that's passing in an un-encoded field name.
```suggestion
    /**
     * File upload fields append "-fileUpload" to the field's fieldKey
     * @param fieldIdentifier Identifier for the field; name ({@link String}) or fieldKey ({@link FieldKey})
     * @return FieldKey with expected suffix
     */
    public static FieldKey fileUploadFieldKey(CharSequence fieldIdentifier)
    {
        return FieldKey.fromFieldKey(FieldKey.fromName(fieldIdentifier) + "-fileUpload");
    }
```

src/org/labkey/test/components/ui/entities/EntityBulkUpdateDialog.java
```diff
@@ -271,7 +272,8 @@ public EntityBulkUpdateDialog removeFile(CharSequence fieldIdentifier)
 
     public FileUploadField getExistingFileField(String fieldIdentifier)
     {
-        return enableAndWait(fieldIdentifier, elementCache().fileField(fieldIdentifier));
+        String identifier = FileAttachmentContainer.fileUploadIdentifier(fieldIdentifier.toString());
+        return enableAndWait(identifier, elementCache().fileField(identifier));
```
labkey-tchad: This should take a `CharSequence`, not a `String`.
Also, what is the distinction between `getFileField` and `getExistingFileField`? (Maybe more of a question for Xing since she [added](https://github.com/LabKey/testAutomation/pull/2463) the latter)

labkey-tchad (changes requested):

src/org/labkey/test/components/ui/entities/EntityBulkUpdateDialog.java
cnathe: I believe that the `getExistingFileField` is to get the file card for an existing file and the `getFileField` is to get the file upload input when no file exists for the field. I'm renaming `getExistingFileField` to `getExistingFileCard`, maybe that will help a bit.

labkey-tchad (approved):

PR #2537, opened 2025-07-07
Avoid generating fields names with reserved substitution format patterns

labkey-tchad:
#### Rationale
As more tests are being converted to use random field names, they are failing periodically because we generate random field names that conflict with reserved formats. The reserved formats are colons followed by strings that tests often use for field names (e.g. ":Date" or ":First"). This change replaces the ':' with another random character if it generates something that might cause trouble.

#### Related Pull Requests
- N/A

#### Changes
- Avoid generating fields names with reserved substitution format patterns

### DIFF
--- a/src/org/labkey/test/util/TestDataGenerator.java
+++ b/src/org/labkey/test/util/TestDataGenerator.java
@@ -580,6 +580,17 @@ public class TestDataGenerator
         String chars = ALL_ILLEGAL_QUERY_KEY_CHARACTERS + " %()=+-[]_|*`'\":;<>?!@#^" + NON_LATIN_STRING;// + WIDE_PLACEHOLDER ;
 
         String randomFieldName = randomName(part, numStartChars, numEndChars, chars, exclusion);
+
+        // Avoid generating fields names with reserved substitution format patterns. e.g. ":Date" or ":First"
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -580,6 +580,17 @@ public static String randomFieldName(@NotNull String part, int numStartChars, in
         String chars = ALL_ILLEGAL_QUERY_KEY_CHARACTERS + " %()=+-[]_|*`'\":;<>?!@#^" + NON_LATIN_STRING;// + WIDE_PLACEHOLDER ;
 
         String randomFieldName = randomName(part, numStartChars, numEndChars, chars, exclusion);
+
+        // Avoid generating fields names with reserved substitution format patterns. e.g. ":Date" or ":First"
+        if (numStartChars > 0 && randomFieldName.charAt(numStartChars - 1) == ':' &&
+                StringUtils.isAlpha(part.substring(0, 4))) // The shortest pattern is four characters (see org.labkey.api.util.SubstitutionFormat.getFormatNames)
+        {
+            String regenExclusion = Objects.requireNonNullElse(exclusion, "") + ":";
```
labkey-danield, line 588: Interesting, didn't know about this. Learn something new every day.

labkey-danield (approved):

PR #2549, opened 2025-07-09
Add regression test for CSP log view

labkey-tchad:
#### Rationale
[[Story](https://github.com/LabKey/kanban/issues/901)]

#### Related Pull Requests
- 

#### Changes
- Add regression test for CSP log view

### DIFF
--- a/src/org/labkey/test/tests/CrawlerTest.java
+++ b/src/org/labkey/test/tests/CrawlerTest.java
@@ -1,11 +1,13 @@
 package org.labkey.test.tests;
 
+import org.assertj.core.api.Assertions;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Assume;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
+import org.labkey.remoteapi.SimpleGetCommand;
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/tests/CrawlerTest.java
```diff
@@ -117,18 +119,25 @@ public void testEnforceCsp() throws Exception
...
+        CspLogUtil.checkNewCspWarnings(getArtifactCollector()); // throws CspWarningDetectedException
     }
 
     @Test (expected = CspLogUtil.CspWarningDetectedException.class)
-    public void testCspWarning()
+    public void testCspWarning() throws Exception
     {
         Assume.assumeFalse("Can't test for CSP report", TestProperties.isCspCheckSkipped());
 
         _cspConfigHelper.setEnforceCsp(false);
```
labkey-adam, line 130: Is it time to set this to true (or not call it at all)?

labkey-adam (approved):

src/org/labkey/test/tests/CrawlerTest.java
labkey-tchad: This test is specifically testing the warning. There is a separate `testCspEnforce`.

PR #2550, opened 2025-07-10
Fix random failures in the ListTest and SampleTypeFolderExportImportTest

labkey-danield:
#### Rationale
Fixing two test issues. The failure in SampleTypeFolderExportImportTest appears to be random, and may not really need this change, but I think checking is still a good thing to do.

#### Related Pull Requests
- None.

#### Changes
- In ListTest, wait for the DataRegions that contain the list history.
- For SampleTypeFolderExportImportTest.testExportImportSampleTypesWithAssayRuns validate that file has been uploaded.

### DIFF
--- a/src/org/labkey/test/tests/SampleTypeFolderExportImportTest.java
+++ b/src/org/labkey/test/tests/SampleTypeFolderExportImportTest.java
@@ -43,6 +43,7 @@ import org.labkey.test.pages.admin.FolderManagementPage;
 import org.labkey.test.params.FieldDefinition;
 import org.labkey.test.params.experiment.DataClassDefinition;
 import org.labkey.test.params.experiment.SampleTypeDefinition;
+import org.labkey.test.util.ArtifactCollector;
 import org.labkey.test.util.DataRegionTable;
 import org.labkey.test.util.LogMethod;
 import org.labkey.test.util.PortalHelper;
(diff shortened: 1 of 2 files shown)


XingY (approved):

src/org/labkey/test/tests/SampleTypeFolderExportImportTest.java
```diff
@@ -557,7 +557,7 @@ public void testExportImportSampleTypesWithAssayRuns() throws Exception
         // create a test sampleType
         List<FieldDefinition> testFields = SampleTypeAPIHelper.sampleTypeTestFields(true);
         SampleTypeDefinition testSampleType = new SampleTypeDefinition(testSamples).setFields(testFields)
-                .addParentAlias("SelfParent"); // to derive from samles in the current type
+                .addParentAlias("SelfParent"); // to derive from samples in the current type
```
labkey-chrisj, line 561: thanks for fixing that

src/org/labkey/test/tests/SampleTypeFolderExportImportTest.java
```diff
@@ -578,6 +578,11 @@ public void testExportImportSampleTypesWithAssayRuns() throws Exception
         sourceSamplesTable.clickEditRow(1);
         waitForElementToBeVisible(Locator.tagWithAttribute("input", "type", "file"));
         setFormElement(Locator.tagWithAttribute("input", "type", "file"), SAMPLE_TXT_FILE);
+        // setFormElement doesn't check that the form element is set.
+        // Because this test uses random field names, we should validate that the file was actually uploaded. If the
+        // file is missing later in the test, we can be sure it was present at this point.
+        Assert.assertFalse("File not uploaded.",
```
labkey-chrisj: Consider maybe using an awaiting assert here, if file-upload taking a second isn't a failure-worthy situation.

labkey-chrisj (approved):

PR #2559, opened 2025-07-13
Optional features ui coverage

labkey-chrisj:
#### Rationale
This change adds UI coverage for optional features and experimental features.
Recent changes have refactored the implementation for exp features, optional features, and deprecated features under the singular optionalFeaturesService implementation.  This coverage seeks to validate the UI continues to work as expected.

#### Related Pull Requests
https://github.com/LabKey/platform/pull/6828

#### Changes
New page class to wrap all three of the optional features pages

### DIFF
--- a/src/org/labkey/test/pages/core/admin/OptionalFeaturesPage.java
+++ b/src/org/labkey/test/pages/core/admin/OptionalFeaturesPage.java
@@ -13,7 +13,7 @@ import java.util.Map;
 import java.util.Set;
 
 /*
-    Wraps Optional Features, Experimental Featueres, Deprecated Features pages linked
+    Wraps Optional Features, Experimental Features, Deprecated Features pages linked
     from Admin Console
  */
 public class OptionalFeaturesPage extends LabKeyPage<OptionalFeaturesPage.ElementCache>
--- a/src/org/labkey/test/pages/core/admin/OptionalFeaturesPage.java
+++ b/src/org/labkey/test/pages/core/admin/OptionalFeaturesPage.java
@@ -1,0 +1,106 @@
+package org.labkey.test.pages.core.admin;
+
+import org.labkey.test.Locator;
+import org.labkey.test.WebDriverWrapper;
+import org.labkey.test.WebTestHelper;
+import org.labkey.test.components.html.Checkbox;
+import org.labkey.test.pages.LabKeyPage;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/*
+    Wraps Optional Features, Experimental Featueres, Deprecated Features pages linked
+    from Admin Console
+ */
+public class OptionalFeaturesPage extends LabKeyPage<OptionalFeaturesPage.ElementCache>
+{
+    public OptionalFeaturesPage(WebDriver driver)
+    {
+        super(driver);
+    }
+
+    public static OptionalFeaturesPage beginAt(WebDriverWrapper webDriverWrapper, OptionalFeatureType featureType)
+    {
+        webDriverWrapper.beginAt(WebTestHelper.buildURL("admin", "/", "optionalFeatures", Map.of("Type", featureType.toString())));
+        return new OptionalFeaturesPage(webDriverWrapper.getDriver());
+    }
+
+    @Override
+    protected void waitForPage()
+    {
+        waitFor(()-> elementCache().listGroupLoc.findWhenNeeded(getDriver()).isDisplayed(),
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/pages/core/admin/OptionalFeaturesPage.java
```diff
@@ -0,0 +1,106 @@
...
+import org.labkey.test.pages.LabKeyPage;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/*
+    Wraps Optional Features, Experimental Featueres, Deprecated Features pages linked
```
labkey-adam:
```suggestion
    Wraps Optional Features, Experimental Features, Deprecated Features pages linked
```

src/org/labkey/test/tests/AdminConsoleTest.java
```diff
@@ -226,6 +230,88 @@ public void testSiteBannerAPIConfiguration() throws Exception
                 .verifyFalse("expect banner not to be shown", bannerLoc.isDisplayed(getDriver()));
     }
 
+    @Test
+    public void testUIOptionalFeatures()
+    {
+        goToAdminConsole();
+        waitAndClickAndWait(Locator.linkWithText("optional features"));
+        var optionalFeaturesPage = new OptionalFeaturesPage(getDriver());
+        var featureIds = List.of("extendedMetrics", "StageFileUploads");
```
labkey-adam: 👍 Makes it easy to add and remove these, when needed

labkey-adam (changes requested):
Functionally, this looks great... exactly what we discussed. However, you can drastically reduce the test method code duplication to improve maintenance, etc. Create a method like:
```
private void testUIOptionalFeatures(String linkText, OptionalFeatureType type, List<String> featureIds)
{
    goToAdminConsole();
    waitAndClickAndWait(Locator.linkWithText(linkText));
    var optionalFeaturesPage = new OptionalFeaturesPage(getDriver());
    ...
}
```
The two test methods each then become a single line that calls this method.

labkey-adam (approved):
Not a big deal, but if you passed in the link text to the shared method, you could avoid a couple more lines of redundant code. The `test*()` methods could literally be one-liners.

Also, not a big deal... but in my review I had added a suggestion to fix a comment typo.

src/org/labkey/test/tests/AdminConsoleTest.java
```diff
@@ -247,7 +247,7 @@ public void testUIExperimentalFeatures()
         waitAndClickAndWait(Locator.linkWithText("experimental features"));
```
labkey-adam: Should remove this line as well

src/org/labkey/test/pages/core/admin/OptionalFeaturesPage.java
```diff
@@ -0,0 +1,106 @@
...
+                for (WebElement el : listItemLabelLoc.findElements(listGroupElement)) {
+                    WebElement idEl = Locator.tagWithAttribute("type", "checkbox")
+                            .findElement(el);
+                    WebElement labelEl = Locator.tagWithClass("span", "toggle-label-text")
+                            .findElement(el);
+                    _listItems.put(idEl.getText(), labelEl.getText());
+                }
+            }
+            return _listItems;
+        }
```
labkey-danield: This doesn't belong in the ElementCache, it should be broken out to a private method in the class. The Locator doesn't need to be public it is used only in the `getListItems` method.

labkey-danield (changes requested):
The `getListItems` method should be moved out of the `ElementCache`.

src/org/labkey/test/pages/core/admin/OptionalFeaturesPage.java
```diff
@@ -0,0 +1,106 @@
...
+
+    public Map<String, String> getFeatureMap()
+    {
+        return elementCache().getListItems();
+    }
+
+    public Set<String> getFeatureIds()
+    {
+        return getFeatureMap().keySet();
+    }
```
labkey-danield: Not called and should be removed.

src/org/labkey/test/pages/core/admin/OptionalFeaturesPage.java
labkey-danield: Actually this code can be removed altogether. The method `getListItems` is only called by `getFeatureMap()` which is only called by `getFeatureIds()`, which is not called.

This method never gets executed.

labkey-danield (changes requested):
You have unused code in `OptionalFeaturesPage` that should be removed.

labkey-danield (approved):

PR #2571, opened 2025-07-18
TestDataGenerator randomFieldName and randomDomainName usage of ValidateDomainAndFieldNamesAction

cnathe:
#### Rationale
We've been slowly handling various cases on the test code side to account for randomly generated domain and field names that end up being invalid (because of invalid character sequences, length, etc.). This PR changes that approach so that we instead call a server API (ValidateDomainAndFieldNamesAction) to do those checks on the generated name and regenerate the name if it is invalid.

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/2571
- https://github.com/LabKey/platform/pull/6859
- https://github.com/LabKey/limsModules/pull/1531
- https://github.com/LabKey/premiumModules/pull/313

#### Changes
- new FieldInfo(TestDataGenerator.randomFieldName() -> FieldInfo.random()
- TestDataGenerator randomString support for REPEAT_PLACEHOLDER to add a repeated character
- TestDataGenerator randomString support for ALL_CHARS_PLACEHOLDER to add all of the charSet to the generated string
- DataClassAPIHelper.dataClassTestFields update to use FieldInfo.random()
- PropertyController.ValidateDomainAndFieldNameAction and usage in TestDataGenerator.randomFieldName and randomDomainName

### DIFF
--- a/src/org/labkey/test/tests/ExportOptionsMetadataOnlyTest.java
+++ b/src/org/labkey/test/tests/ExportOptionsMetadataOnlyTest.java
@@ -67,8 +67,8 @@ public class ExportOptionsMetadataOnlyTest extends BaseWebDriverTest
 
         DataClassDefinition testType = new DataClassDefinition(dataClassName).setFields(DataClassAPIHelper.dataClassTestFields());
         TestDataGenerator testDgen = DataClassAPIHelper.createEmptyDataClass(getProjectName(), testType);
-        testDgen.addCustomRow(Map.of("Name", "class1", "intColumn", 1, "decimalColumn", 1.1, "stringColumn", "one"));
-        testDgen.addCustomRow(Map.of("Name", "class2", "intColumn", 2, "decimalColumn", 2.2, "stringColumn", "two"));
+        testDgen.addCustomRow(Map.of("Name", "class1"));
+        testDgen.addCustomRow(Map.of("Name", "class2"));
         testDgen.insertRows();
 
         log("Export data class design only");
--- a/src/org/labkey/test/util/SampleTypeHelper.java
+++ b/src/org/labkey/test/util/SampleTypeHelper.java
@@ -30,6 +30,7 @@ import org.labkey.test.pages.experiment.CreateSampleTypePage;
 import org.labkey.test.pages.experiment.UpdateSampleTypePage;
 import org.labkey.test.params.FieldDefinition;
 import org.labkey.test.params.experiment.SampleTypeDefinition;
+import org.labkey.test.util.data.TestDataUtils;
 import org.openqa.selenium.WebDriver;
 
 import java.io.File;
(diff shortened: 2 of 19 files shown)


src/org/labkey/test/tests/ExportOptionsMetadataOnlyTest.java
```diff
@@ -67,8 +67,8 @@ public void testDataClassExportOptions() throws IOException, CommandException
 
         DataClassDefinition testType = new DataClassDefinition(dataClassName).setFields(DataClassAPIHelper.dataClassTestFields());
         TestDataGenerator testDgen = DataClassAPIHelper.createEmptyDataClass(getProjectName(), testType);
-        testDgen.addCustomRow(Map.of("Name", "class1", "intColumn", 1, "decimalColumn", 1.1, "stringColumn", "one"));
-        testDgen.addCustomRow(Map.of("Name", "class2", "intColumn", 2, "decimalColumn", 2.2, "stringColumn", "two"));
+        testDgen.addCustomRow(Map.of("Name", "class1"));
+        testDgen.addCustomRow(Map.of("Name", "class2"));
```
labkey-danield, line 71: Is this just cleaning up data that adds no value to the test?

src/org/labkey/test/util/SampleTypeHelper.java
```diff
@@ -81,18 +82,11 @@ public static SampleTypeHelper beginAtSampleTypesList(WebDriverWrapper dWrapper,
...
-        rows.add(String.join("\t", data.get(0).keySet()));
-        data.forEach(dataMap -> {
-            StringBuilder row = new StringBuilder();
-            data.get(0).keySet().forEach(key -> {
-                row.append(dataMap.get(key));
-                row.append("\t");
-            });
-            rows.add(row.substring(0, row.lastIndexOf("\t")));
-        });
-        return String.join("\n", rows);
```
labkey-danield, line 95: Hopefully there aren't too many more methods like this out there.

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -71,8 +74,10 @@
 public class TestDataGenerator
 {
     public static final String WIDE_CHAR = "\uD83D\uDC7E"; // 👾
-    private static final char WIDE_PLACEHOLDER = '\u03A0'; // 'Π' - Wide character can't be picked from the string with 'charAt'
-    private static final String NON_LATIN_STRING = "\u0438\uC548\u306F"; // "и안は"
+    public static final char WIDE_PLACEHOLDER = '\u03A0'; // 'Π' - Wide character can't be picked from the string with 'charAt'
+    public static final char REPEAT_PLACEHOLDER = '\u22EF'; // '⋯' - Used to indicate that the char will be repeated
```
labkey-danield, line 78: Thanks. I was tempted on one of the previous PRs that made `WIDE_CHAR` public to suggest making `WIDE_PLACEHOLDER` and `NON_LATIN_STRING` public as well.

I also wonder how we decide where to put public constants like these as well as the next few lines. We have some constant values in TestDataUtils. BaseWebDriverTest has several constants as well. I think there might be value to consider a future PR that consolidates all of these type of constants into one location.

labkey-danield (approved):

src/org/labkey/test/util/SampleTypeHelper.java
cnathe: We've likely found most of them at this point!

src/org/labkey/test/tests/ExportOptionsMetadataOnlyTest.java
cnathe: Yes, these other columns like "intColumn" are now randomly generated names so when I went to update them in the data rows I noticed that the values aren't checked anywhere so it seemed cleaner to just remove the related data row values.

src/org/labkey/test/util/TestDataGenerator.java
cnathe: Agreed!

PR #2577, opened 2025-07-23
Track search work by container; purge on delete

labkey-jeckels:
#### Rationale
Instead of waiting for the indexer to drain its queue before deleting a container, we can purge that work from the queue.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6872

#### Changes
- Track search work based on the container for the data
- Stop waiting for indexing to complete before deleting containers in tests


labkey-jeckels:
Replaced by https://github.com/LabKey/testAutomation/pull/2599

PR #2578, opened 2025-07-24
Misc test updates related to recent fuzz testing changes

cnathe:
#### Rationale

- FileAttachmentColumnTest update to comment out randomFieldName usage 
  - until [53505](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53505) is fixed
- InlineImagesListTest update to not use randomFieldName for list field labels
  - the test is checking for column width in the exported Excel file so fuzzing the label isn't helpful here
- TextChoiceSampleTypeTest update for how tsv import values are quoted in bulk import
  - remove surrounding quotes added by the test case and instead include a double quote in the value

### DIFF
--- a/src/org/labkey/test/tests/FileAttachmentColumnTest.java
+++ b/src/org/labkey/test/tests/FileAttachmentColumnTest.java
@@ -93,7 +93,9 @@ public class FileAttachmentColumnTest extends BaseWebDriverTest
     private final String RESULT_FILE_COL = "resultFile";
     private final String OTHER_RESULT_FILE_COL = "otherResultFile";
     private final String STUDY_DATASET_NAME = "ogreSpiteLevels";
-    private static final FieldInfo LIST_ATTACHMENT_FIELD = new FieldInfo(TestDataGenerator.randomFieldName("File / Attachment"), ColumnType.Attachment);
+    // Issue 53505: enable use of randomFieldName when issue is fixed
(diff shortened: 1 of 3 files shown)


src/org/labkey/test/tests/FileAttachmentColumnTest.java
```diff
@@ -93,7 +93,9 @@ public class FileAttachmentColumnTest extends BaseWebDriverTest
     private final String RESULT_FILE_COL = "resultFile";
     private final String OTHER_RESULT_FILE_COL = "otherResultFile";
     private final String STUDY_DATASET_NAME = "ogreSpiteLevels";
-    private static final FieldInfo LIST_ATTACHMENT_FIELD = new FieldInfo(TestDataGenerator.randomFieldName("File / Attachment"), ColumnType.Attachment);
+    // Issue 53505: enable use of randomFieldName when issue is fixed
+    //private static final FieldInfo LIST_ATTACHMENT_FIELD = new FieldInfo(TestDataGenerator.randomFieldName("File / Attachment"), ColumnType.Attachment);
```
labkey-danield, line 97: If you haven't done so already, can you add a comment to [53505](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53505) referencing this test and letting whoever fixes know they need to uncomment this line of code?

labkey-danield (approved):

src/org/labkey/test/tests/FileAttachmentColumnTest.java
cnathe: done

PR #2580, opened 2025-07-24
Add ability to paste HTML with actionPaste

labkey-alan:
#### Rationale
We need to be able to test what happens when a users paste's an HTML payload into our ELN editor.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-premium/pull/794
- https://github.com/LabKey/limsModules/pull/1539

#### Changes
- Add ability to paste HTML with actionPaste

### DIFF
--- a/src/org/labkey/test/HtmlFragmentSelection.java
+++ b/src/org/labkey/test/HtmlFragmentSelection.java
@@ -1,0 +1,45 @@
+package org.labkey.test;
+
+import org.jetbrains.annotations.NotNull;
+
+import java.awt.datatransfer.DataFlavor;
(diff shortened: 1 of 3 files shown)


labkey-tchad (approved):

src/org/labkey/test/HtmlFragmentSelection.java
```diff
@@ -0,0 +1,45 @@
+package org.labkey.test;
```
labkey-tchad: Consider putting this in the `util` package
```suggestion
package org.labkey.test.util;
```

PR #2584, opened 2025-07-25
Additional AuditEvents and helper methods for checking file audit logs

labkey-susanh:
#### Rationale
See [related PR](https://github.com/LabKey/platform/pull/6869)

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6869

#### Changes
- Add `FILE_SYSTEM_EVENT` and `ASSAY_AUDIT_EVENT` to `AuditEvents` enum
- Add new variant of `checkAuditEventValuesForTransactionId` for expected rows with different values.


src/org/labkey/test/util/AuditLogHelper.java
```diff
@@ -156,6 +158,18 @@ public void checkAuditEventValuesForTransactionId(String containerPath, AuditEve
         }
     }
 
+    public void checkAuditEventValuesForTransactionId(String containerPath, AuditEvent auditEventName, Integer transactionId, int rowCount, List<Map<String, Object>> expectedValues) throws IOException, CommandException
+    {
+        List<String> columnNames = expectedValues.get(0).keySet().stream().map(Object::toString).toList();
+        List<Map<String, Object>> events = getAuditLogsForTransactionId(containerPath, auditEventName, columnNames, transactionId, ContainerFilter.CurrentAndSubfolders);
+        assertEquals("Unexpected number of events for transactionId " + transactionId, rowCount, events.size());
+        for (int i = 0; i < rowCount; i++)
```
cnathe: do you need the `rowCount` param here? could you just use the `expectedValues` list size?

cnathe (approved):

src/org/labkey/test/util/AuditLogHelper.java
labkey-susanh: For current usages, it's not needed. I was leaving open the possibility that you may not want to verify the data in all the rows, but ended up not using it. Removing `rowCount`.

PR #2585, opened 2025-07-28
Extra Windows Helpers

labkey-martyp:
#### Rationale
Test helper functions for EHR link crawler test.

#### Related Pull Requests
- TBD

#### Changes
- Add a close extra windows function that doesn't log everytime.
- Throw in openLinkInNewWindow if window doesn't open


src/org/labkey/test/WebDriverWrapper.java
```diff
@@ -2872,7 +2893,11 @@ public void clickAndWait(Locator l)
 
     public void openLinkInNewWindow(WebElement link)
     {
+        int winCount = getDriver().getWindowHandles().size();
         link.sendKeys(Keys.chord(WebDriverUtils.MODIFIER_KEY, Keys.ENTER));
+        if (getDriver().getWindowHandles().size() < (winCount + 1))
+            throw new IllegalStateException("Link did not open new window in tab.");
+
         switchToWindow(1);
         waitForDocument();
```
labkey-tchad: `switchToWindow` already waits for the new window. However, this should be updated to switch to the new window instead of always using index 1.
```suggestion
        int winCount = getDriver().getWindowHandles().size();
        link.sendKeys(Keys.chord(WebDriverUtils.MODIFIER_KEY, Keys.ENTER));

        switchToWindow(winCount);
        waitForDocument();
```

src/org/labkey/test/WebDriverWrapper.java
```diff
@@ -1070,6 +1070,27 @@ protected void closeExtraWindows()
...
+            }
+            catch (UnhandledAlertException uae)
+            {
+                Optional.ofNullable(getAlertIfPresent()).ifPresent(Alert::accept);
+            }
+        }
+        if (windows.size() > 1)
+        {
+            getDriver().switchTo().window(windows.get(0));
+        }
```
labkey-tchad: Not a fan of adding a duplicate method that differs only in logging behavior.
I would agree that the existing logging isn't especially useful though. Maybe remove the `@LogMethod` annotation from `closeExtraWindows` and have it log how many windows it closes and skip logging altogether if there aren't any windows to close.
```suggestion
        if (windows.size() > 1)
        {
            log("Closed %s tabs/windows".formatted(windows.size() - 1));
            getDriver().switchTo().window(windows.get(0));
        }
```

src/org/labkey/test/WebDriverWrapper.java
labkey-martyp: So I'm trying to test all anchor tags that could function like links, whether they have a href or a click event handler attached later. And there are many anchor tags on our pages that are not links. So I'm using this function to determine if an anchor tag is a link by trying to click on it and see if a new window pops up. If so, it's a real link. If not, it throws which is caught in the crawler test and it moves on to the next anchor tag. I incorporated your suggestion above, but also created a new function, openLinkInNewWindowOrThrow more designed for this use case.

src/org/labkey/test/WebDriverWrapper.java
labkey-martyp: This was removed due to your suggestion in the ehrModules PR.

src/org/labkey/test/WebDriverWrapper.java
```diff
@@ -2872,8 +2893,21 @@ public void clickAndWait(Locator l)
...
     {
+        int winCount = getDriver().getWindowHandles().size();
         link.sendKeys(Keys.chord(WebDriverUtils.MODIFIER_KEY, Keys.ENTER));
-        switchToWindow(1);
+        switchToWindow(winCount);
+        waitForDocument();
+    }
+
+    public void openLinkInNewWindowOrThrow(WebElement link)
+    {
```
labkey-tchad, line 2882: This method is redundant. `switchToWindow` will throw if the specified window index doesn't exist. That being said, I wouldn't hate if it threw a more informative error (currently, you'd just get an `IndexOutOfBoundsException`)

labkey-tchad (changes requested):

src/org/labkey/test/WebDriverWrapper.java
labkey-martyp: yeah switchToWindow waits 10 secs before throwing that IndexOutOfBoundsException. Since we're using this as a test to discover anchor tags that are links, there will be a relative high number of failures. So openLinkInNewWindowOrThrow speeds up that check for this use case and throws a more explicit exception. If you think it's ok to refactor openLinkInNewWindow to more quickly check and throw then we could use it here without slowing down this test.

labkey-tchad: That's still a pretty misleading method name. Maybe just add timeout parameters to `openLinkInNewWindow` and `switchToWindow`

labkey-tchad (approved):

src/org/labkey/test/WebDriverWrapper.java
labkey-martyp: Ok, that would be a pretty big refactor, switchToWindow is used in a lot of places. Maybe I'm pushing too hard to use this general functionality for a specific use case it doesn't quite match. I'll close this PR and do a more focused function in ehrModules.

labkey-martyp:
Closing this PR for more specific implementation in the test.

PR #2586, opened 2025-07-28
TestDataGenerator random domain and field name max start/end char increase

cnathe:
#### Rationale
See changes from previous PR. This change to increase the random int values was kept out so that we could see what test failures were related to the last set of changes vs which are from this change.

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/2571
- https://github.com/LabKey/testAutomation/pull/2586
- https://github.com/LabKey/limsModules/pull/1552

#### Changes
- random domain and field name max start/end char increase

### DIFF
--- a/src/org/labkey/test/params/FieldDefinition.java
+++ b/src/org/labkey/test/params/FieldDefinition.java
@@ -123,8 +123,7 @@ public class FieldDefinition extends PropertyDescriptor
             }
         }
 
-        // Multiple spaces in the UI are collapsed into a single space
-        return buf.toString().replaceAll("\\s+", " ");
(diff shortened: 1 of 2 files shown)


labkey-tchad (approved):
Consider adding a maximum number of tries to those loops. If somebody accidentally passes an invalid `part` in (e.g. ":Date") they would fall into an infinite loop.

cnathe:
> Consider adding a maximum number of tries to those loops. If somebody accidentally passes an invalid `part` in (e.g. ":Date") they would fall into an infinite loop.

Good call. Done, max tries added.

PR #2589, opened 2025-07-29
coverage for Issue 52739

labkey-chrisj:
#### Rationale
This adds test coverage for [Issue 52739](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52739)
Related github issue https://github.com/LabKey/kanban/issues/949

#### Related Pull Requests
n/a

#### Changes
New test method in `PivotQueryTest.java`


labkey-jeckels (changes requested):
The bug was in how we handled the SQL for an invalid pivot query. The query here looks like it's been fixed and is running successfully.

To ensure that we cover the original issue, please remove `F1` from the `SELECT` list and ensure that you get a reasonable error message instead of a `NullPointerException`. It's fine to also check the corrected version of the query as you're doing here.

labkey-jeckels (approved):
Thanks for the update. This looks fine. FWIW, uou could also have used the executeSql API which would be a little more efficient.

src/org/labkey/test/tests/PivotQueryTest.java
```diff
@@ -182,6 +188,69 @@ public void testPivotQueryChartingTextFieldMeasure()
         assertTextPresent("The y-axis measure '" + MEASURE_COLUMN + "' had 1 value(s) that could not be converted to a number and are not included in the plot");
     }
 
+    // coverage for Issue 52739
+    @Test
+    public void testBadPivotQuery()
+    {
+        String datasetName = TestDataGenerator.randomDomainName("D2", DomainUtils.DomainKind.StudyDatasetVisit);
+        String textFieldName = TestDataGenerator.randomFieldName("F1", ":,;'\"");
```
labkey-danield: Is there a particular reason you are excluding these `:,;'\"` characters?
TestDataGenerator.randomFieldName will select a valid field name for a given domain:
`    public static String randomFieldName(@NotNull String part, @Nullable Integer numStartChars, @Nullable Integer numEndChars, @Nullable String exclusion, @Nullable DomainUtils.DomainKind domainKind)
`

labkey-danield (approved):

src/org/labkey/test/tests/PivotQueryTest.java
labkey-chrisj: I used those in an earlier test like this one, probably shouldn't here. I'll pull those out.

PR #2594, opened 2025-07-31
AuditLogHelper.checkAuditEventDiffCountForLastTransaction debugging for unexpected number of events case

cnathe:
#### Rationale
The SMProMoveAliquotTest.testMoveAliquotToParentProject test intermittent failure indicates that it is getting 3 audit events for the last transactionId when it only expects 2. I have a hunch that it is looking at the wrong transaction events. This PR adds some logging in this failure case.

#### Changes
- AuditLogHelper.checkAuditEventDiffCountForLastTransaction debugging for unexpected number of events case

### DIFF
--- a/src/org/labkey/test/util/AuditLogHelper.java
+++ b/src/org/labkey/test/util/AuditLogHelper.java
@@ -248,10 +248,14 @@ public class AuditLogHelper
     {
         Integer transactionId = getLastTransactionId(containerPath, auditEventName);
         List<Filter> transactionFilter = List.of(new Filter("TransactionId", transactionId, Filter.Operator.EQUAL));
-        int eventCount = getAuditLogsFromLKS(containerPath, auditEventName, List.of("NewRecordMap"), transactionFilter, null, ContainerFilter.CurrentAndSubfolders).getRows().size();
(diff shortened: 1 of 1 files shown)


labkey-danield (approved):

labkey-tchad (approved):
I think your hunch might be correct. Even if it isn't in this case, the fact that it's possible that it might be indicates that this probably isn't a reliable way to perform this check.

It might be prudent to change how the check works at a more fundamental level. This method should take a runnable that triggers the event that is expected to be audited. That way it can know the previous state and wait for a change after triggering the event instead of assuming that the event has already completed. Something like:
```
checkAuditEventDiffCountForLastTransaction(/*existing params*/, Runnable runnable)
{
    Integer initialTransactionId = getLastTransactionId(containerPath, auditEventName);
    runnable.run();
    // Wait for transaction with id > initialTransactionId and verify counts
}
```

cnathe:
@labkey-tchad I like your idea of adding a Runnable and checking / waiting for the new transactionId to show up in the audit event before checking. I'm going to get this small change merged and then we can look into using the `Runnable` in a separate PR.

PR #2597, opened 2025-08-01
Add selenium test for importing attachment as string values

XingY:
#### Rationale
Selenium test for Issue 53498: Sample Manager: Attachment field is lost after export/import round trip. We are no longer accepting string values for attachment fields for import. The exception is list archive import, which we do still allow attachment name being provided.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6894
- https://github.com/LabKey/limsModules/pull/1563
- https://github.com/LabKey/testAutomation/pull/2597

#### Changes
- add test for importing attachment as string from tsv file - should fail
- add test for importing list archive with attachment - should be successful

### DIFF
--- a/src/org/labkey/test/tests/InlineImagesListTest.java
+++ b/src/org/labkey/test/tests/InlineImagesListTest.java
@@ -57,8 +57,8 @@ import static org.junit.Assert.assertTrue;
 @BaseWebDriverTest.ClassTimeout(minutes = 5)
 public class InlineImagesListTest extends BaseWebDriverTest
 {
-    private final static String PROJECT_NAME = "InlineImagesListTestProject";
-    private final static String IMPORT_PROJECT_NAME = "InlineImagesListImportTestProject";
+    private final static String PROJECT_NAME = "Inline Images List Test Project";
+    private final static String IMPORT_PROJECT_NAME = "Inline Images List Import Test Project";
 
     protected final static String LIST_NAME = TestDataGenerator.randomDomainName("InlineImagesList");
     protected final static String LIST_KEY_NAME = TestDataGenerator.randomFieldName("Key");
@@ -376,11 +376,11 @@ public class InlineImagesListTest extends BaseWebDriverTest
         try
         {
             String expectedError = "Row 1: Can't upload '" + attachmentValue + "' to field " + LIST_ATTACHMENT01_NAME + " with type Attachment.";
-            isElementPresent(Locator.tagWithClass("div", "labkey-error").withText(expectedError));
+            checker().withScreenshot("import_error").verifyTrue("Invalid attachment error not as expected", isElementPresent(Locator.tagWithClass("div", "labkey-error").withText(expectedError)));
         }
         catch(NoSuchElementException nse)
         {
-            checker().fatal().error("Invalid attachment error not present.");
+            checker().error("Invalid attachment error not present.");
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/InlineImagesListTest.java
```diff
@@ -327,5 +336,51 @@ public final void testList() throws Exception
...
+    {
+        String pasteData = TestDataUtils.tsvStringFromRowMaps(List.of(Map.of(LIST_KEY_NAME, key, LIST_ATTACHMENT01_NAME, attachmentValue)),
+                List.of(LIST_KEY_NAME, LIST_ATTACHMENT01_NAME), true);
+        log(pasteData);
+        listImportPage.setText(pasteData);
+        listImportPage.submitExpectingError();
+        try
+        {
+            String expectedError = "Row 1: Can't upload '" + attachmentValue + "' to field " + LIST_ATTACHMENT01_NAME + " with type Attachment.";
+            isElementPresent(Locator.tagWithClass("div", "labkey-error").withText(expectedError));
```
labkey-susanh: Shouldn't there be an assertion attached to this? The catch will fail if there is no error, but will succeed if the error is something else, which may hide the failure for the attachment field.

src/org/labkey/test/tests/InlineImagesListTest.java
```diff
@@ -327,5 +336,51 @@ public final void testList() throws Exception
...
+        }
+        catch(NoSuchElementException nse)
+        {
+            checker().fatal().error("Invalid attachment error not present.");
```
labkey-susanh: Is this fatal? Seems like testing of the other import scenarios could still proceed (though will likely also fail).

src/org/labkey/test/tests/InlineImagesListTest.java
```diff
@@ -52,6 +57,9 @@
 @BaseWebDriverTest.ClassTimeout(minutes = 5)
 public class InlineImagesListTest extends BaseWebDriverTest
 {
+    private final static String PROJECT_NAME = "InlineImagesListTestProject";
+    private final static String IMPORT_PROJECT_NAME = "InlineImagesListImportTestProject";
```
labkey-susanh: For good measure, might want to change these to have spaces between words as that is more realistic.

labkey-susanh (changes requested):

src/org/labkey/test/tests/InlineImagesListTest.java
XingY: Good catch. fixed.

src/org/labkey/test/tests/InlineImagesListTest.java
```diff
@@ -327,5 +336,51 @@ public final void testList() throws Exception
...
+        listImportPage.setText(pasteData);
+        listImportPage.submitExpectingError();
+        try
+        {
+            String expectedError = "Row 1: Can't upload '" + attachmentValue + "' to field " + LIST_ATTACHMENT01_NAME + " with type Attachment.";
+            checker().withScreenshot("import_error").verifyTrue("Invalid attachment error not as expected", isElementPresent(Locator.tagWithClass("div", "labkey-error").withText(expectedError)));
+        }
+        catch(NoSuchElementException nse)
+        {
+            checker().error("Invalid attachment error not present.");
```
labkey-susanh, line 383:
```suggestion
            checker().withScreenshot().error("Invalid attachment error not present.");
```

labkey-susanh (approved):

PR #2603, opened 2025-08-05
Fix simpleQueryTest.js (RhinoService$TestCase) to use canonical audit…

labkey-adam:
… column names

#### Rationale
Don't use legacy column names


labkey-klum (approved):

labkey-adam:
Dailies, BVTs, and EHR BVTs are all passing

PR #2606, opened 2025-08-06
MoveToContainerDialog verifyAuditEvents to use Runnable to wait for next transactionId before doing audit diff checks

cnathe:
#### Rationale
See related PR for rationale. 

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1578

#### Changes
- AuditLogHelper.getLastTransactionId try/catch
- Add capital to NON_LATIN_STRING (for BiologicsReportTest export chart pdf scenario)

### DIFF
--- a/src/org/labkey/test/util/AuditLogHelper.java
+++ b/src/org/labkey/test/util/AuditLogHelper.java
@@ -33,6 +33,8 @@ import java.util.stream.Stream;
 import static java.lang.Integer.parseInt;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.fail;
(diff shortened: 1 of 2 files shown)


labkey-tchad (commented):
Am I remembering correctly that the error was happening when no rows are returned?
Consider waiting for an event to occur. Ignoring the exception here will likely just cause a different failure down the line.

cnathe:
> Am I remembering correctly that the error was happening when no rows are returned? Consider waiting for an event to occur. Ignoring the exception here will likely just cause a different failure down the line.

The error was when the code was getting events from the previous transaction instead of what should have been the new audit events from the move action. I updated this code being referenced to throw the exception instead of just returning null.

labkey-tchad (approved):

PR #2608, opened 2025-08-06
Add studyPublishTest to windows pg suite

labkey-chrisj:
#### Rationale
This is to investigate whether we are missing coverage on Windows; following up on curious things I'm seeing locally

#### Related Pull Requests
n/a

#### Changes
Add StudyPublishTest to the Windows PG suite


labkey-chrisj:
Ah, we're getting windows coverage for this test  on 25.7 dailyE SQL
No need for this then

PR #2609, opened 2025-08-07
Add withTextIgnoreCase, use it to verify file root message

labkey-chrisj:
#### Rationale
Locally I've experienced issues setting a pipeline file root, the problem is that the test waits for a labkeyMessage with case-sensitive text notifying the user that the pipeline root was set to a particular path- this adds a `withTextIgnoreCase `overload to `Locators `to accompany existing `startsWithIgnoreCase `and `containsIgnoreCase`, and uses it to verify in a case-insensitive way that the specified path was set.

#### Related Pull Requests
n/a

#### Changes

- add `Locator.withTextIgnoreCase`
- use it to verify `setPipelineRoot`


labkey-tchad (changes requested):
This is just the same as `containingIgnoreCase`.

labkey-tchad (approved):

PR #2614, opened 2025-08-08
Use Apache CSVFormat to quote lookup values for pasting

labkey-tchad:
#### Rationale
Introducing an explicit method for quoting lookup values for pasting into an editable grid.
Updating tests to use `EditableGrid.quoteForPaste` and `TestDataUtils.stringFromRows` will allow us to deprecate and remove `TestDataUtils.TsvQuoter` which is redundant and rarely used for TSVs anymore.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1593

#### Changes
- Use Apache CSVFormat to quote lookup values for pasting
- Deprecate `TestDataUtils.TsvQuoter` in favor of `EditableGrid.quoteForPaste`
- Update `SampleTypeTest` to use test components

### DIFF
--- a/src/org/labkey/test/tests/PivotQueryTest.java
+++ b/src/org/labkey/test/tests/PivotQueryTest.java
@@ -201,18 +201,16 @@ public class PivotQueryTest extends ReportTest
         datasetDesigner.getFieldsPanel().addFields(List.of(textField));
         var viewDatasetPage = datasetDesigner.clickSave()
                 .clickViewData();
-        TestDataUtils.TsvQuoter _tsvQuoter = new TestDataUtils.TsvQuoter();
-        String bulkData = """
-                ParticipantId	date	[F1]
-                1	7/28/2025	this
-                2	7/29/2025	that
-                3	7/30/2025	the other
-                4	7/31/2025	and more
-                5	8/1/2025	but wait
-                6	8/2/2025	still more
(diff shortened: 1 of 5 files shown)


src/org/labkey/test/tests/PivotQueryTest.java
```diff
@@ -201,18 +201,16 @@ public void testBadPivotQuery()
...
-        String bulkData = """
-                ParticipantId	date	[F1]
-                1	7/28/2025	this
-                2	7/29/2025	that
-                3	7/30/2025	the other
-                4	7/31/2025	and more
-                5	8/1/2025	but wait
-                6	8/2/2025	still more
-                """.replace("[F1]", _tsvQuoter.quoteValue(textFieldName));
+        List<List<String>> bulkData = List.of(
```
labkey-chrisj, line 204: Do we expect to need to convert many paste-bound texts in this way?  I expect there are a lot of tests out there using plain formatted TSVs

labkey-chrisj (approved):

src/org/labkey/test/tests/PivotQueryTest.java
labkey-tchad: There are definitely tons of TSV Strings getting created across our tests. I don't think we need to proactively update things that aren't a problem but we should use these helpers going forward.
That being said, we can't use them universally for editable grid pasting.

PR #2625, opened 2025-08-12
Move unescapeBackslashes to a deprecated feature

XingY:
#### Rationale
`_unescapeBackslashes` in TabLoader in app is now default to false. This PR updates the default in the TabLoader class used by test to match that the change in app.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6922
- https://github.com/LabKey/testAutomation/pull/2625
- https://github.com/LabKey/limsModules/pull/1603

#### Changes
- include backslash in charset used to generate field names and string values
- get rid of `writeRowsToTsvEscapeBackslash` and `tsvStringFromRowMapsEscapeBackslash` util as backslash should no longer be quoted


src/org/labkey/test/tests/SampleTypeNameExpressionTest.java
```diff
@@ -74,7 +74,7 @@ public class SampleTypeNameExpressionTest extends BaseWebDriverTest
     private static final String PROJECT_NAME = "SampleType_Name_Expression_Test";
 
     private static final String DEFAULT_SAMPLE_PARENT_VALUE = "SS" +
-            EscapeUtil.escapeForNameExpression(TestDataGenerator.randomString(3));
+            EscapeUtil.escapeForNameExpression(TestDataGenerator.randomString(3, "_)(}{")); // '_' is used as delimiter to get batchRandomId and (){} is used to close the defaultValue()
```
labkey-danield: I just merged my change for this.
I don't really care which one is used, they are both the same.

labkey-danield (approved):

PR #2629, opened 2025-08-14
Issue 53498: reject string attachment values for query insert/update api

XingY:
#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6928
- https://github.com/LabKey/limsModules/pull/1606
- https://github.com/LabKey/testAutomation/pull/2629

#### Changes
-


src/org/labkey/test/BaseWebDriverTest.java
```diff
@@ -2739,6 +2742,56 @@ public void dismissReleaseBanner(String productName, boolean dismiss)
             executeScript("localStorage.removeItem('" + dismissBannerKey + "')");
     }
 
+    protected void verifyQueryAPI(String schema, String dataType, Map<String, Object> row, boolean isInsert, String... errorMsg)
```
labkey-nicka, line 2745: nit: Seems better to use `SelectRowsCommand` for this sort of thing rather than relying on browser/JavaScript.

labkey-nicka (approved):

PR #2631, opened 2025-08-14
Fix tab management and random data generation

labkey-tchad:
#### Rationale
1. `doAndWaitForWindow` was implemented incorrectly. It needs to switch to the named window before calling `doAndMaybeWaitForPageToLoad` so that the page load is correctly detected.
2. When updating `TestDataGenerator.getDefaultDataSupplier` to generate random data for textChoice fields, I assumed that we weren't generating random data for other field types with a `conceptURI`. We are generating random values for some `Sample` columns so I've restored the old behavior for other column types.

#### Related Pull Requests
- #2616 
- #2615

#### Changes
- Switch to target window before waiting for page load
- Fix random data generation for sample fields

### DIFF
--- a/src/org/labkey/test/WebDriverWrapper.java
+++ b/src/org/labkey/test/WebDriverWrapper.java
@@ -148,6 +148,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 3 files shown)


labkey-tchad:
Verify fix on TC: https://teamcity.labkey.org/buildConfiguration/LabKey_257Release_Premium_CommunitySqlserver_DailyASqlserver/3625493

labkey-danield (approved):

PR #2632, opened 2025-08-14
Issue 53699: BufferUnderflow and length validator test case

labkey-nicka:
#### Rationale
Add regression test coverage for [Issue 53699](https://www.labkey.org/issues/home/Developer/issues/details.view?issueId=53699).

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6931

#### Changes
- Regression test for `BufferUnderflowException` and length validation using multibyte characters.


labkey-jeckels (approved):
@labkey-nicka please take a quick peek at my commit that adds the prefix to ensure it's hitting different byte offsets.

labkey-nicka:
> @labkey-nicka please take a quick peek at my commit that adds the prefix to ensure it's hitting different byte offsets.

I've verified the `BufferUnderflowException` occurs when running the test without the associated platform changes. Subsequently, verified test case passes with changes as well as in CI.

image

PR #2633, opened 2025-08-14
Fix Locator.xq to handle trailing quotes

labkey-tchad:
#### Rationale
[Issue 53700: Fuzz-generated names or identifiers with consecutive "" quotes](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53700)

Selenium's `Quotes.escape` doesn't handle multiple trailing quotes. This change reverts back to our old method instead of delegating to the broken method.

#### Related Pull Requests
- N/A

#### Changes
- Fix Locator.xq to handle trailing quotes


labkey-chrisj (approved):
Thanks!

PR #2648, opened 2025-08-23
Issue 53562: Cap max blob size in database configuration option

labkey-jeckels:
#### Rationale
We don't want admins to sign up for BLOB sizes that are likely to yield OutOfMemoryErrors. We can do better at validating other site settings too.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6956

#### Changes
- Try setting some bogus site setting values, ensure they're rejected


labkey-nicka (commented):
Expanded test coverage here a bit to cover the other fields you fixed up plus additional range checks.

labkey-nicka (approved):

PR #2649, opened 2025-08-25
Add fuzz names for domains to Selenium Tests - LKB/FM

labkey-tchad:
#### Rationale
https://github.com/LabKey/kanban/issues/1025

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1626

#### Changes
- Create `SchemaKey` helper to encode schema keys (similar to `FieldKey`)
- Update `ReclickingWebElement` to move mouse to the corner more aggressively
- Add methods to `EscapeUtil` to encode app resource paths
- Move `TestDataGenerator.FileRowIterator` to its own file and make more generic


src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -594,7 +594,8 @@ public static String randomDomainName(@Nullable String namePart, @Nullable Integ
         }
 
         // Multiple spaces in the UI are collapsed into a single space. If we need to test for handling of multiple spaces, we'll not use this generator
-        domainName = domainName.replaceAll("\\s+", " ");
+        // Capitalize to make it easier to match grid header labels
+        domainName = StringUtils.capitalize(domainName).replaceAll("\\s+", " ");
```
labkey-danield: I understand the why, but this i not the place for this.
We have customer domain that do begin with a lower case values: cDNA, cfDNA plasma, iGAS Isolates, iPSC Inventory and so on.

labkey-danield (approved):
I'm going approve this PR, but I need you to remove the change that capitalizes the random domain field names. We have customer data showing that they do create domains starting with lower case.

Could the changes to FieldSelectionDialog be cherry picked into 25.9? If so I think that would help with some of the failures I've seen there and in develop.

src/org/labkey/test/util/TestDataGenerator.java
labkey-tchad: Sure. My thinking was that this is reasonable because the product code is inconsistent about domain name capitalization. I'll open a product issue instead of just having the tests work around the issue (as I often remind others to do).

PR #2652, opened 2025-08-26
Results domain may be different for module assays

labkey-klum:
#### Rationale
Ran into an issue using the test helpers to update the `SignalDataRawTest`. It turns out that the built in assays and module assays have a slight difference in the name of the results domain when viewed in the domain editor. For the built in assays it is : "Results Fields" and a module assay it is : "Result Fields".


labkey-chrisj (approved):
seems reasonable

PR #2657, opened 2025-08-28
Regression for manage views exception

labkey-klum:
#### Rationale
[tracking issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53630)

This adds a simple regression test for 53630 by going to the site users table and opening the manage views page. I'm open to moving this test to a more appropriate class, it is independent of study. I could move it to the `DataViewsTest` but that is also study-based and one of those old style monolithic tests.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6971
- https://github.com/LabKey/testAutomation/pull/2657

### DIFF
--- a/src/org/labkey/test/util/DataRegionTable.java
+++ b/src/org/labkey/test/util/DataRegionTable.java
@@ -169,10 +169,7 @@ public class DataRegionTable extends DataRegion
 
     public ManageViewsPage openManageViews()
     {
-        if (!getCustomizeView().isPanelExpanded())
-        {
-            getViewsMenu().clickSubMenu(false, "Manage Views");
-        }
+        getViewsMenu().clickSubMenu(false, "Manage Views");
         return new ManageViewsPage(getDriver());
     }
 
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/util/DataRegionTable.java
```diff
@@ -166,6 +167,15 @@ public CustomizeView openCustomizeGrid()
         return getCustomizeView();
     }
 
+    public ManageViewsPage openManageViews()
+    {
+        if (!getCustomizeView().isPanelExpanded())
+        {
+            getViewsMenu().clickSubMenu(false, "Manage Views");
+        }
```
labkey-tchad: I don't think we want this to no-op if the customize view panel is open.

labkey-tchad (approved):
One suggestion.
I think this is a fine home for this test though.

src/org/labkey/test/util/DataRegionTable.java
labkey-klum: agreed

PR #2658, opened 2025-08-29
Fix domain/field name validation for trial tests

labkey-tchad:
#### Rationale
We can't validate random field and domain names when running against trial instances. We can do a test-side validation to get largely the same result.

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/2627

#### Changes
- Fix domain/field name validation for trial tests
- Capitalize random domain names

### DIFF
--- a/src/org/labkey/test/util/TestDataGenerator.java
+++ b/src/org/labkey/test/util/TestDataGenerator.java
@@ -37,6 +37,7 @@ import org.labkey.remoteapi.query.RowsResponse;
 import org.labkey.remoteapi.query.SelectRowsResponse;
 import org.labkey.remoteapi.query.Sort;
 import org.labkey.serverapi.reader.TabLoader;
+import org.labkey.test.TestProperties;
 import org.labkey.test.WebTestHelper;
 import org.labkey.test.params.FieldDefinition;
 import org.labkey.test.util.data.ColumnNameMapper;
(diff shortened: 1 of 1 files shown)


labkey-tchad:
Verify on TeamCity: https://teamcity.labkey.org/buildConfiguration/LabKey_HostingAndEvaluationInfrastructure_AutomatedTest_AutomatedTestForTrunk_TrialSmIntegrationTest/3643468

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -644,6 +646,31 @@ public static String randomFieldName(@NotNull String part, @Nullable Integer num
 
     private static boolean isDomainAndFieldNameInvalid(DomainUtils.DomainKind domainKind, @Nullable String domainName, @Nullable String fieldName)
     {
+        if (TestProperties.isTrialServer()) // WebTestHelper.getRemoteApiConnection() won't work against trial server
```
labkey-jeckels: I assume this doesn't work due to authentication issues.

Could we make the `ValidateDomainAndFieldNames` API available to guests? I see it's also a `MutatingApiAction` which surprised me. @cnathe are we close to being able to make this a `ReadOnlyApiAction` and/or open to guests?

cnathe: Having `ValidateDomainAndFieldNames` be a `MutatingApiAction` instead of `ReadOnlyApiAction` was inadvertant, copy-paste / typo on my part. I'll get that switched and a PR open.

labkey-danield: Do we need this change if/when Cory's change is merged?

labkey-tchad: Not this change, but we would need to change `isDomainAndFieldNameInvalid` to use a guest connection instead of `WebTestHelper.getRemoteApiConnection()`.

PR #2659, opened 2025-08-29
Full support for null and non-null issue required properties in API

labkey-jeckels:
#### Rationale
We haven't been properly handling nulls in issue API calls for custom fields, especially when they're required.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6974/files

#### Changes
- Ensure that Java API call sends nulls
- New test coverage to try various combinations of not sending properties and sending with nulls

### DIFF
--- a/src/org/labkey/remoteapi/issues/IssueModel.java
+++ b/src/org/labkey/remoteapi/issues/IssueModel.java
@@ -1,6 +1,7 @@
 package org.labkey.remoteapi.issues;
 
 import org.json.JSONObject;
+import org.json.JSONParserConfiguration;
 import org.labkey.api.collections.CaseInsensitiveHashMap;
 
 import java.io.File;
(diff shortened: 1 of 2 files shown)


labkey-jeckels:
@labkey-klum as a repro case, here's a wiki I created to bulk resolve/close issues. It operates on the selected issues from the grid view. It failed on labkey.org due to problems with required custom fields. Repro steps:

https://www.labkey.org/home/Developer/issues/wiki-page.view?name=Bulk%20Issue%20Resolver

1. Create an issue definition.
2. Add a custom field, making it required.
3. Insert an issue through the UI.
4. Select the issue in the grid view.
5. Go to the wiki (copied locally). Click to resolve/close.
6. Ensure the issue gets resolved and then closed, and then custom field's value is preserved.

labkey-klum (changes requested):
It looks like your `platform` changes are breaking the `IssueAPITest`. The test appears to be relying on the previous behavior where `null` values were being ignored during validation. 

The helper method : `basicIssueModel` probably shouldn't be used much in the update case.

labkey-jeckels:
Sure enough. Updates pushed. I will likely need to retarget to 25.9 but let's see what TC thinks first.

labkey-klum (approved):

PR #2674, opened 2025-09-05
Avoid known issues with random name generation

labkey-tchad:
#### Rationale
There are some known issues with random field and domain name generation.

[Issue 53706: List key field name length is limited to 64 characters](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53706). We should avoid long list key names until this issue is resolved. Limiting the length of names containing "key" should be sufficient to prevent test failures due to this issue.

Numerous SM Starter and SM Trial tests fail because they can't validate names using the `validateDomainAndFieldNames` API; either because the initial user hasn't been created or there isn't a browser session available to piggy-back. I've reimplemented the test-side name validation that we did before that API was created. We can revisit this this later but I'd like to get those tests passing; making the API work reliably would be a much more involved update. (The existing API validation can be enabled via the new `webtest.remote.domain.validation` test property)

#### Related Pull Requests
- #2571 

#### Changes
- Avoid creating long list key field names
- Stop using the server to validate random field and domain names

### DIFF
--- a/src/org/labkey/test/util/TestDataGenerator.java
+++ b/src/org/labkey/test/util/TestDataGenerator.java
@@ -37,6 +37,7 @@ import org.labkey.remoteapi.query.RowsResponse;
 import org.labkey.remoteapi.query.SelectRowsResponse;
 import org.labkey.remoteapi.query.Sort;
 import org.labkey.serverapi.reader.TabLoader;
+import org.labkey.test.TestProperties;
 import org.labkey.test.WebTestHelper;
 import org.labkey.test.params.FieldDefinition;
 import org.labkey.test.util.data.ColumnNameMapper;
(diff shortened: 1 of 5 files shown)


labkey-jeckels:
FYI, I just opened a PR to develop (due to an `exp` upgrade script already added since the 25.9 branch) to support  > 64 character key field names.

https://github.com/LabKey/platform/pull/7018

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -642,7 +644,22 @@ public static String randomFieldName(@NotNull String part, @Nullable Integer num
         return randomFieldName;
     }
 
-    private static boolean isDomainAndFieldNameInvalid(DomainUtils.DomainKind domainKind, @Nullable String domainName, @Nullable String fieldName)
+    private static boolean isDomainAndFieldNameInvalid(DomainKind domainKind, @Nullable String domainName, @Nullable String fieldName)
+    {
+        if (fieldName != null && fieldName.length() > 64 && fieldName.toLowerCase().contains("key")) // Not guaranteed but likely a list key
+            return true; // Issue 53706: List key field name length is limited to 64 characters
```
cnathe, line 650: As @labkey-jeckels mentioned, this issue will be fixed in develop very soon, so this code can be removed there.

labkey-danield: I think this is like picking your poison. This will keep the noise down in 25.9 but we will have to remember to remove this code when merging from 25.9 into develop.
Hopefully having the issue listed here will be a clue when merging into develop to not take this line.

labkey-danield (approved):

PR #2678, opened 2025-09-09
Update Amounts and Units for better sorting and filtering behavior

cnathe:
#### Rationale
See platform PR for rationale.

This PR adds some new test cases for the amounts/units updates, especially in LKS grid cases.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6994

#### Changes
- SampleTypeTest test cases for amounts and units (error when missing a…mount or unit, successful conversion, success insert/update/import)

### DIFF
--- a/src/org/labkey/test/tests/SampleTypeTest.java
+++ b/src/org/labkey/test/tests/SampleTypeTest.java
@@ -32,6 +32,7 @@ import org.labkey.remoteapi.query.SelectRowsCommand;
 import org.labkey.remoteapi.query.SelectRowsResponse;
 import org.labkey.test.BaseWebDriverTest;
 import org.labkey.test.Locator;
+import org.labkey.test.SortDirection;
 import org.labkey.test.TestFileUtils;
 import org.labkey.test.WebTestHelper;
 import org.labkey.test.categories.Daily;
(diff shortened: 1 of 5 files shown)


src/org/labkey/test/tests/SampleTypeTest.java
```diff
@@ -1922,4 +2068,36 @@ public BrowserType bestBrowser()
...
+
+    public enum TestMetricUnit implements MetricUnit
+    {
+        G("g", "g (grams)"),
+        MG("mg", "mg (milligrams)"),
+        KG("kg", "kg (kilograms)"),
+        ML("mL", "mL (milliliters)"),
+        UL("uL", "uL (microliters)"),
+        L("L", "L (liters)"),
+        UNIT("unit", "unit");
```
labkey-danield, line 2080: There is InventoryMetricUnit enum class in the Inventory module. Maybe that should be pulled out to the TestAutomation module? 

I don't think you need to change this, just tossing an idea out.

labkey-danield (approved):

src/org/labkey/test/tests/SampleTypeTest.java
cnathe: Yes, I think `InventoryMetricUnit` should be moved to testAutomation next to the interface it implements. But I am going to hold off an make that change in a separate PR since there are a lot of usages and I want to reduce the churn on this current PR.

PR #2681, opened 2025-09-10
Update tests to stop using controller-first URLs

labkey-tchad:
#### Rationale
Controller-first URLs are [now deprecated](https://github.com/LabKey/platform/pull/6990); tests should also stop using them. Specifically, tests should use `WebTestHelper.buildURL` to create URLs that match the preferred server format.

#### Related Pull Requests
- 

#### Changes
- Update many tests to use WebTestHelper to build URLs
- Only log the first time a particular controller-first URL is found

### DIFF
--- a/src/org/labkey/test/tests/ContainerContextTest.java
+++ b/src/org/labkey/test/tests/ContainerContextTest.java
@@ -33,6 +33,7 @@ import org.labkey.test.TestTimeoutException;
 import org.labkey.test.WebTestHelper;
 import org.labkey.test.categories.Daily;
 import org.labkey.test.categories.Data;
+import org.labkey.test.pages.query.ExecuteQueryPage;
 import org.labkey.test.pages.reports.ScriptReportPage;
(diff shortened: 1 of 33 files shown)


src/org/labkey/test/tests/ContainerContextTest.java
```diff
@@ -336,7 +337,9 @@ public void testSimpleModuleTables() throws Exception
 
         // Verify Issue 16243: Details URL creating URLs with null container unless the container column is actually added to current view
         log("** Removing container column and rechecking lookup URLs...");
-        beginAt("/query/" + getProjectName() + "/executeQuery.view?schemaName=vehicle&query.queryName=EmissionTest&query.sort=RowId");
+        ExecuteQueryPage queryPage = ExecuteQueryPage.getPageFactory("vehicle", "EmissionTest")
```
labkey-danield: Is this queryPage object used anyplace?

labkey-danield (approved):

PR #2685, opened 2025-09-11
Update components for conditional format testing

labkey-chrisj:
#### Rationale
This change makes minor changes to existing components to support better/easier testing of conditional formatting

#### Related Pull Requests
https://github.com/LabKey/limsModules/pull/1662

#### Changes
-make `DetailTable `aware of format pills
-make `GridRow `aware of format pills
-allow `ConditionalFormatDialog `to take a `ConditionalFormat`


labkey-danield (commented):
Looks ok.
Are you going to be adding tests here in testAutomation for the conditional format dialog?

labkey-chrisj:
> Looks ok. Are you going to be adding tests here in testAutomation for the conditional format dialog?

My plan was to add those in the current test, and do verifications both in-app and in-lks

labkey-danield (approved):

src/org/labkey/test/components/domain/ConditionalFormatPanel.java
```diff
@@ -24,51 +25,112 @@ public ConditionalFormatPanel(WebElement element, ConditionalFormatDialog dialog
...
+                setFirstCondition(filterCondition.getOperator());
+                if (filterCondition.getValue() != null)
+                    setFirstValue(filterCondition.getValue().toString());
+            }
+            if (i==1)
+            {
+                setSecondCondition(filterCondition.getOperator());
+                if (filterCondition.getValue() != null)
+                    setSecondValue(filterCondition.getValue().toString());
+            }
```
labkey-danield, line 44: Isn't this really just an if / else condition? Isn't the limit is two conditions at most? If i != 0 then you have to be setting the second one.

PR #2689, opened 2025-09-15
Add helper to get rowCount from file

labkey-chrisj:
#### Rationale
This change adds a method to testFileUtils that returns the number of lines in a file, without loading the whole file into memory

#### Related Pull Requests
https://github.com/LabKey/platform/pull/7040


src/org/labkey/test/TestFileUtils.java
```diff
@@ -118,6 +120,15 @@ public static String getFileContents(Path path)
         }
     }
 
+    public static int getFileRowCount(final File file) throws IOException
+    {
+        BufferedReader reader = new BufferedReader(new FileReader(file));
+        int lines = 0;
+        while (reader.readLine() != null) lines++;
+        reader.close();
+        return lines;
```
labkey-tchad: Use try-with-resources to ensure the reader is closed even if an exception is thrown.
```suggestion
        try (BufferedReader reader = new BufferedReader(new FileReader(file)))
        {
            int lines = 0;
            while (reader.readLine() != null) lines++;
            return lines;
        }
```

labkey-tchad (approved):

src/org/labkey/test/TestFileUtils.java
```diff
@@ -118,6 +120,16 @@ public static String getFileContents(Path path)
         }
     }
 
+    public static int getFileRowCount(final File file) throws IOException
```
labkey-danield: Maybe should return a long?

src/org/labkey/test/TestFileUtils.java
```diff
@@ -118,6 +120,16 @@ public static String getFileContents(Path path)
...
+    {
+        try (BufferedReader reader = new BufferedReader(new FileReader(file)))
+        {
+            int lines = 0;
+            while (reader.readLine() != null) lines++;
+            return lines;
+        }
```
labkey-danield: How about:

        try (Stream<String> lines = Files.lines(file.toPath())) {
            return lines.count();
        }

PR #2693, opened 2025-09-17
Remove waitForIndexerBackground from BaseWebDriverTest.waitForPendingRequests

labkey-danield:
#### Rationale
This is no longer needed, and is causing issues with tests (socket timeout failures)

#### Related Pull Requests
- None

#### Changes
- Remove waitForIndexerBackground from BaseWebDriverTest.waitForPendingRequests

### DIFF
--- a/src/org/labkey/test/BaseWebDriverTest.java
+++ b/src/org/labkey/test/BaseWebDriverTest.java
@@ -100,7 +100,6 @@ import org.labkey.test.util.UIPermissionsHelper;
 import org.labkey.test.util.core.webdav.WebDavUploadHelper;
 import org.labkey.test.util.ext4cmp.Ext4FieldRef;
 import org.labkey.test.util.query.QueryUtils;
-import org.labkey.test.util.search.SearchAdminAPIHelper;
 import org.labkey.test.util.selenium.WebDriverUtils;
 import org.openqa.selenium.By;
 import org.openqa.selenium.ElementClickInterceptedException;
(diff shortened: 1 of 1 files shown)


labkey-tchad (approved):
I thought Josh suggested using the less aggressive `waitForIndexer`, not to remove it altogether?

labkey-danield:
> I thought Josh suggested using the less aggressive `waitForIndexer`, not to remove it altogether?

This was his message in the Disco chat:
Josh Eckels, Mon 10:29 AM
@Dan Duffek I looked at the SMPipelineStatusTest failure. In the run I checked, it took 65 seconds for the search queue to clear. The test is timing out after 60 seconds. Because the test is calling waitForIndexerBackground(), it's going to wait for the indexing to finish based on the current crawling pass, not just the things that the test actively inserted or updated
Now that the indexer tosses out work connected with containers that are being deleted, my recommendation is to simply remove the call to waitForIndexerBackground()in waitForPendingRequests()

labkey-jeckels:
> > I thought Josh suggested using the less aggressive `waitForIndexer`, not to remove it altogether?
> 
> This was his message in the Disco chat: Josh Eckels, Mon 10:29 AM @dan Duffek I looked at the SMPipelineStatusTest failure. In the run I checked, it took 65 seconds for the search queue to clear. The test is timing out after 60 seconds. Because the test is calling waitForIndexerBackground(), it's going to wait for the indexing to finish based on the current crawling pass, not just the things that the test actively inserted or updated Now that the indexer tosses out work connected with containers that are being deleted, my recommendation is to simply remove the call to waitForIndexerBackground()in waitForPendingRequests()

I'm not sure we need `waitForIndexerBackground()` at all. `waitForIndexer()` is good for situations where the test makes a data change and wants to see if it's indexed correctly. There should be no need for either before deleting a container, which is the only usage I see for `waitForPendingRequests()`

labkey-jeckels (approved):

PR #2698, opened 2025-09-17
Selenium tests - Workflow: Change Required Template Fields to be Required at the start of a Job

XingY:
#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1854
- https://github.com/LabKey/labkey-ui-premium/pull/819
- https://github.com/LabKey/limsModules/pull/1674
- https://github.com/LabKey/testAutomation/pull/2698

#### Changes
- Add util for inlineEdit to set value without hitting Enter key

### DIFF
--- a/src/org/labkey/test/components/ui/edit/EditInlineField.java
+++ b/src/org/labkey/test/components/ui/edit/EditInlineField.java
@@ -40,7 +40,9 @@ public class EditInlineField extends WebDriverComponent<EditInlineField.ElementC
         // 'setFormElement' calls 'WebElement.clear()' which can close the edit-in-place input
         getWrapper().actionClear(input);
 
-        if (!skipEnter)
+        if (skipEnter)
+            input.sendKeys(value);
+        else
         {
             input.sendKeys(value, Keys.ENTER);
             getWrapper().shortWait().until(ExpectedConditions.stalenessOf(input));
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/components/ui/edit/EditInlineField.java
```diff
@@ -33,14 +33,23 @@ public WebDriver getDriver()
...
+    public void setValue(String value, boolean skipEnter)
     {
         open();
         WebElement input = elementCache().input;
         // 'setFormElement' calls 'WebElement.clear()' which can close the edit-in-place input
         getWrapper().actionClear(input);
-        input.sendKeys(value, Keys.ENTER);
-        getWrapper().shortWait().until(ExpectedConditions.stalenessOf(input));
+
+        if (!skipEnter)
```
labkey-susanh: If `skipEnter` is true, this is effectively just clearing the input, isn't it? That seems odd. Seems like it would be better to have a `clearValue` method.

labkey-susanh (changes requested):

src/org/labkey/test/components/ui/edit/EditInlineField.java
XingY: Updated method. I meant to also set the new value after clearing the existing value.

labkey-susanh (approved):

PR #2700, opened 2025-09-19
Issue 53416: Don't allow * as a field name

labkey-tchad:
#### Rationale
`TestDataGenerator.isNameInvalidLocal` should match real name restrictions as much as possible.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/7050

#### Changes
- Prevent random generation of illegal field name "*"


labkey-susanh (approved):
I don't love that we have this code repeated, but I understand.

PR #2715, opened 2025-09-30
update test to function as expected when run after other tests

labkey-chrisj:
#### Rationale
This change updates a recently-added test to deal appropriately with the case where it is not the first test to run.
Initially this test used the wikiHelper to both save and initiate edits, and wikiHelper.editWikiPageI() assumes the user is at the project root with the wiki as default-shown in the webpart-- it is looking for a span (the edit-pencil icon) to edit.

To make the test more robust, I've organized it to explicitly navigate to the wiki page for the current test's wiki, and to click on the 'Edit' link on that view.  Hopefully this forces the scenario to be repeatable regardless of whether other wikis are present on the test agent.


labkey-chrisj:
Run including this test [here](https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailySuites_DailyEPostgres?branch=wiki_undelete_test_fix), showing success

labkey-tchad (approved):

PR #2723, opened 2025-10-02
Add explicit permissions to GitHub workflows

labkey-tchad:
#### Rationale
Cleaning up several security warnings.
We should validate that the workflow permission changes work correctly before we roll out to other repositories.

#### Related Pull Requests
- N/A

#### Changes
- Add explicit permissions to GitHub workflows
- Fix ZipSlip in `TestFileUtils`
- Explicitly cast numbers

### DIFF
--- a/src/org/labkey/test/TestFileUtils.java
+++ b/src/org/labkey/test/TestFileUtils.java
@@ -632,11 +632,13 @@ public abstract class TestFileUtils
              TarArchiveInputStream inputStream = new ArchiveStreamFactory().createArchiveInputStream("tar", is))
         {
             TarArchiveEntry entry;
+            Path normalizedOutputPath = outputDir.toPath().normalize();
+
             while ((entry = inputStream.getNextEntry()) != null)
             {
                 final File outputFile = new File(outputDir, entry.getName());
 
-                if (!outputFile.toPath().normalize().startsWith(outputDir.toPath()))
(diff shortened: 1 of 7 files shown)


src/org/labkey/test/TestFileUtils.java
```diff
@@ -635,6 +635,10 @@ private static List<File> unTar(final File inputFile, final File outputDir) thro
             while ((entry = inputStream.getNextEntry()) != null)
             {
                 final File outputFile = new File(outputDir, entry.getName());
+
+                if (!outputFile.toPath().normalize().startsWith(outputDir.toPath()))
```
labkey-jeckels: This should prevent the zip slip concern, but probably best to stash `outputDir.toPath().normalize()` outside the loop and use that.

labkey-jeckels (approved):

labkey-jeckels:
@labkey-tchad what steps in the merge/release process will end up testing the workflow permissions?

labkey-tchad:
> @labkey-tchad what steps in the merge/release process will end up testing the workflow permissions?

The next time we merge changes forward from 25.10 should verify them both.

labkey-jeckels:
> > @labkey-tchad what steps in the merge/release process will end up testing the workflow permissions?
> 
> The next time we merge changes forward from 25.10 should verify them both.

@labkey-klum heads up on this change. Hopefully it won't have any impact on the merge process, but if you see problems with testAutomation, let us know

PR #2724, opened 2025-10-02
increment number to account for attachment as well as wiki

labkey-chrisj:
#### Rationale
[WikiTest.testSteps](https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyPostgres_2/3684940?buildTab=tests&name=org.labkey.test.tests.wiki.WikiTest.testSteps) is failing because when new test method testDeleteUndeleteAttachment() leaves a saved attachment in the newly-created wiki, that attachment is indexed separately and appears as a search result when testSteps() searches and expects to find a specific number of results.

This change increments that counter by one to account for the new attachment in addition to the new wiki created by the test


labkey-chrisj:
run results here https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailySuites_DailyEPostgres?branch=yet_another_wiki_fix

labkey-danield (approved):

PR #2726, opened 2025-10-06
Add EscapeUtil.escapeForExcelSheetName

labkey-danield:
#### Rationale
Add EscapeUtil.escapeForExcelSheetName to replace : / with _ when looking at sheet names in excel.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1709

#### Changes
- EscapeUtil.escapeForExcelSheetName

### DIFF
--- a/src/org/labkey/test/util/EscapeUtil.java
+++ b/src/org/labkey/test/util/EscapeUtil.java
@@ -212,9 +212,15 @@ public class EscapeUtil
     }
 
     private static final Pattern nameExpressionNeedsEscaping = Pattern.compile("([\\\\$/&}~,.])");
-    public static String escapeForNameExpression(String name)
+    public static String escapeForNameExpression(String value)
     {
-        return nameExpressionNeedsEscaping.matcher(name).replaceAll("\\\\$1");
+        return nameExpressionNeedsEscaping.matcher(value).replaceAll("\\\\$1");
+    }
+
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/util/EscapeUtil.java
```diff
@@ -212,9 +212,22 @@ public static String getMarkupEscapedValue(String value)
...
+     * Escapes invalid characters in a string to ensure it can be used as a valid Excel sheet name.
+     * Replaces characters matching the {@code excelPageNeedsEscaping} pattern with an underscore ("_").
+     *
+     * @param value the input string to be escaped
+     * @return the escaped string that can safely be used as an Excel sheet name
+     */
+    public static String escapeForExcelSheetName(String value)
+    {
+        return excelPageNeedsEscaping.matcher(value).replaceAll("_");
     }
```
labkey-tchad, line 232: There's an existing, more thorough method for this (it's what is used on the server side). `org.apache.poi.ss.util.WorkbookUtil#createSafeSheetName(java.lang.String, char)`

```suggestion
    public static String createSafeSheetName(String value)
    {
        return WorkbookUtil.createSafeSheetName(value, '_');
    }
```

Also, this isn't escaping (escaping is reversible). Consider moving this method to `ExcelHelper`.

labkey-danield: Changed.

labkey-tchad (approved):

PR #2727, opened 2025-10-06
Base classes for upgrade pipeline

labkey-tchad:
#### Rationale
https://github.com/LabKey/kanban/issues/1267
[Spec](https://docs.google.com/document/d/1a872oOxuwXB8tJTA6xjzagt2_QjcE0GShsQvqL7GI4I/edit?tab=t.0)

Adding a base class and simple proof of concept test for a TeamCity upgrade pipeline.

#### Related Pull Requests
- 

#### Changes
- Base classes for upgrade pipeline
- Update `TestUser` to handle pre-existing users
- Make some test property getters public

PR #2728, opened 2025-10-06
Base classes for upgrade pipeline

labkey-tchad:
#### Rationale
https://github.com/LabKey/kanban/issues/1267
[Spec](https://docs.google.com/document/d/1a872oOxuwXB8tJTA6xjzagt2_QjcE0GShsQvqL7GI4I/edit?tab=t.0)

Adding a base class and simple proof of concept test for a TeamCity upgrade pipeline.

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/2727

#### Changes
- Base classes for upgrade pipeline
- Update `TestUser` to handle pre-existing users
- Make some test property getters public

### DIFF
--- a/src/org/labkey/test/tests/upgrade/BaseUpgradeTest.java
+++ b/src/org/labkey/test/tests/upgrade/BaseUpgradeTest.java
@@ -1,0 +1,91 @@
+package org.labkey.test.tests.upgrade;
+
+import org.jetbrains.annotations.NotNull;
+import org.junit.Assume;
+import org.junit.BeforeClass;
+import org.junit.Rule;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+import org.labkey.test.BaseWebDriverTest;
+import org.labkey.test.TestProperties;
+import org.labkey.test.util.TestLogger;
+
+import java.util.Arrays;
+import java.util.List;
+
+public abstract class BaseUpgradeTest extends BaseWebDriverTest
+{
+
+    protected static final boolean isUpgradeSetupPhase = TestProperties.getBooleanProperty("webtest.upgradeSetup", true);
+    protected static final double previousVersion = TestProperties.getDoubleProperty("webtest.upgradePreviousVersion", 0.0);
+
+    @Override
+    protected boolean skipCleanup(boolean afterTest)
+    {
+        return afterTest || !isUpgradeSetupPhase;
+    }
+
+    @BeforeClass
+    public static void setupProject() throws Exception
+    {
+        BaseUpgradeTest currentTest = BaseWebDriverTest.getCurrentTest();
+
+        if (isUpgradeSetupPhase)
+        {
+            currentTest.doSetup();
+        }
+        else
+        {
+            TestLogger.info("Skipping setup for %s. Verifying upgrade.". formatted(currentTest.getClass().getSimpleName()));
+        }
+    }
+
+    protected abstract void doSetup() throws Exception;
+
+    @Rule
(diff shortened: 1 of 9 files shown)


src/org/labkey/test/util/TestUser.java
```diff
@@ -102,6 +119,15 @@ public String getPassword()
...
 
+    /**
+     * Create a non-impersonating API connection for the user.
+     * @return new Connection
+     */
+    public Connection getUserConnection()
+    {
+        return new Connection(WebTestHelper.getBaseURL(), getEmail(), getPassword());
+    }
+
```
labkey-danield, line 130: This change isn't in the one for develop. I'm assuming that is because it isn't needed there. That is, the user gets created in the older version which runs first.

src/org/labkey/test/tests/upgrade/BaseUpgradeTest.java
```diff
@@ -0,0 +1,91 @@
...
+        return Arrays.asList();
+    }
+
+    protected @interface EariestVersion {
+        double value();
+    }
+
+    protected @interface LatestVersion {
+        double value();
+    }
```
labkey-danield, line 93: Maybe add a small comment for these to avoid any confusion.
Does LatestVersion == current version (in develop or current release branch)?

src/org/labkey/test/tests/upgrade/BasicUpgradeTest.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.assertFalse;
+import static org.labkey.test.util.PermissionsHelper.READER_ROLE;
+
+@Category({})
+public class BasicUpgradeTest extends BaseUpgradeTest
+{
+    private static final TestUser USER = new TestUser("basic_upgrade_reader@basicupgradetest.test");
```
labkey-danield, line 23: Do you think "basic_upgrade_reader@basicupgradetest.test" should be a test, or TC,  parameter?

src/org/labkey/test/tests/upgrade/BasicUpgradeTest.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    private void queryData(Connection connection) throws IOException, CommandException
+    {
+        List<Map<String, Object>> rows = new SelectRowsCommand("samples", SAMPLE_TYPE)
+            .execute(connection, getProjectName())
+            .getRows();
+        assertFalse("Expected at least one row", rows.isEmpty());
+    }
+
```
labkey-danield, line 86: Since this is a "conventional" selenium test, this will also check the LK log to make sure there were no server errors as part of the upgrade. That's an accurate statement, correct?

labkey-danield (approved):

src/org/labkey/test/util/TestUser.java
labkey-tchad: This bit was backported from an existing change on develop.

src/org/labkey/test/tests/upgrade/BasicUpgradeTest.java
labkey-tchad: There's no more reason to parameterize this than any of the other values that are passed between the tests (e.g. project name or sample type name).

src/org/labkey/test/tests/upgrade/BasicUpgradeTest.java
labkey-tchad: That is correct. The first test that runs will check for startup/upgrade errors.

PR #2737, opened 2025-10-08
refactor testLongName to use FieldInfo

labkey-chrisj:
#### Rationale
Recently, [ListTest.testLongName](https://teamcity.labkey.org/buildConfiguration/LabKey_257Release_Community_DailyPostgres2/3691902?buildTab=tests&status=failed&name=org.labkey.test.tests.list.ListTest.testLongName) failed because it couldn't find a UI field based on the namePart given to generate a random field.  This change refactors the test to use a FieldInfo and refer to it for field information

#### Related Pull Requests
n/a

### DIFF
--- a/src/org/labkey/test/tests/list/ListTest.java
+++ b/src/org/labkey/test/tests/list/ListTest.java
@@ -545,7 +545,7 @@ public class ListTest extends BaseWebDriverTest
                 .getField(fieldWithDefault.getName())
                 .clickAdvancedSettings()
                 .clickDefaultValuesLink();
-        var input = Locator.tagContainingText("td", fieldWithDefault.getName()).followingSibling("td")
+        var input = Locator.tagContainingText("td", fieldWithDefault.getLabel()).followingSibling("td")
                 .descendant("input").findElement(page.getDriver());
         setFormElement(input, "42");
         clickButton("Save Defaults");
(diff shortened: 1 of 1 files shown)


labkey-susanh (approved):

labkey-tchad (approved):

src/org/labkey/test/tests/list/ListTest.java
```diff
@@ -533,26 +533,27 @@ public void testNameTrimming()
...
         listEditPage = _listHelper.goToEditDesign(listName);
         var page = listEditPage.getFieldsPanel()
                 .expand()
-                .getField(fieldWithDefault)
+                .getField(fieldWithDefault.getName())
                 .clickAdvancedSettings()
                 .clickDefaultValuesLink();
-        var input = Locator.tagContainingText("td", "With Default").followingSibling("td").descendant("input").findElement(page.getDriver());
+        var input = Locator.tagContainingText("td", fieldWithDefault.getName()).followingSibling("td")
+                .descendant("input").findElement(page.getDriver());
```
labkey-tchad: I think this does want the label.
```suggestion
        var input = Locator.tagContainingText("td", fieldWithDefault.getLabel()).followingSibling("td")
                .descendant("input").findElement(page.getDriver());
```
The failure here is because in 25.7 there is an off-by-one error in the random name generator that sometimes trims the first character from the name part. The failure screenshot shows the field "`안(и@ith Default?`"; note the missing "W".

labkey-tchad (changes requested):

src/org/labkey/test/tests/list/ListTest.java
labkey-chrisj: Thank you!
I'd spotted the discrepancy between label and namePart, didn't remember that this was resolved later.
I'm assuming the more-general fix would be painful to backport? If you think it's worth doing, I'm willing to do that as well

labkey-tchad (approved):

labkey-chrisj:
waiting for a successful TC run before merging https://teamcity.labkey.org/buildConfiguration/LabKey_257Release_Community_DailySuites_DailyDPostgres/3692836

PR #2750, opened 2025-10-16
Multi chart rendering

labkey-alan:
#### Rationale
This PR adds the necessary infrastructure to test for multiple charts on a grid

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1867
- https://github.com/LabKey/platform/pull/7113
- https://github.com/LabKey/limsModules/pull/1721
- https://github.com/LabKey/testAutomation/pull/2750

#### Changes
- BootstrapMenu: add menuItemIsDisabled
- QueryChartDialog: update usages of getChartPanel
- QueryChartPanelFinder: add required name parameter
- QueryGrid:
    - Add createChartIsDisbled
    - Add chartIsDisabled
    - Add chartIsSelected
    - Add name argument to closeChart, getChartPanel

### DIFF
--- a/src/org/labkey/test/components/html/BootstrapMenu.java
+++ b/src/org/labkey/test/components/html/BootstrapMenu.java
@@ -20,6 +20,7 @@ import org.labkey.test.WebDriverWrapper;
 import org.labkey.test.components.react.BaseBootstrapMenu;
 import org.labkey.test.util.LogMethod;
 import org.labkey.test.util.LoggedParam;
+import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 
@@ -120,6 +121,9 @@ public class BootstrapMenu extends BaseBootstrapMenu
 
     public boolean menuItemIsDisabled(String text)
     {
+        if (findVisibleMenuItemOrNull(text) == null)
+            throw new NoSuchElementException("Menu item not found: " + text);
+
         return findDisabledMenuItemOrNull(text) != null;
     }
 
--- a/src/org/labkey/test/components/html/BootstrapMenu.java
+++ b/src/org/labkey/test/components/html/BootstrapMenu.java
@@ -118,6 +118,11 @@ public class BootstrapMenu extends BaseBootstrapMenu
         clickSubMenu(wait ? getWrapper().getDefaultWaitForPage() : 0, subMenuLabels);
     }
 
+    public boolean menuItemIsDisabled(String text)
+    {
+        return findDisabledMenuItemOrNull(text) != null;
+    }
+
     @Override
     protected Locator getToggleLocator()
     {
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/components/ui/grids/QueryGrid.java
```diff
@@ -729,12 +730,40 @@ public QueryChartDialog createChart()
         return new QueryChartDialog("Create Chart", getDriver(), this);
     }
 
+    public boolean createChartIsDisabled()
+    {
+        elementCache().chartsMenu.expand();
+        return elementCache().chartsMenu.menuItemIsDisabled("Create Chart");
```
cnathe: couldn't this just call `chartIsDisabled("Create Chart")`?

cnathe (approved):

src/org/labkey/test/components/html/BootstrapMenu.java
```diff
@@ -118,6 +118,11 @@ public void clickSubMenu(boolean wait, @LoggedParam String ... subMenuLabels)
         clickSubMenu(wait ? getWrapper().getDefaultWaitForPage() : 0, subMenuLabels);
     }
 
+    public boolean menuItemIsDisabled(String text)
+    {
+        return findDisabledMenuItemOrNull(text) != null;
+    }
```
labkey-tchad, line 128: I'm not sure this is quite right. It will return true if the menu item is missing, which is not what I would expect. Consider throwing if the menu item isn't there.

```suggestion
    public boolean menuItemIsDisabled(String text)
    {
        if (findVisibleMenuItemOrNull(text) == null)
            throw new NoSuchElementException("Menu item not found: " + text);
        return findDisabledMenuItemOrNull(text) != null;
    }
```

src/org/labkey/test/components/react/QueryChartPanel.java
```diff
@@ -100,11 +100,13 @@ public static class QueryChartPanelFinder extends WebDriverComponentFinder<Query
...
         private final QueryGrid _queryGrid;
         private final Locator.XPathLocator _baseLocator = Locator.tagWithClass("div", "chart-panel");
+        private final String _name;
 
-        public QueryChartPanelFinder(WebDriver driver, QueryGrid queryGrid)
+        public QueryChartPanelFinder(WebDriver driver, QueryGrid queryGrid, String name)
         {
             super(driver);
             _queryGrid = queryGrid;
+            _name = name;
```
labkey-tchad, line 109: The chart name shouldn't be mandatory. The finder should find all of the charts by default and there should be a `withName` method to specify the name.
`QueryChartPanel` should probably have a `getName` method too.

labkey-tchad (approved):

src/org/labkey/test/components/react/QueryChartPanel.java
labkey-alan: But if we only ever want to find a chart by name, then why should it find all by default? Put another way, why make a default that isn't used? Same goes for QueryChartPanel, why add a getName method that isn't used?

labkey-tchad: That's a fair point, I was mostly commenting on the typical design of component finders. If we ever have other needs for this, refactoring the finder is easy enough.

PR #2769, opened 2025-10-28
Ignore test failures caused by labkey.org maintenance

labkey-tchad:
#### Rationale
Several tests check links to labkey.org. When labkey.org is undergoing maintenance, these tests fail. Such failures are not informative and just cause noise on TeamCity. We can safely skip such tests during maintenance windows.

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/2769
- https://github.com/LabKey/limsModules/pull/1765

#### Changes
- Ignore test failures caused by labkey.org maintenance

### DIFF
--- a/src/org/labkey/test/util/WebServicesUtil.java
+++ b/src/org/labkey/test/util/WebServicesUtil.java
@@ -98,7 +98,7 @@ public class WebServicesUtil
 
     /**
      * Checks whether the current page is the maintenance page for labkey.org.
-     * This is usually indicated by a 502 response code but we will accept any 5xx response.
+     * This is usually indicated by a {@code 502}-{@code 504} status code.
      * Allows tests that check links to labkey.org to not fail during maintenance periods.
      *
      * @param driver The WebDriver instance to check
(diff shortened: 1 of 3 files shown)


src/org/labkey/test/util/WebServicesUtil.java
```diff
@@ -91,4 +95,27 @@ static public Socket openSocket(String hostIp, int port) throws IOException
...
+     * @return true if the current page is the maintenance page for labkey.org, false otherwise.
+     */
+    public static boolean isLabKeyDotOrgMaintenance(WebDriver driver)
+    {
+        try
+        {
+            URL url = new URL(Objects.requireNonNull(driver.getCurrentUrl()));
+            String title = driver.getTitle();
+            int responseCode = Integer.parseInt(Objects.requireNonNull(title).substring(0, 3));
+            return url.getHost().endsWith("labkey.org") && responseCode >= 500 && responseCode < 600;
```
labkey-danield: Is there any concern that this might be too generous? For example should it not ignore 511 errors?

labkey-danield (approved):

src/org/labkey/test/util/WebServicesUtil.java
labkey-tchad: My thought was that any 500 error is a server-side error and these tests aren't responsible for validating labkey.org. I suppose there's no reason to be this broad though. I'll narrow it down to 502-504. I think I've seen all of those at various times.

PR #2776, opened 2025-11-04
Fix various intermittent test issues

labkey-tchad:
#### Rationale
- Some tests will leave file watchers enabled if they fail. These can show up as memory leaks to subsequent tests. We can disable all pipeline triggers via API to prevent this.
- Tooltips for the editable grid headers are very sensitive. When attempting to inspect error popovers, the mouse often passes over the header. Occasionally, this triggers a header tooltip which blocks the cell we actually want to inspect. We can move the mouse in a path that will not pass over the grid header to avoid this.
- The query metadata form can take more than two seconds to load. We can give it a little extra time.

#### Related Pull Requests
- N/A

#### Changes
- Disable pipeline triggers after failed tests
- Avoid mousing over editable grid headers
- Wait a little longer for DomainFormPanel to load

### DIFF
--- a/modules/simpletest/resources/scripts/validationTest/actionUrlTest.js
+++ b/modules/simpletest/resources/scripts/validationTest/actionUrlTest.js
@@ -21,7 +21,7 @@ function doTest()
 
     if (contextPath.length > 0) {
         var baseUrl = LABKEY.ActionURL.getBaseURL();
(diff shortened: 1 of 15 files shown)


labkey-susanh (approved):
Approving in Dan's stead, pending, of course, successful test runs.

PR #2779, opened 2025-11-05
Handle paths, not just file names, when extracting zips in test code

labkey-jeckels:
#### Rationale
We need to handle paths when extracting from a zip file, not just simple file names

#### Changes
- Accept paths too


labkey-jeckels:
Test to verify: https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_ProductSuites_Biologics_LimsStarter_LimsStarterB/3726062

labkey-tchad (approved):

PR #2780, opened 2025-11-08
Fixes for Windows failures after FileLike refactor

labkey-jeckels:
#### Rationale
https://teamcity.labkey.org/buildConfiguration/LabKey_2511Release_Premium_CommunitySqlserver_DailyASqlserver/3728152

#### Changes
- Choose a pipeline root that encompasses all of the files being referenced

### DIFF
--- a/src/org/labkey/test/tests/ExpTest.java
+++ b/src/org/labkey/test/tests/ExpTest.java
@@ -87,12 +87,12 @@ public class ExpTest extends BaseWebDriverTest
                 });
         clickButton("Setup");
         // Don't upload file. Uploading file creates 'exp.Data' entries that we don't want.
-        setPipelineRoot(TestFileUtils.getSampleData("xarfiles/expVerify").getAbsolutePath());
+        setPipelineRoot(TestFileUtils.getSampleData("xarfiles").getAbsolutePath());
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
Verification in FB: https://teamcity.labkey.org/buildConfiguration/LabKey_2511Release_Premium_CommunitySqlserver_DailyASqlserver/3728516

labkey-tchad (approved):
Why is this only failing on Windows; shouldn't Linux fail in the same way?
Does this indicate that there is a backwards compatibility issue with XAR file import?

labkey-jeckels:
> Why is this only failing on Windows; shouldn't Linux fail in the same way? Does this indicate that there is a backwards compatibility issue with XAR file import?

Yes, it should, but there was a problem related to `/` vs `\` so we skipped an intended check on non-Windows. Fixed in the related platform PR.

labkey-jeckels:
> Why is this only failing on Windows; shouldn't Linux fail in the same way? Does this indicate that there is a backwards compatibility issue with XAR file import?

As for backwards compatibility, yes, there's a potential concern, but hopefully not in the real world. The test was setting pipeline root in a place that didn't cover the whole tree of imported files, which shouldn't be common.

PR #2806, opened 2025-12-08
Avoid some UI test issues.

labkey-danield:
#### Rationale
Addressing some failures in SsampleTypeRenameTest

[TeamCity run](https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyPostgres/3762837).

#### Related Pull Requests
- None

#### Changes
- Dismiss the grid menu.
- Refresh page to check for link to sample type.

### DIFF
--- a/src/org/labkey/test/tests/SampleTypeRenameTest.java
+++ b/src/org/labkey/test/tests/SampleTypeRenameTest.java
@@ -21,6 +21,7 @@ import org.labkey.test.util.SampleTypeHelper;
 import org.labkey.test.util.TestDataGenerator;
 import org.labkey.test.util.exp.SampleTypeAPIHelper;
 import org.labkey.test.util.search.SearchAdminAPIHelper;
+import org.openqa.selenium.Keys;
 import org.openqa.selenium.WebElement;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/SampleTypeRenameTest.java
```diff
@@ -173,6 +179,9 @@ public void testCustomViewWithSampleTypeRename()
                 .verifyTrue(String.format("Doesn't look like custom view '%s' was saved. Fatal error.", customViewName),
                 menuItems.contains(customViewName));
 
+        // Dismiss the menu so it doesn't get in the way.
+        menu.findVisibleMenuItems().get(0).sendKeys(Keys.ESCAPE);
```
labkey-tchad: Would the menu's collapse method work?
```suggestion
        // Dismiss the menu so it doesn't get in the way.
        menu.collapse();
```

labkey-tchad (approved):

src/org/labkey/test/tests/SampleTypeRenameTest.java
labkey-danield: Changed.

PR #2807, opened 2025-12-08
Update GridBar.setAliquotView to be more reliable

labkey-danield:
#### Rationale
Update GridBar.setAliquotView to be more reliable

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1843

#### Changes
- Do nothing if already on the current view.
- Wait for grid update after setting the view.

### DIFF
--- a/src/org/labkey/test/components/ui/grids/GridBar.java
+++ b/src/org/labkey/test/components/ui/grids/GridBar.java
@@ -302,6 +302,11 @@ public class GridBar extends WebDriverComponent<GridBar.ElementCache>
      */
     public void setAliquotView(AliquotViewOptions view)
     {
+
+        // Do nothing if currently on the requested menu item.
+        if (getCurrentAliquotView().equals(view))
+            return;
+
         // Need to identify where we are. The menu text is contextual to the page.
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/components/ui/grids/GridBar.java
```diff
@@ -356,8 +362,8 @@ else if(onSamplePage)
                 break;
         }
 
-        doMenuAction(currentButtonText, Arrays.asList(menuChoice));
-
+        String finalMenuChoice = menuChoice;
+        _queryGrid.doAndWaitForUpdate(()->doMenuAction(currentButtonText, Arrays.asList(finalMenuChoice)));
```
labkey-tchad: Creating this extra `final` variable isn't necessary. `menuChoice` can be made final by removing its initial assignment and adding a `default` branch to the view switch.

labkey-tchad (approved):

src/org/labkey/test/components/ui/grids/GridBar.java
labkey-danield: Changed.

PR #2815, opened 2025-12-13
Set headless flag earlier in test initialization

labkey-tchad:
#### Rationale
Forgot to check the test results from my previous "fix". AWT locks in headless mode at some point early so we need to set the flag as early as we can.

#### Related Pull Requests
- #2814 

#### Changes
- Set headless flag earlier in test initialization


labkey-tchad:
TeamCity verification: https://teamcity.labkey.org/buildConfiguration/bt311/3771615?buildTab=overview

labkey-danield (approved):

PR #2834, opened 2026-01-05
Add convenience methods for conditionalizing upgrade checks

labkey-tchad:
#### Rationale
Some upgrade checks need to be adjusted depending on what version was being upgraded from.

#### Related Pull Requests
- N/A

#### Changes
- Add convenience methods for conditionalizing upgrade checks

### DIFF
--- a/src/org/labkey/test/tests/upgrade/BaseUpgradeTest.java
+++ b/src/org/labkey/test/tests/upgrade/BaseUpgradeTest.java
@@ -70,6 +70,16 @@ public abstract class BaseUpgradeTest extends BaseWebDriverTest
         return Arrays.asList();
     }
 
+    protected boolean isUpgradingFromBefore(String version)
+    {
+        return !isUpgradingFrom(version, null);
+    }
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/tests/upgrade/BaseUpgradeTest.java
```diff
@@ -70,6 +70,16 @@ public List<String> getAssociatedModules()
         return Arrays.asList();
     }
 
+    protected boolean isUpgradingFromBefore(String version)
```
labkey-susanh: Javadoc here would be helpful.

labkey-danield (approved):

labkey-susanh (approved):

src/org/labkey/test/tests/upgrade/BaseUpgradeTest.java
labkey-tchad: Done. Also added to other methods and related classes

PR #2843, opened 2026-01-13
Add action to avoid tooltip and fix flaky

DariaBod:
#### Rationale
Added action to avoid flaky test because of tooltip.

#### Related Pull Requests
- 

#### Changes
- moveToElement(title) to avoid tooltip from table header

[Run](https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_SampleManager_SampleManagerPostgres/3804544?buildTab=tests&status=failed)

[New run](https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_SampleManager_SampleManagerPostgres?branch=test_editablegrid&buildTypeTab=overview)

### DIFF
--- a/src/org/labkey/test/components/ui/grids/EditableGrid.java
+++ b/src/org/labkey/test/components/ui/grids/EditableGrid.java
@@ -26,7 +26,6 @@ import org.openqa.selenium.By;
 import org.openqa.selenium.Keys;
 import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.NotFoundException;
-import org.openqa.selenium.TimeoutException;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.interactions.Actions;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/components/ui/grids/EditableGrid.java
```diff
@@ -992,8 +992,12 @@ public void selectCellRange(WebElement startCell, WebElement endCell)
     private void dragToCell(WebElement elementToDrag, WebElement destinationCell)
     {
         var size = destinationCell.getSize();
+        Locator.XPathLocator selectionHandleLoc = Locator.byClass("no-margin-top detail__header--name");
+        WebElement title = selectionHandleLoc.findElement(getDriver());
```
labkey-tchad: I don't think all editable grids have this sort of header. Also, components should usually find elements only within themself (`findElement(this)` instead of `findElement(getDriver())`).

That being said, there is an existing method in `EditableGrid` to dismiss the tooltip.

```suggestion
        dismissPopover();
```

labkey-tchad (changes requested):

labkey-danield (approved):

labkey-tchad (approved):

PR #2853, opened 2026-01-21
Make uiTests task work with Gradle 9.3

labkey-tchad:
#### Rationale
Something in Gradle 9.3 changed how it detects JUnit tests. Gradle only recognizes test classes that meet [certain criteria](https://docs.gradle.org/current/userguide/java_testing.html#sec:test_detection). Adding a `RunWith` annotation to our test runner is the most straightforward.
Additionally `scanForTestClasses` needs to be enabled, though this seems to be a direct contradiction of the Gradle documentation.
```
Execution failed for task ':server:testAutomation:ciTestsPostgres18'. org.gradle.api.internal.exceptions.MarkedVerificationException: There are test sources present and no filters are applied, but the test task did not discover any tests to execute. This is likely due to a misconfiguration. Please check your test configuration. If this is not a misconfiguration, this error can be disabled by setting the 'failOnNoDiscoveredTests' property to false.
```

#### Related Pull Requests
- https://github.com/LabKey/server/pull/1261

#### Changes
- Make uiTests task work with Gradle 9.3
- Use lazy task configuration

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ plugins {
     id 'maven-publish'
 }
 
-import org.labkey.gradle.task.RunUiTest
+import org.labkey.gradle.task.RunTestSuite
 import org.labkey.gradle.util.BuildUtils
 import org.labkey.gradle.util.GroupNames
 import org.labkey.gradle.util.PomFileHelper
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -170,8 +171,15 @@ project.tasks.register("convertHarToStressXml", JavaExec) {
...
-project.tasks.uiTests.dependsOn(initPropertiesTask)
-project.parent.parent.tasks.ijConfigure.dependsOn(initPropertiesTask)
+project.tasks.named("uiTests").configure {
+    dependsOn(initPropertiesTask)
+}
+project.parent.parent.tasks.named("ijConfigure").configure {
+    dependsOn(initPropertiesTask)
+}
+project.tasks.withType(RunTestSuite).configureEach { it ->
+    scanForTestClasses = true // Required for Gradle 9.3
```
labkey-susanh, line 181: I think this is required because we are using an older version of JUnit. If I'm understanding correctly, "JUnit Platform" was introduced with JUnit 5. So, could make a note here about possibly removing this if we update our version of JUnit.

labkey-susanh (approved):
Gradle is usually pretty good about warning about breaking changes like this. Probably I missed something in previous release notes...

PR #2857, opened 2026-01-22
Rlabkey test for makeFilter

labkey-klum:
#### Rationale
Automated tests for LabKey filter generation for both legacy and future formats.

### DIFF
--- a/data/api/rlabkey-api-query.xml
+++ b/data/api/rlabkey-api-query.xml
@@ -625,4 +625,47 @@
             ]]>
         </response>
     </test>
+    <test name="make filters" type="post">
+        <url>
+            <![CDATA[
+                library(Rlabkey)
+                makeFilter(
+                    c("foo/Reagent", "CONTAINS", "this or that"),
+                    c("Temp_C", "GREATER_THAN", "12"),
+                    c("SystolicBloodPressure", "MISSING", ""))
+            ]]>
+        </url>
+        <response>
+            <![CDATA[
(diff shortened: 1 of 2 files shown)


data/api/rlabkey-api-query.xml
```diff
@@ -625,4 +625,47 @@
...
+        <response>
+            <![CDATA[
+                [1] "query.foo%2FReagent~contains=this%20or%20that"
+                [2] "query.Temp_C~gt=12"
+                [3] "query.SystolicBloodPressure~isblank="
+            ]]>
+        </response>
+    </test>
+<!--
+    Uncomment when Rlabkey 3.4.5 is available
```
cnathe: The tests on TeamCity should run with the latest code in the package, so you should be able to uncomment this new test case now.

cnathe (changes requested):

data/api/rlabkey-api-query.xml
cnathe: nevermind. I didn't see the comment from the other PR until now.

cnathe (approved):

data/api/rlabkey-api-query.xml
labkey-klum: No you are right, the tests will build and install what is in the repo (not CRAN). So I don't need to wait for the publish. It does mean that this may fail locally for some devs but I don't think that is a concern.

PR #2861, opened 2026-01-28
Add test to check package-lock.json files for suspicious dependencies

labkey-tchad:
#### Rationale
By requiring all NPM dependencies to come from registry.npmjs.org or labkey.jfrog.io, we can reduce the risk of downloading malicious code.

#### Related Pull Requests
- N/A

#### Changes
- Add test to check package-lock.json files for suspicious dependencies


labkey-alan (approved):
Looks good to me.

PR #2865, opened 2026-01-30
Update test dependencies

labkey-martyp:
#### Rationale
Bump some dependencies

#### Changes
- assertj
- jetty
- selenium

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,15 +2,15 @@ apacheCommonsCsvVersion=1.14.1
 
 aspectjVersion=1.9.25.1
 
-assertjVersion=3.27.6
+assertjVersion=3.27.7
 
 awaitilityVersion=4.3.0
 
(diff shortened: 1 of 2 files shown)


labkey-adam (approved):

labkey-adam:
I think Selenium upgrades often require reworking tests and helpers. I'll let @labkey-tchad comment on that. Other bumps look fine to me.

labkey-martyp:
@labkey-tchad there do appear to be a bunch of test failures associated with this but it looks mostly due to reading dropdown values in the domain designer. Let me know if you'd like me to investigate or back out any of these updates.

labkey-tchad (approved):
There was a slight change to how `ExpectedConditions.attributeToBe` handles blank attributes. I've updated the two places that use that method. It should be good now.

PR #2866, opened 2026-01-31
Test Automation for Issue 783

labkey-danield:
#### Rationale
Automation to cover [Server lockup when updating data class domain design](https://github.com/LabKey/internal-issues/issues/783)

#### Related Pull Requests
- None.

#### Changes
- Added SampleTypeDesignerStressTest.
- Added code to FieldDefinition to get the ValueExpression.
- Added code to DomainFormPanel to set the ValueExpression if it is set in a field definition.

### DIFF
--- a/src/org/labkey/test/tests/SampleTypeTest.java
+++ b/src/org/labkey/test/tests/SampleTypeTest.java
@@ -38,6 +38,7 @@ import org.labkey.test.WebTestHelper;
 import org.labkey.test.categories.Daily;
 import org.labkey.test.components.CustomizeView;
 import org.labkey.test.components.assay.AssayConstants;
+import org.labkey.test.components.bootstrap.ModalDialog;
 import org.labkey.test.components.domain.AdvancedSettingsDialog;
 import org.labkey.test.components.domain.BaseDomainDesigner;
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/tests/SampleTypeTest.java
```diff
@@ -2012,6 +2015,77 @@ private void verifySampleAmountUnitValues(String name, String expectedAmount, St
...
+
+        sampleTypeHelper.goToSampleType(sampleTypeName);
+
+        for (int i = 0; i < 5; i++)
+        {
+            waitAndClickAndWait(Locator.lkButton("Edit Type"));
+            UpdateSampleTypePage domainDesignerPage = new UpdateSampleTypePage(getDriver());
+            DomainFormPanel fieldsPanel = domainDesignerPage.getFieldsPanel();
+
+            for (FieldDefinition fd : fields)
```
XingY: Could we also add a new column during the update?

labkey-danield: Test is updated to add new columns, and not change types of the existing columns.

XingY (approved):

PR #2868, opened 2026-02-03
Make CSP error logging validation more robust

labkey-jeckels:
#### Rationale
WikiCspTest has been failing in a feature branch because the error is logged a little later than expected.

https://teamcity.labkey.org/buildConfiguration/LabKey_2511Release_Community_ModuleSuites_WikiSqlserver/3826615

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/7373

#### Changes
- Ensure unique container path to avoid log de-duplication
- Wait for the CSP error to log before proceeding
- Make sure the error gets logged

### DIFF
--- a/src/org/labkey/test/tests/wiki/WikiCspTest.java
+++ b/src/org/labkey/test/tests/wiki/WikiCspTest.java
@@ -21,7 +21,9 @@ import java.util.List;
 @BaseWebDriverTest.ClassTimeout(minutes = 2)
 public class WikiCspTest extends BaseWebDriverTest
 {
-    private static final String PROJECT_NAME = TRICKY_CHARACTERS_FOR_PROJECT_NAMES + "WikiCspTest";
+    // Add something unique to make sure log de-duping doesn't suppress the logging. Otherwise, running the test multiple
+    // times against the same server won't actually log the CSP error after the first hit
+    private static final String PROJECT_NAME = TRICKY_CHARACTERS_FOR_PROJECT_NAMES + "WikiCspTest" + System.currentTimeMillis();
     private static final String WIKI_PAGE_TITLE = "TOC_with_inline";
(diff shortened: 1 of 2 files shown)


labkey-jeckels:
@labkey-tchad the suite passed after this test update. 

https://teamcity.labkey.org/buildConfiguration/LabKey_2511Release_Community_ModuleSuites_WikiSqlserver/3832854

I am pretty confident this was an intermittent problem, but it failed twice in the FB before this change.

src/org/labkey/test/tests/wiki/WikiCspTest.java
```diff
@@ -21,7 +21,9 @@
 @BaseWebDriverTest.ClassTimeout(minutes = 2)
 public class WikiCspTest extends BaseWebDriverTest
 {
-    private static final String PROJECT_NAME = TRICKY_CHARACTERS_FOR_PROJECT_NAMES + "WikiCspTest";
+    // Add something unique to make sure log de-duping doesn't suppress the logging. Otherwise, running the test multiple
+    // times against the same server won't actually log the CSP error after the first hit
+    private static final String PROJECT_NAME = TRICKY_CHARACTERS_FOR_PROJECT_NAMES + "WikiCspTest" + System.currentTimeMillis();
```
labkey-tchad: We disable deduping in other tests by enabling debug logging for the CSP warning endpoint.
You can do so by calling `CspConfigHelper.debugCspWarnings();` somewhere in the test setup.

labkey-tchad (approved):
One suggestion

src/org/labkey/test/tests/wiki/WikiCspTest.java
labkey-jeckels: Good idea. I switched to this approach. I'll merge once it passes on TeamCity.

PR #2872, opened 2026-02-05
Tests for multi choice value

DariaBod:
#### Rationale
Some changes to support multi choice

#### Related Pull Requests
- [Tests for multi choice value#1971
](https://github.com/LabKey/limsModules/pull/1971)

#### Changes
- Added setAllowMultipleSelections method to check “Allow multiple selections” checkbox 
- Changed method initFilterColumn for work with filters for List values 
- Added method testMultiChoiceValues() to ListTest (scope: create, edit) 
- Created new shuffleSelect() for random amount of random values from list 
- Changed method selectOptionByText() for multi choice values. If we need value to be selected we first check if it’s selected and then click


src/org/labkey/test/components/domain/DomainFormPanel.java
```diff
@@ -236,6 +236,10 @@ else if (validator instanceof FieldDefinition.TextChoiceValidator textChoiceVali
                         throw new IllegalArgumentException("TextChoice fields cannot have additional validators.");
                     }
                     fieldRow.setTextChoiceValues(textChoiceValidator.getValues());
+                    if(textChoiceValidator.getMultipleSelections())
+                    {
+                        fieldRow.setAllowMultipleSelections(textChoiceValidator.getMultipleSelections());
+                    }
```
labkey-tchad: This would make it impossible to disable multi-select through this method.
This should either call `setAllowMultipleSelections` unconditionally or the condition should check whether `textChoiceValidator.getMultipleSelections() != null`.

src/org/labkey/test/components/ui/search/FilterFacetedPanel.java
```diff
@@ -48,6 +49,15 @@ public void selectValue(String value)
...
     }
 
+    /**
+     * Select a single facet value by clicking its label. Should replace all existing selections.
+     * @param value desired value
+     */
+    public void selectFilter(String value)
+    {
+        elementCache().filterTypeSelects.select(value);
+    }
```
labkey-tchad, line 60: Update this comment to describe this method (looks leftover from the `selectValue`). The comment should also mention that this is only relevant to multi-value text choice fields.

This could also take a `Filter.Operator` instead of a `String` for some extra type-safety.

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
```diff
@@ -234,15 +240,18 @@ public String filterColumnExpectingError(CharSequence columnIdentifier, Filter.O
...
         GridFilterModal filterModal = new GridFilterModal(getDriver(), this);
         if (operator != null)
         {
-            if (operator.equals(Filter.Operator.IN) && value instanceof List<?>)
+            if (listOperators.contains(operator) && value instanceof List<?>)
             {
                 List<String> values = (List<String>) value;
                 filterModal.selectFacetTab().selectValue(values.get(0));
                 filterModal.selectFacetTab().checkValues(values.toArray(String[]::new));
+                filterModal.selectFacetTab().selectFilter(operator.getDisplayValue());
```
labkey-tchad: I think this will (or could) break functionality for other column types; the facet panel usually doesn't have a filter type.
Also, this wouldn't work for "Is Empty" or "Is Not Empty".

src/org/labkey/test/pages/query/UpdateQueryRowPage.java
```diff
@@ -186,7 +185,7 @@ WebElement findField(String name)
         {
             if (!fieldMap.containsKey(name))
             {
-                fieldMap.put(name, Locator.name(EscapeUtil.getFormFieldName(name)).findElement(this));
+                fieldMap.put(name, Locator.nameContaining(EscapeUtil.getFormFieldName(name)).findElement(this));
```
labkey-tchad: Use a more precise locator and add a comment explaining why this can't match the name exactly.
```suggestion
                // Multi-value text choice fields are prepended with "[]"
                fieldMap.put(name, Locator.tag("*").attributeStartsWith("name", EscapeUtil.getFormFieldName(name)).findElement(this));
```

src/org/labkey/test/pages/DatasetInsertPage.java
```diff
@@ -82,7 +82,7 @@ private void tryInsert(Map<String, String> values)
     {
         for (Map.Entry<String, String> entry : values.entrySet())
         {
-            WebElement fieldInput = Locator.name(EscapeUtil.getFormFieldName(entry.getKey())).findElement(getDriver());
+            WebElement fieldInput = Locator.nameContaining(EscapeUtil.getFormFieldName(entry.getKey())).findElement(getDriver());
```
labkey-tchad: Use a more precise locator and add a comment explaining why this can't match the name exactly. (See `UpdateQueryRowPage`)

src/org/labkey/test/params/FieldDefinition.java
```diff
@@ -475,6 +475,13 @@ public FieldDefinition setTextChoiceValues(List<String> values)
         return this;
     }
 
+    public FieldDefinition setMultiChoiceValues(List<String> values)
+    {
+        Assert.assertEquals("Invalid field type for text choice values.", ColumnType.TextChoice, getType());
+        setValidators(List.of(new FieldDefinition.TextChoiceValidator(values).setMultipleSelections()));
+        return this;
+    }
```
labkey-tchad, line 483: There should be a separate `ColumnType` for Multi Choice. That would make our API helpers able to deal with Multi-choice columns and it more closely matches the product design. It would also make `setMultiChoiceValues` redundant.

src/org/labkey/test/tests/list/ListTest.java
```diff
@@ -1688,6 +1689,42 @@ public void testAutoIncrementKeyEncoded()
...
+                .setMultiChoiceValues(tcValues));
+        _listHelper.goToList(encodedListName);
+
+        DataRegionTable table = new DataRegionTable("query", getDriver());
+        table.clickInsertNewRow();
+        String valuesToChoose = tcValues.subList(1, 3).stream()
+                .sorted()
+                .collect(Collectors.joining(" "));
+        Locator loc = Locator.nameContaining(EscapeUtil.getFormFieldName(columnName));
+        selectOptionByText(loc, valuesToChoose);
```
labkey-tchad: `clickInsertNewRow` returns a page object that should have methods for setting fields.

src/org/labkey/test/components/domain/DomainFormPanel.java
labkey-danield: This will only set it if textChoiceValidator.getMultipleSelections is true, correct? This is ok if you are creating a field, but wouldn't work as expected if the test is editing an existing field and wants to change to field from a multi-choice to a single choice. 
Perhaps a better check would be:
if(null != textChoiceValidator.getMultipleSelections())

src/org/labkey/test/components/ui/grids/EditableGrid.java
```diff
@@ -507,11 +507,14 @@ public WebElement setCellValue(int row, CharSequence columnIdentifier, Object va
...
         if (value instanceof List)
         {
+
             // If this is a list assume that it will need a lookup.
             List<String> values = (List) value;
 
             ReactSelect lookupSelect = elementCache().lookupSelect(gridCell);
 
+            lookupSelect.clearSelection();
+
```
labkey-danield: Will this work if a test is trying to add to an existing selections?

src/org/labkey/test/tests/list/ListTest.java
```diff
@@ -1688,6 +1689,42 @@ public void testAutoIncrementKeyEncoded()
         _listHelper.deleteList();
     }
 
+    @Test
+    public void testMultiChoiceValues()
+    {
+        // setup a list with an auto-increment key and multi text choice field
+        String encodedListName = "multiChoiceList";
+        String keyName = "'><script>alert(\":(\")</script>'";
+        String columnName = "MultiChoiceField";
```
labkey-tchad: Consider using randomly generated field and list names. (I'm not sure whether ListHelper can handle the tricky characters but it's worth a try)

```suggestion
        String encodedListName = TestDataGenerator.randomDomainName("multiChoiceList", DomainUtils.DomainKind.IntList);
        String keyName = TestDataGenerator.randomFieldName("'><script>alert(\":(\")</script>'");
        String columnName = TestDataGenerator.randomFieldName("MultiChoiceField");
```

src/org/labkey/test/tests/list/ListTest.java
```diff
@@ -1688,6 +1689,42 @@ public void testAutoIncrementKeyEncoded()
...
+        DataRegionTable table = new DataRegionTable("query", getDriver());
+        table.clickInsertNewRow();
+        String valuesToChoose = tcValues.subList(1, 3).stream()
+                .sorted()
+                .collect(Collectors.joining(" "));
+        Locator loc = Locator.nameContaining(EscapeUtil.getFormFieldName(columnName));
+        selectOptionByText(loc, valuesToChoose);
+
+        clickButton("Submit");
+        checker().verifyEquals("Multi choice value not as expected", valuesToChoose, table.getDataAsText(0, columnName));
```
labkey-tchad: `checker()` doesn't take screenshots automatically. If you're doing a single check, `checker().withScreenshot().verifyX` will trigger a failure screenshot. If you're doing several checks, you should call `checker().screenShotIfNewError();` after the checks.
```suggestion
        checker().withScreenshot().verifyEquals("Multi choice value not as expected", valuesToChoose, table.getDataAsText(0, columnName));
```

src/org/labkey/test/tests/list/ListTest.java
```diff
@@ -1688,6 +1689,42 @@ public void testAutoIncrementKeyEncoded()
...
+
+        table.clickEditRow(0);
+        valuesToChoose = tcValues.subList(1, 3).stream()
+                .sorted()
+                .collect(Collectors.joining(" "));
+        selectOptionByText(loc, valuesToChoose);
+        clickButton("Submit");
+
+        // verify the multi choice value is persisted
+        checker().verifyEquals("Multi choice value not as expected", valuesToChoose, table.getDataAsText(0, columnName));
```
labkey-tchad: Take a screenshot.
```suggestion
        checker().withScreenshot().verifyEquals("Multi choice value not as expected", valuesToChoose, table.getDataAsText(0, columnName));
```

labkey-danield (commented):
Added a few comments.

src/org/labkey/test/WebDriverWrapper.java
```diff
@@ -4059,6 +4059,13 @@ public void selectOptionByText(Locator locator, String text)
 
     public void selectOptionByText(WebElement selectElement, String value)
     {
+        if(Boolean.parseBoolean(selectElement.getAttribute("multiple"))) {
+            List<WebElement> elems = selectElement.findElements(Locator.tag("option"));
+            elems.forEach(element->{
+                if(value.contains(element.getAttribute("value")) ^ element.isSelected()) element.click();
+            });
+            return;
+        }
```
labkey-tchad: This is less precise than we like for broadly available helpers; it would get confused by selection options that have overlapping text. I would suggest having a separate method for multi-select (`selectOptionsByText(WebElement selectElement, List<String> values)`.

src/org/labkey/test/tests/list/ListTest.java
```diff
@@ -1688,6 +1689,42 @@ public void testAutoIncrementKeyEncoded()
         _listHelper.deleteList();
     }
 
+    @Test
+    public void testMultiChoiceValues()
+    {
```
labkey-tchad, line 1695: We should only run the test on Postgres.
```suggestion
    public void testMultiChoiceValues()
    {
        Assume.assumeTrue("Multi-choice text fields are only supported on PostgreSQL", WebTestHelper.getDatabaseType() == WebTestHelper.DatabaseType.PostgreSQL);
```

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
DariaBod: I’m planning to add support for 'Is Empty' and 'Is Not Empty' as well.

Regarding the functionality for other types: I’ve already found one test that was failing, so I added a check to the initFilterColumn method. It now checks if the filter type selector is present on the page before trying to interact with it. After this change, the test is passing.

Does this approach sound okay to you?

src/org/labkey/test/params/FieldDefinition.java
XingY: Trey is correct that we should use different `ColumnType` for multi choice. Though not obvious on UI, MVTC and TC are very different fields with distinct field definition. It uses rangeUri:
`ColumnType MultiValueTextChoice = new ColumnTypeImpl("Text Choice (Multi-Value)", "http://cpas.fhcrc.org/exp/xml#multiChoice");`

src/org/labkey/test/components/ui/grids/EditableGrid.java
labkey-tchad: I would expect a method called `setCellValue` to overwrite any existing value.

src/org/labkey/test/components/ui/search/FilterFacetedPanel.java
```diff
@@ -48,6 +52,23 @@ public void selectValue(String value)
...
+    public boolean isFiltersPresented()
+    {
+        return waitFor(() -> isVisible(elementCache().filterTypeSelects), 1000);
+    }
+
+    /**
+     * Select a filer by clicking its label. Right now this method relevant only for multi-value text choice.
+     * @param operator desired filter value
+     */
+    public void selectFilter(Filter.Operator operator)
```
labkey-tchad: A slightly more descriptive method name.
```suggestion
    public void selectArrayFilterOperator(Filter.Operator operator)
```

src/org/labkey/test/components/ui/search/FilterFacetedPanel.java
```diff
@@ -123,6 +144,8 @@ protected class ElementCache extends Component<?>.ElementCache
     {
         protected final Input filterInput =
                 Input(Locator.id("filter-faceted__typeahead-input"), getDriver()).findWhenNeeded(this);
+        protected final ReactSelect filterTypeSelects =
+                new ReactSelect.ReactSelectFinder(getDriver()).index(0).findWhenNeeded(this);
```
labkey-tchad: Making name consistent with method rename suggestion.
```suggestion
        protected final ReactSelect arrayFilterOperatorSelect =
                new ReactSelect.ReactSelectFinder(getDriver()).index(0).findWhenNeeded(this);
```

src/org/labkey/test/components/ui/search/FilterFacetedPanel.java
```diff
@@ -48,6 +52,23 @@ public void selectValue(String value)
         elementCache().findCheckboxLabel(value).click();
     }
 
+    /**
+     * Check that filter choosing option exists on the page.
+     */
+    public boolean isFiltersPresented()
+    {
+        return waitFor(() -> isVisible(elementCache().filterTypeSelects), 1000);
```
labkey-tchad: We shouldn't wait for things that might not be present, especially when the most common state is for it to not be present.

src/org/labkey/test/pages/query/UpdateQueryRowPage.java
```diff
@@ -108,6 +114,19 @@ public UpdateQueryRowPage setField(String fieldName, String value)
...
+    {
+        WebElement field = elementCache().findField(fieldName, true);
+        List<WebElement> options = field.findElements(By.tagName("option"));
+       //unselect all options that selected but shouldn't be selected
+        options.forEach(option ->
+        {
+            if (!values.contains(option.getText()) && option.getAttribute("selected") != null) option.click();
+        });
+        values.forEach(value -> selectOptionByText(field, value));
+        return this;
```
labkey-tchad: WebDriver's `Select` class handles multi-select.
```suggestion
        WebElement field = new Select(elementCache().findField(fieldName, true));
        field.deselectAll();
        values.forEach(value -> field.selectByVisibleText(value));
        return this;
```

src/org/labkey/test/pages/query/UpdateQueryRowPage.java
```diff
@@ -108,6 +114,19 @@ public UpdateQueryRowPage setField(String fieldName, String value)
         return this;
     }
 
+    public UpdateQueryRowPage setField(String fieldName, List<String> values)
```
labkey-tchad, line 118: This can replace the existing `setMultiValueField` that was added last week. It only has one usage and doesn't actually support multiple values.

src/org/labkey/test/components/domain/DomainFormPanel.java
```diff
@@ -236,6 +236,17 @@ else if (validator instanceof FieldDefinition.TextChoiceValidator textChoiceVali
...
+                }
+                else if (validator instanceof FieldDefinition.MultiValueTextChoiceValidator multiValueTextChoiceValidator)
+                {
+                    // MultiValueTextChoice is a field type; implemented using a special validator. TextChoice field cannot have other validators.
+                    if (validators.size() > 1)
+                    {
+                        throw new IllegalArgumentException("TextChoice fields cannot have additional validators.");
+                    }
+                    fieldRow.setTextChoiceValues(multiValueTextChoiceValidator.getValues());
+                    fieldRow.setAllowMultipleSelections(true);
```
labkey-tchad: `MultiValueTextChoiceValidator` is redundant. We can just check the column type to decide whether to allow multiple selections.
```suggestion
                    fieldRow.setAllowMultipleSelections(fieldDefinition.getType() == FieldDefinition.ColumnType.MultiValueTextChoice);
                }
```

src/org/labkey/test/params/FieldDefinition.java
```diff
@@ -1145,6 +1153,49 @@ public List<String> getValues()
 
     }
 
+    /**
+     * TextChoice is implemented using a validator, however it is more 'limited' in scope. The user does not name a TextChoice
+     * validator or add a description or error message. A TextChoice is a lot like a look-up field, but it is not linked
+     * to an external data source. The user only provides the list of (string) values that the field will display in the dropdown.
+     * Another difference is that there can only be one TextChoice on a field, whereas you can have multiple validators on a field.
+     */
+    public static class MultiValueTextChoiceValidator extends FieldValidator<MultiValueTextChoiceValidator>
```
labkey-tchad: `MultiValueTextChoiceValidator` is redundant and should be removed.

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
```diff
@@ -234,18 +243,31 @@ public String filterColumnExpectingError(CharSequence columnIdentifier, Filter.O
 
     private GridFilterModal initFilterColumn(CharSequence columnIdentifier, Filter.Operator operator, Object value)
     {
+        List<Filter.Operator> listOperators = List.of(IN, CONTAINS_ALL, CONTAINS_ANY, CONTAINS_EXACTLY, CONTAINS_NONE,
+                DOES_NOT_CONTAIN_EXACTLY);
```
labkey-tchad: `IN` isn't actually a list operator (this component just uses it as a signal to use the checkboxes). It shouldn't be grouped with them. This list should also be static and contain the empty operators.

```suggestion
public static final List<Filter.Operator> ARRAY_OPERATORS = List.of(CONTAINS_ALL, CONTAINS_ANY, CONTAINS_EXACTLY, CONTAINS_NONE, DOES_NOT_CONTAIN_EXACTLY, IS_EMPTY, IS_NOT_EMPTY);

private GridFilterModal initFilterColumn(CharSequence columnIdentifier, Filter.Operator operator, Object value)
{
```

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
```diff
@@ -234,18 +243,31 @@ public String filterColumnExpectingError(CharSequence columnIdentifier, Filter.O
...
+                filterPanel.checkValues(values.toArray(String[]::new));
+                if (filterPanel.isFiltersPresented())
+                {
+                    filterPanel.selectFilter(operator);
+                }
+            }
+            else if (value == null)
+            {
+                filterModal.selectFacetTab().selectFilter(operator);
             }
```
labkey-tchad, line 265: Now that I have a better understanding of these new filter operators, I have a better idea for this method. We don't actually need to check whether the operator select is present; for the new operators, the select will always be present.
Something like this:
```
if (operator.equals(Filter.Operator.IN) && value instanceof List<?> || ARRAY_OPERATORS.contains(operator))
{
    FilterFacetedPanel filterPanel = filterModal.selectFacetTab();
    if (ARRAY_OPERATORS.contains(operator))
    {
        filterPanel.selectFilter(operator);
    }
    if (value != null)
    {
        List<String> values = (List<String>) value;
        filterPanel.selectValue(values.get(0));
        filterPanel.checkValues(values.toArray(String[]::new));
    }
}
```

labkey-tchad (changes requested):
I think Xing is going to work on the Java remote API. This PR is blocked by that update.

XingY:
> I think Xing is going to work on the Java remote API. This PR is blocked by that update.

https://github.com/LabKey/labkey-api-java/pull/88

src/org/labkey/test/components/ui/grids/ResponsiveGrid.java
```diff
@@ -234,18 +243,30 @@ public String filterColumnExpectingError(CharSequence columnIdentifier, Filter.O
 
     private GridFilterModal initFilterColumn(CharSequence columnIdentifier, Filter.Operator operator, Object value)
     {
+        List<Filter.Operator> listOperators = List.of(IN, ARRAY_CONTAINS_ALL, ARRAY_CONTAINS_ANY, ARRAY_CONTAINS_EXACT, ARRAY_CONTAINS_NONE,
+                ARRAY_CONTAINS_NOT_EXACT, ARRAY_ISEMPTY, ARRAY_ISNOTEMPTY);
```
labkey-tchad: `IN` isn't an array operator and won't have the same UI in the filter dialog. Also, make the List of operators static so that it doesn't get recreated with each call to this method.
```suggestion

private static final List<Filter.Operator> ARRAY_OPERATORS = List.of(ARRAY_CONTAINS_ALL, ARRAY_CONTAINS_ANY, ARRAY_CONTAINS_EXACT, ARRAY_CONTAINS_NONE,
                ARRAY_CONTAINS_NOT_EXACT, ARRAY_ISEMPTY, ARRAY_ISNOTEMPTY);
                
    private GridFilterModal initFilterColumn(CharSequence columnIdentifier, Filter.Operator operator, Object value)
    {
```

labkey-tchad (approved):
One small change you missed.

labkey-danield (approved):

PR #2881, opened 2026-02-10
Issue 53534: Support Microsoft's Graph Protocol API for authentication via OAuth2 token

labkey-bpatel:
#### Rationale
The related PRs adds Microsoft Graph as an alternative mail transport using OAuth2 authentication. Since Dumbster is an SMTP-only mail recorder, it should only start when SMTP is the active transport.

#### Related Pull Requests
- https://github.com/LabKey/server/pull/1276
- https://github.com/LabKey/platform/pull/7381

#### Changes
- Start Dumbster when SMTP is the active transport.


modules/dumbster/src/org/labkey/dumbster/DumbsterModule.java
```diff
@@ -65,7 +67,7 @@ public WebPartView<?> getWebPartView(@NotNull ViewContext portalCtx, @NotNull Po
     @Override
     public void doStartup(ModuleContext moduleContext)
     {
-        if (AppProps.getInstance().isMailRecorderEnabled())
+        if (MailHelper.getActiveProvider() instanceof SmtpTransportProvider && AppProps.getInstance().isMailRecorderEnabled())
```
labkey-tchad, line 70: Maybe add this check to `DumbsterManager.start` to prevent it from being started by other means (e.g. `DumbsterController.SetRecordEmailAction`). It should also log some sort of error.

labkey-tchad (changes requested):

modules/dumbster/src/org/labkey/dumbster/DumbsterModule.java
labkey-bpatel: Good point. Done!

labkey-tchad (approved):

PR #2887, opened 2026-02-16
GitHub Issue 465: Add auditing capabilities for grid views

XingY:
#### Rationale
Add test coverage for create/update/delete named/default private / shared custom grid views.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/7421
- https://github.com/LabKey/testAutomation/pull/2887

#### Changes
-

### DIFF
--- a/src/org/labkey/test/tests/component/GridPanelViewTest.java
+++ b/src/org/labkey/test/tests/component/GridPanelViewTest.java
@@ -226,24 +226,38 @@ public class GridPanelViewTest extends GridPanelBaseTest
 
         log("Verify audit event for shared default view save.");
         // if the current default view is a private view, 2 events will be generated, one for the creation of the new shared default view, and one for deletion of the existing view.
-        // If the current default view is already a shared view, then only 1 event will be generated for the update of the default view.
+        // If the current default view is already a shared view, then only 1 event will be generated for the update of the default view. If no current view, a view will be created.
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/tests/component/GridPanelViewTest.java
```diff
@@ -207,9 +218,34 @@ private void resetDefaultView(String sampleTypeName, List<String> columns)
...
+        log("Get baseline audit row ID before saving the shared default view.");
+        Integer defaultViewBaselineRowId = _auditLogHelper.getLatestAuditRowId("GridViewAuditEvent");
+
         // Save as default view for everyone.
         cv.saveCustomView("", true);
 
+        log("Verify audit event for shared default view save.");
+        List<Map<String, Object>> defaultViewAuditRows = getGridViewAuditEvents(defaultViewBaselineRowId);
+        checker().verifyTrue("Expected at least 1 audit event for shared default view save.",
+                !defaultViewAuditRows.isEmpty());
```
cnathe, line 232: is there ever a valid case where there would be more than 1 audit event row? if not, could this just be changed to an assert that the list length is 1?

cnathe (approved):

src/org/labkey/test/tests/component/GridPanelViewTest.java
XingY: I've added comment that explains why there could be 1 or 2 events. If the current default view for this user is "private", saving a "shared" default view will actually be 2 events. One for creating the shared, one for deleting the old private view.

PR #2890, opened 2026-02-20
Ensure consistent XML parser config

labkey-jeckels:
#### Rationale
We want to be using our centrally defined XML parser configuration

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/7440

#### Changes
- Use preconfigured XML factories
- Minor cleanup

### DIFF
--- a/src/org/labkey/test/util/PipelineAnalysisHelper.java
+++ b/src/org/labkey/test/util/PipelineAnalysisHelper.java
@@ -17,6 +17,7 @@ package org.labkey.test.util;
 
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.labkey.api.util.XmlBeansUtil;
 import org.labkey.test.BaseWebDriverTest;
 import org.labkey.test.Locator;
 import org.labkey.test.TestFileUtils;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/util/PipelineAnalysisHelper.java
```diff
@@ -78,7 +78,7 @@ public void runPipelineAnalysis(@LoggedParam String importAction, String[] files
         for (String file : files)
         {
             String directory = file.substring(0, file.lastIndexOf("/") + 1);
-            String fileName = file.substring(file.lastIndexOf("/") + 1, file.length());
+            String fileName = file.substring(file.lastIndexOf("/") + 1);
```
labkey-danield, line 81: Was the `file.lenght()` parameter removed because it was redundant / not needed?
I'm fine with it, I just want to make sure I'm not missing something.

labkey-danield (approved):

src/org/labkey/test/util/PipelineAnalysisHelper.java
labkey-jeckels: Correct, it's not needed. IntelliJ suggested it to clean up lint. If you look at `String.substring(int)`, you'll see that it just calls the two-arg version with `length()` as the second argument.

PR #2894, opened 2026-02-24
Add waiting for isChecked for DataViewsTest

DariaBod:
#### Rationale
Fix for test DataViewsTest.testSteps add waiting for isChecked

### DIFF
--- a/src/org/labkey/test/tests/DataViewsTest.java
+++ b/src/org/labkey/test/tests/DataViewsTest.java
@@ -471,7 +471,9 @@ public class DataViewsTest extends ParticipantListTest
 
         // check if top category sorts Alphabetical
         openCustomizePanel(ORIGINAL_WEBPART_TITLE);
-        RadioButton().withLabel("Alphabetical").find(getDriver()).check();
+        var alphabeticalRadio = RadioButton().withLabel("Alphabetical").find(getDriver());
+        alphabeticalRadio.check();
+        waitFor(alphabeticalRadio::isChecked, "Radio button 'Alphabetical' did not become checked after clicking.", 2000);
         clickButton("Save", 0);
         _ext4Helper.waitForMaskToDisappear();
         assertTextPresentInThisOrder(CATEGORIES[0], "Abbreviated Demographics", "Alt ID mapping", "APX-1: Abbreviated Physical Exam",


labkey-tchad (commented):
This doesn't seem related to the failure on TeamCity:
```
java.lang.AssertionError: Text 'Type:' was not present
  at org.labkey.test.WebDriverWrapper.assertTextPresent(WebDriverWrapper.java:1731)
  at org.labkey.test.WebDriverWrapper.assertTextPresent(WebDriverWrapper.java:1720)
  at org.labkey.test.WebDriverWrapper.waitForText(WebDriverWrapper.java:2684)
  at org.labkey.test.WebDriverWrapper.waitForText(WebDriverWrapper.java:2678)
  at org.labkey.test.tests.DataViewsTest.refreshDateTest(DataViewsTest.java:302)
  at org.labkey.test.tests.DataViewsTest.doVerifySteps(DataViewsTest.java:95)
  at org.labkey.test.tests.StudyBaseTest.runUITests(StudyBaseTest.java:150)
  at org.labkey.test.tests.StudyBaseTest.testSteps(StudyBaseTest.java:360)
```
Was this just an error you were hitting locally?

DariaBod:
@labkey-tchad 
I'm having trouble reproducing the TeamCity error locally. But this problem appears every second run.
Should I just ignore this because we have no failures on TeamCity?

labkey-tchad:
> @labkey-tchad I'm having trouble reproducing the TeamCity error locally. But this problem appears every second run. Should I just ignore this because we have no failures on TeamCity?

I definitely think it's worth making this more reliable for folks running the test locally. I was just commenting that rationale in the PR description isn't accurate.
If you aren't having any luck reproducing the failure, I think it's fine to just merge this fix. Tooltip checks like that can be hard to nail down and it looks like it last failed in this way last October.

labkey-danield (approved):
Echoing what Trey said that I think it is worthwhile making this test more reliable when running locally. 

I would check in this change, and let's see how it does on TeamCity.

If the failure is very infrequent, once every couple of months, I don't think it is worth spending any more time on it. There are other intermittent failures that happen more often, especially in the Apps, that would be better to investigate and fix.

PR #2896, opened 2026-02-25
GitHub Issue 875: Standard Assay Multi-File Transform Import Skips First Data Row

cnathe:
#### Rationale
[#875](https://github.com/LabKey/internal-issues/issues/875) Luminex Multi-File Transform Import Skips First Data Row

See related PR for rationale. This PR updates the Standard assay test case to fuzz whether or not a no-op R transform script is added to the assay design so that we test both code paths for an assay import (with and without transform script).

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/7456
- https://github.com/LabKey/commonAssays/pull/986

#### Changes
- GpatAssayTest to randomlyAddTransformScript for testing GitHub Issue 875


labkey-klum (approved):
I forgot about that option, looks good.

labkey-danield (approved):

PR #2897, opened 2026-02-26
Fix for export tests

DariaBod:
#### Rationale
Fix for export tests, changed the " " symbol processing.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/2013


labkey-tchad (changes requested):
TestDataUtils.parseMultiValueText should return a `List`. Turning the List into a String loses information and returns possibly ambiguous values.
`"a,b"` could be one value or two, there's no way to be certain.

PR #2903, opened 2026-03-03
Add sample finder test for multi choice field

DariaBod:
#### Rationale
Add sample finder test for multi choice field

#### Related Pull Requests
- [https://github.com/LabKey/limsModules/pull/2021](https://github.com/LabKey/limsModules/pull/2021)

#### Changes
- create TestArrayDataUtils to move duplicated methods for multi choice


src/org/labkey/test/util/data/TestArrayDataUtils.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.labkey.remoteapi.query.Filter;
+
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class TestArrayDataUtils
+{
```
labkey-tchad, line 20: Consider moving `TestDataUtils.parseMultiValueText` and `ResponsiveGrid.ARRAY_OPERATORS` into this class.

src/org/labkey/test/util/data/TestArrayDataUtils.java
```diff
@@ -0,0 +1,43 @@
...
+        return switch (type)
+        {
+            case ARRAY_CONTAINS_ALL -> actualValues.containsAll(searchValues);
+            case ARRAY_CONTAINS_ANY -> searchValues.stream().anyMatch(actualValues::contains);
+            case ARRAY_CONTAINS_EXACT -> actualValues.size() == searchValues.size() && actualValues.containsAll(searchValues);
+            case ARRAY_CONTAINS_NONE -> searchValues.stream().noneMatch(actualValues::contains);
+            case ARRAY_CONTAINS_NOT_EXACT -> !(actualValues.size() == searchValues.size() && actualValues.containsAll(searchValues));
+            case ARRAY_ISEMPTY -> actualValues.isEmpty();
+            case ARRAY_ISNOTEMPTY -> !actualValues.isEmpty();
+            default -> true;
```
labkey-tchad: Should this default to `true` for non-array operators? What does the server do?
Unless this matches the actual product behavior, we should just throw an `IllegalArgumentException` for non-array operators.

src/org/labkey/test/util/data/TestArrayDataUtils.java
```diff
@@ -0,0 +1,43 @@
...
+    {
+        return sourceMap.entrySet().stream()
+                .filter(entry -> isMatch(entry.getValue(), searchValues, filterType))
+                .collect(Collectors.toMap(
+                        Map.Entry::getKey,
+                        e -> e.getValue().stream().sorted().collect(Collectors.joining(", ")),
+                        (e1, e2) -> e1,
+                        LinkedHashMap::new
+                ));
+    }
```
labkey-tchad: Avoid losing information. `filterMap` should return the filtered map in the same format that it was passed in, `Map<String, List<String>>`. 

Maybe create a separate method to sort and join the values in a standard format. Something like:
```
public static String formatValues(List<String> values)
{
    return TextUtils.normalizeSpace(values.stream().sorted().collect(Collectors.joining(", "));
}
```

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -491,6 +499,16 @@ public static String randomString(int size)
         return randomString(size, null);
     }
 
+    public static List<String> randomTextChoice(int size)
+    {
+        List<String> textChoices = new ArrayList<>();
+        for (int i = 0; i < size; i++)
+        {
+            textChoices.add(randomString(randomInt(1, 25)));
+        }
```
labkey-tchad: This should trim the generated values and ensure that there aren't any duplicate or blank values.

src/org/labkey/test/components/ui/search/FilterFacetedPanel.java
```diff
@@ -74,9 +77,15 @@ public boolean isChecked(String value)
...
+       // waitForElement(elementCache().findCheckbox(values[0]));
+       // waitFor(() -> elementCache().findCheckbox(values[0]).isDisplayed(), 10000);
         for (String value : values)
         {
-            elementCache().findCheckbox(value).check();
+            try{
+            elementCache().findCheckbox(value).check();}
+            catch (Exception e){
+                elementCache().findCheckbox(value).check();
+            }
```
labkey-tchad: Consider making `ElementCache.findCheckbox` wait for the checkbox instead of attempting to find it twice here.

labkey-tchad (approved):

src/org/labkey/test/util/data/TestArrayDataUtils.java
```diff
@@ -0,0 +1,105 @@
...
+                .map(entry -> Map.entry(entry.getKey(), (List<String>) entry.getValue()))
+                .filter(entry -> isMatch(entry.getValue(), searchValues, filterType))
+                .collect(Collectors.toMap(
+                        Map.Entry::getKey,
+                        e -> e.getValue().stream()
+                                .sorted(Comparator
+                                        .comparing((String s) -> s.substring(0, 1).toLowerCase())
+                                        .thenComparing(s -> s.substring(0, 1))
+                                        .thenComparing(s -> s))
+                                .collect(Collectors.toList()),
```
labkey-tchad, line 58: Add some documentation explaining this special sorting.

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -183,6 +183,14 @@ else if (fieldDefinition.getType().equals(FieldDefinition.ColumnType.TextChoice)
                         else
                             entityData.put(key, textChoices.get(textChoiceIndex));
                     }
+                    else if (fieldDefinition.getType().equals(FieldDefinition.ColumnType.MultiValueTextChoice))
+                    {
+                        FieldDefinition.TextChoiceValidator validator =
+                                (FieldDefinition.TextChoiceValidator) fieldDefinition.getValidators().getFirst();
+                        List<String> values = shuffleSelect(validator.getValues());
+                        Collections.sort(values);
```
labkey-tchad: We shouldn't pre-sort the generated data. The server should be able to handle any order the values are input. The only time the test should sort the values is when generating the expected display string.
```suggestion
```

labkey-tchad (approved):
A couple more requests

src/org/labkey/test/components/ui/search/FilterFacetedPanel.java
DariaBod: This method should be the same as it was. Used this for debugging and forgot to delete it, my bad

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -491,6 +498,16 @@ public static String randomString(int size)
...
     }
 
+    public static List<String> randomTextChoice(int size)
+    {
+        List<String> textChoices = new ArrayList<>();
+        for (int i = 0; i < size; i++)
+        {
+            textChoices.add(randomString(randomInt(1, 25)));
+        }
+        return textChoices;
```
labkey-tchad: This is likely to be used to generate Lists that shouldn't be changed. Use `Collections.unmodifiableList` to prevent modification.
```suggestion
        return Collections.unmodifiableList(textChoices);
```

PR #2918, opened 2026-03-20
LKSM: Multivalue text choice automation: Assays

DariaBod:
#### Rationale
Added method for MVTC formatting and couple of other helper methods.

#### Related Pull Requests
- [https://github.com/LabKey/limsModules/pull/2052](https://github.com/LabKey/limsModules/pull/2052)
- [https://github.com/LabKey/platform/pull/7511](https://github.com/LabKey/platform/pull/7511)

### DIFF
--- a/src/org/labkey/test/pages/assay/AssayRunsPage.java
+++ b/src/org/labkey/test/pages/assay/AssayRunsPage.java
@@ -113,6 +113,7 @@ public class AssayRunsPage extends LabKeyPage<AssayRunsPage.ElementCache>
         elementCache().manageMenu.doMenuAction("Edit Assay Design");
     }
 
+    //Method for Standard assays only.
     public AssayImportPage clickImportData()
     {
         clickAndWait(Locator.linkWithText("Import Data"));
--- a/src/org/labkey/test/pages/assay/AssayRunsPage.java
+++ b/src/org/labkey/test/pages/assay/AssayRunsPage.java
@@ -113,6 +113,12 @@ public class AssayRunsPage extends LabKeyPage<AssayRunsPage.ElementCache>
         elementCache().manageMenu.doMenuAction("Edit Assay Design");
     }
 
+    public AssayImportPage clickImportData()
+    {
+        clickAndWait(Locator.linkWithText("Import Data"));
+        return new AssayImportPage(getDriver());
+    }
+
     @Override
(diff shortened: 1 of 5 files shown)


src/org/labkey/test/util/data/TestArrayDataUtils.java
```diff
@@ -45,19 +47,35 @@ public static <T> Map<String, List<String>> filterMap(Map<String, T> map, List<S
...
     }
 
+    /**
+     * Standard alphabetical sort that accounts for symbols and numbers.
+     * Uppercase letters are positioned before lowercase letters.
+     */
+    public static List<String> sortValues(List<String> values)
+    {
+        return values.stream()
+                .filter(s -> !s.isEmpty())
```
labkey-tchad: This should probably throw an exception if one of the values is empty. If an empty value ends up in here, something probably went wrong upstream.
```suggestion
                .peek(s -> { if (s.isEmpty()) throw new IllegalArgumentException("Empty values aren't allowed: " + values);})
```

src/org/labkey/test/pages/assay/AssayRunsPage.java
```diff
@@ -113,6 +113,12 @@ public void clickEditAssayDesign()
         elementCache().manageMenu.doMenuAction("Edit Assay Design");
     }
 
+    public AssayImportPage clickImportData()
```
labkey-tchad, line 117: I think this method didn't exist before because many assay types have custom import pages that aren't compatible with the standard `AssayImportPage`. I think it's fine to have this method but it should be documented that it only works for Standard assays.

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -504,7 +504,7 @@ public static List<String> randomTextChoice(int size)
         Set<String> textChoices = new LinkedHashSet<>();
         while (textChoices.size() < size)
         {
-            String generated = randomString(randomInt(1, 25)).trim();
+            String generated = randomString(randomInt(1, 25), "%<>").trim();
```
labkey-tchad: Why can't text choices include these characters? Any special exclusions should be documented.

labkey-tchad (changes requested):

labkey-tchad (approved):

PR #2922, opened 2026-03-25
26.3 fb investigate windows failures

labkey-danield:
#### Rationale
The [AssayTransformMissingParentDirTest.testMissingParentDirectoryRegression](https://teamcity.labkey.org/buildConfiguration/LabKey_263Release_Premium_CommunitySqlserver_DailyCSqlserver/3903946?buildTab=tests&status=failed&name=AssayTransformMissingParentDirTest.testMissingParentDirectoryRegression) is failing on windows because the expected error message is not shown.
The expected error message is not shown because the directory that contains the R script is not deleted as expected. It looks like windows holds onto the directory a little longer.
This avoids the issue on windows by using a retry mechanism to delete the directory.

#### Related Pull Requests
- None.

#### Changes
- If the directory is locked (AccessDenied) use a retry method to try and delete it again.


labkey-jeckels (approved):
Might want to attempt the delete multiple times with delays in-between if it fails since the original failure looks to be caused by Windows locking the file from it still being open by some process. But sounds like this is passing consistently already so feel free to delay that for future troubleshooting, if needed.

PR #2923, opened 2026-03-25
Encode form name

labkey-matthewb:
#### Rationale
Support an encoding scheme for INPUT.name to avoid problems with multipart/form-data

#### Related Pull Requests
- https://github.com/LabKey/labkey-api-js/pull/211
- https://github.com/LabKey/labkey-ui-components/pull/1961
- https://github.com/LabKey/labkey-ui-premium/pull/916
- https://github.com/LabKey/limsModules/pull/2061
- https://github.com/LabKey/platform/pull/7515
- https://github.com/LabKey/testAutomation/pull/2923

#### Changes
-

### DIFF
--- a/src/org/labkey/test/util/EscapeUtil.java
+++ b/src/org/labkey/test/util/EscapeUtil.java
@@ -282,7 +282,7 @@ public class EscapeUtil
      */
     public static String getFormFieldName(String columnName, @Nullable String prefix)
     {
-        String name = prefix + columnName;
+        String name = Objects.toString(prefix,"") + columnName;
 
         final String escapeChar = "%";
         final String problemChars = "\\\"";
--- a/src/org/labkey/test/util/EscapeUtil.java
+++ b/src/org/labkey/test/util/EscapeUtil.java
@@ -15,6 +15,7 @@
  */
 package org.labkey.test.util;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.text.StringEscapeUtils;
 import org.apache.poi.ss.util.WorkbookUtil;
 import org.eclipse.jetty.util.URIUtil;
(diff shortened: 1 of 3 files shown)


src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -670,16 +670,16 @@ public static String randomFieldName(@NotNull String part, @Nullable Integer num
...
         {
-            randomFieldName = randomName(part, getNumChars(numStartChars, 5), getNumChars(numEndChars, 50), chars, exclusion);
+            randomFieldName = randomName(part, getNumChars(numStartChars, 5), getNumChars(numEndChars, 48), chars, exclusion);
             if (++currentTries >= MAX_RANDOM_TRIES)
                 throw new IllegalStateException("Failed to generate a valid field name after " + MAX_RANDOM_TRIES + " tries. Last generated name: " + randomFieldName);
         }
 
         TestLogger.log("Generated random field name for domainKind " + _domainKind + ": " + randomFieldName);
-        return randomFieldName.name();
+        return randomFieldName.name() + "\"\'";
```
labkey-tchad: Is this temporary to test in your branch or intentionally left in?
If we want this included in the long or medium-term, these characters should be appended to the names generated on lines 673 and 676 so that they are included in the name validation (`isDomainAndFieldNameInvalid`)

src/org/labkey/test/util/EscapeUtil.java
```diff
@@ -271,20 +272,25 @@ public static String getFormFieldName(String columnName, boolean multiValue)
...
 
-        return prefix == null ? fieldName.toString() : prefix + fieldName;
+        final String escapeChar = "%";
+        final String problemChars = "\\\"";
+        final String unclean = escapeChar + problemChars;
+        if (!StringUtils.containsAny(name, unclean))
+            return name;
+        var ret = FIELD_ENCODED_PREFIX + encode(name);
+        assert !StringUtils.containsAny(ret, problemChars);
+        return ret;
```
labkey-tchad, line 294: How does this work with multi-value columns that need name encoding?
The `getFormFieldName` on line 270 would return something like `%_[]quf_tricky%22stuff` but I presume the actual name is `[]%_quf_tricky%22stuff`.
If so, that method should be changed to work correctly.

labkey-tchad (changes requested):

src/org/labkey/test/util/TestDataGenerator.java
labkey-matthewb: This should be reverted

src/org/labkey/test/util/EscapeUtil.java
labkey-matthewb: The decoding happens very early, before anyone doing binding sees the value.   So the %_ will be stripped and the binding code will see `[]quf_tricky"stuff`

I opted to not be tricky and make this specific to columns names or part of the field name.  This processing happens to the entire name.

labkey-tchad (approved):

PR #2928, opened 2026-03-31
LKSM: Multivalue text choice automation: Sample Finder MVFK

DariaBod:
#### Rationale
Make MVTC selections more predictable using bitmaskSelect.

#### Related Pull Requests
- [https://github.com/LabKey/limsModules/pull/2080](https://github.com/LabKey/limsModules/pull/2080)

### DIFF
--- a/src/org/labkey/test/params/FieldDefinition.java
+++ b/src/org/labkey/test/params/FieldDefinition.java
@@ -498,11 +498,6 @@ public class FieldDefinition extends PropertyDescriptor
         _namePart = namePart;
     }
 
-    public String getNamePart()
-    {
-        return _namePart;
-    }
-
     public boolean isNamePartMatch(String namePart)
     {
         return _namePart != null && _namePart.equals(namePart);
--- a/src/org/labkey/test/params/FieldDefinition.java
+++ b/src/org/labkey/test/params/FieldDefinition.java
@@ -498,6 +498,11 @@ public class FieldDefinition extends PropertyDescriptor
         _namePart = namePart;
     }
 
+    public String getNamePart()
+    {
+        return _namePart;
+    }
+
     public boolean isNamePartMatch(String namePart)
     {
         return _namePart != null && _namePart.equals(namePart);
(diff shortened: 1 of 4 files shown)


src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -188,7 +190,10 @@ else if (fieldDefinition.getType().equals(FieldDefinition.ColumnType.MultiValueT
                     {
                         FieldDefinition.TextChoiceValidator validator =
                                 (FieldDefinition.TextChoiceValidator) fieldDefinition.getValidators().getFirst();
-                        List<String> values = shuffleSelect(validator.getValues());
+                        // Use i + 1 as a bitmask so each row gets a deterministic, non-empty subset of choices,
+                        // consistent with how other field types (Integer, Boolean, TextChoice) use i for predictable data.
+                        // Salt with queryName hash so different entity types produce distinct MVTC values at the same row index.
+                        List<String> values = bitmaskSelect(validator.getValues(), i + 1, queryName.hashCode());
```
labkey-tchad, line 196: Other than string fields, all of the other data types have predictable values based only on the row index. Why do MVTC values need to be distinct between different queries?

DariaBod: The main problem is the testMVTCWithMPlusNAncestors() test. Since MVTC values are based only on a row index, they are the same for Sources, Parents, and Samples with that index. This means sorting and filtering results are also identical, which I think results in poor test coverage.

src/org/labkey/test/params/FieldDefinition.java
```diff
@@ -498,6 +498,11 @@ public void setNamePart(String namePart)
         _namePart = namePart;
     }
 
+    public String getNamePart()
+    {
+        return _namePart;
+    }
+
```
labkey-tchad: Add `getNamePart` to `FieldInfo` instead. It's guaranteed to be non-null there.

labkey-tchad (approved):

PR #2929, opened 2026-04-01
Fix PortTab helper.

labkey-danield:
#### Rationale
Remove the role attribute from the locator.

#### Related Pull Requests
- [Improve accessibility - remove inappropriate role="presentation" on tabs](https://github.com/LabKey/platform/pull/7534)

#### Changes
- Remove the role attribute from the locator.

### DIFF
--- a/src/org/labkey/test/components/labkey/PortalTab.java
+++ b/src/org/labkey/test/components/labkey/PortalTab.java
@@ -222,7 +222,7 @@ public class PortalTab extends WebDriverComponent<PortalTab.ElementCache>
         }
 
         static public Locator.XPathLocator container = Locator.tagWithClass("ul", "lk-nav-tabs");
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):
If this ends up messing with finding the right element, we could introduce a new class or other marker to help the test.

labkey-danield:
> If this ends up messing with finding the right element, we could introduce a new class or other marker to help the test.

Ok, I'll keep that in mind. I think this test component could be cleaned up as well. Several of the locators it uses are inefficient, but I didn't want to take that on as part of this PR.

labkey-tchad:
I'm pretty sure that's going to end up finding extra elements. When the page is in admin mode, the menus within the portal tabs are also `li`s

labkey-danield:
Well, let's see how it runs.

PR #2937, opened 2026-04-06
test setRequired()

labkey-matthewb:
#### Rationale
Test that field (non-MVTC) can't be set to required if null is present

regression test for change in https://github.com/LabKey/platform/pull/7549

#### Related Pull Requests
[- ](https://github.com/LabKey/platform/pull/7549)

#### Changes
-

### DIFF
--- a/src/org/labkey/test/tests/list/ListTest.java
+++ b/src/org/labkey/test/tests/list/ListTest.java
@@ -1396,6 +1396,45 @@ public class ListTest extends BaseWebDriverTest
         assertTextBefore(newFieldName, origFieldName);
     }
 
+
+    @Test
+    public void requiredFieldsTest()
+    {
+        log("Test changing required property of field");
+        String listName = "requiredColList";
+        String fieldA = "c$a";
+        String fieldB = "c_b";
+
+        _listHelper.createList(PROJECT_VERIFY, listName, "key",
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/list/ListTest.java
```diff
@@ -1396,6 +1396,45 @@ public void doRenameFieldsTest()
...
+
+        _listHelper.createList(PROJECT_VERIFY, listName, "key",
+                new FieldDefinition(fieldA, ColumnType.String).setDescription("first column").setRequired(false),
+                new FieldDefinition(fieldB, ColumnType.String).setDescription("second column").setRequired(false)
+        );
+
+        // insert a row with a NULL value and NON-NULL value
+        Map<String, String> row = new HashMap<>();
+        row.put(fieldA, "not null");
+        row.put(fieldB, "");
```
XingY, line 1416: Should there be a fieldC, that has a mixed value of empty and not empty, that also cannot be set to required?

XingY (approved):
Looks like we already have a DomainDesignerTest that tries to test some of the required toggle scenarios.

src/org/labkey/test/tests/list/ListTest.java
labkey-matthewb: Adding a second row to cover that makes sense.

PR #2946, opened 2026-04-10
Add TestFileUtils.deleteDirWithRetry method.

labkey-danield:
#### Rationale
Revert AssayTransformMissingParentDirTest.testMissingParentDirectoryRegression to (mostly) its original state.
Move the delete code into TestFileUtils.deleteDirWithRetry, this might be useful to help debug failures in the future.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/7578

#### Changes
- Revert changes to the test.
- Move the delete and retry code into TestFileUtils.


src/org/labkey/test/TestFileUtils.java
```diff
@@ -445,6 +451,80 @@ public static void deleteDir(File dir)
...
+                    else {
+                        pb = new ProcessBuilder("ps", "-ef");
+                    }
+
+                    try {
+                        LOG.info("Lock diagnostic...");
+                        pb.redirectErrorStream(true);
+
+                        // Tried to use "try (Process p = pb.start()) {" without the finally block but our build
+                        // system didn't like that and complained that Process doesn't implement closeable (it does).
```
labkey-jeckels, line 506: `Process` started implementing `Closeable` in Java 26. We're using Java 25.

src/org/labkey/test/TestFileUtils.java
```diff
@@ -445,6 +451,80 @@ public static void deleteDir(File dir)
...
+                dir.setWritable(true, false);
+
+                // Wrap in a try to close the stream.
+                try (Stream<Path> files = Files.walk(dir.toPath())) {
+                    files.forEach(p -> p.toFile().setWritable(true, false));
+                }
+
+                FileUtils.deleteDirectory(dir);
+                LOG.info(String.format("Deletion of directory %s was successful.", dir));
+                break;
```
labkey-jeckels, line 482: Consider making this `return`. If you do that, you can eliminate the `if (attempt == 10)` statement below and move its body outside of the loop.

src/org/labkey/test/TestFileUtils.java
```diff
@@ -445,6 +451,80 @@ public static void deleteDir(File dir)
...
+            } catch (IOException ioException) {
```
labkey-jeckels, line 483: `File.walk().forEach()` can throw `UncheckedIOException`. Probably good to catch that here too.

labkey-jeckels (approved):

src/org/labkey/test/TestFileUtils.java
```diff
@@ -445,6 +451,80 @@ public static void deleteDir(File dir)
...
+                        LOG.info("Lock diagnostic...");
+                        pb.redirectErrorStream(true);
+
+                        // Tried to use "try (Process p = pb.start()) {" without the finally block but our build
+                        // system didn't like that and complained that Process doesn't implement closeable (it does).
+                        Process p = pb.start();
+                        try {
+                            String output = new String(p.getInputStream().readAllBytes(), StringUtilsLabKey.DEFAULT_CHARSET);
+                            LOG.info("Running processes:\n" + output);
+                        }
```
labkey-tchad, line 511: Consider try-with-resources for the stream
```suggestion
                        try (InputStream is = p.getInputStream()) {
                            String output = new String(is.readAllBytes(), StringUtilsLabKey.DEFAULT_CHARSET);
                            LOG.info("Running processes:\n" + output);
                        }
```

src/org/labkey/test/TestFileUtils.java
labkey-tchad, line 483: We should just rethrow permissions exceptions. We haven't actually run into errors with those.
```suggestion
            } catch (AccessDeniedException adException) {
                throw adException;
            } catch (IOException ioException) {
```

src/org/labkey/test/TestFileUtils.java
```diff
@@ -445,6 +451,80 @@ public static void deleteDir(File dir)
...
+                        finally {
+                            // Don't leak the process resource.
+                            p.destroy();
+                        }
+
+                    } catch (IOException diagnosticException) {
+                        LOG.warn("Failed to run lock diagnostic: " + diagnosticException.getMessage());
```
labkey-tchad, line 518: Loggers can include the entire stack-trace and I think it would be useful to do so.
```suggestion
                        LOG.warn("Failed to run lock diagnostic: " + diagnosticException.getMessage(), diagnosticException);
```

src/org/labkey/test/TestFileUtils.java
```diff
@@ -445,6 +451,80 @@ public static void deleteDir(File dir)
...
+        // Sometimes on Windows the directory could be locked, maybe by an external process, or the child directory is
+        // readonly. Use a retry mechanism to set the writeable flag and then try to delete the parent directory.
+        for (int attempt = 1; attempt <= 10; attempt++) {
+            try {
+                dir.setWritable(true, false);
+
+                // Wrap in a try to close the stream.
+                try (Stream<Path> files = Files.walk(dir.toPath())) {
+                    files.forEach(p -> p.toFile().setWritable(true, false));
+                }
```
labkey-tchad, line 478: I don't think the tests should be in the business of changing file permissions (unless that's something we are explicitly testing at some point).
If the server (or other process) is leaving a file in an unwritable state, that is something we probably wouldn't want to ignore.

src/org/labkey/test/TestFileUtils.java
```diff
@@ -445,6 +451,80 @@ public static void deleteDir(File dir)
...
+
+                FileUtils.deleteDirectory(dir);
+                LOG.info(String.format("Deletion of directory %s was successful.", dir));
+                break;
+            } catch (IOException ioException) {
+                LOG.warn(String.format("IOException trying to delete directory %s. Error: %s. Waiting 10s and retrying. Attempt %d of 10.",
+                        dir, ioException.getMessage(), attempt));
+                if (attempt == 10) {
```
labkey-tchad, line 486: Put the max attempt count in a constant to ensure that this stays in sync with the for loop.
e.g.
```suggestion
                if (attempt == MAX_ATTEMPTS) {
```

labkey-tchad (approved):
I don't think we need to cast too wide of a net here. File locks on Windows are the only thing we've had trouble with (unless I'm forgetting something). We shouldn't add workarounds for hypothetical file permission issues that we don't understand.

PR #2960, opened 2026-04-21
LKSM: Multivalue text choice automation: Type Conversion

DariaBod:
#### Rationale
Added tests for type conversion MVTC->TC, TC->MVTC, MVTC->Text for for saved custom grid views with all array type filters.
Also added test to verify MVTC cannot convert to TC if any row has >1 values, cross folder scenario, if rows with multi value is in child/sibling folder only.

#### Related Pull Requests
- [https://github.com/LabKey/limsModules/pull/2136](https://github.com/LabKey/limsModules/pull/2136)

#### Changes
-


XingY (approved):

src/org/labkey/test/components/domain/DomainFieldRow.java
```diff
@@ -767,11 +767,19 @@ public DomainFieldRow clickRemoveOntologyConcept()
     // behind the scenes. Because of that the validator aspect of the TextChoice field is hidden from the user (just like
     // it is in the product).
 
-    public void setAllowMultipleSelections(Boolean allowMultipleSelections)
+    public DomainFieldRow setAllowMultipleSelections(Boolean allowMultipleSelections)
```
labkey-tchad, line 769: Add a parameter to explicitly determine whether a confirmation dialog is expected. This is consistent with the `setType` method.
```suggestion
    public DomainFieldRow setAllowMultipleSelections(Boolean allowMultipleSelections, boolean confirmDialogExpected)
```
Without this change, the component might fail to dismiss a dialog that appears a little too slowly.

src/org/labkey/test/tests/component/GridPanelViewTest.java
```diff
@@ -992,7 +1038,7 @@ public void testFieldInsertionOrder() throws Exception
                 customizeModal.isAvailableFieldSelected(columnToAdd));
 
         log("Validate that the order of the fields in the 'Shown in Grid' column are as expected.");
-        expectedFields = List.of(COL_NAME, COL_STRING1, COL_STRING2, COL_INT, COL_BOOL);
+        expectedFields = List.of(COL_NAME, COL_STRING1, COL_STRING2, COL_INT, COL_BOOL, COL_MULTITEXTCHOICE);
```
labkey-tchad: The list of expected fields should be conditional
```suggestion
        expectedFields = MULTI_CHOICE_ENABLED
                ? List.of(COL_NAME, COL_STRING1, COL_STRING2, COL_INT, COL_BOOL, COL_MULTITEXTCHOICE)
                : List.of(COL_NAME, COL_STRING1, COL_STRING2, COL_INT, COL_BOOL);

```

src/org/labkey/test/tests/component/GridPanelViewTest.java
```diff
@@ -190,11 +236,11 @@ private void createSampleType(String sampleTypeName, String samplePrefix, int nu
      * @param sampleTypeName Name of the sample type with the default view to change.
      * @param columns The columns to show in the default view. Will be added in the order of the list.
      */
-    private void resetDefaultView(String sampleTypeName, List<String> columns) throws Exception
+    private void resetDefaultView(String projectPath, String sampleTypeName, List<String> columns) throws Exception
```
labkey-tchad: Why add this path parameter? It is never anything but `getProjectName()`.

src/org/labkey/test/tests/component/GridPanelViewTest.java
```diff
@@ -96,6 +111,20 @@ public class GridPanelViewTest extends GridPanelBaseTest
 
     private final AuditLogHelper _auditLogHelper = new AuditLogHelper(this);
 
+    private record GridConversionResult(boolean isDropped, @Nullable String expectedFilterText)
+    {
+        static GridConversionResult kept(String filterText) { return new GridConversionResult(false, filterText); }
+        static GridConversionResult dropped()               { return new GridConversionResult(true, null); }
+    }
```
labkey-tchad, line 128: Add more documentation. Claude is great at this.
```suggestion
    /**
     * Describes the expected state of a saved grid view's filter pill after a field type conversion.
     *
     * @param isDropped          {@code true} if the filter is expected to be silently dropped by the conversion
     *                           (i.e., no filter pill should be visible); {@code false} if the filter should survive.
     * @param expectedFilterText the substring expected to appear in the filter pill after conversion;
     *                           {@code null} when {@code isDropped} is {@code true}.
     */
    private record GridConversionResult(boolean isDropped, @Nullable String expectedFilterText)
    {
        /** The filter survives conversion and its pill should contain {@code filterText}. */
        static GridConversionResult kept(String filterText) { return new GridConversionResult(false, filterText); }
        /** The filter is silently dropped by the conversion; no filter pill should be visible. */
        static GridConversionResult dropped()               { return new GridConversionResult(true, null); }
    }
```

src/org/labkey/test/tests/component/GridPanelViewTest.java
```diff
@@ -96,6 +111,20 @@ public class GridPanelViewTest extends GridPanelBaseTest
...
+
+    private record GridMVTCCase(String viewName, Filter.Operator op, String[] filterVals,
+                                GridConversionResult conversionResult,
+                                Map<String, String> expectedSampleMap) {}
+
+    private record GridTCCase(String viewName, Filter.Operator op, String[] filterVals,
+                              String expectedAfterConversionText,
+                              Map<String, String> expectedSampleMap) {}
```
labkey-tchad, line 156: Add documentation to these as well

src/org/labkey/test/tests/component/GridPanelViewTest.java
```diff
@@ -1393,6 +1439,373 @@ public void testWarningOnInvalidDateFilter() throws Exception
...
+    private void changeMVTCFieldToText()
+    {
+        UpdateSampleTypePage updatePage = beginAtSampleTypesList(this, getProjectName())
+                .goToEditSampleType(DEFAULT_VIEW_SAMPLE_TYPE);
+        updatePage.getFieldsPanel()
+                .getField(COL_MULTITEXTCHOICE)
+                .expand()
+                .setType(FieldDefinition.ColumnType.String, true);
+        updatePage.clickSave();
+    }
```
labkey-tchad, line 1740: If any of these new tests fails while the column type has been changed, it could cause all of the subsequent MVTC tests to fail because the type is not as expected.

Consider having these type changer methods not depend on a certain initial type and alter their behavior based on the current type. Then each of the conversion tests could switch the column to MVTC at the start instead of having each of them switch back to MVTC at the end.

labkey-tchad (approved):
Several suggestions. (Claude `/review-pr` was useful)

PR #2969, opened 2026-04-28
Fix for SMEditableGridTest.testDragFillReadOnlyColumn

DariaBod:
#### Rationale
Test [SMEditableGridTest.testDragFillReadOnlyColumn](https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_SampleManager_SampleManagerPostgres/3933361?buildTab=tests&name=dragFill) failing with the error "Drag is initializing in the wrong column".

#### Related Pull Requests
- [https://github.com/LabKey/limsModules/pull/2158](https://github.com/LabKey/limsModules/pull/2158)

#### Changes
- Added retry after dragToCell

### DIFF
--- a/src/org/labkey/test/components/ui/grids/EditableGrid.java
+++ b/src/org/labkey/test/components/ui/grids/EditableGrid.java
@@ -1023,7 +1023,16 @@ public class EditableGrid extends WebDriverComponent<EditableGrid.ElementCache>
     {
         dismissPopover();
         Locator.XPathLocator selectionHandleLoc = Locator.byClass("cell-selection-handle");
-        WebElement selectionHandle = selectionHandleLoc.findElement(startCell);
+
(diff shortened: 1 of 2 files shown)


labkey-tchad (approved):

labkey-tchad (commented):
This retry doesn't know enough to replicate the initial state. It assumes that a single cell is selected, which is not always the case.
In order to be able to do a retry, a method would the start and end point of the selection area. Most calls to `dragFill` are preceded by a call to `selectCellRange`. You could overload `dragFill` to include the `selectCellRange` step and retry if it fails. Something like this:
```
public void dragFill(WebElement selectStart, WebElement selectEnd, WebElement dragEnd)
{
    selectCellRange(selectStart, selectEnd);
    dragFill(selectEnd, dragEnd);
    /*
     check result and retry if necessary
     */
}
```

PR #2974, opened 2026-05-04
Fix for SampleTypeNameExpressionTest

DariaBod:
#### Rationale
[SampleTypeNameExpressionTest.testNameExpressionPreview, SampleTypeNameExpressionTest.testErrorsAndWarnings](https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyPostgres_2/3969885?buildTab=tests&status=failed&name=expression) failed with message "Unable to find element: xpath=//span[text()='Naming ']//div[@class='overlay-trigger']". Element xpath changed.

#### Related Pull Requests
- 

#### Changes
- Changed xpath

### DIFF
--- a/src/org/labkey/test/components/ui/domainproperties/EntityTypeDesigner.java
+++ b/src/org/labkey/test/components/ui/domainproperties/EntityTypeDesigner.java
@@ -436,7 +436,8 @@ public abstract class EntityTypeDesigner<T extends EntityTypeDesigner<T>> extend
 
         public final WebElement helpTarget(String divLabelText)
         {
-            return Locator.xpath(String.format("//div[text()='%s']//div[@class='overlay-trigger']", divLabelText)).findWhenNeeded(this);
+            String element = getThis().getClass().getSimpleName().contains("SampleType")? "div" : "span";
+            return Locator.xpath(String.format("//%s[text()='%s']//div[@class='overlay-trigger']", element, divLabelText)).findWhenNeeded(this);
         }
 
         // Tool tips exist on the page, outside the scope of the domainDesigner, so scope the search accordingly.
--- a/src/org/labkey/test/components/ui/domainproperties/EntityTypeDesigner.java
+++ b/src/org/labkey/test/components/ui/domainproperties/EntityTypeDesigner.java
@@ -436,7 +436,7 @@ public abstract class EntityTypeDesigner<T extends EntityTypeDesigner<T>> extend
 
         public final WebElement helpTarget(String divLabelText)
         {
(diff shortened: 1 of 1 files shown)


labkey-susanh (approved):

labkey-tchad (changes requested):
The class name shouldn't be functionally significant like this.

Also, I don't think this is the correct fix. The DOM structure isn't dependent on the entity type, i depends on whether it is in app or not. If you look at the sample type designer in LKS, it matches the old Locator. The new Locator works in LKSM and LKB.

Platform is still using `@labkey/components` without the [accessibility improvements](https://github.com/LabKey/labkey-ui-components/pull/1991) behind this.

@labkey-susanh When do we bump the components version in platform modules?

labkey-susanh:
> The class name shouldn't be functionally significant like this.
> 
> Also, I don't think this is the correct fix. The DOM structure isn't dependent on the entity type, i depends on whether it is in app or not. If you look at the sample type designer in LKS, it matches the old Locator. The new Locator works in LKSM and LKB.
> 
> Platform is still using `@labkey/components` without the [accessibility improvements](https://github.com/LabKey/labkey-ui-components/pull/1991) behind this.
> 
> @labkey-susanh When do we bump the components version in platform modules?

I can update platform to the new version in this branch if that works. Otherwise, I can do it in next accessibility branch. Let me know what's preferable.

labkey-susanh:
> > The class name shouldn't be functionally significant like this.
> > Also, I don't think this is the correct fix. The DOM structure isn't dependent on the entity type, i depends on whether it is in app or not. If you look at the sample type designer in LKS, it matches the old Locator. The new Locator works in LKSM and LKB.
> > Platform is still using `@labkey/components` without the [accessibility improvements](https://github.com/LabKey/labkey-ui-components/pull/1991) behind this.
> > @labkey-susanh When do we bump the components version in platform modules?
> 
> I can update platform to the new version in this branch if that works. Otherwise, I can do it in next accessibility branch. Let me know what's preferable.

The platform versions are also being updated with [Cory's PR](https://github.com/LabKey/platform/pull/7639), which will include these changes.

labkey-tchad:
> > I can update platform to the new version in this branch if that works. Otherwise, I can do it in next accessibility branch. Let me know what's preferable.
> 
> The platform versions are also being updated with [Cory's PR](https://github.com/LabKey/platform/pull/7639), which will include these changes.

It depends on the ETA of Cory's update. We could probably afford to mute the test failure for a couple of days.
Daria's monitoring these tests for May; I'll leave it up to her whether she'd like it updated sooner so that we don't have to mute the failure on TeamCity.

DariaBod:
[Cory's changes](https://github.com/LabKey/platform/pull/7639) will naturally fix these tests.

PR #2975, opened 2026-05-04
Refining the fuzz generator

DariaBod:
#### Rationale
Fuzz generator made our tests to fail intermittent because of several things: double spaces, repeat characters, Selenium special characters.

#### Related Pull Requests
- [https://github.com/LabKey/limsModules/pull/2178](https://github.com/LabKey/limsModules/pull/2178)

#### Changes
- ~Backslash excluded because Selenium can mess if read it from UI~
- Attempts count added to randomTextChoice
- containsKey+put → computeIfAbsent
- Random() and Math.random() changed to ThreadLocalRandom


src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -507,8 +503,11 @@ public static String randomString(int size)
     public static List<String> randomTextChoice(int size)
     {
         Set<String> textChoices = new LinkedHashSet<>();
+        int attempts = 0;
         while (textChoices.size() < size)
         {
+            if (++attempts >= MAX_RANDOM_TRIES)
+                throw new IllegalStateException("Failed to generate " + size + " unique text choices after " + MAX_RANDOM_TRIES + " attempts");
```
labkey-tchad: Consider making the max tries scale with the requested size. This is more of a future proofing kind of thing; in case we decide to introduce some stress testing with a large (100+) number of text choices.
```suggestion
        int attempts = 0;
        final int maxTries = Math.max(MAX_RANDOM_TRIES, size * 2);
while (textChoices.size() < size)
{
            if (++attempts >= maxTries)
                throw new IllegalStateException("Failed to generate " + size + " unique text choices after " + maxTries + " attempts");
```

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -507,8 +503,11 @@ public static String randomString(int size)
...
             String generated = randomString(randomInt(1, 25), ";").trim();
```
labkey-tchad: I missed this in the last PR. Why are we excluding `;` from text choices?
If it is a valid character, we should include it.

labkey-tchad (changes requested):
I think we should focus on whitespace and repeated characters for now.
I'm not ready to exclude backslashes from our fuzz testing. It is one of the mostly likely characters to cause problems and should be a prime target for fuzz testing.
If there is a bug in Selenium, we need to [open an issue](https://github.com/SeleniumHQ/selenium/issues).

src/org/labkey/test/util/TestDataGenerator.java
DariaBod: https://github.com/LabKey/internal-issues/issues/1096#issuecomment-4331544773
It breaks MVTC conversion tests. I can create a separate method without semicolon for conversion tests. What do you think?

labkey-tchad: I'd say we should have a separate method for those tests. Or maybe have the conversion tests adjust their expectations depending on whether the choices have semicolons or not.

labkey-tchad (approved):

PR #2988, opened 2026-05-08
Fix spaces added inside randomString method

DariaBod:
#### Rationale
Just changed randomString method instead of insertSpaces.

Note: check [build](https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_NightlyBiologicsPostgres?branch=fixSpacesInDataGenerator#all-projects) before merge.

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/src/org/labkey/test/util/TestDataGenerator.java
+++ b/src/org/labkey/test/util/TestDataGenerator.java
@@ -535,7 +535,7 @@ public class TestDataGenerator
 
     /** After this many consecutive non-space characters, {@link #randomString} forces the next char to ' '
      * (provided ' ' is in the effective charset). */
-    private static final int FORCE_SPACE_AFTER = 8;
+    private static final int FORCE_SPACE_AFTER = 15;
 
     public static String randomString(int size, @Nullable String exclusion, @Nullable String charSet)
     {
--- a/src/org/labkey/test/util/TestDataGenerator.java
+++ b/src/org/labkey/test/util/TestDataGenerator.java
@@ -563,8 +563,6 @@ public class TestDataGenerator
             String picks = charSetFrom;
             if (val.length() == 0 || isLast || prev == ' ')
                 picks = picks.replace(" ", "");
-            if (prev == ' ')
-                picks = picks.replace("-", "");
             if (picks.isEmpty())
                 picks = charSetFrom; // pathological charset, fall back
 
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -535,32 +533,69 @@ public static String randomString(int size, @Nullable String exclusion)
...
+            int randIndex = ThreadLocalRandom.current().nextInt(picks.length());
+            char c = picks.charAt(randIndex);
             if (c == REPEAT_PLACEHOLDER)
             {
-                randIndex = ThreadLocalRandom.current().nextInt(charSetFrom.length());
-                c = charSetFrom.charAt(randIndex);
-                int repeatCount = randomInt(2, 5); // repeat between 2 and 5 times
+                randIndex = ThreadLocalRandom.current().nextInt(picks.length());
+                c = picks.charAt(randIndex);
+                int repeatCount = randomInt(2, 5);
```
labkey-tchad: The regex at the end would catch them but we should explicitly avoid repeated spaces here.
```suggestion
                c = picks.charAt(randIndex);
                int repeatCount = (c == ' ') ? 0 : randomInt(2, 5);
```

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -535,32 +533,69 @@ public static String randomString(int size, @Nullable String exclusion)
...
+                nonSpaceRun = 0;
+                continue;
+            }
+
+            // Random pick. Filter to keep invariants: no leading/trailing space, no double space, no " -" pattern.
+            String picks = charSetFrom;
+            if (val.length() == 0 || isLast || prev == ' ')
+                picks = picks.replace(" ", "");
+            if (prev == ' ')
+                picks = picks.replace("-", "");
```
labkey-tchad: This restriction is only valid for domain names and `randomDomainName` already prevents this ` -` pattern.
```suggestion
```

src/org/labkey/test/util/TestDataGenerator.java
```diff
@@ -535,32 +533,69 @@ public static String randomString(int size, @Nullable String exclusion)
         return randomString(size, exclusion, CHARSET_STRING);
     }
 
+    /** After this many consecutive non-space characters, {@link #randomString} forces the next char to ' '
+     * (provided ' ' is in the effective charset). */
+    private static final int FORCE_SPACE_AFTER = 8;
```
labkey-tchad: 8 feels a little short, I'd go with something like 20.
The problems we're seeing are with 50+ character long strings without a break.

labkey-tchad (approved):
A few suggestions

src/org/labkey/test/util/TestDataGenerator.java
labkey-danield: A limit of 20 is ok, but a bit on the high end. Looking at the list of "real name of things" data the average number of character in a name with spaces is closer to 12 to 15.
For names without spaces, and using underscores, the lengths are in the 20+ range.

labkey-danield (approved):

PR #2995, opened 2026-05-13
PackageLockJsonTest update to allow for transitive dependencies with "npm:" alias prefix

cnathe:
#### Rationale
  This PR updates PackageLockJsonTest to allow transitive dependency version strings that use npm's package alias syntax (`npm:<package>@<range>`). The original approach used a hardcoded allowlist of three specific aliases from
  `@isaacs/cliui`; the new approach allows any version string starting with `npm:`, which is more general and won't need updating when new npm aliases appear in lock files.

#### Changes
- update to allow for transitive dependencies with "npm:" alias prefix

### DIFF
--- a/src/org/labkey/test/tests/PackageLockJsonTest.java
+++ b/src/org/labkey/test/tests/PackageLockJsonTest.java
@@ -28,6 +28,7 @@ import java.util.Set;
 public class PackageLockJsonTest
 {
     private static final Set<String> ALLOWED_DEPENDENCY_HOSTS = Set.of("registry.npmjs.org", "labkey.jfrog.io");
+    private static final File SERVER_MODULES_DIR = new File(TestFileUtils.getLabKeyRoot(), "server/modules");
 
     private final List<String> errors = new ArrayList<>();
     private final File moduleDir;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/PackageLockJsonTest.java
```diff
@@ -148,7 +146,7 @@ private void verifyPackage(String packageName, JSONObject packageJson, File pack
             else
             {
                 String tVer = transitiveDeps.optString(tDep);
-                if (tVer == null || tVer.contains(":") && !ALLOWED_NONSTANDARD_VERSIONS.contains(tVer)) // URL, file, or workspace dependency
+                if (tVer == null || (tVer.contains(":") && !tVer.startsWith("npm:"))) // URL, file, or workspace dependency
```
labkey-tchad: Not sure it really matters but maybe make this trim off the "npm:" instead of just skipping any versions starting with it.

```suggestion
                if (tVer == null || tVer.replaceAll("^npm:", "").contains(":")) // URL, file, or workspace dependency
```

labkey-tchad (approved):

labkey-alan (approved):

labkey-tchad (approved):
I piggy-backed some logging improvements but this change looks ok to me.

src/org/labkey/test/tests/PackageLockJsonTest.java
cnathe: Done

PR #3001, opened 2026-05-15
Update tests that relied on obsolete actions.

labkey-klum:
#### Rationale
A recent PR removed some internal actions like `query.internalNewView` and `reports.reportInfo`. I've updated those tests to rely on alternate ways to accomplish the same actions.

### DIFF
--- a/src/org/labkey/test/BaseWebDriverTest.java
+++ b/src/org/labkey/test/BaseWebDriverTest.java
@@ -24,9 +24,9 @@ import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.hc.core5.http.HttpStatus;
 import org.awaitility.Awaitility;
+import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.json.JSONObject;
-import org.jetbrains.annotations.NotNull;
 import org.junit.Assume;
 import org.junit.AssumptionViolatedException;
 import org.junit.ClassRule;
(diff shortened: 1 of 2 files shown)


src/org/labkey/test/BaseWebDriverTest.java
```diff
@@ -1686,20 +1686,17 @@ public String getBaseURL()
...
-        {
-            strFields.append("&");
-            strFields.append(fields[i]);
-        }
-        setFormElement(Locator.name("ff_columnList"), strFields.toString());
-        clickButton("Save");
+            setFormElement(Locator.name("quf_Name"), viewName);
+        setFormElement(Locator.name("quf_Columns"), String.join("&", fields));
+        setFormElement(Locator.name("quf_Flags"), "0");
+        clickButton("Submit");
```
labkey-tchad: Use the page class returned by `clickInsertNewRow` instead of hard-coded locators.

src/org/labkey/test/tests/TimeChartImportTest.java
```diff
@@ -316,13 +315,24 @@ public void verifyMaskedPtidOnPublishStudy()
         {
             clickTab("Clinical and Assay Data");
             waitAndClickAndWait(Locator.linkWithText(chartInfo.getName()));
-            beginAt(WebTestHelper.buildURL("reports",
-                getProjectName() + "/" + VISIT_STUDY_FOLDER_NAME + "/" + publishFolderName,
-                "reportInfo", Map.of("reportId", getUrlParam("reportId"))));
-            waitForText("Report Debug Information");
+        }
```
labkey-tchad: The for-loop above (over VISIT_CHARTS) no longer does anything interesting. Might as well remove it.

src/org/labkey/test/tests/TimeChartImportTest.java
```diff
@@ -316,13 +315,24 @@ public void verifyMaskedPtidOnPublishStudy()
...
+
+        // verify the report descriptor XML does not contain the masked ptid
+        goToSchemaBrowser();
+        DataRegionTable drt = viewQueryData("core", "Reports");
+        CustomizeView customizeView = drt.openCustomizeGrid();
+        customizeView.addColumn("DescriptorXML");
+        customizeView.applyCustomView();
+
+        drt = new DataRegionTable("query", getDriver());
+        int xmlIdx = drt.getColumnIndex("DescriptorXML");
```
labkey-tchad, line 323: Not sure how many reports get created by this test but you might want to increase the grid size to make sure nothing is hiding on later pages.
```suggestion
        drt = new DataRegionTable("query", getDriver());
        drt.setPageSize(100);
        int xmlIdx = drt.getColumnIndex("DescriptorXML");
```

labkey-tchad (approved):
Some suggestions

PR #3004, opened 2026-05-18
Fix for testUpdateAndDeleteWithCommentsAndFlags

DariaBod:
#### Rationale
Test failed because of wrong `"` processing. Also was filed issue because of silent operation fails in the app.

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/src/org/labkey/test/tests/SampleTypeTest.java
+++ b/src/org/labkey/test/tests/SampleTypeTest.java
@@ -966,17 +966,17 @@ public class SampleTypeTest extends BaseWebDriverTest
             sample.replace(field01.getName(), fieldValue.toUpperCase());
         }
 
-        List<String> fileData = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/SampleTypeTest.java
labkey-tchad: We should use the dedicated helper methods from `TestDataUtils` instead of having individual tests deal with file formatting.

```
List<Map<String, Object>> fileData = new ArrayList<>();
for(Map<String, String> sample : sampleData)
{
    fileData.add(Map.of(sample.get("Name"), sample.get(field01.getName())));
}

String fileName = "SampleTypeTest_UpdateSamples.tsv";
File importFile = TestDataUtils.writeDataToFile(fileName, fileData);
```

labkey-tchad (approved):
Consider using `TestDataUtils`

labkey-danield (approved):

PR #3012, opened 2026-05-21
Exercise big ints in RlabkeyTest

bbimber:
This is related to this PR:

https://github.com/LabKey/labkey-api-r/pull/126

which is designed to make Rlabkey support big int values.

### DIFF
--- a/data/lists/rlabkey.lists/AllTypes.tsv
+++ b/data/lists/rlabkey.lists/AllTypes.tsv
@@ -1,5 +1,5 @@
 DisplayFld	TextFld	IntFld	DoubleFld	DateTimeFld	BooleanFld	LongTextFld	AttachmentFld	RequiredText	RequiredInt	Category
-Row 1	some text here	99	987.654	2/14/2010	FALSE	Could be some very long text		I have to fill this out	42	PLAIN
+Row 1	some text here	99999999999	987.654	2/14/2010	FALSE	Could be some very long text		I have to fill this out	42	PLAIN
 Row 2 Empty Vals					FALSE			Cannot be empty	55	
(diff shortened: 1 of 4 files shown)


data/lists/rlabkey.lists/AllTypes.tsv
```diff
@@ -1,5 +1,5 @@
 DisplayFld	TextFld	IntFld	DoubleFld	DateTimeFld	BooleanFld	LongTextFld	AttachmentFld	RequiredText	RequiredInt	Category
-Row 1	some text here	99	987.654	2/14/2010	FALSE	Could be some very long text		I have to fill this out	42	PLAIN
+Row 1	some text here	99999999999	987.654	2/14/2010	FALSE	Could be some very long text		I have to fill this out	42	PLAIN
```
cnathe, line 2: in order for this to work for me locally, I had to update the related lists.xml file in testAutomation/data/lists/rlabkey.lists/lists.xml so that the "IntFId" `<datatype>` at line 70 was "bigint"

cnathe (changes requested):

data/lists/rlabkey.lists/lists.xml
```diff
@@ -66,7 +66,7 @@
         <columnTitle>Text Field</columnTitle>
       </column>
       <column columnName="IntFld">
-        <datatype>integer</datatype>
+        <datatype>bigint</datatype>
```
cnathe, line 69: To get the Rlabkey test passing now with this change, I needed to update line 261 of rlabkey-api-query.xml to:
`HTTP request was unsuccessful. Status code = 400, Error message = Could not convert value 'string' (String) for Long field 'IntFld'`

cnathe (changes requested):

data/lists/rlabkey.lists/lists.xml
bbimber: I will get this running locally and look into this. I want to understand what triggered the change.

I have noticed that something in either the Rlabkey side or the server side converts columns that are 100% NA (not a string), into the string 'NA'. I have long wondered whether this is a bug in Rlabkey. This change might be exposing that here.

cnathe:
I think this new test case can just go to develop instead of 26.3 at this point in the release cycle. I'll open a new PR with these changes targeting develop since I have to create a new PR for the labkey-api-r version bump anyway.

cnathe:
https://github.com/LabKey/testAutomation/pull/3078

PR #3020, opened 2026-05-26
Update copyrights and licenses

labkey-adam:
#### Rationale
Updates for 2026: https://github.com/LabKey/internal-issues/issues/1184

### DIFF
--- a/src/org/labkey/serverapi/collections/ArrayListMap.java
+++ b/src/org/labkey/serverapi/collections/ArrayListMap.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2003-2016 Fred Hutchinson Cancer Research Center
+ * Copyright (c) 2003-2026 Fred Hutchinson Cancer Research Center
  *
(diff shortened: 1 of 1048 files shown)


src/org/labkey/serverapi/collections/ArrayListMap.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2003-2016 Fred Hutchinson Cancer Research Center
+ * Copyright (c) 2003-2026 Fred Hutchinson Cancer Research Center
```
labkey-klum, line 2: Is this correct : FHCRC?

labkey-adam: Yes, the predecessor file was authored while we were still Hutch employees, so our agreement is to tag it with FHCRC copyright. You'll see a lot more examples of this in `platform`.

labkey-klum (approved):

PR #3066, opened 2026-06-22
Enable Optional Feature "Derive Samples in LabKey Server UI" in Test

labkey-danield:
#### Rationale
Get the test passing.

#### Related Pull Requests
- 

#### Changes
- Enable optional feature

### DIFF
--- a/src/org/labkey/test/tests/SampleTypeFolderExportImportTest.java
+++ b/src/org/labkey/test/tests/SampleTypeFolderExportImportTest.java
@@ -99,12 +99,27 @@ public class SampleTypeFolderExportImportTest extends BaseWebDriverTest
         return BrowserType.CHROME;
     }
 
-    @BeforeClass
-    public static void setupProject()
+    @Override
+    protected void doCleanup(boolean afterTest)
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/tests/SampleTypeFolderExportImportTest.java
```diff
@@ -119,6 +120,8 @@ private void doSetup()
         portalHelper.addWebPart("Sample Types");
         portalHelper.addWebPart("Experiment Runs");
 
+        OptionalFeatureHelper.setOptionalFeature(createDefaultConnection(), "deriveSamplesNotInApp", true);
```
labkey-susanh: Is there a need for a call to resetOptionalFeature in the cleanup?

labkey-susanh (approved):

src/org/labkey/test/tests/SampleTypeFolderExportImportTest.java
labkey-danield: Updated to reset it.

PR #3072, opened 2026-06-24
Mutating SQL test

labkey-adam:
#### Rationale
Simple test that a GET request invoking mutating SQL is flagged and the deprecated feature flag allows it

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/7791

### DIFF
--- a/src/org/labkey/test/tests/SecurityTest.java
+++ b/src/org/labkey/test/tests/SecurityTest.java
@@ -100,7 +100,7 @@ public class SecurityTest extends BaseWebDriverTest
     @BeforeClass
     public static void setupProject()
     {
-        ((SecurityTest)getCurrentTest()).doSetup();
+        ((SecurityTest) getCurrentTest()).doSetup();
     }
 
     protected void doSetup()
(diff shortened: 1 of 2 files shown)


labkey-matthewb (approved):

labkey-tchad (approved):

src/org/labkey/test/tests/SecurityTest.java
```diff
@@ -551,4 +551,25 @@ public void loginSelfRegistrationDisabledTest()
...
+    {
+        // Ensure that a GET request invoking mutating SQL is flagged
+        String feature = "AllowMutatingSqlViaGet";
+        Connection conn = createDefaultConnection();
+        OptionalFeatureHelper.disableOptionalFeature(conn, feature);
+        beginAt(buildURL("test", "executeMutatingSql"));
+        assertTextPresent("MUTATING SQL executed as part of handling action: GET org.labkey.devtools.TestController$ExecuteMutatingSqlAction");
+        checkExpectedErrors(2);
+
+        // Turn on the deprecated feature flag and ensure that GET request can invoke mutating SQL
```
labkey-tchad: Consider POSTing to the action to verify that it does _not_ throw an exception.

labkey-tchad (approved):

src/org/labkey/test/tests/SecurityTest.java
labkey-tchad: Actually, POSTing throws an exception because of the action's super-type, so maybe not.

labkey-adam: Right, it's not quite as easy as POSTing to the existing action. We flag any action that accepts GET, even if POST is the current method. So I had to add a new action that accepts POST.

PR #3074, opened 2026-06-24
Leave browser windows open after test failures

labkey-tchad:
#### Rationale
Starting with Geckodriver v0.37.0, shutting down the driver service also shuts down the browser.
We need to leave the FirefoxDriverService running after test failures to allow manual inspection of the browser state.
I'm putting this behavior behind a flag because it will also leave a lot of zombie 'geckodriver' processes that will need to be cleanup up manually.

 * https://github.com/mozilla/geckodriver/issues/2247
 * https://bugzilla.mozilla.org/show_bug.cgi?id=2046321

#### Related Pull Requests
- N/A

#### Changes
- Leave geckodriver running after test failures

### DIFF
--- a/src/org/labkey/test/BaseWebDriverTest.java
+++ b/src/org/labkey/test/BaseWebDriverTest.java
@@ -44,9 +44,9 @@ import org.labkey.junit.rules.TestWatcher;
 import org.labkey.remoteapi.CommandException;
 import org.labkey.remoteapi.CommandResponse;
 import org.labkey.remoteapi.Connection;
-import org.labkey.remoteapi.admin.ClearCachesCommand;
 import org.labkey.remoteapi.SimpleGetCommand;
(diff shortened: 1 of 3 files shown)


labkey-danield (approved):
Are there any considerations for TeamCity? Do we need to add a `killall` command at the end of the run? I don't think TeamCity leaves the browsers up on failure, but I cannot remember.

labkey-tchad:
> Are there any considerations for TeamCity? Do we need to add a `killall` command at the end of the run? I don't think TeamCity leaves the browsers up on failure, but I cannot remember.

There are several guardrails to prevent any problems on TeamCity. Like you said, we don't leave browser windows open. I also made it so the new property is ignored when running on TeamCity.
The `ciTests` (used to run tests on TeamCity) automatically kills all firefox and and geckodriver processes after it runs.

PR #3080, opened 2026-06-29
Fix for notification menu 'View all activity'

DariaBod:
#### Rationale
Dropdown menu click isn't working sometimes because of re-rendering.

#### Related Pull Requests
- 

#### Changes
- added retry

### DIFF
--- a/src/org/labkey/test/components/ui/notifications/ServerNotificationMenu.java
+++ b/src/org/labkey/test/components/ui/notifications/ServerNotificationMenu.java
@@ -5,6 +5,7 @@ import org.labkey.test.WebDriverWrapper;
 import org.labkey.test.components.Component;
 import org.labkey.test.components.WebDriverComponent;
 import org.labkey.test.components.ui.pipeline.ImportsPage;
+import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.StaleElementReferenceException;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/components/ui/notifications/ServerNotificationMenu.java
```diff
@@ -157,10 +158,19 @@ public void clickMarkAll()
...
+        WebDriverWrapper.waitFor(() -> {
+            try
+            {
+                expand();
+                elementCache().viewAllLink().click();
+                return true;
+            }
+            catch (StaleElementReferenceException | NoSuchElementException retry)
+            {
+                return false;
```
labkey-danield, line 171: Should this do a `collapse()` before retrying?

I also wonder if part of the issue might be:
`        public final WebElement menuContent = Locator.byClass("navbar-menu__content").refindWhenNeeded(this);
`
in the element cache. Maybe this should be a locator, or put in a method so it is found everytime it is referenced.

labkey-danield (commented):
I'm ok with retrying but I wonder if the call to `elementCache().viewAllLink()` might be part of the problem. Specifically it's dependency on `menuContent` in the elementCache.

src/org/labkey/test/components/ui/notifications/ServerNotificationMenu.java
DariaBod: Switched toggle to refindWhenNeeded so it recovers on re-render, and kept the retry for the transient not-yet-rendered case.

labkey-danield (approved):

PR #3086, opened 2026-07-01
Remove duplicate test code - SearchResultsPage

labkey-tchad:
#### Rationale
([Epic](https://github.com/LabKey/kanban/issues/1911))

These are some test library improvements I needed for the related limsModules refactor

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/2244

#### Changes
- Add `Locator.union` that will deal with xpath or css locators at runtime
- Improve synchronization of buttons in `ManageViewsDialog`
- Overload `WebTestHelper.parseUrlQueryString` to optionally decode the values

#### Tasks 📍
- [x] Claude Code Review
- [x] TeamCity verification


src/org/labkey/test/Locator.java
```diff
@@ -1019,6 +1019,40 @@ public static String cq(String value)
...
+    public static Locator union(Locator... locators)
+    {
+        if (locators.length < 1)
+            throw new IllegalArgumentException("Specify at least one locator");
+        if (locators.length == 1)
+            return locators[0];
+
+        List<CssLocator> cssLocators = Arrays.stream(locators).map(locator -> {
+            CssLocator cssLocator = null;
+            if (locator instanceof CssLocator)
```
DariaBod: This method is getting a bit long. What do you think about extracting this logic into a helper method, like this:
```
private static CssLocator asRawCssLocator(Locator locator)
{
    CssLocator cssLocator = null;
    if (locator instanceof CssLocator css)
        cssLocator = css;
    else if (locator instanceof XPathCSSLocator xc)
        cssLocator = xc.getCssLoc();

    return (cssLocator != null && CssLocator.isRawCssLocator(cssLocator)) ? cssLocator : null;
}
```
But it's totally up to you, I don't mind leaving it as is. Just a suggestion.

DariaBod (approved):

src/org/labkey/test/Locator.java
labkey-tchad: I like your suggestion. Thank you.

PR #3108, opened 2026-07-15
Refactor MailHelper to be more provider-agnostic

labkey-adam:
## Rationale
Dumbster changes to support a somewhat more agnostic MailHelper

## Related Pull Requests
- https://github.com/LabKey/platform/pull/7848

### DIFF
--- a/modules/dumbster/src/org/labkey/dumbster/model/DumbsterManager.java
+++ b/modules/dumbster/src/org/labkey/dumbster/model/DumbsterManager.java
@@ -96,9 +96,8 @@ public class DumbsterManager implements ShutdownListener
             return false;
         }
 
-        // Install our own SMTP provider pointed at the local capture server and make it the active provider, rather
-        // than mutating another provider's session state. All outgoing email is captured regardless of how the server's
-        // real email transport (SMTP, Microsoft Graph, etc.) is configured.
+        // Dumbster uses its own SMTP configuration and sets the MailHelper active provider to capture all outgoing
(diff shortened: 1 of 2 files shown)


modules/dumbster/src/org/labkey/dumbster/model/DumbsterManager.java
```diff
@@ -75,56 +77,59 @@ public boolean start()
...
-            }
-            catch (IOException ignored) {}
+            _log.error("Failed to connect mail recorder. Port {} may be in use.", port);
+            _server = null;
+            return false;
         }
 
+        // Install our own SMTP provider pointed at the local capture server and make it the active provider, rather
+        // than mutating another provider's session state. All outgoing email is captured regardless of how the server's
+        // real email transport (SMTP, Microsoft Graph, etc.) is configured.
```
labkey-bpatel: I could be misreading this - but I read these two sentences like a  contradiction on first pass: sentence 1 says we install our own SMTP provider, sentence 2 says we capture email regardless of transport including Graph (implies Dumbster handles Graph traffic, but of course we're just replacing the active provider).

labkey-bpatel (approved):
Please see my comment regarding a comment

modules/dumbster/src/org/labkey/dumbster/model/DumbsterManager.java
labkey-adam: AI-generated comment that could be clearer. I'll clean it up.

labkey-adam: Rewritten

PR #3128, opened 2026-07-27
GH Issue 1257: Check for duplicates among field names, import aliases, and parent import aliases

labkey-susanh:
## Rationale
Issue [1257](https://github.com/LabKey/internal-issues/issues/1257):

## Related Pull Requests
- https://github.com/LabKey/platform/pull/7884

## Changes
- Adding tests

### DIFF
--- a/src/org/labkey/test/tests/SampleTypeParentColumnTest.java
+++ b/src/org/labkey/test/tests/SampleTypeParentColumnTest.java
@@ -724,7 +724,7 @@ public class SampleTypeParentColumnTest extends BaseWebDriverTest
                 String.join("\n", errors));
         updatePage.removeParentAlias(0);
 
-        log("Now add a valid parent column and check that you cannot not add a field in the sample type with the same name.");
+        log("Now add a valid parent column and check that you cannot add a field in the sample type with the same name.");
         updatePage.addParentAlias(GOOD_PARENT_NAME, SampleTypeDesigner.CURRENT_SAMPLE_TYPE);
         updatePage.clickSave();
 
@@ -750,9 +750,9 @@ public class SampleTypeParentColumnTest extends BaseWebDriverTest
         updatePage.clickCancel();
 
         // GH Issue 1257
+        clickFolder(SUB_FOLDER_NAME);
         log("Check that you cannot add a field with an import alias that conflicts with the parent import alias");
-        waitAndClickAndWait(Locator.lkButton("Edit Type"));
-        updatePage = new UpdateSampleTypePage(getDriver());
+        updatePage = sampleHelper.goToEditSampleType(SAMPLE_TYPE_NAME);
         updatePage.getFieldsPanel().addField("DupeAliasCheck")
                 .setImportAliases(ALIAS_NAME_CONFLICT);
         errors = updatePage.clickSaveExpectingErrors();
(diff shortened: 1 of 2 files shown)


labkey-danield (approved):

src/org/labkey/test/tests/SampleTypeParentColumnTest.java
```diff
@@ -751,7 +749,17 @@ public void testAliasNameConflictsWithFieldName()
...
 
-        log("Validated name conflicts.");
+        // GH Issue 1257
+        clickFolder(SUB_FOLDER_NAME);
+        log("Check that you cannot add a field with an import alias that conflicts with the parent import alias");
+        updatePage = sampleHelper.goToEditSampleType(SAMPLE_TYPE_NAME);
+        updatePage.getFieldsPanel().addField("DupeAliasCheck")
+                .setImportAliases(GOOD_PARENT_NAME);
+        errors = updatePage.clickSaveExpectingErrors();
+        errorMsgExpectedTxt = "Field DupeAliasCheck has an import alias " + GOOD_PARENT_NAME + " that conflicts with a parent alias header.";
```
cnathe: any reason we need to add some checks for the other cases here like having multiple fields that have the same import alias or having an import alias that matches a parent alias? Or maybe some of that could be covered with junit tests (not sure what our validation coverage looks like there for domains).

labkey-susanh: I didn't include that here because this test class is specifically for `SampleTypeParentColumnTest`. So, tests aren't completely done for this branch.

src/org/labkey/test/tests/SampleTypeTest.java
```diff
@@ -301,7 +301,69 @@ public void testCustomProperties()
...
+                containsString("You have 2 field errors."));
+        checker().screenShotIfNewError("duplicateImportAlias");
+
+        log("Aliases differing only by case are still duplicates.");
+        createPage = new CreateSampleTypePage(this.getDriver());
+        fieldsPanel.getField(fieldTwo).setImportAliases(sharedAlias.toUpperCase());
+        checker().verifyThat("Expected an error when two fields share an import alias differing only by case",
+                String.join("\n", createPage.clickSaveExpectingErrors()),
+                containsString("You have 2 field errors."));
+        checker().screenShotIfNewError("duplicateImportAliasIgnoringCase");
```
labkey-danield, line 335: FYI, you can give a screenshot name to the .withScreenshot call:

`checker().withScreenshot("duplicateImportAlias").verifyThat("Expected an error when two fields share an import alias",
`

src/org/labkey/test/tests/SampleTypeTest.java
```diff
@@ -301,7 +301,69 @@ public void testCustomProperties()
...
+
+        log("Verify there can be no import aliases that collide with field names (case-insensitive).");
+        createPage = new CreateSampleTypePage(this.getDriver());
+        fieldsPanel.getField(fieldTwo).setImportAliases("");
+        fieldsPanel.getField(fieldOne).setImportAliases(fieldTwo);
+        checker().verifyThat("Expected an error when an import alias matches another field's name",
+                String.join("\n", createPage.clickSaveExpectingErrors()),
+                containsString("Import alias '" + fieldTwo + "' on field '" + fieldOne + "' conflicts with a field name."));
+        checker().screenShotIfNewError("importAliasConflictsWithFieldName");
```
labkey-danield, line 344: Same comment as above.

labkey-danield (approved):

PR #3131, opened 2026-07-28
Parse report-to CSP reports

labkey-adam:
## Rationale
Now that TeamCity is using a Firefox ESR that supports the `report-to` directive, our test framework should be able to parse report-to CSP reports. Once TeamCity supports report-to smoothly, we'll look at pursuing https://github.com/LabKey/internal-issues/issues/900

### DIFF
--- a/src/org/labkey/test/tests/elispotassay/ElispotAssayTest.java
+++ b/src/org/labkey/test/tests/elispotassay/ElispotAssayTest.java
@@ -757,7 +757,7 @@ public class ElispotAssayTest extends AbstractAssayTest
         List<String> expected4FMedians = Arrays.asList("0.0", "0.0", "6666.7", "0.0");
         List<String> expected6FMeans = Arrays.asList("0.0", "0.0", "0.0", "0.0");
         List<String> expected6FMedians = Arrays.asList("0.0", "0.0", "0.0", "0.0");
-        Bag<List<String>> expectedRows = new HashBag<>(DataRegionTable.collateColumnsIntoRows(
+        MultiSet<List<String>> expectedRows = new HashMultiSet<>(DataRegionTable.collateColumnsIntoRows(
                 expectedPtids,
                 expected2FMeans,
                 expected2FMedians,
@@ -766,7 +766,7 @@ public class ElispotAssayTest extends AbstractAssayTest
                 expected6FMeans,
                 expected6FMedians));
 
-        Bag<List<String>> actualRows = new HashBag<>(table.getRows(
+        MultiSet<List<String>> actualRows = new HashMultiSet<>(table.getRows(
                 "ParticipantID",
                 "Atg2FMean",
                 "Atg2FMedian",
--- a/src/org/labkey/test/util/CspLogUtil.java
+++ b/src/org/labkey/test/util/CspLogUtil.java
@@ -123,7 +123,7 @@ public class CspLogUtil
                     throw new RuntimeException("Failed to read recent CSP violations.", e);
                 }
 
-                boolean foundVioloation = false;
+                boolean foundViolation = false;
                 MultiValuedMap<Crawler.ControllerActionId, String> violations = new HashSetValuedHashMap<>();
                 List<CspReport> cspReports = new ArrayList<>();
                 StringBuilder sb = new StringBuilder();
@@ -137,11 +137,11 @@ public class CspLogUtil
                     {
                         cspReports.add(new CspReport(sb.toString()));
                         sb = new StringBuilder();
-                        foundVioloation = true;
+                        foundViolation = true;
                     }
                 }
 
-                if (!foundVioloation)
+                if (!foundViolation)
                 {
                     throw new AssertionError("Detected CSP violations but unable to parse log file: " + recentWarningsFile.getAbsolutePath());
                 }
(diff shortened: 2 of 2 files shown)


labkey-adam:
@labkey-tchad let me know if there's a good way to test this. Seems like it should have been failing since we added support for report-to.

labkey-tchad:
> @labkey-tchad let me know if there's a good way to test this. Seems like it should have been failing since we added support for report-to.

The tests are fine because our standard CSP still includes `report-uri` for backwards compatibility.
Firefox doesn't support report-to and Chrome requires HTTPS for it.
Few of our tests run with HTTPS on TeamCity and none run on Chrome so the tests don't need to support the new report format just yet. I'd be open to making the `CspReport` constructor understand both formats to future proof it though.

labkey-adam:
> > @labkey-tchad let me know if there's a good way to test this. Seems like it should have been failing since we added support for report-to.
> 
> The tests are fine because our standard CSP still includes `report-uri` for backwards compatibility. Firefox doesn't support report-to and Chrome requires HTTPS for it. Few of our tests run with HTTPS on TeamCity and none run on Chrome so the tests don't need to support the new report format just yet. I'd be open to making the `CspReport` constructor understand both formats to future proof it though.

Current Firefox versions now support `report-to`, so the related PR eliminates `report-uri` and unconditionally adds `report-to`. With that change, the tests should only see report-to-style JSON going forward. See the related issue (which requested this change), and update it if you foresee a problem.

labkey-adam:
Note that Claude is somewhat concerned about how well this helper will work with `report-to`. The `report-uri` reports were sent synchronously (or nearly so), but report-to reports are queued and sent asynchronously.

It's also worried about throwing due to reports that don't contain "body" or "documentUrl," but that seems unlikely.

labkey-tchad:
> Current Firefox versions now support `report-to`, so the related PR eliminates `report-uri` and unconditionally adds `report-to`. With that change, the tests should only see report-to-style JSON going forward. See the related issue (which requested this change), and update it if you foresee a problem.

We would have to switch all of our suites on TeamCity to use HTTPS and run on the latest Firefox version.
Luckily, the new ESR just dropped, which includes the 'report-to' support. I'm working on a PR to update the TeamCity agents.
We already have some TeamCity suites running with HTTPS. It shouldn't be too much trouble to make the change universal. I'm not positive that our agent build is actually installing the certificate so that it's recognized OS-wide. I know we install it so that Java recognizes it but we might just be using various self-signed-cert flags to make other things work.

That being said, I don't understand the rush to remove `report-uri`. Firefox has only supported it since March, and the ESR release the supports it has been out for about a week. Sure, that accounts for a fraction of a fraction of browser share but I don't see the harm in retaining it.
I'm also not sure how many folks are still running their dev servers with HTTP; I bet it's a lot.

labkey-jeckels:
> > Current Firefox versions now support `report-to`, so the related PR eliminates `report-uri` and unconditionally adds `report-to`. With that change, the tests should only see report-to-style JSON going forward. See the related issue (which requested this change), and update it if you foresee a problem.
> 
> We would have to switch all of our suites on TeamCity to use HTTPS and run on the latest Firefox version. Luckily, the new ESR just dropped, which includes the 'report-to' support. I'm working on a PR to update the TeamCity agents. We already have some TeamCity suites running with HTTPS. It shouldn't be too much trouble to make the change universal. I'm not positive that our agent build is actually installing the certificate so that it's recognized OS-wide. I know we install it so that Java recognizes it but we might just be using various self-signed-cert flags to make other things work.
> 
> That being said, I don't understand the rush to remove `report-uri`. Firefox has only supported it since March, and the ESR release the supports it has been out for about a week. Sure, that accounts for a fraction of a fraction of browser share but I don't see the harm in retaining it. I'm also not sure how many folks are still running their dev servers with HTTP; I bet it's a lot.

Even last year, we got a pen test complaints about using `report-uri` since it's deprecated (despite Firefox not supporting `report-to` yet).

I'd like to get more dev and TeamCity servers running HTTPS, so I'm happy to use this as a nudge in that direction.

That said, there's not a rush to switch. I hadn't checked the Firefox ESR schedule. Given analytics, I'm not worried about losing reports from any meaningful number of users. But let's do whatever homework needed to make this smooth on TeamCity and dev machines, and defer this to a future milestone if needed.

labkey-adam:
> 
> Even last year, we got a pen test complaints about using `report-uri` since it's deprecated (despite Firefox not supporting `report-to` yet).
> 
> I'd like to get more dev and TeamCity servers running HTTPS, so I'm happy to use this as a nudge in that direction.
> 
> That said, there's not a rush to switch. I hadn't checked the Firefox ESR schedule. Given analytics, I'm not worried about losing reports from any meaningful number of users. But let's do whatever homework needed to make this smooth on TeamCity and dev machines, and defer this to a future milestone if needed.

Okay, for now I'll back out the report-to / report-uri changes from the product PRs. And adjust the testAutomation PR to handle both report types. Once we're comfortable that TeamCity can handle both we can revisit.

labkey-tchad (approved):
The ElispotAssayTest change was an accident, I presume? Regardless, it's in a comment so won't hurt anything.
Looks good otherwise.

labkey-adam:
> The ElispotAssayTest change was an accident, I presume? Regardless, it's in a comment so won't hurt anything. Looks good otherwise.

On purpose. I migrated from Bag -> MultiSet everywhere else, and Claude insisted I do it here as well. Even though it's a comment.

PR #3134, opened 2026-07-31
Add Sources to Workflow - Automation

DariaBod:
## Rationale
Just addition to limsModule PR.

## Related Pull Requests
- https://github.com/LabKey/limsModules/pull/2361

## Changes
- Method for getting the text of the standalone action buttons on the grid bar.

### DIFF
--- a/src/org/labkey/test/components/ui/grids/GridBar.java
+++ b/src/org/labkey/test/components/ui/grids/GridBar.java
@@ -260,18 +260,14 @@ public class GridBar extends WebDriverComponent<GridBar.ElementCache>
      */
     public List<String> getButtonText()
     {
-        List<String> buttonText = new ArrayList<>();
-        for (WebElement button : BootstrapLocators.button().findElements(this))
-        {
-            String cssClass = button.getAttribute("class");
-            if (cssClass == null || !cssClass.contains("dropdown-toggle"))
-            {
-                String text = button.getText().trim();
-                if (!text.isEmpty())
-                    buttonText.add(text);
-            }
-        }
-        return buttonText;
+        return BootstrapLocators.button().findElements(this).stream()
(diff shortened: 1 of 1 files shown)


src/org/labkey/test/components/ui/grids/GridBar.java
```diff
@@ -252,6 +252,28 @@ public List<String> getMenuButtonsText()
...
+            String cssClass = button.getAttribute("class");
+            if (cssClass == null || !cssClass.contains("dropdown-toggle"))
+            {
+                String text = button.getText().trim();
+                if (!text.isEmpty())
+                    buttonText.add(text);
+            }
+        }
+        return buttonText;
+    }
```
labkey-danield, line 268: You could use stream and filter for this, but I don't think that makes it any more readable.

labkey-danield (approved):

src/org/labkey/test/components/ui/grids/GridBar.java
```diff
@@ -252,6 +252,24 @@ public List<String> getMenuButtonsText()
...
+     *
+     * @return the (non-empty) labels of the standalone buttons.
+     */
+    public List<String> getButtonText()
+    {
+        return BootstrapLocators.button().findElements(this).stream()
+                .filter(button -> {
+                    String cssClass = button.getAttribute("class");
+                    return cssClass == null || !cssClass.contains("dropdown-toggle");
+                })
```
labkey-tchad: You can include the unwanted class in the locator rather than iterating through all of the buttons.
```suggestion
        return BootstrapLocators.button().withoutClass("dropdown-toggle").findElements(this)
                .stream()
```

labkey-tchad (approved):

PR #159, opened 2019-11-12
19.3 fb issue38940

labkey-chrisj:
restores field-level error and warning validation test coverage in Domain Designer.
adds synchronizing wait-for-error/warning methods to domainFieldRow

PR #3139, opened 2026-08-03
GitHub Issue #159: AssayExportImportTest fix to use assay designer UI to add transform script

cnathe:
## Rationale
The changes for this issue https://github.com/LabKey/internal-issues/issues/159 require that assay transform scripts live in the LabKey container's `@scripts` directory. This PR fixes the AssayExportImportTest to upload the script to the expected location using the assay designer UI.

## Related Pull Requests
- https://github.com/LabKey/platform/pull/7895

## Changes
- AssayExportImportTest to upload transform script to @script dir using assay designer UI


labkey-tchad (approved):
Does the Java remote API need an update for this?

cnathe:
> Does the Java remote API need an update for this?

I don't think so. The SaveProtocolCommand in the java API could still use the setProtocolTransformScripts helper if it was providing a file path within the container's `@scripts`. So it should still work under those conditions. However, any usages that were just referring to a different path will stop working (this is the only usage I found in the tests).

